Issue #21, opened 2021-09-23
CI/CD


foundrytom:
Due to the unique way in which GitHub billing works, we can't actually enable `Actions` until this is a public repo.

# What

- [x] Run `pytest` via CI for PRs.
- [x] #103 (see #25)
- [x] PR status check that blocks merge unless DCO is satisfied for each commit of the PR (#66)
- [ ] #101

PR #91, opened 2021-11-22
[CI] Run `pytest` for Pull Requests

foundrytom:
Part of #21, ensures `pytest` is run for all PRs, preventing merges on failures.

There is a [mirror PR](https://github.com/foundrytom/OpenAssetIO/pull/2) that shows the action running.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,35 @@
+# A simple workflow to run tests using pytest against supported Python versions
+
+name: Tests
+on:
+  pull_request:
+    branches-ignore:
+      - docs
+    paths-ignore:
+      - 'docs/**'
+
+jobs:
+  pytest:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -r tests/requirements.txt
+
+    - name: Build
+      run: |
+        python -m pip install .
+
+    - name: Test
+      run: |
+        pytest
+


.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
+# A simple workflow to run tests using pytest against supported python versions
```
TomFoundry, line 1: [consider]
```suggestion
# A simple workflow to run tests using pytest against supported Python versions
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -r tests/requirements.txt
+    - name: Build
+      run: |
+        python -m pip install .
```
TomFoundry, line 21: [question] Why is this command part of a separate `Build` step, rather than part of `Install dependencies`?

TomFoundry (approved):

.github/workflows/test.yml
foundrytom: Just as this installs our code, instead of dependencies of the test infra

.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
+# A simple workflow to run tests using pytest against supported Python versions
+
+name: Tests
+on: [ pull_request ]
```
feltech, line 4: [consider][minor] [allow/ignore](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-paths) appropriate paths, so we don't run pytest if only `doc` changes. We could only "allow" the `python` directory, but that's perhaps less robust in case we add python files elsewhere, we'd have to update this rule.

feltech (approved):

.github/workflows/test.yml
foundrytom: Ahh, good catch, updated [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/91/commits/82f2b0d0c1e48eb55fa5050f319a19c48f3d6235) - dummy PR [here](https://github.com/foundrytom/OpenAssetIO/pull/4), showing the check is skipped for the `docs` branch.

feltech: Good idea to do that too, though not quite what I meant - I was talking about skipping based on paths in the main repo.  Also the dummy PR you link still runs pytest :(

foundrytom: Doh! Im sure I looked at that and it exited early! 🙃 For some reason, despite the PR base being `docs` it's insisting on merging into `main` 🤷:
```
HEAD is now at 96827ef Merge 82f2b0d0c1e48eb55fa5050f319a19c48f3d6235 into e8f82dd47e83d5e7a62d163d3ae7b63d3b239491
```

`e8f82` is `HEAD` on `main`....

foundrytom: It _seems_ like something to do with the run definition changes when you push, not just re-run. I pushed [this](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/91/commits/7e996aa0e1747b3cd6193cf4c45450ba876fadd8) commit, and it seems to have properly skipped the check [here](https://github.com/foundrytom/OpenAssetIO/pull/2), and they seem to be running [here](https://github.com/foundrytom/OpenAssetIO/pull/5).

foundrytom: It's definitely mixing up runs. I made a fresh PR -> `docs` on my fork, and its auto-linking the check to the results from the old PR 😠

feltech:
Looks good, squash and merge when ready.

PR #99, opened 2021-11-24
[Build] Use `npm ci` instead of `npm install .` for docs build

foundrytom:
We should be using the pinned versions. `npm install` doesn't respect package-lock.json. In addition, updates picked up during `make deploy` would then cause the deploy to fail as there are un-staged changes to package-lock.json.

`npm ci` (https://docs.npmjs.com/cli/v6/commands/npm-ci) does what we want.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -111,7 +111,7 @@ src/styles.css: src/styles.scss $(SASS)
 #
 
 $(SASS):
-	npm install .
+	npm ci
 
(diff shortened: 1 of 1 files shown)


feltech:
For info, looks like `npm install` behaviour changed in 2017 by way of https://github.com/npm/npm/issues/16866

feltech (approved):

PR #103, opened 2021-11-25
[Docs] Continuous deployment of GitHub Pages from `main`

foundrytom:
For #21, sets up CD that continually deploys `main` to GitHub pages.

This required a little refactoring of the `Makefile` to split out the deployment logic. It was getting a bit complex for single line shell foo.

There are runs for a deployment [here](https://github.com/foundrytom/OpenAssetIO/runs/4326630988?check_suite_focus=true), and one where there is no change to the built docs [here](https://github.com/foundrytom/OpenAssetIO/runs/4326658874?check_suite_focus=true).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,46 @@
+name: "Docs"
+
+concurrency:
+  group: docs
+  cancel-in-progress: true
+
+on:
+  push:
+    branches: [ main ]
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+    - name: Checkout source
+      uses: actions/checkout@v2
+      with:
+        path: src
+
+      # The checkout action ensures the correct auth tokens
+      # are set in the repo so we can push. We will pass
+      # this path to the deploy stage to use instead of
+      # its own temporary checkout.
+    - name: Checkout docs branch
+      uses: actions/checkout@v2
+      with:
+        path: target
+        ref: docs
+
+    - name: Build
+      run: |
+        # TODO [tc] We should really publish the Docker
+        # image to the GitHub registry
+        make -C src/doc
+
+    - name: Deploy
+      run: |
+        # Global means we don't need hop around the checkouts to set this
+        git config --global user.name "GitHub Actions"
+        git config --global user.email "github-actions@github.com"
+        # Pass the path to the actions/checkout@v2 docs branch checkout
+        OAIO_DOCS_REPO_DIR=`pwd`/target make -C src/doc deploy
--- a/doc/deploy.sh
+++ b/doc/deploy.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+
+# Copyright 2013-2021 [The Foundry Visionmongers Ltd]
+# SPDX-License-Identifier: Apache-2.0
+
+# A script to deploy an existing documentation build in
+# the current working directory to the target branch of
+# the specifed local repository clone.
+#
+# The script expects the build to exist at `./html`,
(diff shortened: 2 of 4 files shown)


doc/deploy.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+
+# Copyright 2013-2021 [The Foundry Visionmongers Ltd]
```
feltech, line 3: [question] Are these two lines something we should be including in all our sources?

doc/deploy.sh
```diff
@@ -0,0 +1,102 @@
...
+rm -rf "${target_dir}"
+# Copy in the new build
+cp -r ./html "${target_dir}"
+
+# See if there are any resulting changes in the target
+# branch, commit and push if so.
+
+# Ensure we restore the working dir if any actions in the
+# target repo fail
+trap restore_cwd EXIT
```
feltech, line 80: [question] This should only be necessary if you're `source`ing the script, is that something we envision? Also perhaps `pushd`/`popd` is of interest.

doc/deploy.sh
```diff
@@ -0,0 +1,102 @@
...
+echo "Updated files:"
+git status -s
+
+git add .
+git commit -s -m "Update docs to ${commit_hash}"
+git push
+
+# We're done, and now need to go back to the original dir
+restore_cwd
+trap - EXIT
```
feltech, line 99: [consider] This shouldn't be necessary, `EXIT` trap will happen during normal script exit.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,46 @@
...
+      # its own temporary checkout.
+    - name: Checkout docs branch
+      uses: actions/checkout@v2
+      with:
+        path: target
+        ref: docs
+
+    - name: Build
+      run: |
+        # TODO [tc] We should really publish the Docker
```
feltech, line 36: [plan] Indeed! Testing this on my fork and it's taking a looong time to run.

feltech (approved):
Awesome! Tested on my fork, took a while but works really well.

doc/deploy.sh
foundrytom: Yeah. We need to make sure there is an Apache 2 licence, the `# SPDX-License-Identifier: Apache-2.0` was noted by a contact a the Linux Foundation as being satisfactory to indicate that license.

We should stick with the existing pattern we have in the `.py` files for now though (short Apache 2 string instead of SPDX).

doc/deploy.sh
foundrytom: Doh! Of course yes. No not intending a `source`. Something was stuck in my mind that one needed to do such house keeping. But as you say, it is redundant. Removed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/103/commits/a47f8d5cbf805d9450cb6ae5b7cb2c51eefa1b41).

doc/deploy.sh
foundrytom: Also removed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/103/commits/a47f8d5cbf805d9450cb6ae5b7cb2c51eefa1b41.

.github/workflows/docs.yml
foundrytom: Added https://github.com/TheFoundryVisionmongers/OpenAssetIO/issues/109

doc/deploy.sh
feltech: OK, I've added a review comment to my own PR to address this in a test file I added: https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/98/files#r758354644

feltech (approved):

feltech:
Squash and merge when ready.

PR #94, opened 2021-11-23
[Docs] Misc documentation fixes

foundrytom:

PR #104, opened 2021-11-26
[CI] Build docs for Pull Requests to catch markup errors

foundrytom:
Part of #21, builds docs as for pull requests, and adds the Doxygen problem matcher. 

This should allow us to catch docs build errors through the status checks. 

For now, this is a known-fail until we fix the remaining Doxygen warnings due to the UI and other sundry parameter errors. #94  and #98 resolve many of the warnings reported on `main`, bar the UI ones.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,3 +33,27 @@ jobs:
       run: |
         pytest
 
+  docs:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout source
+      uses: actions/checkout@v2
+
+    - name: Test
+      run: |
+        echo "::add-matcher::./.github/workflows/problem-matchers/doxygen.json"
+        # Problem matches don't actually error a step, so we have to
+        # manually check for warnings/errors at the end. Capture the
+        # output so we can parse it later.
+        make -C doc 2>&1 | tee doxygen-log.txt
+        # Remove to prevent the error summary from the grep below from
+        # being matched too.
+        echo "::remove-matcher owner=doxygen::"
+        # Fail the job if we have Doxygen warning/error lines in the
+        # output We don't suppress the output as it forms an easier
+        # to digest error summary.
+        # NB: This is the same regex as doxygen.json, adapted to work
+        # with GNU grep.
+        ! grep -E "^/src/([^:]+):([0-9]+): ?([a-zA-Z]+): ?(.*)$" doxygen-log.txt
+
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -33,3 +33,23 @@ jobs:
...
+        echo "::add-matcher::./.github/workflows/problem-matchers/doxygen.json"
+        # Problem matches don't actually error a step, so we have to manually
+        # check for warnings/errors at the end. Capture the output so we can
+        # parse it later.
+        make -C doc 2>&1 | tee doxygen-log.txt
+        echo "::remove-matcher owner=doxygen::"
+        # Fail the job if we have Doxygen warning/error lines in the output
+        # Note: Was unable to get the regex from the problem matcher to match
+        # with grep, when run in GitHub actions (worked fine locally).
+        ! grep -E "/src/.+:[0-9]+: ?(warning|error):" doxygen-log.txt
```
foundrytom, line 54: [note] Deliberately left the output from grep visible, as it gives a summary of the failures in one place.

.github/workflows/test.yml
```diff
@@ -33,3 +33,23 @@ jobs:
...
+      uses: actions/checkout@v2
+
+    - name: Test
+      run: |
+        echo "::add-matcher::./.github/workflows/problem-matchers/doxygen.json"
+        # Problem matches don't actually error a step, so we have to manually
+        # check for warnings/errors at the end. Capture the output so we can
+        # parse it later.
+        make -C doc 2>&1 | tee doxygen-log.txt
+        echo "::remove-matcher owner=doxygen::"
```
feltech, line 50: [question] Is removing the matcher necessary?  

I checked a couple of the repos under [Examples in the docs](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md#examples) and they don't remove their matcher.

One example I found of `remove-matcher` was to disable the built-in eslint matcher set up by `setup-npm` in https://github.com/actions/setup-node/issues/72

Is it perhaps that matchers live on beyond the step/action?

foundrytom: Added a note [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/104/commits/6764e7a1b770840bfe3c2d281170e56c9d30dcd4) on this.

foundrytom:
@feltech Fixed the regex based on our chat earlier - good call on `\d`. Added a note [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/104/commits/00ce1f4459c37b6d304acb61016e1161d0087273).

feltech (approved):
All good, just a shame about the mismatched line numbers :(

feltech:
If there's really nothing we can do about the line numbers, then LGTM, squash and merge when ready.

foundrytom:
> If there's really nothing we can do about the line numbers, then LGTM, squash and merge when ready.

I think it'll get a lot better when we're in `cpp`, but there still seems to be some pre-processing happening. One for a rainy day or a look into over Christmas.

PR #105, opened 2021-11-26
[CI] Run `pylint` for Pull Requests

feltech:
Part of #21. Set up environment then run `pylint` for warnings and
errors only (for now). Uses a custom Javascript action that reports via
GitHub annotations a summary, followed by specific issues associated
with their corresponding source line.

### DIFF
--- a/.github/actions/pylint/README.md
+++ b/.github/actions/pylint/README.md
@@ -30,3 +30,14 @@ with:
   pylint-rcfile: "./pyproject.toml"
   pylint-paths: "python tests"
 ```
+
+## Building
+
+To bundle the source for use as a GitHub Action, assuming the current
+working directory contains `package.json` run
+```shell
+npm install
+npm run build
+```
+This will create a bundled build under `dist`, which is referenced by
+the `action.yml`.
--- a/.github/actions/pylint/README.md
+++ b/.github/actions/pylint/README.md
@@ -1,0 +1,32 @@
+# Pylint javascript action
+
+This action executes [Pylint](https://pylint.pycqa.org/en/latest) on the
+codebase and reports a summary in the action log and annotations on
+source lines of a PR.
+
+It is assumed that `pylint` is already installed in the environment.
+
+## Inputs
+
+## `pylint-paths`
+
+**Required** The paths to instruct `pylint` to scan, space-separated.
+
+## `pylint-rcfile`
+
+Pylint configuration file location. Specifically, the value of 
+`pylint`'s `--rcfile=` command-line option.
+
+## `pylint-disable`
+
+Pylint checks to disable. Specifically, the value of `pylint`'s 
+`--disable=` command-line option.
+
+## Example usage
+```yaml
+uses: ./.github/actions/pylint
+with:
+  pylint-disable: "C,I,R"  # Only errors and warnings.
+  pylint-rcfile: "./pyproject.toml"
+  pylint-paths: "python tests"
+```
--- a/.github/actions/pylint/index.js
+++ b/.github/actions/pylint/index.js
@@ -18,11 +18,11 @@ const exec = require('@actions/exec');
  * Execute pylint and report issues, if any found.
  */
 async function run() {
-    let pylintOutput = "";  // Pylint process output.
     const pylintDisable = core.getInput('pylint-disable') || "";
     const pylintRCFile = core.getInput('pylint-rcfile') || "";
     const pylintPaths = core.getInput('pylint-paths');
 
+    let pylintOutput = "";  // Pylint process output.
     try {
         // Run pylint
         await exec.exec("pylint", [
--- a/.github/actions/pylint/index.js
+++ b/.github/actions/pylint/index.js
@@ -24,7 +24,10 @@ async function run() {
 
     let pylintOutput = "";  // Pylint process output.
     try {
-        // Run pylint
+        // Run pylint via `exec()`, which throws an exception if the
+        // process exits with a non-zero exit code. Here, we assume a
+        // non-zero exit code (and hence exception) means pylint has
+        // detected issues in the code that we need to report.
         await exec.exec("pylint", [
             `--disable=${pylintDisable}`,
             `--rcfile=${pylintRCFile}`,
--- a/.github/actions/pylint/index.js
+++ b/.github/actions/pylint/index.js
@@ -1,0 +1,120 @@
+const core = require('@actions/core');
+const exec = require('@actions/exec');
+
+/**
+ * Execute pylint and report issues, if any found.
+ */
+async function run() {
+    let pylintOutput = "";  // Pylint process output.
+    const pylintDisable = core.getInput('pylint-disable') || "";
+    const pylintRCFile = core.getInput('pylint-rcfile') || "";
+    const pylintPaths = core.getInput('pylint-paths');
+
+    try {
+        // Run pylint
+        await exec.exec("pylint", [
+            `--disable=${pylintDisable}`,
+            `--rcfile=${pylintRCFile}`,
+            `--output-format=json`,
+            ...pylintPaths.split(" ")
+        ], {
+            listeners: {
+                stdout: (data) => pylintOutput += data.toString()
+            }
+        });
+    } catch (pylintError) {
+        try {
+            reportResults(JSON.parse(pylintOutput));
+        } catch (reportingError) {
+            // The exception thrown by `exec` doesn't include the exit
+            // code as a field (we'd have to parse the error message).
+            // So report the pylint process error here as well, in case
+            // that was the root cause for `reportResults` failing.
+            console.error(pylintError);
+            console.error(reportingError);
+        }
+    }
+}
+
+/**
+ * Create GitHub annotations and set the action to failed.
+ *
+ * Report a summary followed by per-line annotations.
+ *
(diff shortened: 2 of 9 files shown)


.github/actions/pylint/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "pylint-oaio-gh-action",
+    "version": "1.0.0",
+    "description": "Run pylint as a GitHub action",
+    "main": "index.js",
+    "scripts": {
+        "build": "ncc build index.js --license licenses.txt"
```
foundrytom, line 7: [fix] I think we need a `README.md` that explains the development process here. Alternately, if we hosted this in a separate `REPO` is there an off-the-shelf GitHub workflow that will build/release actions based on tags?

foundrytom (changes requested):
WIP review, as the GitHub Actions Docs currently have some varnish cache issues so I can't look some things up 🙃

.github/actions/pylint/index.js
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Execute pylint and report issues, if any found.
+ */
+async function run() {
+    let pylintOutput = "";  // Pylint process output.
+    const pylintDisable = core.getInput('pylint-disable') || "";
+    const pylintRCFile = core.getInput('pylint-rcfile') || "";
+    const pylintPaths = core.getInput('pylint-paths');
+
+    try {
```
foundrytom, line 13: [fix] Be good to add a little note here on the expected exception/error handling from `exec.exec` as its not immediately obvious from the docs. As JS isn't our every day language, would help future maintainers.

.github/actions/pylint/index.js
```diff
@@ -0,0 +1,120 @@
...
+        // Run pylint
+        await exec.exec("pylint", [
+            `--disable=${pylintDisable}`,
+            `--rcfile=${pylintRCFile}`,
+            `--output-format=json`,
+            ...pylintPaths.split(" ")
+        ], {
+            listeners: {
+                stdout: (data) => pylintOutput += data.toString()
+            }
```
foundrytom, line 23: [question] Do we ever need to worry about `stderr`?

foundrytom (changes requested):
Great stuff @feltech, particularly the annotations - great to be able to see them in context, in the code.

.github/actions/pylint/index.js
```diff
@@ -0,0 +1,120 @@
...
+    const pylintWarnings = pylintMessages.filter(message => message.type === 'warning');
+    const pylintConvention = pylintMessages.filter(message => message.type === 'convention');
+    const pylintRefactor = pylintMessages.filter(message => message.type === 'refactor');
+    const pylintInfo = pylintMessages.filter(message => message.type === 'info');
+
+    const prAnnotationMsg = `\n${pylintErrors.length} error\n\n` +
+        `${pylintWarnings.length} warning\n\n` +
+        `${pylintConvention.length} convention\n\n` +
+        `${pylintRefactor.length} refactor\n\n` +
+        `${pylintInfo.length} info`;
```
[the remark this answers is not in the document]
feltech: Good point. Perhaps worth adding the value of the `disable=` argument too, so people reading it won't scratch their head wondering why other severity levels aren't reported at all.

.github/actions/pylint/index.js
```diff
@@ -0,0 +1,120 @@
...
+}
+
+/**
+ * Parse a Pylint message into a GitHub annotation message.
+ *
+ * @param message Pylint message encoded as JSON.e
+ * @returns string Message to display on annotation.
+ */
+function annotationMessage(message) {
+    if (message.obj) {
```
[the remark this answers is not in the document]
feltech: Good catch.

.github/actions/pylint/index.js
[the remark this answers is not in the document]
foundrytom: Sounds ideal!

.github/actions/pylint/index.js
```diff
@@ -0,0 +1,120 @@
...
+        title: "Linting summary"
+    };
+
+    core.setFailed("Pylint linter issues detected");
+
+    if (pylintErrors.length) {
+        core.error(prAnnotationMsg, prAnnotationProps);
+    } else {
+        core.warning(prAnnotationMsg, prAnnotationProps);
+    }
```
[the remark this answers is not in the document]
feltech: I considered that, the problem is you don't get to customize the title. But yeah I agree it would be nice to free up an annotation, so will roll them together.

.github/actions/pylint/index.js
feltech: I don't think so. We can add that too, but I'm wary of "just in case" code. The one case I can think of that does print to stderr is if you use an unrecognized command line switch, e.g.
```
$ pylint --fgsffds 1>/dev/null
Usage: pylint [options]

pylint: error: no such option: --fgsffds
```
but we don't care about that.

foundrytom: Sounds fine as is, thanks

.github/actions/pylint/package.json
feltech: I'll address the issues in this repo first, then when it's looking OK we can abandon it and move the code to a new repo. Sound sensible?

foundrytom: A fine plan indeed :)

.github/actions/pylint/index.js
```diff
@@ -0,0 +1,120 @@
+const core = require('@actions/core');
```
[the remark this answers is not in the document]
feltech: Added in 08642ca

.github/actions/pylint/index.js
```diff
@@ -0,0 +1,120 @@
...
+const exec = require('@actions/exec');
+
```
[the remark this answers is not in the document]
feltech: Added a module docstring in 958330b

.github/actions/pylint/index.js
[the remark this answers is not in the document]
feltech: Updated in e3af06d

.github/actions/pylint/index.js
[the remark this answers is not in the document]
feltech: Simplified in bfa4dcb

.github/actions/pylint/index.js
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Execute pylint and report issues, if any found.
+ */
+async function run() {
+    let pylintOutput = "";  // Pylint process output.
```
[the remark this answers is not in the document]
feltech: Moved in 9c8f7c1

.github/actions/pylint/index.js
feltech: Added a comment in d3e8a3f

.github/actions/pylint/package.json
feltech: Added a README in a454ff8 based on the template in [the docs](https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action#creating-a-readme)

.github/actions/pylint/index.js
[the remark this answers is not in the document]
feltech: Updated as part of e3af06d whilst addressing https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/105#discussion_r758654438

.github/actions/pylint/package.json
feltech: Updated the README to actually address your comment in 2f51b2f :roll_eyes:

.github/actions/pylint/index.js
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Parse a Pylint message into GitHub annotation properties.
+ *
+ * @param message Pylint message decoded from JSON.
+ * @returns core.AnnotationProperties Github annotation properties.
+ */
+function annotationProperties(message) {
+    return {
+        title: `Linter ${message.type} ${message['message-id']} (${message.symbol})`,
```
feltech, line 117: [question] Annoyingly, the annotation title doesn't show up in the action log. The title only seems to show when displayed in the action "Summary" or when displayed alongside the source.

Do we care?

.github/actions/pylint/README.md
```diff
@@ -30,3 +30,14 @@ with:
   pylint-rcfile: "./pyproject.toml"
   pylint-paths: "python tests"
 ```
+
+## Building
+
+To bundle the source for use as a GitHub Action, assuming the current
```
foundrytom, line 36: [question] Right now, the build is in the source tree and checked in (which is less common for us). Is this going to change when we release it properly, or does all that still have to happen? If so, be good to explain that in a little more detail - as it might be less common for people with the general background of the project (python/cpp).

.github/actions/pylint/README.md
```diff
@@ -30,3 +30,14 @@ with:
...
+working directory contains `package.json` run
+```shell
+npm install
```
foundrytom, line 39: [consider] Is it worth adding that installing `npm` is a pre-requisite?

.github/actions/pylint/index.js
foundrytom: The appearance as-is is still pretty good - and a million times better than the previous state. So I'm happy as is.

PR #112, opened 2021-11-30
[CI] Suppress summary output from grep when checking for build errors

foundrytom:
#104 introduced PR checks for the Doxygen docs build. 

Due to the unique way in which problem matchers work, we require manual effort to actually fail the job.  We'd been using `! grep` for this purpose, and printing its output as a summary of failures. However, some of the Doxygen warnings are multi-line, and so the summary is less helpful. 

 This switches grep to quiet mode. The reader can always click-through from the annotation to the correct line in the log anyway, which makes it easier to see the remaining lines of the message until (if) we figure out a multi-line matcher.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,13 +47,9 @@ jobs:
         # manually check for warnings/errors at the end. Capture the
         # output so we can parse it later.
         make -C doc 2>&1 | tee doxygen-log.txt
-        # Remove to prevent the error summary from the grep below from
-        # being matched too.
         echo "::remove-matcher owner=doxygen::"
         # Fail the job if we have Doxygen warning/error lines in the
-        # output We don't suppress the output as it forms an easier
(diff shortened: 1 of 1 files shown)


feltech (approved):
I'm happy with this, the Doxygen logs aren't so long as to make scrolling/searching them difficult.

The [multiline matcher](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md#multiline-matching) syntax doesn't look too onerous, but that's for another PR. 

Assuming that's fixed one day, using the final `grep` just to control error status makes our lives easier, since then there's no need to also update the `grep` pattern to support multiline.

feltech:

foundrytom:
> Assuming that's fixed one day, using the final `grep` just to control error status makes our lives easier, since then there's no need to also update the `grep` pattern to support multiline.

Yeah, that's a really good point.

Issue #86, opened 2021-11-18
Remove `thumbnailSpecification`


feltech:
# What

Handle thumbnails for entities generically, rather than through the existing special-case of the `thumbnailSpecification` API.

# Why

Currently we have a special case to handle thumbnails, whereby all `EntitySpecification`s include a `thumbnailPath` attribute, so that when `register`ing an entity the manager can also pull in the thumbnail.  

In order to determine whether the manager wants a thumbnail, and if so what are the desired attributes (e.g. width and height), `thumbnailSpecification` must be called by the host, passing the `Specification` of the entity that is about to be published.

However, we probably already have mechanisms of accomplishing these steps in a generic way.  One proposal could be

* Add a `kWantsThumbnail` bit to the `managementPolicy`, to determine if the `Specification` of the target entity that is about to be published also needs a thumbnail.
* `preflight` the target entity, passing a `ThumbnailSpecification`. The return value of `preflight` is then the writeable entity reference for the thumbnail, say `thumbnailRef`.
* Get the required thumbnail properties via `entityAttributes([thumbnailRef], ["thumbnailWidth", "thumbnailHeight", ...], ...)` (based on previous discussion of how we would like to redesign `getEntityMetadata`)
* Write the thumbnail to the path returned by `resolveEntityReference([thumbnailRef], ...)`
* `register` the thumbnail.

foundrytom:
This also has the added benefit that the mechanism is truly bi-directional. The host uses the standard publishing semantic to register a thumbnail, but can also use the standard `getRelatedReferences` mechanism with a `ThumbnailSpecification` to request a thumbnail from the manager for any given entity.

PR #116, opened 2021-12-02
[Core] Remove `ManagerInterface.thumbnailSpecification`

foundrytom:
Implements the approach proposed in #86. It's just plain simple better. Thumbnails no longer need bespoke API, and now just use the standard query/publish semantics applicable to any other entity.

This updates the documentation and adds an example of the correct implementation in a host.


doc/src/Thumbnails.dox
```diff
@@ -0,0 +1,57 @@
...
+ * to look up a thumbnail for an existing entity.
+ *
+ * @section thumbnails_creation_during_publish Requesting Thumbnail Creation During Publishing
+ *
+ * In many situations, it is difficult for a @ref manager to
+ * retrospectively generate a thumbnail for any given @ref entity. It
+ * usually requires first-class knowledge of how the specific entity
+ * type can be rendered to into an image.
+ *
+ * For image-based entities, this is potentially straight forward. The
```
feltech, line 17: [fix]
```suggestion
 * For image-based entities, this is potentially straightforward. The
```

doc/src/Thumbnails.dox
```diff
@@ -0,0 +1,57 @@
...
+ * - Call @ref openassetio.managerAPI.ManagerInterface.ManagerInterface.getEntityMetadata
+ *   "getEntityMetadata" on the returned reference to determine the desired width/height/format
+ *   for the thumbnail.
+ * - Generate a thumbnail and @ref register it to the same reference.
+ *
+ * @note The thumbnail generation process may be carried out
+ * asynchronously by the host and registration may occur at an arbitrary
+ * time in the future.
+ *
+ * @see example_generating_a_thumbnail
```
feltech, line 42: [fix] This ref doesn't render.

python/openassetio/constants.py
```diff
@@ -48,6 +48,10 @@
 ## will determine the URL to use for writing data for new entities
 kWillManagePath = 8
 
+## If True, and supported by a @ref host, a thumbnail will be generated by the
```
feltech, line 51: [fix] 
```suggestion
## If True, and supported by a @ref host, a thumbnail will be generated by the
```
Subtle NBSP (how did you manage that?), but does affect the rendered output.

doc/src/Thumbnails.dox
```diff
@@ -0,0 +1,57 @@
+/*!
+ * @page thumbnails Thumbnails
+ *
+ * @section thumbnails_overview Overview
+ *
+ * OAIO provides a mechanism for a @ref manager to request a thumbnail
+ * from the @ref host when a new @ref entity is created, and for a host
+ * to look up a thumbnail for an existing entity.
+ *
+ * @section thumbnails_creation_during_publish Requesting Thumbnail Creation During Publishing
```
feltech, line 10: [consider][minor] This title gets cut off (well, causes a horizontal scrollbar) in the sidebar. Very minor quibble.

feltech (changes requested):

doc/src/Thumbnails.dox
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @note The thumbnail generation process may be carried out
+ * asynchronously by the host and registration may occur at an arbitrary
+ * time in the future.
+ *
+ * @see example_generating_a_thumbnail
+ *
+ * @todo Document the attributes used to determine width/hight/format,
+ * these should be the same as the fields of the
+ * ThumbnailSpecification.
```
feltech, line 46: [question] What is the interaction between width/height/format in the Specification vs. the Attributes?  I can understand the attributes are what the manager wants. So what's the use case for having these fields also in the specification?  Is the idea perhaps that a host can specify a list of formats and resolutions it supports, then the attributes given by the manager are selected from that list?

foundrytom: Yeah that thing, and when (in the future) a host requests a thumbnail from a manager via `getRelatedReferences`, they can request a specific format/size.

doc/src/Examples.dox
```diff
@@ -202,4 +202,48 @@
  * with open(os.path.join(os.path.expanduser('~'), 'history', 'a') as f:
  *     f.write(f"{final_ref}\n")
  * @endcode
+ *
+ * @section example_generating_a_thumbnail Generating a Thumbnail During Publish
```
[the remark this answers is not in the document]
foundrytom: Saving this for later, as we have other long titles.

doc/src/Thumbnails.dox
foundrytom: Saving this for later, as we have other long titles.

doc/src/Thumbnails.dox
foundrytom: Fixed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/116/commits/f16b1413aaa134ce202f67352f1fa6e597ebf15a).

python/openassetio/constants.py
foundrytom: Fixed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/116/commits/34e2fdf021f156a3c1c9328c61b4027d3b1378b1)

feltech (approved):

PR #117, opened 2021-12-02
[Core] Rename `metadata` to `attributes`

### LINKED ISSUE #3: RFC 02: Rename Metadata
## Proposal

Rename `metadata` to `properties` or `attributes` or similar.

## Rationale

Metadata, as a term, is too closely aligned with “ancillary information”, as we support (and encourage) its use for primary asset information, it needs to be named accordingly.

## Background

`OAIO` supports interactions with "non file-based" assets. An example of which being querying or creating shots under a sequence. There is no real meaningful "file" for a shot (compared to say, a geometry cache or image sequence). A coarse generalisation may define a shot as being made up of a 'name' and a frame range + handles. Rather than forcing this data into a sidecar file, `OAIO` models these kind of assets using simple key/value pairs, with an empty `primaryString`.

Tools that wished to query the frame range for a shot, would simple call `getEntityMetadata( shotRef, ... )` and look for the well-known `startFrame` and `endFrame` keys. This works well in practice (lets skip conversations about where 'well known keys' are defined for now). 

The issue is that this isn't really 'metadata', it's actually the _data_ for that shot. So the terminology `metadata` is misleading.  This also leads on to #4.

foundrytom:
Closes #3. Using the term "metadata" was fundamentally misleading. There
is nothing "meta" about the data held in an entity's associated key/value
pairs. It is considered part of the entity itself - as a companion to
the primary string.

Metadata also conjured up thoughts around timestamps, permissions etc. when raised in discussions around the purpose of the API.

We considered 'data', but it was too ambiguous with the 'data' that may be referenced by the primary string when it is an URL.

We considered 'properties', and enjoyed a journey down a philosophical rabbit hole reading Wikipedia articles on the philosphy of attributes and properties and ended up coming up none the wiser.

We settled on attributes, probably for no specific reason. The computing version of the attributes wikipedia entry says this:
 
> "In computing, an attribute is a specification that defines a property of an object"

So attributes felt more fitting as within OAIO we desire to anchor concrete meaning and usage to this data. I'm sure it'll be the wrong way around which ever one we pick. 🙃

## Review notes

- The first commit adjusts method names, but doesn't update documentation or docstrings.
- The other commits update docs/etc.

### DIFF
--- a/doc/src/Glossary.dox
+++ b/doc/src/Glossary.dox
@@ -76,7 +76,7 @@
  *
  * @li name
  * @li @ref primary_string
- * @li @ref metadata
+ * @li @ref attributes
  *
  * This makes it relatively straightforward for an @ref
  * asset_management_system to provide support for any entity that a @ref
@@ -314,42 +314,44 @@
  *   @see openassetio.managerAPI.ManagerInterface.ManagerInterface.thawState
  *
  *
- * @section metadata Metadata
+ * @section attributes Attributes
  *
- * Within the core API, Metadata is used to persist additional properties
- * of an @ref entity. This is particularly useful when there is no
- * meaningful @ref primary_string for any given type of entity.
+ * Within the core API, Attributes are used to persist additional
+ * properties of an @ref entity. This is particularly useful when there
+ * is no meaningful @ref primary_string for any given type of entity.
  *
- * The @ref manager must guarantee to store and recall any metadata set
- * by the @ref host. They may freely re-map any keys to their local
- * names internally, as long as the mapping is reversed for queries.
+ * The @ref manager must guarantee to store and recall any attributes set
+ * by the @ref host. They may freely re-map any attribute names to their
+ * local names internally, as long as the mapping is reversed for
+ * queries.
  *
- * Metadata is limited to **string** keys, and **plain old data** typed values.
+ * Attributes names are strings, and their values are simple **plain
+ * old data** types (ie. int, float, string, bool).
  *
  * > Note: The length of a value is yet to be defined, but ideally it
  * > will be something suitably large - this will only be restricted if
  * > common host applications and/or managers have recurring practical
  * > limits.
  *
- * A simple example of where metadata brings great value is something
- * as humble as an ImageSpecification entity that resolves to an image
- * sequence.
- *  - When a host wishes to read the sequence, if the entity's metadata
- *  has the well-known key for colorspace, then it can ensure the data
- *  is correctly interpreted, without relying on fragile mechanisms such
- *  as the file extension.
- *  - When a host wishes to write image data, if the entity's metadata
- *  has the well-known keys for the file format and associated encoding
- *  options, the data can be reliably written with the correct format,
- *  bit-depth and compression.
+ * A simple example of where entity attributes bring great value is
+ * something as humble as an ImageSpecification entity that resolves to
+ * an image sequence.
+ *  - When a host wishes to read the sequence, if the entity has a
+ *  colorspace attribute then it can ensure the data is correctly
+ *  interpreted, without relying on fragile mechanisms such as the file
+ *  extension.
+ *  - When a host wishes to write image data, the entity's file format
+ *  attribute (along with any associated encoding options) can be used
+ *  to ensure the data is reliably written with the correct format,
+ *  bit-depth and compression, etc.
  *
  * Shots are another type of entity that makes use of this mechanism. As
  * there is no real meaningful @ref primary_string, all relevant data
  * about its frame range, handles, etc. are persisted via the entity's
- * metadata.
+ * attributes.
  *
  * The @ref Specification of any given entity is used to determine the
- * applicable specific set of well-known OAIO metadata keys.
+ * applicable specific set of well-known OAIO attribute names.
  *
  *
  * @section meta_version Meta-version
@@ -377,7 +379,7 @@
  * host and one of these systems in order to query or persist data.
  *
  * Asset managers will generally re-map the well-known OAIO concepts
- * (@ref Specification, @ref metadata) into their own internal
+ * (@ref Specification, @ref attributes) into their own internal
  * structures, and back again when queried.
  *
  * Managers are made available via a @ref ManagerPlugin, that ultimately
@@ -578,7 +580,7 @@
  *
  * The best way to think of a Specification in general use is simply as
  * a glorified 'type string'. The fact that is also can contain
- * key-value metadata is in some ways irrelevant to understanding its
+ * key-value pairs is in some ways irrelevant to understanding its
  * main purpose. A Manager guarantees to be able to 'filter' Entities
  * based on a Specification, and a Host always Registers a new Entity by
  * identifying it with a Specification.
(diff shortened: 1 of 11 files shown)


feltech:
[consider] The commit message is missing a trailing ```````

doc/src/Glossary.dox
```diff
@@ -314,42 +314,44 @@
...
- * The @ref manager must guarantee to store and recall any metadata set
- * by the @ref host. They may freely re-map any keys to their local
- * names internally, as long as the mapping is reversed for queries.
+ * The @ref manager must guarantee to store and recall any attributes set
+ * by the @ref host. They may freely re-map any attribute names to their
+ * local names internally, as long as the mapping is reversed for
+ * queries.
  *
- * Metadata is limited to **string** keys, and **plain old data** typed values.
+ * Attributes names are strings, and their values are simplme **plain
```
feltech, line 328: [fix]
```suggestion
 * Attributes names are strings, and their values are simple **plain
```

python/openassetio/hostAPI/Manager.py
```diff
@@ -259,7 +259,7 @@ def prefetch(self, references, context):
         a chance to retrieve information in advance.
 
         The prefetch calls instructs the manager to retrieve any
-        information needed to either resolve, or fetch metadata for the
+        information needed to either resolve, or populate attributes for the
         supplied list of entities.
```
feltech, line 263: [consider] 72 char
```suggestion
        information needed to either resolve, or populate attributes for
        the supplied list of entities.
```

python/openassetio/hostAPI/Manager.py
```diff
@@ -557,105 +557,105 @@ def entityDisplayName(self, entityRefs, context):
...
-        metadata, and no meaningful @ref primary_string.
+        Retrieve @ref attributes for each given entity. This may contain
+        well-known attributes that you can then use to further customize
+        your handling of the entity. Some types of entity may only have
+        attributes, and no meaningful @ref primary_string.
 
-        There are some well-known keys defined in the core API - @see
-        openassetio.constants. These have universally agreed meaning.
+        There are some well-known attribute names defined in the core
+        API - @see openassetio.constants. These have universally agreed
```
feltech, line 568: [consider]/[plan] Not caused by this PR, but this doesn't render as it appears to be expected

image

python/openassetio/hostAPI/Manager.py
```diff
@@ -557,105 +557,105 @@ def entityDisplayName(self, entityRefs, context):
...
+        As a host, you can also advertize well-known attribute names of
+        your own as part of any first-class asset based workflows you
+        may have. For example, a compositor may choose to consume the
+        `colorspace` attribute (if present) and adjust the input space
+        of an image reader node accordingly.
 
-        These new keys should be clearly explained in your
+        These new attribute names should be clearly explained in your
         documentation. A manager may then be able to provide this
         information, based on introspection of your identifier (@see
```
feltech, line 579: [consider]/[plan]  Another case unrelated to this PR where the rendered output doesn't match what the text clearly expects
image

python/openassetio/hostAPI/Manager.py
```diff
@@ -557,105 +557,105 @@ def entityDisplayName(self, entityRefs, context):
...
-        the return of @ref getEntityMetadata for those keys will be the same.
-        Managers may remap keys internally to their own native names,
-        but a set/get should be transparent.
+        A Manager guarantees that it will round-trip attributes, such that
+        the return of @ref getEntityAttributes for attributes with those names
+        will be the same as set. Managers may remap attributes internally to
+        their own native names, but a set/get should be transparent.
 
-        If any value is 'None' it should be assumed that that key should
+        If any value is 'None' it should be assumed that that attribute should
```
feltech, line 606: [consider] 72 char

python/openassetio/hostAPI/Manager.py
```diff
@@ -557,105 +557,105 @@ def entityDisplayName(self, entityRefs, context):
...
 
-        @param data List[dict] The metadata to set for each referenced
+        @param data List[dict] The attributes to set for each referenced
         entity.
 
         @param context Context The calling context.
 
-        @param merge bool If true, then each entity's existing metadata
+        @param merge bool If true, then each entity's existing attributes
         will be merged with the new data (the new data taking
```
feltech, line 617: [consider] 72 char
```suggestion
        @param merge bool If true, then each entity's existing
        attributes will be merged with the new data (the new data taking
```

python/openassetio/hostAPI/Manager.py
feltech: Actually, can skip this given the method is scheduled to be removed.

python/openassetio/hostAPI/Manager.py
feltech: Actually, can skip this given the method is scheduled to be removed.

python/openassetio/hostAPI/Manager.py
```diff
@@ -557,105 +557,105 @@ def entityDisplayName(self, entityRefs, context):
...
         @param defaultValue `primitive` If not `None`, this value will
-        be returned in the case of the specified key not being set for
+        be returned in the case of the specified name not being set for
         an entity.
 
         @param context Context The calling context.
 
         @return `Union[primitive, KeyError]` The value for the specific
-        key, or `KeyError` if not found and no defaultValue is supplied.
+        attribute, or `KeyError` if not found and no defaultValue is supplied.
```
feltech, line 649: [consider] 72 char
```suggestion
        attribute, or `KeyError` if not found and no defaultValue is
        supplied.
```

doc/src/Transactions.dox
```diff
@@ -35,7 +35,7 @@
  *    manager.startTransaction(context)
  *    try:
  *        # Create the shot first, so we can then publish the clips to it
- *        shot_ref = manager.register([shot.name], [shot.metadata],
+ *        shot_ref = manager.register([shot.name], [shot.attributes],
  *                                    [shot.spec], [entity_ref], context)[0]
```
feltech, line 39: [consider]/[plan] Not caused by this PR. This doesn't match the signature of `register`, `attributes` should be a keyword argument.

doc/src/Transactions.dox
```diff
@@ -82,7 +82,7 @@
  *    """
  *    with coordinator.scopedActionGroup(context):
  *        # Create the shot so we can publish
- *        shot_ref = manager.register([shot.name], [shot.metadata],
+ *        shot_ref = manager.register([shot.name], [shot.attributes],
  *                                    [shot.spec], [entity_ref], context)[0]
```
feltech, line 86: [consider]/[plan] Not caused by this PR. This doesn't match the signature of register, attributes should be a keyword argument.

feltech (changes requested):

python/openassetio/hostAPI/Manager.py
foundrytom: #128

python/openassetio/hostAPI/Manager.py
foundrytom: #128

doc/src/Transactions.dox
foundrytom: #129

doc/src/Transactions.dox
foundrytom: #129

feltech (approved):

feltech:
All good, squash and merge when ready

PR #120, opened 2021-12-03
WIP [Core] Fix cyclic import in python modules

foundrytom:
`managerAPI.Host` was importing `HostInterface`, which meant the `hostAPI` package was initialized, which hoists `Manager`, this in turn was importing `HostSession` which was importing `Host`...

...all in the aid of previously removed Python 3 type hints (#64), which meant they were unused imports anyway!

Definitely looking forward to this all being back in C++ and getting rid of all these hoists.

Has the added bonus of fixing 3 of the `pylint` errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 __pycache__
 *.egg-info
 .idea
+.vim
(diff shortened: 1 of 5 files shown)


foundrytom:
See #123 - somehow messed up the branches.

PR #123, opened 2021-12-03
[Core] Fix cyclic dependencies in module imports

foundrytom:
`managerAPI.Host` was importing `HostInterface`, which meant the `hostAPI` package was initialized, which hoists `Manager`, this in turn was importing `HostSession` which was importing `Host`...

...all in the aid of previously removed Python 3 type hints (#64), which meant they were unused imports anyway!

Definitely looking forward to this all being back in C++ and getting rid of all these hoists.

Adds test coverage to keep an eye on this.

### DIFF
--- a/tests/openassetio/test_imports.py
+++ b/tests/openassetio/test_imports.py
@@ -1,0 +1,128 @@
+#
+#   Copyright 2013-2021 [The Foundry Visionmongers Ltd]
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


feltech:
For info: one trick we used in my previous life was to put imports that are only used for type hints under an `if False:` condition. This is sufficient for PyCharm, though linters don't like it!

tests/openassetio/test_imports.py
```diff
@@ -0,0 +1,130 @@
...
+"""
+These test ensure that all modules are importable. This can help catch
+dependencies between packages due to the hoisting required to hide
+duplicate namespaces to match the future C++ implementation appearance.
+"""
+
+import sys
+import pytest
+
+# pylint: disable=no-self-use
```
feltech, line 26: [consider][minor] Multiple checks can be comma-separated after the `disable=` to avoid so many lines of code.

feltech (approved):

tests/openassetio/test_imports.py
foundrytom: Good idea, updated [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/123/commits/d37a29b404890fa441a056eda90a7de75fbe1602).

feltech (approved):

PR #125, opened 2021-12-03
[Core] Make `ManagerInterface.identifier` static

foundrytom:
It doesn't make much sense to allow the manager's identifier to vary. It has to be consistent with the one reported by the ManagerPlugin. Having this be an instance method makes having a single source of truth a little ugly to no tangible benefit.

Leaving the host side, as it makes more sense to be an instance method, as an application may have different modes served from the same binary (eg: PLE).

Considered making this a read-only property, but keeping it as a method for symmetry with `HostInterface` (which has to be a method in c++).

### DIFF
--- a/python/openassetio/managerAPI/ManagerInterface.py
+++ b/python/openassetio/managerAPI/ManagerInterface.py
@@ -186,8 +186,8 @@ class ManagerInterface(object):
     #
     # @{
 
-    @abc.abstractmethod
-    def identifier(self):
+    @staticmethod
+    def identifier():
         """
         Returns an identifier to uniquely identify a specific asset
         manager. This may be used by a host to persist the users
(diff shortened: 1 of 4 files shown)


python/openassetio/managerAPI/ManagerInterface.py
```diff
@@ -186,8 +186,8 @@ class ManagerInterface(object):
     #
     # @{
 
-    @abc.abstractmethod
-    def identifier(self):
```
feltech, line 190: [fix] `Examples.dox` needs updating

feltech (changes requested):

python/openassetio/managerAPI/ManagerInterface.py
foundrytom: Good catch, thanks! Fixed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/125/commits/d7f45bd639b5653fce4468ce049bd8f321a37e5c).

feltech (approved):

PR #126, opened 2021-12-03
[Core] Remove `@auditApiCall` from `HostInterface`

foundrytom:
Auditing should only be added to the wrapper class, rather than the interface. This was inadvertently left in from the old API structure, where there was no wrapper on the host side.


feltech (approved):

feltech:

PR #140, opened 2021-12-09
[CI] Ensure Doxygen problemMatcher catches all warnings

foundrytom:
Inadvertently missed off preprocessing warnings that include a prefix before the file path.

https://github.com/foundrytom/OpenAssetIO/actions/runs/1559423183 is a run prior to the second commit, that shows the improved regex catches the existing failure.

### DIFF
--- a/.github/workflows/problem-matchers/doxygen.json
+++ b/.github/workflows/problem-matchers/doxygen.json
@@ -10,9 +10,9 @@
       "owner": "doxygen",
       "pattern": [
         {
-          "regexp": "^.*?\\/src\\/([^:]+):(\\d+): ?(\\w+): ?(.*)$",
+          "regexp": "^.*?\\/src\\/([^:]+):\\d+: ?\\w+: ?(.*)$",
           "file": 1,
-          "message": 4
+          "message": 2
         }
       ]
     }
--- a/.github/workflows/problem-matchers/doxygen.json
+++ b/.github/workflows/problem-matchers/doxygen.json
@@ -10,7 +10,7 @@
       "owner": "doxygen",
       "pattern": [
         {
-          "regexp": "^\\/src\\/([^:]+):(\\d+): ?(\\w+): ?(.*)$",
+          "regexp": "^.*?\\/src\\/([^:]+):(\\d+): ?(\\w+): ?(.*)$",
           "file": 1,
           "message": 4
         }
--- a/doc/src/Thumbnails.dox
+++ b/doc/src/Thumbnails.dox
@@ -30,8 +30,8 @@
  * If supported by the host, it will then:
  * - Call @ref preflight with the target entity and a @needsref ThumbnailSpecification
  *   that describes supported formats.
- * - Call @ref openassetio.managerAPI.ManagerInterface.ManagerInterface.getEntityMetadata
- *   "getEntityMetadata" on the returned reference to determine the desired width/height/format
+ * - Call @ref openassetio.managerAPI.ManagerInterface.ManagerInterface.getEntityAttributes
+ *   "getEntityAttributes" on the returned reference to determine the desired width/height/format
  *   for the thumbnail.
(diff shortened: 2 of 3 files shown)


.github/workflows/problem-matchers/doxygen.json
```diff
@@ -10,7 +10,7 @@
       "owner": "doxygen",
       "pattern": [
         {
-          "regexp": "^\\/src\\/([^:]+):(\\d+): ?(\\w+): ?(.*)$",
+          "regexp": "^.*?\\/src\\/([^:]+):(\\d+): ?(\\w+): ?(.*)$",
```
rickcoles-fn, line 13: If the bit before the `/src/` is optional does it need to be grouped, i.e. `"^(.+)?\\/src\\/ ..."`?

foundrytom: They're only grouped here so they can be referenced in the subsequent lines and fed back into the annotation. The two different line formats look like this:

```
Parsing file /src/python/openassetio/plug/src/doc/src/Thumbnails.dox:33: warning: unable to resolve reference to `openassetio.managerAPI.ManagerInterface.ManagerInterface.getEntityMetadata' for \ref command
/src/python/openassetio/constants.py:148: warning: unable to resolve reference to `DocumentLocale' for \ref command
```

We don't care about the `Parsing file...` bit.

rickcoles-fn: I deleted my original comment because I subsequently I thought I'd misread `.*?` as `.+?`, which is why I asked about grouping, since I believe the latter wouldn't have worked as expected, at least not reliably.

But then I realised you just changed the `+` to an `*`, presumably in response to my comment!

foundrytom:
@rickcoles-fn You highlighted a good point, I've added another commit [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/140/commits/934a1ad7afccf6bc2019098b9d2432f4ffb34998), that removes the unnecessary capture groups. I've broken the docs so you can see it in action [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/runs/4507971415?check_suite_focus=true#step:3:2155).

doc/src/Thumbnails.dox
```diff
@@ -30,7 +30,7 @@
  * If supported by the host, it will then:
  * - Call @ref preflight with the target entity and a @needsref ThumbnailSpecification
  *   that describes supported formats.
- * - Call @ref openassetio.managerAPI.ManagerInterface.ManagerInterface.getEntityMetadata
+ * - Call @ref openassetio.managerAPI.ManagerInterface.ManagerInterface.getEntityAttributes
  *   "getEntityMetadata" on the returned reference to determine the desired width/height/format
```
rickcoles-fn: I'm a bit confused by this. I see `ManagerInterface` has a `getEntityAttributes` and not a `getEntityMetadata` method, what's going on with the `"getEntityMetadata"` appearing on the next line?

Btw, searching the codebase shows up `getEntityMetadata` in a few other places - should they all be fixed up?

foundrytom:
@rickcoles-fn Not sure why GitHub won't let me reply in-line, but good spot, thanks. I've fixed that [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/140/commits/2ad4d6a52817b3ea9d6f7b68f5c48b30902dd64b). Essentially, the name in quotes on the next line is the display version of that link, so needed to be updated too.

I also added #146 to address the other incorrect usage of `getEntityMetadata`. These were all victim of a mid-air-collision when we merged #117 - as it was up in review for quite a while.

rickcoles-fn:
Squash ready for merge. Happy to review the result or for you to merge immediately.

foundrytom:
> Squash ready for merge. Happy to review the result or for you to merge immediately.

Thanks @rickcoles-fn, squashed and merged 👍

Issue #96, opened 2021-11-24
Example Manager plugin(s)


foundrytom:
## What

As a developer working with OpenAssetIO I need a sample implementation(s) of the `ManagerInterface` that can be used via the `PluginSystemManagerFactory`. They need to demonstrate canonical behaviour expected from a manager.

## Why

- As a host developer:
  - I need a working asset management system to perform testing of my implementation of OpenAssetIO.
  - I need a working asset management system for UX and feature design exploration.
- As an asset management system developer, I need a reference implementation to help me better understand expected behaviour or best-practice implementation patterns.
- As a developer on the OpenAssetIO team I need to be able to write working example host code.
 
## Acceptance Criteria

- [x] #131
- [ ] OpenAssetIO example code updated to work using this plugin.
- [ ] The manager(s) implement canonical behaviour based on the OpenAssetIO API contracts for:
  - [ ] A 'librarian', that isn't capable of resolving/querying future references
  - [ ] A 'path manager' that can resolve and answer attribute queries for future entities.
- [ ] The manager(s) can be configured to initialise with well-known states, containing multiple entities, relationships etc.
- [ ] The implementation is suitable for use as a code reference for other developers.

foundrytom:
This need to be done in conjunction with #95.

elliotcmorris:
Closing due to being out of date. We have BAL now, are looking at Ayon, etc. Other tickets capture this well enough.

PR #143, opened 2021-12-10
[Core] Remove `kField_EntityReferencesMatchRegex`

foundrytom:
Technically, saying "some calls won't need it" is a bit of a lie. In theory, if this was provided, we could avoid ever calling `isEntityRefrence`. Shall we gloss over that for now, to make sure that people do still implement that method. If feels ripe for fragility otherwise?

### DIFF
--- a/python/openassetio/hostAPI/Manager.py
+++ b/python/openassetio/hostAPI/Manager.py
@@ -146,7 +146,6 @@ class Manager(Debuggable):
           @li openassetio.constants.kField_SmallIcon (upto 32x32)
           @li openassetio.constants.kField_Icon (any size)
           @li openassetio.constants.kField_EntityReferencesMatchPrefix
-          @li openassetio.constants.kField_EntityReferencesMatchRegex
 
         Keys will always be str, and Values will be int, bool, float or
         str.
@@ -410,8 +409,7 @@ class Manager(Debuggable):
         @see resolveEntityReference
 
         @todo Make use of
-        openassetio.constants.kField_EntityReferenceMatchPrefix or
-        openassetio.constants.kField_EntityReferenceMatchRegex if
+        openassetio.constants.kField_EntityReferencesMatchPrefix if
         supplied, especially when bridging between C/python.
         """
         # We need to add support here for using the supplied prefix match string,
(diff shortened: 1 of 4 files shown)


python/openassetio/hostAPI/Manager.py
```diff
@@ -410,8 +409,7 @@ def isEntityReference(self, tokens, context):
         @see resolveEntityReference
 
         @todo Make use of
-        openassetio.constants.kField_EntityReferenceMatchPrefix or
-        openassetio.constants.kField_EntityReferenceMatchRegex if
+        openassetio.constants.kField_EntityReferenceMatchPrefix if
```
rickcoles-fn, line 412: There's an existing typo here, should be Reference**s**.

foundrytom: Good spot. Fixed that [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/143/commits/301e748e609fbbff62f99067362466bf4cb75fc0).

rickcoles-fn: I noticed it's also wrong in Examples.dox.

rickcoles-fn: In fact. that needs updating anyway to talk about what's now the specific key, rather than 'keys'.

rickcoles-fn:
Out of interest, what sort of thing are you expecting would be used for `kField_EntityReferencesMatchPrefix`?

python/openassetio/hostAPI/Manager.py
foundrytom: That would explain way I missed it before when grepping for the key name with a `s`... Fixed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/143/commits/f967ea46458bfb0938aed856c9d4a8e06b8657f2).

foundrytom:
> Out of interest, what sort of thing are you expecting would be used for `kField_EntityReferencesMatchPrefix`?

We advise that most managers use standardised URI for their entity references, and so then often look something like, for #96:

```
sam://some/asset?params=value
```

So, `info()` could return:

```
{
    kField_EntityReferencesMatchPrefix: "sam://",
   ...
}
```

Which allows the `c++` implementation of `Manager.isEntityReference` to short-circuit and not bother going into python if required.

rickcoles-fn (approved):

Issue #137, opened 2021-12-09
Doxygen `problemMatcher` not catching `warnings` in Docs test output


foundrytom:
See https://github.com/TheFoundryVisionmongers/OpenAssetIO/runs/4456474519?check_suite_focus=true#step:3:2156 and #104.

This check should have failed!

PR #144, opened 2021-12-13
[Docs] Fix and conform `@see` references

### LINKED ISSUE #128: Incorrect use of `@see` in Doxygen docs
## What

Replace inline uses of `@see` with `See @ref`.

## Why

Then used inline, the `@see` formatting forces a block-display heading and nested text (that also doesn't link without an explicit `@ref`)

image

_Originally posted by @feltech in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/117#discussion_r761892705_

foundrytom:
Closes #128. Limited in scope to only `@see` refs, not all, fixes:

- Incorrect use of `@see` for in-line references in copy.
- `@see` without `@ref` - which don't error if the ref is not found.
- `@needsref` for UI references
- Short display names for `@see` class references.


foundrytom:
Rebased to `main` now #137 is merged.

PR #155, opened 2021-12-16
Don't use `OAIO`/`oaio`

### LINKED ISSUE #153: Do not use `OAIO` as an abbreviation for `OpenAssetIO`
## What

Remove all uses of the `OAIO` acronym from the repository, documentation and discontinue colloquial use when talking about `OpenAssetIO`.

- Use the title in full in in written copy.
- Use "Open Asset IO" or "the asset API" in colloquial dialog.

## Why

An unfortunate side-effect of the emergent standard of `Open<Project>IO` for project naming, is a large number of collisions in the acronym space.

For example https://github.com/AheadIO/OpenAnnotationIO - `OAIO`, and OpenAssetIO - `OAIO` 🙃 . They got here first, and so should be the ones to use the abbreviation.

foundrytom:
Closes #153.

There is a prior CI run on my fork [here](https://github.com/foundrytom/OpenAssetIO/runs/4547119924?check_suite_focus=true) showing the test catching usage in hidden files, docs and code.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -43,4 +43,4 @@ jobs:
         git config --global user.name "GitHub Actions"
         git config --global user.email "github-actions@github.com"
         # Pass the path to the actions/checkout@v2 docs branch checkout
-        OAIO_DOCS_REPO_DIR=`pwd`/target make -C src/doc deploy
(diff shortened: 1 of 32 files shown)


foundrytom:
After discussion with @rickcoles-fn, squashed to avoid the codebase being conceptually broken between steps.

PR #156, opened 2021-12-16
[BUildPublish docs Docker

foundrytom:


foundrytom:
Humm, despite what [this page](https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#default-permissions-and-access-settings-for-containers-modified-through-workflows) says, still getting auth failure to pull the image.

foundrytom:
Created  #157 to sort out the passing check for a failed build.

foundrytom:
Using the published container, the docs test is now down to `40s` or so from several minutes, whoop.

doc/Makefile
```diff
@@ -19,25 +19,68 @@ default: container-build
...
-container-build: docker-image
-	docker run --rm -v `pwd`/../:/src openassetio-doc-build bash -c 'make -C /src/doc html'
+container-build:
+	docker run --rm -v `pwd`/../:/src $(CONTAINER_TAG) bash -c 'make -C /src/doc html'
+
+# Targets for building/tagging/publishing the container
 
 docker-image: Dockerfile
-	docker build . -t openassetio-doc-build
+	docker build . -t $(CONTAINER_NAME)
```
feltech, line 70: [consider] We could use `CONTAINER_TAG` throughout, then there is no need for a separate `tag-docker` step.

I hit this because when I tried to run `make` it failed because I didn't have the tagged image.  So I tried `make docker-image` (because I didn't read the updated docs :roll_eyes:) and of course it still didn't work because the image wasn't tagged.

I think the `CONTAINER_NAME` image is now superfluous, since it's never used except for tagging, which  "copies" the image (also see comment re. `clean-docker`).

doc/Makefile
```diff
@@ -19,25 +19,68 @@ default: container-build
...
 
 clean-docker:
-	docker image rm openassetio-doc-build --force
+	docker image rm $(CONTAINER_NAME) --force
```
feltech, line 73: [consider] This removes the `$CONTAINER_NAME:latest` image, but not the `$CONTAINER_TAG` image.

It looks like there is (still) no built-in way to delete all images for a given "repository" (`CONTAINER_PATH`). Simplest solution is to delete the current `CONTAINER_TAG` image and assume the tag hasn't been bumped since the image was last built. Otherwise shell grepping is required.

doc/Makefile
```diff
@@ -19,25 +19,68 @@ default: container-build
...
+
+tag-docker: docker-image
+	docker tag $(CONTAINER_NAME) $(CONTAINER_TAG)
+
+publish-docker: tag-docker
+	echo $(GITHUB_PAT) | docker login ghcr.io --username $(GITHUB_USERNAME) --password-stdin
+	docker push $(CONTAINER_TAG)
+	# Update the 'latest' tag
+	docker tag $(CONTAINER_TAG) $(CONTAINER_PATH):latest
```
feltech, line 82: [question] Do we need to care about the `latest` tag?

doc/Makefile
feltech: I checked and deleting the "parent" `CONTAINER_NAME` image then rebuilding is almost a no-op, since docker just re-tags the `CONTAINER_TAG` image as `CONTAINER_NAME` - the image ID is identical.

feltech (approved):

doc/Makefile
foundrytom: Yeah, otherwise you have to specify an explicit version, which means you can't do:
```
docker run -v `pwd`/../:/src ghcr.io/thefoundryvisionmongers/openassetio-doc-build bash -c 'make -C /src/doc html'
```

feltech: As in, if you want to do it manually? (since `make container-build` uses `CONTAINER_TAG`)

doc/Makefile
foundrytom: Thats a much better idea, thanks. [Done](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/156/commits/4cf8f40fa1c23d331ea63d10cb5e6dfed211988c).

doc/Makefile
foundrytom: This is better now with the above change to always using `CONTAINER_TAG`. You can use `clean-docker` to remove your local build of the current image version - which could be enough for now?

doc/Makefile
foundrytom: Yeah (we list the above in `README.md` and it saves the maintenance overhead.

feltech:
Neat, squash and merge when ready!

feltech (approved):

Issue #101, opened 2021-11-25
`pylint` checks


foundrytom:
- [x] Runs on Pull requests
  - Anything run by CI can be run locally with minimal changes.
  - Merges aren't blocked for pylint failures but results are shown as warnings.
- [x] Pylint rules updated accordingly for the current codebase
  - Decide whether to keep Python 2 to 3 rules e.g. use f-strings, don't inherit from `object`.
- [ ] #176
- [x] Issues raised for current known fails.

feltech:
Searching the [GitHub Marketplace for "pylint"](https://github.com/marketplace?type=actions&query=pylint+) reveals a few third-party solutions.  

Unfortunately the nicest-looking solutions don't have a mechanism to add a preamble (i.e. `pip install -r requirements.txt`), so would likely lead to lots of false positives, especially going forward if/when we make more use of third-party libraries (and potentially if we require a C++ bindings build).

The solutions that do have such a mechanism are essentially thin Docker containers that pre-install `pylint` and expect a shell command as input (so usage would be something like `pip install -r requirements.txt && pylint`).

Perhaps then it's worth just building up an action similar to #91

foundrytom:
- Tests: Check regex for functions/classes to only allow snake case if it begins with `test_` [experiment]
- Add `# pylint: disable=` declarations to existing unit tests

PR #165, opened 2021-12-20
[Tests] Remove code duplication

foundrytom:
Highlighted as part of #101, we were duplicating the module cache cleanup code.

@feltech - know any `pytest` tricks to auto-use an existing fixture without the dance here?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,37 @@
+#
+#   Copyright 2013-2021 The Foundry Visionmongers Ltd
+#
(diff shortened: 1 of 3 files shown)


feltech (approved):
I think this is about as clean as it gets. 

* Perhaps we could have a common utility function that can be called by each fixture, rather than a common fixture dependency, but that would entail `__init__.py`s and additional `import`s, at a minimum.
* Alternatively just explicitly add the fixture to all tests that need it.

PR #168, opened 2022-01-05
[Docs] Add notes on doxygen formatting for method params

foundrytom:
Capturing the discussion around https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/152#discussion_r778299871

Longer-term we should probably split the documentation advice out at some point as there are a bunch of other conventions we follow.


foundrytom:
Added #169.

feltech (approved):
Grr, PyCharm is usually good at removing trailing whitespace, but several times now I've seen you've had to clean them up in markdown I've written. I'll look into that.

feltech:
Rebase and merge when ready.

PR #170, opened 2022-01-05
[CI] Fix `GITHUB_ORG` used for docs deploy

foundrytom:
This was inadvertently missed in 33f1f99 in #156.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -33,9 +33,7 @@ jobs:
 
     - name: Build
       run: |
-        # TODO [tc] We should really publish the Docker
-        # image to the GitHub registry
-        make -C src/doc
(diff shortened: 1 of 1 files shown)


feltech (approved):
Good spot. Perhaps adding the variable globally in an `env` section would avoid the need for this: https://docs.github.com/en/actions/learn-github-actions/environment-variables - but since it's just a temporary fix this is fine.

foundrytom:
> Good spot. Perhaps adding the variable globally in an `env` section would avoid the need for this: https://docs.github.com/en/actions/learn-github-actions/environment-variables - but since it's just a temporary fix this is fine.

I just realised too, technically, I actually just put it in the wrong place before, as it's not needed for the `deploy` step. Hopefully we can just remove this entirely soon anyway.

PR #178, opened 2022-01-06
Merge main into feature/96-exampleManager

feltech:
- Fixed some minor conflicts.
- Had to fix a few issues raised by CI (see final few commits).
- Note: the DCO check action fails because the number of commits overwhelms it.


feltech:
Good 'ole CI finding genuine issues! Will fix.

python/openassetio/test/managerValidator/commandLine.py
```diff
@@ -177,7 +177,7 @@ def load(filePath):
 
         @return `dict` Dictionary of test case fixtures.
         """
-        spec = importlib.util.spec_from_file_location("oaioTestFixtures", filePath)
+        spec = importlib.util.spec_from_file_location("TestFixtures", filePath)
```
foundrytom, line 180: [question] If this is a module name, should it be `testFixtures`?

foundrytom:
> Note: the DCO check action fails because the number of commits overwhelms it.

We should stop using that modular approach, and just have a quick checker that doesn't funnel data via the command line...

foundrytom (approved):

PR #180, opened 2022-01-07
[Docs] Add missing `@namespace` to module docstrings

feltech:
A few module docstrings were missing their respective `@namespace` prefix that causes the text to actually get rendered in the Doxygen output for the module - see blank spaces in https://thefoundryvisionmongers.github.io/OpenAssetIO/annotated.html

### DIFF
--- a/python/openassetio/pluginSystem/ManagerPlugin.py
+++ b/python/openassetio/pluginSystem/ManagerPlugin.py
@@ -14,6 +14,7 @@
 #   limitations under the License.
 #
 """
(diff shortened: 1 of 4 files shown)


foundrytom (changes requested):
Thanks @feltech, we're just missing one in `PluginSystemPlugin` still.

In the doc build, `html/annotated.html` is a good test page as it shows the full namespace hierarchy.

foundrytom (approved):
Thanks @feltech, please feel free to squash and merge at your leisure. 🏖️

PR #186, opened 2022-01-18
[Docs] Add extended documentation for the manager test harness

### LINKED ISSUE #177: [Tests] `managerValidator` public API refactor
## What

As part of #139, we introduced the `managerValidator` module/CLI tool that allows a given `ManagerPlugin` implementation to be functionality tested.  As as developer of plugins, I'd like to be able to extend the test suite to run additional tests of my own design, re-using as much of the test harness infrastructure as possible.

## Why

The supplied tests only cover generic OpenAssetIIO API Contract behaviour, and cannot test the business logic of my plugin (eg: asserting that my `managementPolicy` is correct for my specific asset manager). I am already using the `managerValidator` as part of my development cycle, and so being able to easily extend the test cases avoids the need to build a parallel system that covers much of the same functionality.

## Acceptance Criteria

- The public surface area should be the minimum required to allow:
  - A developer to write supplementary modules similar to `validatorSuite.py` and invoke the test harness to run those in addition/instead of the OpenAssetIO suite with minimal coding. Ie: rewrite `__main__.py` in as few a words as possible with the ability to:
    - Provide the fixture values in the same file as the standard test suite.
    - Provide the fixture values in an additional file as per the above
    - Provide fixtures through some other means of their design as a `dict`.
    - Invoke the harness programatically using fixtures derived by the means noted above and a test suite module. 
    - Test failures are raised as exceptions by default, but can be optionally returned as a bool
- The CLI remains unchanged and only runs tests from `validatorSuite.py`, when provided with a path to a fixtures `.py` file.

## Design sketch / pseudo-code

> Note: Names are illustrative

 - `fixturesFromPythonFile(path: str)`
 - `executeSuite(suiteModule: module, fixtures: dict, unittestArgs: List[str], throwOnFail=True)`

```
from openassetio.test.managerValidator import executeSuite, fixturesFromFile, validatorSuite
from my_code.tests import managerSuite

fixtures = fixturesFromFile(pathToFixtures)

executeSuite(validatorSuite, fixtures)
executeSuite(managerSuite, fixtures)
```

foundrytom:
Closes #177.

I was hoping that we could make use of `@copydoc` or similar to assemble the `Testing.dox` page from closer-to-source documentation, however this seemed to refuse to work, claiming the targets didn't exist, even though they work in `@ref`.


python/openassetio/test/manager/_implementation.py
```diff
@@ -14,7 +14,9 @@
 #   limitations under the License.
 #
 """
-Boilerplate classes for initialization and execution of test cases.
+@namespace openassetio.test.manager._implementation
+Private implementation classes for the manager test framework.
+@private
```
feltech, line 19: [question] The `@private` tag on the class/function below means the docstring is not rendered by Doxygen. The `@private` on this `@namespace` seems to have no effect, i.e. `_implementation` is listed in the docs for the `manager` package, but appears nowhere else. Is that all intentional?

feltech (approved):

python/openassetio/test/manager/_implementation.py
foundrytom: I was trying to hide it as much as possible. I wondered If it was worth leaving there to make sure it was obvious to people reading the src too?

doc/src/Testing.dox
```diff
@@ -0,0 +1,108 @@
...
+ *         }
+ *     },
+ *     "Test_displayName": { ... },
+ *     ...
+ * }
+ * @endcode
+ *
+ * @note A comprehensive example fixtures file is provided as part of
+ * the Sample Asset Manager included in the distribution (see
+ * `fixtures.py`)
```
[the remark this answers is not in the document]
foundrytom: It's not in `main` yet, so we can't link to a stable path yet sadly.

python/openassetio/test/manager/_implementation.py
feltech: Yeah that's fine, just wanted to ensure it's intentional.

feltech (approved):

Issue #181, opened 2022-01-10
Core C++


foundrytom:
## What

The initial work that starts the migration of the core OpenAssetIO model to `C++`.

## Why

OpenAssetIO functionality must be available to both `C++` and `Python` hosts, and so the core model and business logic needs to be ported to `C++`, and bound to `Python`.

## Acceptance Criteria

- [x] #183
  - Use established cross-platform conventions for library boilerplate, `linux` builds for now.
  - Minimal `libopenassetio.[a|so]` containing only `ManagerInterface`, with no methods.
  - `include` and `src` top level project directories (public headers in `include`)
  - Optional (on by default) bindings to `Python 3.7` under `_openassetio` using `pybind` via `_openassetio.so`
  - Base class for `openassetio.managerAPI.ManagerInteface` (python) changed to `C++` class.
  - Built using generic `CMake` following industry standard practices for build and dependency management.
- [x] #182
  - Run existing tests using lib
  - lint/static analysis on `C++` code
- [x] #243

foundrytom:
Closing as we have the base requirements in `feature/181-coreCppBuild` now.

PR #189, opened 2022-01-20
[Docs] Add decision record for core language strategy

foundrytom:
Part of #181, documents the high-level strategy for multi-language support. Specifics around the conventions we will use on the ground will come at a later date, once this choice is ratified.

Thanks to @feltech for the comprehensive code sketch.

## Notes for reviewers

The most digestible view of the document is [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/files?short_path=13a4af3#diff-13a4af328cd7c2e775857a05460352d18efd46f463288a24f7a158495bd2d980).

This PR serves to validate the decision itself, not just the correctness of its writeup (given it's me, I'm sure there will be some copy gremlins 🙃). Any parties with comment on the choice itself should feel free to use this to discuss the topic.

### DIFF
--- a/decisions/DR004-Core-language-support.md
+++ b/decisions/DR004-Core-language-support.md
@@ -1,0 +1,485 @@
+# DR004 Core language support
+
+-   **Status:** Decided
+-   **Impact:** High
+-   **Driver:** @foundrytom
+-   **Approver:** @foundrytom @feltech
+-   **Contributors:** @foundrytom @feltech @meshula
+-   **Outcome:** OpenAssetIO will use `C++` as the core language for
+    efficient implementation, with critical business logic exposed
+    through portable `C` API.
+
+## Background
+
+OpenAssetIO strives to provide a universal standard for communications
+between a process that consumes or produces data, and any external
+systems that manage that data.
+
+To be successful in any meaningful way, it needs to be portable to
+which ever programming language(s) are meaningful to its target
+audience.
+
+At the time of writing, the majority of expected integrations are within
+`C++` and/or `Python` environments. Other languages such as `Rust` or
+`go` are gaining traction. Broader reaching applications of OpenAssetIO
+introduce `JavaScript`, `Swift` and others.
+
+The library must be designed to simplify integration into current
+_and_ future programming environments, without the need to re-implement
+critical business logic.
+
+## Relevant data
+
+### Required language support
+
+`C++`, `Python`
+
+### Projected language support
+
+`C`, `JavaScript`, `Java`, `Swift`, `Rust`, `go`
+
+### Bindings
+
+Research has shown that the common denominator when addressing the above
+listed languages is a `C` interface. A standard `C` API surface, with
+clear and manageable ownership semantics would facilitate support in the
+broadest range of future programming environments.
+
+Ideally, any bindings would also allow idiomatic programming within
+the target language, rather than forcing specific design patterns from
+any internals of OpenAssetIO - such as synchronous programming, RAII,
+etc.
+
+### Performance
+
+A core requirement raised by numerous parties is that the "happy path"
+should be as performant as possible. Thus, it is required that any
+mechanism introduced to broaden compatibility doesn't compromise the
+(current) majority deployment scenarios.
+
+Of note, a `C++` host talking to a `C++` manager has been the most
+frequently requested high-performance pathway.
+
+### Development lifecycle
+
+The overhead of maintaining dynamically linked `C++` libraries within a
+plugin-based architecture is often fraught with constantly chasing ABI
+changes and/or conflicts between mutual dependencies.
+
+Foundry has prior art in using `C` function pointer suites as a compiler
+isolation technique on top of `C++` implementations that can help
+de-couple the plugin host and the specifics of any given plugin.
+
+In the constantly evolving landscape defining which dependency versions
+are in use within which versions of common DCC tools, avoiding the need
+to matrix-compile and juggle plugin search paths per-host would be
+massively beneficial.
+
+## Options considered
+
+### Option 1
+
+`C++` and `Python` support only.
+
+**Estimated Cost**: Small
+
+#### Pros
+
+-   Simplifies development.
+
+#### Cons
+
+-   Leaves use within languages other than these a potentially
+    challenging exercise left to the reader.
+-   Compiler isolation support would need to be handled as an add-on
+    through an additional layer such as a `C` suite, `protobuf` or
+    similar.
+
+### Option 2
+
+`C` core library, bound to `C++` and `Python` as appropriate.
+
+**Estimated Cost**: Large
+
+#### Pros
+
+-   Makes future bindings or adoption in `C`-based languages far
+    simpler as all core API functionality is inherently available
+    through idiomatic `C`.
+-   Inherently presents a stable ABI isolation layer between hosts and
+    plugins.
+
+#### Cons
+
+-   The data model of OpenAssetIO _requires_ `map` and `optional` types.
+    The lack of standardization in `C` around these concepts means a
+    local implementation of these would be required. This, and their
+    mapping to C++ and Python introduces run-time and maintenance
+    overhead, as well as increased code complexity and project risk.
+-   `C` is notoriously unsafe with regards to type-safety and memory
+    management. Higher-level systems programming languages, such as
+    `C++` and `Rust`, add well-known idioms and/or protective mechanisms
+    (both compile-time and run-time, with little or no performance
+    overhead).  These can help reduce run-time or programming errors of
+    this nature.
+-   `C` programming idioms around required topics such as inheritance
+    and generic programming use macros. The use of macros makes for
+    awkward, or impossible interoperation with languages such as `C++`
+    where the use of such constructs interfere with clarity and best
+    practices.
+-   Long-term maintenance of an extensive `C` codebase, using advanced
+    techniques such as the above, is potentially more difficult due the
+    deficit of low-level `C` experience in the broader project
+    community.
+-   Introduces additional boilerplate and overhead in the `C++` to `C++`
+    high-performance path.
+
+### Option 3
+
+`C++` core library, with critical business logic exposed via `C`
+interfaces. Bound to `Python`.
+
+**Estimated Cost**: Medium
+
+#### Pros
+
+-   The `C` API is still a first-class citizen within the architecture,
+    aiding future adoption in other languages and the development of
+    optional ABI isolation layers.
+-   The core implementation can leverage the `C++17` standard library,
+    and language features, dramatically reducing code complexity.
+-   The `C++` to `C++` high-performance path is natively supported.
+
+#### Cons
+
+-   Easy to inadvertently 'leak' `C++` specific practices through the
+    `C` API.
+-   The standard library can add 'bloat' when certain conveniences are
+    used.
+
+## Outcome
+
+OpenAssetIO will adopt **Option 3** and implement the API core in `C++`.
+Where this contains language-universal business logic (as opposed to
+language-specific boilerplate), it will be exposed through an idiomatic
+`C` interface that can be used to bind to arbitrary languages at some
+future point.
+
+The appendix contains a sketch illustration of how the implementation of
+a typical API component may look.
+
+## Rationale
+
+This arrangement provides the best possible intersection of the
+requirements outlined in the Relevant data section.
+
+-   The `C++` to `C++` fast path can be implemented directly with
+    minimal middleware, at the expense of ABI coupling/etc.
+-   Compiler/memory isolation via a `C` function pointer suite, or other
+    technique such as `protobuf` can be implemented as required via a
+    shim between a host and a plugins implementation on top of the
+    underlying `C++` core.
+-   Future language bindings should be more readily implementable as the
+    `C` interface is part of the core design, rather than an
+    afterthought.
+-   The core code base can make maximum use of standard library
+    containers and memory management tools, reducing entire categories
+    of programming errors.
+
+
+## Appendix - an implementation sketch
+
+The code segment below illustrates a working sketch of a potential
+approach to a typical core component implementation.
+
+> Note: The exact semantics of string handling, and the extended details
+> of the opaque handle + access API for data types exposed through `C`
+> will be finalized and detailed in future documentation.
+
+```cpp
+#include <cstring>
+#include <memory>
+#include <string>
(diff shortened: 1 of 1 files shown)


decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,460 @@
...
+
+`C++` and `Python` support only.
+
+**Estimated Cost**: Small
+
+#### Pros
+
+-   Simplifies development.
+
+#### Cons
```
feltech, line 90: [consider] No C++ ABI isolation, so subject to compiler and library ABI breaks between the host, OpenAssetIO and manager plugins.

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,460 @@
...
+
+### Option 2
+
+`C` core library, bound to `C++` and `Python` as appropriate.
+
+**Estimated Cost**: Large
+
+#### Pros
+
+-   Makes future bindings or adoption in `C`-based languages far
```
feltech, line 103: [consider] Presents a stable ABI isolation layer between hosts and plugins, so that the OpenAssetIO core, and manager plugins, are less likely to require recompilation when compilers and libraries are updated in host applications.

feltech (changes requested):

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,460 @@
...
+**Estimated Cost**: Medium
+
+#### Pros
+
+-   The `C` API is still a first-class citizen within the architecture,
+    aiding future adoption in other languages.
+-   The core implementation can leverage the `C++17` standard library,
+    dramatically reducing code complexity.
+
+#### Cons
```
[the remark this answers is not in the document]
foundrytom: Is that a con of the choice here, or just a con of that deployment scenario? It feels like more the former? I think we need to treat that scenario as a requirement, rather than an option, so it should probably show up as a con for options 2 instead - i.e. the addition of boilerplate to the happy path.

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,460 @@
...
+
+**Estimated Cost**: Large
+
+#### Pros
+
+-   Makes future bindings or adoption in `C`-based languages far
+    simpler as all core API functionality is inherently available
+    through idiomatic `C`.
+
+#### Cons
```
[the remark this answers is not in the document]
foundrytom: Good call, added [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/a3083492b088223f161091c09b156980224215bb).

decisions/DR004-Core-language-support.md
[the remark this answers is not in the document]
foundrytom: Added a note that Option 3 natively supports the high-perf path [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/5408c60dc1ec565784c68318eff67bd22701d105), and added a note on the potential overhead in Opt 2 [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/5e678a5e6925b1a07500d21264baf2954518f70d).

decisions/DR004-Core-language-support.md
foundrytom: Added [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/ff11aeee4d2e53abf1b3ba5a2e336b354952bbf6).

decisions/DR004-Core-language-support.md
foundrytom: Added [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/73524625a3debd24b95c4827279b2b2c25395361).

foundrytom:
@feltech thanks for extra points, I've added those, and rejigged a few of the existing ones to better reflect the impact of the other options on the two noted topics of ABI/perf. See what you think.

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,476 @@
...
+-   Inherently presents a stable ABI isolation layer between hosts and
+    plugins.
+
+#### Cons
+
+-   The data model of OpenAssetIO _requires_ `map` and `optional` types.
+    Implementing these in pure `C` either requires significant coding,
+    or would add additional third-party dependencies. Either of these
+    add significantly to code complexity, maintenance overhead and
+    project risk.
```
meshula, line 118: I disagree on the "significant coding" or "third-party depencies" point, and the subsequent risk. hash tables are well known solutions for maps, and it's not hard to design an `optional` surrogate. I'd word this point instead as 

The data model of OpenAssetIO _requires_ `map` and `optional` types. The lack of standardization in C around these concepts means a local implementation of these for C is required, and a mapping from this implementation to C++ and Python concepts introduces runtime and maintenance overhead.

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,476 @@
...
+-   C is notoriously unsafe with regards to type-safety and memory
+    management. Higher-level systems programming languages, such as C++
+    and Rust, add protective mechanisms (both compile-time and run-time)
+    with little or no performance overhead.
```
meshula, line 122: C++ does not introduce safety to memory management. It only introduces some idioms, that can be used, to mitigate issues, but foot-guns abound.

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,476 @@
...
+-   C is less expressive than higher-level languages, missing features
+    such as inheritance and generic programming, which would likely
+    increase development time.
```
meshula, line 125: An earlier point in this document is that Foundry has prior art in C function suites, this point seems contradictory. The advertised lack of generic programmings stems from a rejection by the C++ community of macros; perhaps a better wording would be:

C programming idioms around inheritance and generic programming use macros. The use of macros makes for awkward, or impossible interoperation with languages such as C++ where the use of such constructs interfere with clarity and best practices.

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,476 @@
...
+## Rationale
+
+This arrangement provides the best possible intersection of the
+requirements outlined in the Relevant data section.
+
+-   The `C++` to `C++` fast path can be implemented directly with
+    minimal middleware, at the expense of ABI coupling/etc.
+-   Compiler isolation via a `C` function pointer suite, or other
+    technique such as `protobuf` can be implemented as a shim between a
+    host and a plugins implementation.
```
meshula, line 172: Some words about memory management are perhaps necessary. It's not sufficient to simply isolate the compiler. Across the C boundary, there should probably be some consideration that any objects obtained from the interface cannot be externally memory managed. In other words, the objects should be explicitly incompatible with malloc, free, realloc, and the string functions that reallocate.

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,476 @@
...
+-   Future language bindings should be more readily implementable as the
+    `C` interface is part of the core design, rather than an
+    afterthought.
+-   The core code base can make maximum use of standard library
+    containers and memory management tools, reducing entire categories
+    of programming errors.
+
```
meshula, line 179: my previous point being support of this point, of course, but trying to lay it out explicitly

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,476 @@
...
+////////////////////////////////////////////////////////////////////////
+// Default terminology
+
+#undef ns
+#define ns(thing) OpenAssetIO_C_NAMESPACE(terminology_##thing)
+
+extern "C"
+{
+    [[maybe_unused]] static const char * ns(kTerm_Asset) = "asset";
```
meshula, line 326: years of fighting MSVC and string deduplication, and how to decorate extern statics for __declspec(dllexport) make me strongly prefer strings not being inlined like this.

I'd go so far as to say that both char* and std::string are more or less anti-patterns... Moving forward, I'm using the pattern of library specific string classes, with c_str and std::string adaptors.

the reason is that we generally want to impose constraints on allowable strings, such as notions of being well formed. all that goes out the window with the standard types, and all the many ways one can introduce a malformed string into the system become an ongoing maintenance concern.

A typical example is a string to represent a URL. If it's a string, it has to be validated and sanitized at every point of use. If it's a specialized value, it's sanitized/validated on construction, and can subsequently make a promise of being well formed.

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,476 @@
...
+
+namespace openassetio
+{
+inline namespace OpenAssetIO_VERSION
+{
+
+namespace terminology
+{
+
+std::string replaceTerms(const StringMap & terminology, const std::string & sourceStr)
```
meshula, line 368: so like here. I'm sure there's a notion of a well formed terminology string, and a malformed one.

decisions/DR004-Core-language-support.md
luke-titley: A random thought... not sure if this is useful for you guys, but the rust wg wrote a tool (cppmm) which generates a c api for a c++ library. It's what we're using to make rust bindings.

decisions/DR004-Core-language-support.md
luke-titley: +1 to your own string type. It doesn't actually have to support may operations if it can convert to/from std::string and char* and it gives you compiler independence for movability, alignment and size.

decisions/DR004-Core-language-support.md
luke-titley: An example of usd stage in c. It's fairly easy to use and the whole project is in c++.

https://github.com/vfx-rs/usd-bind/blob/main/usd-c/include.in/pxr/usd/usd/usd_stage.h

decisions/DR004-Core-language-support.md
feltech: > An earlier point in this document is that Foundry has prior art in C function suites, this point seems contradictory
Foundry's C function pointer suites (structs) invariably call through to C++ logic. We could clarify that point if its causing confusion.

decisions/DR004-Core-language-support.md
foundrytom: Thanks @luke-titley looks really promising, we'll def factor this in when we start on the `C` exposed code for-real.

decisions/DR004-Core-language-support.md
foundrytom: Reworded [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/8c053b05bd8dda23fb133640698a7900b2401514), also keeps the risk part, which is certainly a consideration from our p.o.v.

decisions/DR004-Core-language-support.md
foundrytom: Re-worded this [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/153040cc29363227f3a6c9fcb81731854dec7f22), to indicate that it is not always infallible checks.

decisions/DR004-Core-language-support.md
foundrytom: Added a little clarification to the section on prior-art [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/175d30a3194aad501b825bb0fd02750d02588de4), and broadened to include memory isolation [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/5c28736890dffce8644ef24127afcec60421130f).

decisions/DR004-Core-language-support.md
foundrytom: Broadened the text to include memory [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/5c28736890dffce8644ef24127afcec60421130f). We will definitely factory in the `malloc` interaction patterns when we finalise the actual approach to strings/etc. in future work.

decisions/DR004-Core-language-support.md
foundrytom: Updated the "cons" of a complex `C` implementation [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/6111631839f7432e8fade7434e28e778be8203ed).

foundrytom:
Many thanks for all the input @meshula, @luke-titley. I think we've addressed relevant comments as noted. 

We have a general policy that a PR is blocked until review comments are resolved as satisfied by their author. I believe GitHub's permission mechanism means that you probably can't resolve your own conversation. If so, please add a 👍 as a reaction if you're happy to consider the thread resolved. We can close them for you and continue with the merge.

 @feltech, I've also conformed some formatting gremlins [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/8efd7e465f2b55f0cc30afed7e814fc7d0a8470b) and [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/4ffe4419845d1796fbf6fc365c9ea3df4d5e09a9).

decisions/DR004-Core-language-support.md
```diff
@@ -0,0 +1,485 @@
...
+
+-   The data model of OpenAssetIO _requires_ `map` and `optional` types.
+    The lack of standardization in `C` around these concepts means a
+    local implementation of these would be required. This, and their
+    mapping to C++ and Python introduces run-time and maintenance
+    overhead, as well as increased code complexity and project risk.
+-   `C` is notoriously unsafe with regards to type-safety and memory
+    management. Higher-level systems programming languages, such as
+    `C++`
+    and Rust, add well-known idioms and/or protective mechanisms (both
```
feltech, line 122: [fix] Elsewhere programming languages are surrounded in backticks, but not "Rust" here.

feltech (approved):

decisions/DR004-Core-language-support.md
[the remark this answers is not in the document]
foundrytom: Lol, no, good spot, thanks. Removed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/d337559264006c1eb061176d56b82ee77a950617).

foundrytom: Should be consistent [now](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/189/commits/b829e705dbea283d8ffd98f06294a71ffc031bd5).

feltech (approved):

Issue #183, opened 2022-01-10
Core C++: project infrastructure


foundrytom:
See description on #181.

## Notes

- PR for C++ build/lib
- PR for Python bIndings

Issue #182, opened 2022-01-10
Core C++: CI


foundrytom:
## What

Reinstate CI for existing checks, and ensure best-practice tests are added for the `C`/`C++` codebase. See #181.

## Why

- #183 required us to disable the existing `pytest` checks as they required the built `cpython` `_openassetio` module.
- We need equivalent checks to our `pytest`/`pylint` validation of the Python code base for the `C`/`C++` code base.

## ACs

- Restrict to `linux` for now. Size macOS/Windows as possible
- Opinionated CI check using `clang-format` using the `Google` preset, with line length adapted to our base length.
- Linting should use the minimal number of third-party tools where possible, flag and new additions. (eg: `clang-tidy`)
- Use public (`Conan`?) releases for CI deps (+ CI cache) or other `CMake` native solutions

## Out of scope

- Windows or macOS CMake builds
- Wheels and other such distributable artifacts
- `setup.py` calling `CMake` from `pip install`, can just use `ctest`

foundrytom:
We need to work towards https://bestpractices.coreinfrastructure.org/en

TheRealJokerMan:
Where applicable, have you considered just utilising CMake's [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) to pull lightweight CMake-based projects?

foundrytom:
@feltech Updated the ACs to capture our earlier selection of the `Google` preset for `clang-format`

foundry-markf:
Note this CMake open issue https://gitlab.kitware.com/cmake/cmake/-/issues/8438 

> Currently, add_dependencies can not be used to create a dependency if one of the specified targets is a cmake "build-in" target, like "test", "install" or "all".

So anything doing this works by luck it seems.

foundrytom:
Closing, as #251 will cover remanning tasks at our leisure.

PR #192, opened 2022-01-25
[Build] Initial skeleton Python bindings

feltech:
Part of #183.

Add CMake infrastructure for building and installing a Python extension module, `_openassetio`, and use that (rather blank) extension module in the Python sources.  

Hence we must disable `pytest` in CI until #182 is done. To validate locally (and possibly useful for upcoming CI work) add some convenience targets for running tests in an installed Python environment.

The exported targets when this project is used as a CMake library are now `OpenAssetIO::openassetio-core` and `OpenAssetIO::openassetio-python`.


src/openassetio-python/CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+# Calculate relative path from site-packages to lib directory.
+file(RELATIVE_PATH
+    python_site_dir_rel_to_lib
+    "${CMAKE_INSTALL_PREFIX}/${OPENASSETIO_PYTHON_SITEDIR}"
+    "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+
+set_target_properties(
+    openassetio-python
+    PROPERTIES
+    INSTALL_RPATH "$ORIGIN/${python_site_dir_rel_to_lib}"
```
CesarSaezFoundry, line 57: [consider] This might only work on Linux, probably is worth to branch out doing the right thing per OS in here (gotta love the rpath fun! :smile: )

cmake/DefaultTargetProperties.cmake
```diff
@@ -83,15 +82,11 @@ function(set_default_target_properties target_name)
...
     # RPATH wrangling.
 
-    # With the 'usual' install path structure in mind, search from the
-    # bin directory (i.e. a binary loading a dynamic library) and then
-    # from the current directory (i.e. dynamic library loading another
-    # dynamic library).
     if (APPLE)
         # TODO(DF): handle RPATH on OSX - e.g. @rpath vs. @loader_path
         message(AUTHOR_WARNING "OSX RPATH not configured")
     else ()
```
CesarSaezFoundry, line 88: [question] Does the project supports Windows?
Perhaps it wouldn't be too painful to set `$ORIGIN` behind an `elseif(UNIX)`

python/openassetio/managerAPI/ManagerInterface.py
```diff
@@ -27,13 +27,15 @@
 # pylint: disable=unused-argument,no-self-use
 
 import abc
+# TODO(DF): Remove pylint disable once CI is fixed.
+import _openassetio  # pylint: disable=import-error
```
CesarSaezFoundry, line 31: [consider] This looks like a good candidate for a relative import!? (unless I'm missing something)

src/openassetio-python/CMakeLists.txt
feltech: Good point, added OS-specific branching and TODO comment in f1ed0ce9

cmake/DefaultTargetProperties.cmake
feltech: Yeah we will definitely be supporting Windows at some point.  Full Windows support is for another ticket, but we've been trying to ensure a reasonable foundation.  I updated the OS-specific branching in 74580c30 (as well as reworded some comments).

python/openassetio/managerAPI/ManagerInterface.py
feltech: The directory structure looks like
```
.
└── site-packages
    ├── openassetio
    │   ├── managerAPI
    │   │   ├── __init__.py
    │   │   ├── ManagerInterface.py
    ├── _openassetio.so
```
i.e. the `_openassetio` Python extension module is installed at the root of `site-packages`, not within the same package, so can't be relatively imported. 

Does this seem reasonable?

CesarSaezFoundry: Oh I see! Sorry my bad, I thought it was all part of the same package. In that case it makes perfect sense :+1:

foundrytom: Question, would it be worth moving it under `openassetio`?

foundrytom (approved):

python/openassetio/managerAPI/ManagerInterface.py
feltech: Moved under `openassetio` in 9b48c9ac, which necessitated a full `pip install` without `--editable`, updated in 025f4a7c.

feltech:
d37fec9f fixes an issue I found where pybind11's CMake module requires the Python interpreter, even if we don't need it

foundrytom (approved):
Awesome, thanks @feltech, happy to squash and merge at your leisure.

PR #204, opened 2022-02-01
[CI] Publish docs artifacts

### LINKED ISSUE #199: Expose docs built through CI
## What

As a developer/tester on the project, I'd like to have access to the docs produced by the `docs` CI check.

## Why

Otherwise docs need to be manually built for QA.

## Acceptance Criteria

- Ability to download build artifacts from docs CI check via GitHub PR page/check page.

steve-adamsATfoundry:
Closes #199
Publishes the documentation html built to allow us to access the docs for QA

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,6 +61,12 @@ jobs:
         # to work with GNU grep.
         ! grep -qE "^.*?/src/[^:]+:[0-9]+: ?[a-zA-Z]+: ?.*$" doxygen-log.txt
 
+    - name: Expose archive docs artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: doxygen-documentation
+        path: doc/html
+
   business-language:
     runs-on: ubuntu-latest
 


.github/workflows/test.yml
```diff
@@ -54,6 +54,12 @@ jobs:
         # to work with GNU grep.
         ! grep -qE "^.*?/src/[^:]+:[0-9]+: ?[a-zA-Z]+: ?.*$" doxygen-log.txt
 
+    - name: Expose archive docs artifact
```
feltech, line 57: [question] Do you think it would be useful to have the artifact even if there are Doxygen errors/warnings, e.g. for debugging? 

By default the next step won't run if the previous step failed.  But that can be tweaked, e.g. see [stackoverflow answer](https://stackoverflow.com/a/58003436/535103) and ["steps" context docs](https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context).

Perhaps though it's not worth the additional effort until we actually need it.

feltech (approved):

.github/workflows/test.yml
foundrytom: I'd be happy to leave this as-is for now, but @steve-adamsATfoundry can we squash these two commits together please?

foundrytom (approved):
Thanks @steve-adamsATfoundry looks good! Unfortunately there have been a few changes to `main` since this was originally authored, so it just needs re-basing so we can merge it. One of the devs can help you, or let me know if you need a hand.

PR #206, opened 2022-02-02
[CI] Run CI on macOS and Windows

foundrytom:
Introduces CI test coverage on `macOS` and `Windows`. This will most certainly make #181, #182, #183 at little more complex, but even basic coverage would be valuable as we don't have a lot of trig-platform development resource available right now, so catching issues early would be beneficial.  If the `C++` side becomes onerous, we can always disable some platforms in the interim.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,11 +24,8 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        # Note: It has been discussed that 3.9 is the first "production
-        # usable" version on Windows this needs further investigation
-        # before we special-case the platform. CY2022 uses 3.9
-        # throughout, so we should probably just target that.
-        python-version: 3.7
+        # CY2022 (https://vfxplatform.com)
+        python-version: 3.9
 
     - name: Install dependencies
       run: |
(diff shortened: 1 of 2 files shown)


foundrytom:
NB: Required CI checks need updating as they were anchored to the platform-free check name.

feltech (commented):
Changes look fine, but just a bit concerned by the
> pytest Expected — Waiting for status to be reported 

which appears to be stuck in the checks

foundrytom:
> Changes look fine, but just a bit concerned by the
> 
> > pytest Expected — Waiting for status to be reported
> 
> which appears to be stuck in the checks

Ahh sorry I'll fix that. I think its that GitHub matches checks by name, and so the existing requirement on the old non-os suffixed check is causing the stall.

foundrytom:
> Ahh sorry I'll fix that. I think its that GitHub matches checks by name, and so the existing requirement on the old non-os suffixed check is causing the stall.

Looks like I can't get rid of the 'expected' test without pushing a rebase. This should be fine to merge as is, then I can update the expected test names.

feltech (approved):

carmenpinto (approved):

PR #208, opened 2022-02-02
[Docs] Fix wording: through not thorough

fnRaihanKibria:


foundrytom (changes requested):
Thanks for fixing this @fnRaihanKibria - just a quick note - we have a specific format for our commit messages, see: https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/main/contributing/COMMITS.md#commit-messages

So in this case, it would be:

`[Docs] Fix wording: through not thorough`

If you could amend the title, LGTM.

fnRaihanKibria:
@foundrytom Fixed up commit!

foundrytom (approved):

foundrytom:
Closing so the branch can be moved to a fork for signing.

PR #213, opened 2022-02-04
[CI] Update ClangFormat configuration and add Github action to enforce it

feltech:
For #182 we want to choose a C++ style and enforce it in CI. We decided to stick to the "Google" preset of ClangFormat, with the only override being line length limit.

This PR updates the `.clang-format` configuration, reformats the C++ files, and adds a Github CI check to enforce it.


.github/workflows/code-quality.yml
```diff
@@ -34,3 +34,13 @@ jobs:
...
           #     https://github.com/PyCQA/pylint/issues/214
           pylint-ignore-paths: "tests/openassetio/pluginSystem/resources/symlinkPath"
+
+  clang-format:
+    name: ClangFormat
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run clang-format style check for C/C++/Protobuf.
+        uses: jidicula/clang-format-action@v4.5.0
```
foundrytom, line 44: [question] To play devils advocate, the GitHub Actions runners have [`clang-format` pre-installed](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md). Do we feel the complexity this action brings (ie. file-by file checking in a separate docker container `run`) gives us significant advantages over a simple script here that calls it directly?

foundrytom (commented):
Awesome. Thanks for updating the `README.md` too. Just one Q around the action.

.github/workflows/code-quality.yml
feltech: Good point, I didn't pick up on the fact it's running a separate docker process for each discovered file.  

There are a few little things to solve - finding the files, running clang-format, parsing the output for errors (it doesn't exit with non-zero if fixes are required) and neatly reporting the results.  

Perhaps all that is relatively trivial - I'll have a play and see if we can do it in a one-liner and avoid the dependency.

foundrytom: Thanks, I just thought too - it would be nice to elevate them to annotations as we did in `fn-pylint-action`.

feltech: Actually fairly trivial to do a one-liner (sans annotations): aca9224

We can look into the (several) other third-party actions to see if they add annotations, or we can roll our own, but perhaps this one-liner is sufficient for now and we can either see how it goes or [plan] to improve it in the future?

foundrytom (approved):
LGTM @feltech, please feel free to drop, squash and merge at your leisure.

PR #218, opened 2022-02-07
[CI] Build and test via CTest

feltech:
For #182, re-enable `pytest` testing, wrapped in a `ctest` target. That is, `cmake` configure the project then run `ctest` in Github CI. 

Only Linux is supported and enabled at the moment in Github CI. Windows and OSX support are left for future work.

The `openassetio-python` CTest target is configured to ensure dependencies are built and installed, including a Python environment for running the tests. 

A [Vagrant](https://www.vagrantup.com/) virtual machine config is included that roughly mirrors the Github Linux CI environment, useful for testing CI changes and onboarding devs.

The [Conan](https://conan.io/) package manager is used to source third-party dependencies from the public [ConanCenter](https://conan.io/center/) repository, for both Github CI and the Vagrant VM.

The README is updated to show both the `ctest` workflow as well as step-by-step.

The address sanitizer (ASan) and undefined behavior sanitizer (UBsan) are enabled for the Github CI action.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,9 @@ jobs:
           -Wdev \
           -DOPENASSETIO_ENABLE_TESTS=ON \
           -DOPENASSETIO_WARNINGS_AS_ERRORS=ON \
-          -DCMAKE_VERBOSE_MAKEFILE=ON
+          -DCMAKE_VERBOSE_MAKEFILE=ON \
+          -DOPENASSETIO_ENABLE_SANITIZER_ADDRESS=ON \
+          -DOPENASSETIO_ENABLE_SANITIZER_UNDEFINED_BEHAVIOR=ON
 
     - name: Build, install and test
       run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,14 +9,16 @@ on:
       - 'docs/**'
 
 jobs:
-  pytest:
+  test:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         # We can't properly align to the VFX Reference Platform as this
         # requires glibc 2.17, which is older than any of the available
         # environments.
-        os: [ubuntu-18.04, windows-2019, macos-11]
+        # TODO(DF): Enable other platforms.
+        # os: [ubuntu-20.04, windows-2019, macos-11]
+        os: [ubuntu-20.04]
 
     steps:
     - uses: actions/checkout@v2
@@ -27,19 +29,33 @@ jobs:
         # CY2022 (https://vfxplatform.com)
         python-version: 3.9
 
+    - name: Cache conan packages
+      id: cache-conan
+      uses: actions/cache@v2
+      with:
+        path: ~/conan
+        key: ${{ runner.os }}-conan
+
     - name: Install dependencies
+      # Configure the system and install library dependencies via conan
+      # packages.
       run: |
-        python -m pip install --upgrade pip
-        python -m pip install -r tests/requirements.txt
+        source resources/build/bootstrap-${{ matrix.os }}.sh
 
-    - name: Build
+    - name: Configure CMake build
       run: |
-        python -m pip install .
+        cmake -S . -B build \
+          --install-prefix $GITHUB_WORKSPACE/dist \
+          --toolchain $HOME/conan/conan_paths.cmake \
+          -Wdev \
+          -DOPENASSETIO_ENABLE_TESTS=ON \
+          -DOPENASSETIO_WARNINGS_AS_ERRORS=ON \
+          -DCMAKE_VERBOSE_MAKEFILE=ON
 
-    - name: Test
-      # TODO(DF): Re-enable once CI work has been done.
+    - name: Build, install and test
       run: |
-        echo "Skipping tests until CI build environment is configured"
+        cd build
+        ctest -VV
 
   docs:
     runs-on: ubuntu-latest
--- a/README.md
+++ b/README.md
@@ -281,6 +281,21 @@ cd build
 ctest
 ```
 
+### Using the ASWF Docker image
+
+The Academy Software Foundation maintains a set of VFX Reference 
+Platform compatible [Docker](https://www.docker.com/) images [here](https://github.com/AcademySoftwareFoundation/aswf-docker).
+
+The CY22 image contains all dependencies currently required for building
+OpenAssetIO. For example, to build and run the tests via a container, 
+from the repository root run
+
+```shell
+docker run -v `pwd`:/src aswf/ci-base:2022.2 bash -c '
+  cd /src && cmake -S . -B build -DOPENASSETIO_ENABLE_TESTS=ON &&\
+  cd build && ctest -VV'
+````
+
 ## Getting involved
 
 - See the [contribution guide](contributing/PROCESS.md)
--- a/README.md
+++ b/README.md
@@ -238,6 +238,48 @@ We can now run the Python tests via `pytest`
 pytest tests
 ```
 
+### Using Vagrant
+
+Included for convenience is a [Vagrant](https://www.vagrantup.com/)
+configuration for creating reproducible build environments. This is
+configured to create a virtual machine that matches the Linux Github CI
+environment as close as is feasible.
+
+The Vagrant VM installs third-party library dependencies using the
+[conan](https://conan.io/) package manager, and sources the dependencies
+from the default public [ConanCenter](https://conan.io/center/)
+repository.
+
+In order to build and run the tests within a Vagrant VM, assuming
+Vagrant is installed and the current working directory is the root of
+the repository, run the following
+
+```shell
+cd ci
+vagrant up
+# Wait a while...
+vagrant ssh
+cmake -S openassetio -B build --install-prefix ~/dist \
+  --toolchain ~/conan/conan_paths.cmake -DOPENASSETIO_ENABLE_TESTS=ON
+```
+
+Then we can run the following steps to build, install and run the tests
+
+```shell
+cmake --build build
+cmake --install build
+cmake --build build --target openassetio-python-venv
+source dist/bin/activate
+pip install ./openassetio
+pip install -r openassetio/tests/requirements.txt
+pytest openassetio/tests
+```
+or alternatively, simply
+
+```shell
+cd build
+ctest
+```
 
 ## Getting involved
 
--- a/README.md
+++ b/README.md
@@ -107,34 +107,6 @@ which appears to overlap with a subset of `OpenAssetIO`s concerns.
 
 ## Getting started
 
-### System requirements
-
-Ubuntu Linux 20.04 is currently the only fully tested platform. Support
-for other platforms is the subject of ongoing work. We further assume
-the following system packages are installed
-
-- CMake 3.21+
-- GCC 9.3
-
-### Library dependencies
-
-Binary builds additionally require the following packages to be
-available to the CMake build system.
-- Python 3.9 (development install)
-- [pybind11](https://pybind11.readthedocs.io/en/stable/) 2.6.1+
-
-We use the CMake build system for compiling the C++ core library and
-its Python bindings. As such, the library dependencies listed above must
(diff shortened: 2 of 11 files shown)


feltech:
For testing, if you wish to trigger ASan you can apply the following to cause a memory leak:
```
diff --git a/src/openassetio-core/managerAPI/ManagerInterface.cpp b/src/openassetio-core/managerAPI/ManagerInterface.cpp
index ddbd689..549141d 100644
--- a/src/openassetio-core/managerAPI/ManagerInterface.cpp
+++ b/src/openassetio-core/managerAPI/ManagerInterface.cpp
@@ -2,6 +2,7 @@
 // Copyright 2013-2022 The Foundry Visionmongers Ltd
 
 #include <openassetio/managerAPI/ManagerInterface.hpp>
+#include <limits>
 
 namespace openassetio
 {
@@ -9,7 +10,10 @@ inline namespace OPENASSETIO_VERSION
 {
 namespace managerAPI
 {
-ManagerInterface::ManagerInterface() {}
+ManagerInterface::ManagerInterface()
+{
+    new int;
+}
 }  // namespace managerAPI
 }  // namespace OPENASSETIO_VERSION
 }  // namespace openassetio
```

To test UBsan you can apply the following to cause an overflow
```
diff --git a/src/openassetio-core/managerAPI/ManagerInterface.cpp b/src/openassetio-core/managerAPI/ManagerInterface.cpp
index ddbd689..0114af3 100644
--- a/src/openassetio-core/managerAPI/ManagerInterface.cpp
+++ b/src/openassetio-core/managerAPI/ManagerInterface.cpp
@@ -2,6 +2,7 @@
 // Copyright 2013-2022 The Foundry Visionmongers Ltd
 
 #include <openassetio/managerAPI/ManagerInterface.hpp>
+#include <limits>
 
 namespace openassetio
 {
@@ -9,7 +10,12 @@ inline namespace OPENASSETIO_VERSION
 {
 namespace managerAPI
 {
-ManagerInterface::ManagerInterface() {}
+ManagerInterface::ManagerInterface()
+{
+    int j = std::numeric_limits<int>::max();
+    j += std::numeric_limits<int>::max();
+    (void)j;
+}
 }  // namespace managerAPI
 }  // namespace OPENASSETIO_VERSION
 }  // namespace openassetio
```

feltech:
The last commit came from internal testing on CentOS 7 where we found an issue with ConanCenter's Python 3.9 package. So the README was clarified to mention that Ubuntu 20.04 is the only currently tested platform.  

This is mainly because Github CI doesn't support other Linux distros. We can work around that using containerisation, but that is left for future work.

README.md
```diff
@@ -109,24 +109,158 @@ which appears to overlap with a subset of `OpenAssetIO`s concerns.
 
 ### System requirements
 
-- `linux` (`macOS` and `Windows` are currently unsupported)
-- `Python 3.7` or later
+Ubuntu Linux 20.04 is currently the only fully tested platform. Support
```
foundrytom, line 112: [consider] Move the vagrant section to `Getting Started` and then move the full build instructions into `INSTALL`, and link from here?

Generally good to have easiest first, then increasing detail after to avoid Wall of Text syndrome. I appreciate that long-term this isn't what we'd want. But as we officially say Ubuntu is the only currently fully supported platform, that's prob OK for now?

foundry-markf: Q: Can you move it to the Wiki instead? with an optional link from the README.md to look at other options.

ci/bootstrap-ubuntu-20.04.sh
```diff
@@ -0,0 +1,30 @@
...
+# action, so a fresh Github VM is left without these system packages.
+sudo apt-get install -y  --no-install-recommends libfontenc-dev libx11-xcb-dev libxaw7-dev \
+    libxcb-dri3-dev libxcb-icccm4-dev libxcb-image0-dev libxcb-keysyms1-dev libxcb-randr0-dev \
+    libxcb-render-util0-dev libxcb-shape0-dev libxcb-sync-dev libxcb-util-dev libxcb-xfixes0-dev \
+    libxcb-xinerama0-dev libxcb-xkb-dev libxcomposite-dev libxcursor-dev libxdamage-dev \
+    libxfixes-dev libxi-dev libxinerama-dev libxmu-dev libxmuu-dev libxpm-dev libxrandr-dev \
+    libxres-dev libxss-dev libxtst-dev libxv-dev libxvmc-dev libxxf86vm-dev
+
+# Install additional build tools.
+sudo pip3 install conan cmake==3.21 ninja
```
foundrytom, line 22: [consider] Should we version pin `conan` and `ninja`? We had a lot of fun before with minor differences in package versions.

foundry-markf: Best to.
Validate wherever you can too.

CMake minimum version check

Conan can do 
e.g. https://github.com/conan-io/conan-center-index/blob/master/recipes/zlib/all/conanfile.py#L4

README.md
```diff
@@ -109,24 +109,158 @@ which appears to overlap with a subset of `OpenAssetIO`s concerns.
...
+dist/bin/pip install ./openassetio
+dist/bin/pip install -r openassetio/tests/requirements.txt
+dist/bin/pytest openassetio/tests
 ```
-pip install -r tests/requirements.txt
-pytest
+or alternatively, simply
+```shell
+cd build
+ctest
```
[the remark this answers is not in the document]
feltech: `ctest -VV` should do it (`test.yml` uses that)

README.md
```diff
@@ -109,24 +109,158 @@ which appears to overlap with a subset of `OpenAssetIO`s concerns.
...
+  --toolchain ~/conan/conan_paths.cmake -DOPENASSETIO_ENABLE_TESTS=ON
+```
+Then we can run the following steps to build, install and run the tests
+```shell
+cmake --build build
+cmake --install build
+cmake --build build --target openassetio-python-venv
+dist/bin/pip install ./openassetio
+dist/bin/pip install -r openassetio/tests/requirements.txt
+dist/bin/pytest openassetio/tests
```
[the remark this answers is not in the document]
feltech: I'm reluctant to allow `pytest` to scan the entire repository root for tests, since the repo directory can get filled with cruft, depending how users configure things (e.g. build directories, venvs, conan packages), slowing down discovery, and possibly even running the tests of third-party packages.

We can add directories to exclude to the `pyproject.toml`?  Though it would be easy to miss one.

README.md
```diff
@@ -109,24 +109,158 @@ which appears to overlap with a subset of `OpenAssetIO`s concerns.
...
+overridden by setting the built-in CMake variables `BUILD_SHARED_LIBS`
+and `CMAKE_INSTALL_PREFIX`, respectively.
+
+If shared library builds are disabled, then the core library will be
+built and installed as a static library, and statically linked into the
+Python module.
+
+In order to build and install the binary artifacts, assuming the current
+working directory is at the repository root, run
+```shell
```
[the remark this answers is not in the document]
feltech: Added spaces in 48aa5f95

README.md
feltech: As discussed offline, I've left this as-is. Also added a section on using the ASWF Docker image as part of 5ed9cdae

README.md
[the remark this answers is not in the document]
feltech: Updated in 1d59523b and 67d11825

feltech:
Not part of any current review comments: 5ed9cda introduced a slight change of behaviour with a few consequences - see the lengthy commit message.

README.md
foundrytom: Awesome, thanks. I am a little worried how long the `README.md` is now, what do you think about splitting the build stuff out into a separate doc (`INSTALL`)?

foundrytom: What do you think about leaving the `ASWF docker` one on the main page, and linking to full build instructions- given the nature of our target audience?

README.md
```diff
@@ -109,26 +109,193 @@ which appears to overlap with a subset of `OpenAssetIO`s concerns.
...
+
+The CY22 image contains all dependencies currently required for building
+OpenAssetIO. For example, to build and run the tests via a container, 
+from the repository root run
+
+```shell
+docker run -v `pwd`:/src aswf/ci-base:2022.2 bash -c '
+  cd /src && cmake -S . -B build -DOPENASSETIO_ENABLE_TESTS=ON &&\
+  cd build && ctest -VV'
+````
```
foundrytom, line 297: [fix] Stray extra backtick (sorry I couldn't figure the escape syntax to make it a suggestion)

README.md
```diff
@@ -109,26 +109,193 @@ which appears to overlap with a subset of `OpenAssetIO`s concerns.
...
+
+```shell
+cmake -S . -B build -DOPENASSETIO_ENABLE_TESTS=ON
 ```
+
+#### Quick start: using `ctest`
+
+The manual steps detailed below are conveniently wrapped by CTest -
+CMake's built-in test runner. In order to execute the tests,
+from the build directory simply run
```
foundrytom, line 191: [fix] ```suggestion
from the `build` directory simply run
```

File: .github/workflows/test.yml [hunk 8]
```diff
@@ -9,14 +9,16 @@ on:
       - 'docs/**'
 
 jobs:
-  pytest:
+  ctest:
```
foundrytom, line 12: [consider] Then we don't have to keep changing the "required checks" in the branch protection rules in the future if we change the test runner again. 
```suggestion
  test:
```

README.md
```diff
@@ -109,26 +109,193 @@ which appears to overlap with a subset of `OpenAssetIO`s concerns.
...
+[conan](https://conan.io/) package manager, and sources the dependencies
+from the default public [ConanCenter](https://conan.io/center/)
+repository.
+
+In order to build and run the tests within a Vagrant VM, assuming
+Vagrant is installed and the current working directory is the root of
+the repository, run the following
+
+```shell
+cd ci
```
foundrytom, line 258: [question] The `ci` dir suggests it is only used for CI. Is it right that this is really "build env"more generally - ie, as we also use it in vagrant, which could easily be used for local development?

src/openassetio-python/CMakeLists.txt
```diff
@@ -67,16 +67,28 @@ endif ()
...
+    # Target to wrap the venv command so it can be executed in isolation.
+    add_custom_target(
+        openassetio-python-venv
+        DEPENDS "${CMAKE_INSTALL_PREFIX}/bin/pip"
+    )
+
+    # Command to install test-specific dependencies (e.g. pytest).
+    add_custom_command(
+        OUTPUT
+        "${CMAKE_INSTALL_PREFIX}/bin/pytest"
```
foundrytom, line 88: [question] `CMake` newbie here. Are these like file targets in a `Makefile`, in that, with our flaky python install in the docker image, will this run every time as `bin/pytest` (and `bin/pip`) won't ever end up existing. Not sure there is much we can do about that though?

README.md
feltech: Good idea, will do.

README.md
feltech: Yeah I guess so. Though it's tightly coupled to CI at the moment, since the OS and hardware (roughly) match and the same bootstrap script is executed in both.

README.md
feltech: Split in 5212172

README.md
feltech: Fixed as part of 5212172

README.md
foundrytom: I wondered if we should rename it? Especially as its use is described in the `README` in the Vagrant section, not sure on a name though `tools`?

README.md
foundrytom: Copy is perfect, how about this for a structural tweak (its a `git diff`)
[suggestion.txt](https://github.com/TheFoundryVisionmongers/OpenAssetIO/files/8048370/suggestion.txt)

feltech: Applied with minor tweaks in 1d59523b

README.md
feltech: Moved in 3394f462

foundrytom (approved):
Awesome work @feltech Feel free to squash and merge at your leisure.

PR #221, opened 2022-02-07
[Test] Provide a suitably configured locale to `FixtureAugmentedTestCase`

foundrytom:
Following on from #219, makes the test harness hosts follow best practice of setting a suitably configured locale in all context's passed to API calls. Doing it in the test loader saves the test author from having to write a lot of boilerplate.

Ultimately, we should probably make `Context.locale` required and enforce that is its not `None` somewhere.

### DIFF
--- a/python/openassetio/test/manager/_implementation.py
+++ b/python/openassetio/test/manager/_implementation.py
@@ -22,6 +22,8 @@ import unittest
 
 from openassetio import hostAPI, pluginSystem, logging
 
+from .specifications import ManagerTestHarnessLocale
+
 
 __all__ = ['createHarness']
 
@@ -128,13 +130,18 @@ class _ValidatorTestLoader(unittest.loader.TestLoader):
         execute.
         """
         testCaseNames = self.getTestCaseNames(testCaseClass)
-        loadedSuite = self.suiteClass(
-            testCaseClass(
+        cases = []
+        for testCaseName in testCaseNames:
+
+            locale = ManagerTestHarnessLocale()
+            locale.testCase = f"{testCaseClass.__name__}.{testCaseName}"
+
+            cases.append(testCaseClass(
                 self.__fixtures.get(testCaseClass.__name__, {}).get(testCaseName),
-                self.__session, testCaseName
-            )
-            for testCaseName in testCaseNames)
-        return loadedSuite
+                self.__session, locale, testCaseName
+            ))
+
+        return self.suiteClass(cases)
 
     def setFixtures(self, fixtures):
         """
--- a/python/openassetio/test/manager/harness.py
+++ b/python/openassetio/test/manager/harness.py
@@ -131,7 +131,7 @@ class FixtureAugmentedTestCase(unittest.TestCase):
     subclasses via protected members.
     """
 
-    def __init__(self, fixtures, session, *args, **kwargs):
+    def __init__(self, fixtures, session, locale, *args, **kwargs):
         """
         Initializes an instance of this class.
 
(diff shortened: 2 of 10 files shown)


tests/openassetio/test/manager/conftest.py
```diff
@@ -75,3 +76,7 @@ def a_fixture_dict():
     return {
         "identifier": "org.some.manager"
     }
+
+@pytest.fixture
```
feltech, line 80: [consider] PEP8 - 2 lines above function

python/openassetio/test/manager/harness.py
```diff
@@ -30,8 +30,14 @@
 
 from . import _implementation
 
+# Hoist our locale from our private implementation (we don't declare
+# it here as otherwise we'd end up with circular imports trying
+# to use it in the harness loader).
+from ._implementation import ManagerTestHarnessLocale
```
feltech, line 36: [question] It looks like the only reason this is hoisted and exported is for the unit tests? I don't think it's so bad if the unit tests access `_implementation`.  Is there another reason?  Do we expect end users will need access to the class via `harness`, perhaps?

feltech (approved):

python/openassetio/test/manager/harness.py
feltech: Maybe the `ManagerTestHarnessLocale` should be defined in the new `specifications`? Not sure.

foundrytom: The `specifications` module is perfect - great idea. It needs to be publicly visible as any new specification introduced by a host must be available to the manager/etc... so they can test against it/know its properties/etc...

python/openassetio/test/manager/_implementation.py
```diff
@@ -22,10 +22,10 @@
 
 from openassetio import hostAPI, pluginSystem, logging
 
-from ..specifications import TestHarnessLocale
+from .specifications import ManagerTestHarnessLocale
```
foundrytom, line 25: @feltech I meant to ask before, I feel we need an official stance on relative vs absolute imports

python/openassetio/test/manager/harness.py
feltech: Not quite what I meant, but this seems reasonable to me.

python/openassetio/test/specifications.py
```diff
@@ -0,0 +1,31 @@
...
+#   limitations under the License.
+#
+"""
+@namespace openassetio.test.specifications
+Specifications for use within openassetio test harnesses.
+"""
+
+from openassetio.specifications import LocaleSpecification
+
+class TestHarnessLocale(LocaleSpecification):
```
feltech, line 23: [consider] PEP8 - 2 spaces above class :grin: 
```suggestion

class TestHarnessLocale(LocaleSpecification):
```

feltech (approved):

python/openassetio/test/manager/_implementation.py
feltech: I don't mind the Python 3 explicit relative imports, but really I also don't mind absolute imports.  

I think on balance relative imports convey slightly more information at a glance (where the file you're looking at sits in relation to its dependencies), and is terser. So I guess I lean toward using them.

One minor disadvantage is the "attempted relative import with no known parent package" error if importing the module on its own. Perhaps this could be annoying if trying to debug something on the command line, but is easily avoided. I've had this erroneously flagged by `pylint` before, but am struggling to reproduce it now.

foundrytom: Cool, shall we say relative by default then? I guess it also means you dont bake in the outer package name into every file, and force it's `__init__`.

foundrytom:
@feltech I've rebased this so it only contains the required commits now. Do you think we should fix the absolute imports in some of these files?

python/openassetio/test/manager/_implementation.py
feltech: Agreed

feltech (approved):

PR #231, opened 2022-02-14
[Core] Remove context passed to `isEntityReference`

foundrytom:
As documented, this method should only be testing if the format of the string is valid, as such, the result of this should _never_ be context dependent. Supplying a context is redundant and just creates ambiguity.


feltech (approved):
Interesting that `Examples.dox` already assumed this, and so was previously incorrect.

Issue #115, opened 2021-12-02
RFC 10: Redesign `managementPolicy() kWillManagePaths` flag


foundrytom:
## What

Redesign the flags returned by `ManagerInterface.managementPolicy` to better represent the permitted interactions after entity `preflight`, and before `registration`.

## Why

Presently, a `ManagerInterface` implementation sets the `kWillManagePaths` bit in its response to `managementPolicy` to indicate to the host whether it can supply the path for a future entity prior its creation. This is generally used by hosts that write data to disk, to ensure that data is written to the correct place. The flag exists as not all asset management systems are capable of doing this, and so a host can't assume they will always be told where to put new data. 

[This example](https://thefoundryvisionmongers.github.io/OpenAssetIO/examples.html#example_publishing_a_file) shows the the host logic required to handle a manager's policy for any given publish.

In many situations, a host may also wish to query attributes of the future entity to determine other aspects of data generation. For example #86.  At present `kWillManagePath` is the only means to determine if a manager is capable of providing information about a future entity. This has several issues:

 - It implies that paths are the only thing that might need to be queried ahead of time.
 - It doesn't indicate whether entity attributes or other calls can also be determined.

foundrytom:
We should so be careful to document how `managementPolicy` and `kReadMultiple` or `kWriteMultiple` interact. Should this be used as a 'can I' mechanism? Probably not, as it doesn't permit any meaningful messaging, and is perhaps conflating two things. We need a separate #9 anyway.

foundrytom:
#492 switched to using traits, which significantly improved the mechanism, and removed the need for a Core API definition of all possible behaviors.

PR #233, opened 2022-02-14
[Core] Remove optional entity ref from `managementPolicy`

### LINKED ISSUE #9: `Manager` permissions API
## What

API methods that allow a host to verify the permissions for manager related actions prior to attempting them.

## Why

Many UI elements need to be conditionally enabled based on the edibility of assets etc. (see Katana LiveGroups for reference). Relying on exceptions during operations post-fact does not allow a sensible user experience to be developed.

The Katana AssetAPI gained a [`checkPermissions`](https://learn.foundry.com/katana/dev-guide/OpsAndOpScript/AssetAPI/OpScript.html#Asset.AssetPlugin.checkPermissions) method in order to facilitate some of this.

Before we can migrate Katana over to OAIO, we will need to add API surface area to allow a host to query whether an action with any given context is expected to succeed.

foundrytom:
A partial precursor to #115, removes the optional `entityRef` from `managementPolicy`. It seemed like a good idea at the time, but in reality, just opens up ambiguity, and only half solves the problem.

#9 covers the fact that we need a broader mechanism for working out if operation X can be done in the context of Y. One of these things could be to register a new entity to a specific target entity. A more flexible mechanism that covers this and the topics in #9 would be far better than shoehorning the check into `managementPolicy` - which has no mechanism for communicating why.

With the optional reference, it was not clear if a host should always call this prior to `preflight`/`register` or not. Technically, it should! When we do resolve #9, that would then mean two methods potentially needed to be called before a registration/preflight, which would be a very poor design indeed.

By making `managementPolicy` entity invariant, it makes its use much clearer as to when it should be called by a host, and much easier for a host to respect its response.

It can now be simply defined as a one-off check for each context access pattern in relation to any given specification to enable or disable host-side behaviours.

### DIFF
--- a/resources/examples/manager/SampleAssetManager/tests/fixtures.py
+++ b/resources/examples/manager/SampleAssetManager/tests/fixtures.py
@@ -19,7 +19,6 @@ Manager test harness test case fixtures for Sample Asset Manager (SAM).
 from openassetio.constants import kField_EntityReferencesMatchPrefix
 
 kIdentifier = "org.openassetio.examples.manager.sam"
-dummy_ref = "sam:///placeholder"
 
 fixtures = {
     "identifier": kIdentifier,
@@ -39,11 +38,5 @@ fixtures = {
                 kField_EntityReferencesMatchPrefix: "sam:///"
             }
         }
-    },
-    "Test_managementPolicy": {
-        "test_calling_with_read_context_and_entity_reference": { "valid_entity_reference": dummy_ref },
-        "test_calling_with_write_context_and_entity_reference": { "valid_entity_reference": dummy_ref },
-        "test_calling_with_read_multiple_context_and_entity_reference": { "valid_entity_reference": dummy_ref },
-        "test_calling_with_write_multiple_context_and_entity_reference": { "valid_entity_reference": dummy_ref }
-    },
+    }
 }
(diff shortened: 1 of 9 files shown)


resources/examples/manager/SampleAssetManager/tests/fixtures.py
```diff
@@ -39,11 +39,5 @@
                 kField_EntityReferencesMatchPrefix: "sam:///"
             }
         }
-    },
-    "Test_managementPolicy": {
-        "test_calling_with_read_context_and_entity_reference": { "valid_entity_reference": dummy_ref },
```
feltech, line 44: [consider][minor] `dummy_ref` is now unused and can be removed.

feltech (approved):

resources/examples/manager/SampleAssetManager/tests/fixtures.py
foundrytom: Bah, sorry. Removed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/233/commits/3c5b12291d307fa5942fe7f81a595a670af7a3de).

feltech (approved):
LGTM, rebase, squash and merge when ready

PR #234, opened 2022-02-15
[Test] Fix locale used in `apiComplianceSuite`

foundrytom:
It is very easy for test case authors to forget to set the correct locale. This saves everyone getting it wrong, as long as they remember to use it.


feltech (approved):
Whoops!  Also added #236 since I noticed we're not asserting if `managementPolicy` returns a valid bitfield.

Rebase, squash and merge when ready.

fnRaihanKibria (approved):
Nothing to add from me.

PR #237, opened 2022-02-16
[Test] Assert range of `managementPolicy` in `test.manager.apiComplianceSuite`

### LINKED ISSUE #236: `apiComplianceSuite` `Test_managementPolicy` should assert on valid return value not just type
Currently the `apiComplanceSuite` test `Test_managementPolicy` asserts the returned value from `managementPolicy` calls is an `int` but does not further assert that the integer is one of the allowed values.

foundrytom:
Closes #236 - we weren't being strict enough when just checking its type.

### DIFF
--- a/python/openassetio/constants.py
+++ b/python/openassetio/constants.py
@@ -65,6 +65,10 @@ kWantsThumbnail = 16
 ## @}
 
 
+## The maximum integer value of a valid management policy bitfield
+kMaxManagementPolicyValue = 31
+
+
 ## @name Field Names
 ## Field names are to be used whenever data is get or set from a dictionary by
 ## key, rather than through an accessor in some wrapper class (eg:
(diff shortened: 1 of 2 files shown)


python/openassetio/constants.py
```diff
@@ -65,6 +65,10 @@
 ## @}
 
 
+## The maximum integer value of a valid management policy bitfield
+kMaxManagementPolicyValue = 31
```
feltech, line 69: [question] How was this number derived?  Naively,
```python
>>> 1 | 3 | 8 | 16
27
```
am I missing something?  If I add in the `7` then I get `31`, but the `7` is for `kStateMask`, which presumably isn't a state in itself? Or is it that we want to reserve 3 bits for future state?

foundrytom:
Closing to revisit properly in the future.

PR #238, opened 2022-02-17
[Docs] Add hierarchical or compositional specifications decision record

carmenpinto:
Specifications are used to define the nature of entities, relationships, or locales in OpenAssetIO. The decision we've just made describes a change from a hierarchical structure for specifications to a trait based, more functional approach. Please give your thoughts on this decision record, your feedback makes our work more relevant to you. 

Easy to read view: [DR007-Hierarchical-or-compositional-traits-for-specifications.md](https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/124790f0d136defff1b292e4b42827d123ff9364/decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md)

### DIFF
--- a/decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
+++ b/decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
@@ -1,0 +1,589 @@
+# DR007 Hierarchical or composition based traits for specifications
+
+- **Status:** Decided
+- **Impact:** High
+- **Driver:** @carmenpinto
+- **Approver:** @feltech
+- **Contributors:** @foundrytom @feltech @fnRaihanKibria
+- **Outcome:** OpenAssetIO will use a trait-composition based typing
+  system instead of a hierarchical approach.
+
+## Background
+
+Specifications are used to define the nature of entities, relationships,
+or locales in OpenAssetIO. These definitions can be used to determine
+acceptable behaviours by a Manager[^1] system to act on the
+specification, or to determine how the Host[^2] interacts through
+OpenAssetIO. Reducing the amount of code required to deal with each
+specification can be achieved by defining guarantees about the
+specifications. How specifications are defined in terms of their
+relation to each other and the degree that these classifications should
+be enforced are discussed in the options below.
+
+The current design of the OpenAssetIO specification system comes from
+prior work, where it was successfully employed in the development of
+real-world editorial and post-production workflows connecting
+commercially available DCCs tools and asset management systems.
+
+## Relevant data
+
+There are a few challenges to defining specifications. The types of
+entities manipulated by Managers continually grow over time as new
+software and file formats and new ways of creating images and scenes are
+added. Specifications have to continue to be able to keep their
+guarantees about their types while new guarantees are added. This makes
+it easier for DCCs, their developers, and users of the pipeline to be
+able to add to the types available without losing shared behaviour
+between specifications - both across DCCs and shared between the
+OpenAssetIO community. These types of entities, relationships and
+locales can't be counted on to be exactly the same, but defining the
+similarities between them so that they can fulfill reliable behaviours
+is important so that the specifications have sharable meaning.
+
+OpenAssetIO requires a mechanism capable of precisely describing
+entities, relationships and locales with the following criteria:
+
+- Hierarchical or Composable (to support coercion of custom definitions
+  to well-known behavioural traits)
+- Simple in structure (to simple data types for interop/serialisation)
+- Robust (compile-time checks for programming errors - property access
+  or invalid method arguments)
+- Discoverable (runtime introspection/help)
+- Definable out of core (by host integrations/managers)
+
+Precedents exist to deal with the criteria above. USD[^3] supports
+coercion of custom definitions to well-known behavioural traits through
+USD schemas. It allows some schemas to be applied to primitives, which
+gives them attributes and behaviours specified by the schema. For
+example, `UsdGeomModelAPI` adds bounding extents to primitives, as well
+as the ability to create a constraint target. Applied schemas can't
+inherit from each other. Schemas that provide a type instead of being
+applied to a primitive can be hierarchical. An example of this
+is `UsdGeomXformable`, which inherits from `UsdGeomImageable`
+. `UsdGeomImageable` provides an interface for computing the visibility
+of the primitive, `UsdGeomXformable` adds the ability to manipulate the
+transform of primitives that are of this type as well as keeping the
+visibility attributes. Applied schemas are used on these typed schemas
+to share functionality.
+
+A case of this is occurring is with mesh lights - a geometry object that
+should have light behaviours. A typed, hierarchical, schema `Light` in
+USD was rewritten as an applied schema `UsdLuxLightAPI` so light
+behaviours could be added to other schemas -
+see [Adapting UsdLux to Accommodate Geometry Lights](https://graphics.pixar.com/usd/release/wp_usdlux_for_geometry_lights.html)
+. The mesh light could be then represented as a primitive of
+the `UsdGeomMesh` schema, with the `UsdLuxLightAPI` applied. This is a
+working example of the necessity of being able to share guaranteed
+behaviours causing the emergence of more trait based specifications.
+
+## Options considered
+
+### Option 1
+
+Hierarchical specifications with a single parent per child. We assume
+that every child fulfills the contract of its parent. For example, that
+all alembic files are a geometry cache, and that all geometry caches
+might have coordinate system based data, and that all geometry caches
+are a type of binary blob. This could be described in a
+specification: `entity.blob.geometry.cache.alembic`. Another example
+could be an exr render, which could be a type of image, which could also
+be a type of blob binary data, represented as: `entity.blob.image.exr` .
+
+#### Pros
+
+- There's only one way to represent the specification
+- Inheritance provides an automatic fall back mechanism
+- Has been proven in real-world, working scenarios
+
+#### Cons
+
+- It's difficult to insert new specifications in between types, requires
+  updating the child type identifier of the new insert in the case of
+  back-end stores. For example, should the schema change between shows,
+  and the entities need to be compared, the entity type would have to be
+  mapped manually for any tools relying on it.
+- This requires knowledge of as many possible entities as possible to be
+  efficiently taxonomied into an efficient and useful hierarchy
+- Sharing behaviours between taxonomically unrelated types - for example
+  a frame range applied to both a geometry cache and a shot - is
+  difficult, it relies on duplicated property definitions on each of
+  specifications, meaning that the trait of “having a frame range"
+  becomes implied though well-known property names, rather than an
+  explicitly stated trait
+- In order to share well-known base specification hierarchies with the
+  OpenAssetIO community, interop is facilitated through the use of
+  common base specifications. If these base hierarchies change, every
+  place where the schema was persisted will no longer be able to
+  interact with each other despite possibly having the same expected
+  behaviour
+
+### Option 2
+
+Trait based compositional specifications. Specifications are made from
+generic traits that guarantee certain behaviours or consumption patterns
+from the Manager or Host. The alembic cache could be represented as the
+following: `entity.geometry.cache`. The exrs rendered out from the
+Option 1 example could also be represented as `entity.image.cache` as
+they are a type of image cache. Managers that are interested in freeing
+up some memory could then query `entity.cache` and schedule them for
+deletion.
+
+#### Pros
+
+- Adding to specifications is easy by adding traits
+- Traits can be shared between specifications easily without needing to
+  hide them as emergent phenomena based on well-known property names
+- Common traits can be shared with the OpenAssetIO community without
+  having to exactly match a whole specification, decreasing the risk of
+  inclusion of very well known traits (such as `blob`)
+  in the core API if required
+
+#### Cons
+
+- Traits can be contradictory
+- Working with traits could be more complex than working with a single
+  parent hierarchy due to multiple combinations of traits being possible
+- Duplication of traited specification is possible
+- Not yet practically validated
+- Changes to specific traits still affects all specifications that
+  incorporate them
+
+## Outcome
+
+We will adopt Option 2.
+
+## Rationale
+
+Using traits changes the meaning of the specification from
+classification to a purpose-driven, exact requirement approach. It
+increases the shareability and extensibility of specifications through
+the modular traits, and avoids hidden traits being expressed through
+“magic” groups of well-known properties where the hierarchy precluded
+them being ancestors. Although it could potentially add complexity in
+terms of mapping traits to specifications, the ability to search by
+exactly the aspect of the specification that is required could reduce
+having to traverse over classifications while guaranteeing the behaviour
+can be applied.
+
+## Appendix
+
+### Implementation sketches
+
+#### Option 1
+
+In this hierarchy, the startFrame and endFrame properties are duplicated
+a few times due these types not being able to share a base.
+
+```
+  specification Blob {
+    class BlobSpecification
+    str mimeType
+  }
+  specification Blob.Geometry {
+    class GeometrySpecification
+  }
+  specification Blob.Geometry.AnimationCache {
+    class AnimationCacheSpecification
+    int startFrame
+    int endFrame
+  }
+  specification Blob.Geometry.Rig {
+    class RigSpecification
+  }
+  specification Blob.Image {
+    class ImageSpecification
+    int channels
+  }
+  specification Blob.Image.RenderCache {
+    class RenderCacheSpecification
+    int startFrame
+    int endFrame
+  }
+  specification Blob.Image.Panoramic {
+    class PanoramicSpecification
+    float angle
+  }
+  specification Container {
+    class ContainerSpecification
+    bool isEmpty
+  }
+  specification Container.Shot {
+    class ShotSpecification
+    int startFrame
+    int endFrame
+  }
+```
+
+```mermaid
+classDiagram
+  direction TB
+  BlobSpecification <-- GeometrySpecification
+  GeometrySpecification <-- AnimationCacheSpecification
+  GeometrySpecification <-- RigSpecification
+  BlobSpecification <-- ImageSpecification
+  ImageSpecification <-- RenderCacheSpecification
+  ImageSpecification <-- PanoramicSpecification
+  ContainerSpecification <-- ShotSpecification
+  
+  class BlobSpecification {
+    str mimeType
+  }
+  class GeometrySpecification {
+  }
+  class AnimationCacheSpecification {
+    int startFrame
+    int endFrame
+    int size
+  }
+  class RigSpecification {
+  }
+  class ImageSpecification {
+    int channels
+  }
+  class RenderCacheSpecification {
+    int startFrame
+    int endFrame
+    int size
+  }
+  class PanoramicSpecification {
+    float angle
+  }
+  class ShotSpecification {
+    int startFrame
+    int endFrame
+  }
+  class ContainerSpecification {
+    bool isEmpty
+  }
+```
+
+If we change the hierarchy to share Sequence, Rig and Panoramic now have
+an unwanted startFrame and endFrame property that might not apply. In
+addition, all sequences have to be containers, or all containers have to
+be sequences.
+
+```mermaid
+classDiagram
+  direction TB
+  BlobSpecification <-- GeometrySpecification
+  SequenceSpecification <-- BlobSpecification
+  GeometrySpecification <-- AnimationCacheSpecification
+  GeometrySpecification <-- RigSpecification
+  BlobSpecification <-- ImageSpecification
+  ImageSpecification <-- RenderCacheSpecification
+  ImageSpecification <-- PanoramicSpecification
+  SequenceSpecification <-- ShotSpecification
+  ContainerSpecification <-- SequenceSpecification : everything has to be a container
+  
+  class BlobSpecification {
+    str mimeType
+  }
+  class SequenceSpecification {
+    int startFrame
+    int endFrame
+  }
+  class GeometrySpecification {
+  }
+  class AnimationCacheSpecification {
+    int size
+  }
+  class RigSpecification {
+  }
+  class ImageSpecification {
+    int channels
+  }
+  class RenderCacheSpecification {
+    int size
+  }
+  class PanoramicSpecification {
+    float angle
+  }
+  class ShotSpecification {
+  }
+  class ContainerSpecification {
+    bool isEmpty
+  }
+```
+
+```mermaid
+classDiagram
+  direction TB
+  BlobSpecification <-- GeometrySpecification
+  SequenceSpecification <-- BlobSpecification
+  GeometrySpecification <-- AnimationCacheSpecification
+  GeometrySpecification <-- RigSpecification
+  BlobSpecification <-- ImageSpecification
+  ImageSpecification <-- RenderCacheSpecification
+  ImageSpecification <-- PanoramicSpecification
+  ContainerSpecification <-- ShotSpecification
+  SequenceSpecification <-- ContainerSpecification : all containers have to be sequences
+  
+  class BlobSpecification {
+    str mimeType
+  }
+  class SequenceSpecification {
+    int startFrame
+    int endFrame
+  }
+  class GeometrySpecification {
+  }
+  class AnimationCacheSpecification {
+    int size
+  }
+  class RigSpecification {
+  }
+  class ImageSpecification {
+    int channels
+  }
+  class RenderCacheSpecification {
+    int size
+  }
+  class PanoramicSpecification {
+    float angle
+  }
+  class ShotSpecification {
+  }
+  class ContainerSpecification {
+    bool isEmpty
+  }
+```
+
+#### Option 2
+
+This lets us specify only the properties that apply to the
+specification, without duplication.
+
+```
+  trait Blob {
+    class BlobTrait
+    str mimeType
+  }
+  trait Geometry {
+    class GeometryTrait
+  }
+  trait Image {
+    class ImageTrait
+    int channels
+  }
+  trait Lens {
+    class LensTrait
+    int angle
+  }
+  trait Sequence {
+    class SequenceTrait
+    int startFrame
+    int endFrame
+  }
+  trait Cache {
+    class CacheTrait
+    int size
+  }
+  trait Container {
+    class ContainerTrait
+    bool isEmpty
+  }
+  
+  specification AnimationCache {
+    class AnimationCacheSpecification
+    traits: blob, geometry, sequence, cache
+  }
+  specification Rig {
+    class RigSpecification
+    traits: blob, geometry
+  }
+  specification RenderCache {
+    class RenderCacheSpecification
+    traits: blob, sequence, image, cache
+  }
+  specification Panoramic {
+    class PanoramicSpecification
+    traits: blob, image, lens
+  }
+```
+
+```mermaid
+classDiagram
+  direction RL
+  RigSpecification *-- GeometryTrait
+  RigSpecification *-- BlobTrait
+  
+  AnimationSpecification *-- BlobTrait
+  AnimationSpecification *-- GeometryTrait
+  AnimationSpecification *-- SequenceTrait
+  AnimationSpecification *-- CacheTrait
+  RenderSpecification *-- ImageTrait
+  RenderSpecification *-- BlobTrait
+  RenderSpecification *-- SequenceTrait
+  RenderSpecification *-- CacheTrait
+  PanoramicSpecification *-- ImageTrait
+  PanoramicSpecification *-- BlobTrait
+  PanoramicSpecification *-- LensTrait
+  ShotSpecification *-- ContainerTrait
+  ShotSpecification *-- SequenceTrait
+  class BlobTrait {
+    str mimeType
+  }
+  class GeometryTrait {
+  }
+  class LensTrait {
+    int angle
+  }
+  class SequenceTrait {
+    int startFrame
+    int endFrame
+  }
+  class CacheTrait {
+  }
+  class ImageTrait {
+    int channels
+  }
+  class ContainerTrait {
+    bool isEmpty
+  }
+  class AnimationSpecification {
+  }
+  class RigSpecification {
+  }
+  class RenderSpecification {
+  }
+  class PanoramicSpecification {
+  }
+  class ShotSpecification {
+  }
+ 
+```
+
+### Checking management policies
+
+Often a Host may want to check management policies for a broader range
+of entity types based on some common shared properties.
+
+#### Option 1
+
+When using management interface functions the code using specifications
+is the same for Option 1 and Option 2.
+
+```python
+# uses specifications from "An implementation sketch of Option 1"
+
+# check if Images and Geometry are managed
+policy = manager.managementPolicy(
+    [ImageSpecification(), GeometrySpecification()], context)
+
+# manager offers to manage Image and Geometry specifications, and anything that descends from those
+policy == [kManaged, kManaged] 
+```
+
+#### Option 2
+
+```python
+# uses traits from "An implementation sketch of Option 2"
+
+# check if Specifications with Image or Geometry traits are managed
+policy = manager.managementPolicy(
+    [ImageSpecification, GeometrySpecification], context)
+
+# manager offers to manage Specifications with ImageTraits or GeometryTraits
+policy == [kManaged, kManaged] 
+```
+
+It could be possible to query combinations of these traits, such as by
+accepting lists of lists in Python.
+
+```python
+# uses traits from "An implementation sketch of Option 2"
+
+# check if Sequenced Images or Sequenced Geometry are managed
+policy = manager.managementPolicy(
+    [[ImageTrait, SequenceTrait], [GeometryTrait, SequenceTrait]],
+    context)
+
+# manager offers to manage Specifications that are Sequenced Images or Sequenced Geometry
+policy == [kManaged, kManaged] 
+```
+
+### Finding all the caches associated with a shot
+
+The resulting entity references could be used to help clear up space on
+disk when they're no longer necessary for work.
+
+Assumes the following set up:
+
+```python
+assets = ["asset://fxPointsCache",  # a geometry cache of point data
+          "asset://precompShadows"]  # rendered image of shadows
+
+shots = ["shot://ABC/01"]
+
+```
+
+#### Option 1
+
+```python
+manager.getRelatedReferences(shots, [RenderCacheSpecification(),
+                                     GeometryCacheSpecification()],
+                             context, session)
+
+# expected output: [["asset://fxPointsCache"], ["asset://precompShadows"]]
+```
+
+#### Option 2
+
+```python
+manager.getRelatedReferences(shots, [CacheTrait], context, session)
+
+# expected output: [["asset://fxPointsCache", "asset://precompShadows"]]
+```
+
+### List all the frame ranges in the host application
+
+In a host using a single shot, finding which entities aren't currently
+rendered with enough frames to cover the shot length could be useful.
+
+In this case, since having a frame range can't be part of the hierarchy
+without also affecting all leaf specifications in the hierarchy, it's
+considerably easier to specify the wanted trait versus having to list
+every possible specification where the frame range attributes might
+exist.
+
+#### Option 1
+
+```python
+specifications = [RenderCacheSpecification(),
+                  GeometryCacheSpecification(),
+                  ShotSpecification()]
+
+for specification in specifications:
+    entityRefs = hostInterface.knownEntityReferences(specification)
+    entityAttributes = manager.getEntityAttributes(entityRefs, context,
+                                                   session)
+
+    for entity in entityAttributes:
+        print("Frame range: ", entity[constants.kField_FrameStart],
+              "-", entity[constants.kField_FrameEnd])
+```
+
+#### Option 2
+
+```python
+entityRefs = hostInterface.knownEntityReferences(SequenceTrait())
+entityAttributes = manager.getEntityAttributes(entityRefs, context,
+                                               session)
+for entity in entityAttributes:
+    print("Frame range: ", entity[constants.kField_FrameStart],
+          "-", entity[constants.kField_FrameEnd])
+```
+
+## Terminology
+
+[^1]: The manager is a system that serves as the source of truth
+regarding the status or location of digital media or other
+organizational aspects of a project.
+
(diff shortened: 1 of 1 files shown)


decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+similarities between them so that they can fulfill reliable behaviours is important so that the
+specifications have sharable meaning.
+
+OpenAssetIO requires a mechanism capable of precisely describing entities, relationships and
+locales with the following criteria:
+
+Hierarchical or Composable (to support coercion of custom definitions to well-known behavioural
+traits) Simple in structure (to simple data types for interop/serialisation) Robust (compile-time
+checks for programming errors - property access or invalid method arguments) Discoverable (runtime
+introspection/help) Definable out of core (by host integrations/managers)
```
foundrytom, line 41: [fix] Formatting seems to be off here. Is this meant to be a bullet list?

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+- proven in real-world, working scenarios
+
+#### Cons
+
+- difficult to insert new specifications in between types, requires updating the child type
+  identifier of the new insert
+- requires knowledge of as many possible entities as possible to be efficiently taxonomied into an
+  efficient and useful hierarchy
+- sharing behaviours between taxonomically unrelated types - for example a frame range applied to
+  both a geometry cache and an image - is difficult, it relies on duplicated property definitions
```
foundrytom, line 67: [consider] Would using a 'shot' as an alternative to 'image' make things clearer? The current implementation shared frame range between geometry and images already via `file.geometry` and `file.image`, where frame range was defined on `file`.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+    # startFrame==endFrame means "not animated"
+    startFrame = Specification.TypedProperty(float, doc="")
+    endFrame = Specification.TypedProperty(float, doc="")
+
+
+class EditorialTrait(specifications.EntityTrait):
+    _type = "editorial"
+
+    startFrame = Specification.TypedProperty(int, doc="")
+    endFrame = Specification.TypedProperty(int, doc="")
```
foundrytom, line 169: [fix] The properties here seem astray? Would they be on a trait definition for a `FrameSequenceTrait`?

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+potentially add complexity in terms of mapping traits to specifications, the ability to search by
+exactly the aspect of the specification that is required could reduce having to traverse over
+classifications while guaranteeing the behaviour can be applied.
+
+## Appendix
+
+### An implementation sketch of Option 2
+
+```python
+from openassetio import Specification, specifications
```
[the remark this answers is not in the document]
foundrytom: eg:
```
trait raster {
   width: int
   height: int
}

trait sequence {
  startFrame: float
  endFrame: float
}

trait blob {
  mimeType: string
}

trait container {
}

---

specification Shot {
    traits: container, sequence
}

specification Image {
    traits: blob, raster, sequence
}

specification ImageBundle {
    traits: container, raster
}
```

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+class ImageCacheSpec(ImageTrait, CacheTrait):
+    _type = ".".join(sorted((ImageTrait._type, CacheTrait)))
+
+
+class GeometryTrait(specifications.EntityTrait):
+    _type = "geometry"
+
+
+class GeometryCacheSpec(GeometryTrait, CacheTrait):
+    _type = ".".join(sorted((GeometryTrait._type, CacheTrait)))
```
foundrytom, line 350: [fix] These declarations seem unused in this example.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+
+
+class Bundle(specifications.RelationshipSpecification):
+    _type = "bundle"
+
+
+manager.getRelatedReferences(shots, Bundle, context, session, resultSpec=[CacheTrait])
```
foundrytom, line 357: [fix] In the current API, you woudln't need the `Bundle` spec as you can query the trait directly:

`getRelatedReferences(shots, [CacheTrait()], context, session)`

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+
+class GeometryCacheSpec(GeometryTrait, CacheTrait):
+    _type = ".".join(sorted((GeometryTrait._type, CacheTrait)))
+
+
+class Bundle(specifications.RelationshipSpecification):
+    _type = "bundle"
+
+
+manager.getRelatedReferences(shots, Bundle, context, session,
```
foundrytom, line 290: [fix] In the current API, you woudln't need the `Bundle` spec as you can query the specifications directly:

`getRelatedReferences(shots, [ImageCacheSpec(), GeometryCacheSpec()], context, session)`

(`resultSpect` is currently singular)

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+
+```python
+
+assets = ["asset://fxPointsCache",  # a geometry cache of point data
+          "asset://precompShadows"]  # rendered image of shadows
+
+shots = ["shot://ABC/01"]
+
+
+class CacheTrait(specifications.EntityTrait):
```
foundrytom, line 333: [consider] Can we de-dupe the declarations of these traits to a section at the top? Then the examples can be a bit closer to gather, which makes it easier to read.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+    _type = ".".join(sorted((BlobTrait._type, DocumentTrait._type, NukeScriptTrait._type)))
+
+
+class KatanaProjectSpecification(BlobTrait, DocumentTrait, KatanaProjectTrait):
+    _type = ".".join(sorted((BlobTrait._type, DocumentTrait._type, KatanaProjectTrait._type)))
+```
+
+### An example of using these specifications
+
+#### Checking management policies
```
foundrytom, line 235: [consider] Is it worth noting that this and the ones below is identical for option 1 and option 2?

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+  identifier of the new insert
+- requires knowledge of as many possible entities as possible to be efficiently taxonomied into an
+  efficient and useful hierarchy
+- sharing behaviours between taxonomically unrelated types - for example a frame range applied to
+  both a geometry cache and an image - is difficult, it relies on duplicated property definitions
+  on each of specifications, meaning that the trait of “having a frame range becomes” implied
+  though well-known property names, rather than an explicitly stated trait.
+- in order to share well-known base specification hierarchies with the OpenAssetIO
+  community,interop is facilitated through the use of common base specifications. If these base
+  hierarchies change, every copy of the specifications will no longer be able to interact with each
```
foundrytom, line 72: [question] Does `copy` mean "places where the schema was persisted", rather than places where some code has produced a derived class?

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+  other despite possibly having the same expected behaviour
+
+### Option 2
+
+Trait based compositional specifications. Specifications are made from generic traits that
+guarantee certain behaviours or consumption patterns from the Manager or Host. The alembic cache
+could be represented as the following: entity.geometry.cache. The exrs rendered out from the Option
```
foundrytom, line 79: [consider]
```suggestion
could be represented as the following: `entity.geometry.cache`. The exrs rendered out from the Option
```

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+1 example could also be represented as entity.image.cache as they are a type of image cache.
```
foundrytom, line 80: [consider]
```suggestion
1 example could also be represented as `entity.image.cache` as they are a type of image cache.
```

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+Managers that are interested in freeing up some memory could then query entity.cache and schedule
```
foundrytom, line 81: [consider]
```suggestion
Managers that are interested in freeing up some memory could then query `entity.cache` and schedule
```

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+#### Cons
+
+- no guarantee that the traits aren't contradictory
+- working with traits could be more complex than working with a single parent hierarchy due to
+  multiple combinations of traits being possible
+- duplication of traited specification is possible
+- not yet practically validated
+- changes to specific traits still affects all specifications that incorporate them
+
+Estimated cost: Small|Medium|Large
```
foundrytom, line 102: [fix] Needs cost selecting

foundrytom (commented):
Great writeup, I think the main thing we need to include is an a/b comparison, and an illustration of how they differ on the manager side of the API too.

The examples are fantastic at explaining how the trait system could work, but as this is a choice between trait composition and the hierarchical approach, we need to present both options in equal detail.

It'd be good to make sure that we pick sample cases that highlight the differences between the two approaches too. For example, The `ImageCacheSpec` and `GeometryCacheSpec` share the same base traits, and so are equally mappable to the hierarchical version `blob.cache.image` and `blob.cache.geometry` ,to the extent that if you look at the code in the examples where it reaches the `MangerInterface`, it is actually identical for both options on the host side. (the last example, if you would just use the `BlobCacheSpecification`).

In the past, one of the biggest categorisation challenges in the hierarchical system was how to handle sequences of files. Should it be:

a) All files are sequences,
b) sequences are a specialisation of files (doesn't work well in the hierarhichcal model as you have to duplicate the entire sub-tree of files that could be either)
c) file is a specialisation of sequence
d) ... something else

This gets even more complicated with other non-file entities also are logically sequences (eg: shot), that resulted in the 'sequence' nature of things becoming an emergent behaviour via well-known property names that happened to be present in two different branches of the specification hierarchy.

 It feels like this is something the trait system makes infinitely more straight forward, understandable, and realisable.

It'd be great if we could use this to demonstrate the concrete differences between the two options, as I think it would give more tangible points to reason about when we share this with people in the broader community - on both sides of the API divide.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+
+#### Pros
+
+- only one way to represent the specification
+- automatic fall back mechanism through inheritance
+- proven in real-world, working scenarios
+
+#### Cons
+
+- difficult to insert new specifications in between types, requires updating the child type
```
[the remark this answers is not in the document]
carmenpinto: It's more about existing persisted copies of the scheme string in back-end stores. Comparisons between historically saved data - such as if you had to create entities based on a schema from a previous show - where the entity type specification has been serialised and saved could require explicit re-mapping.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: I was thinking literally in terms of clones of a shared repo of schemas, but places where the schema was persisted is better

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: How does the manager know which caches are associated with the shots?

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,277 @@
...
+
+#### Checking management policies
+
+When using management interface functions the code using specifications
+is the same for Option 1 and Option 2.
+
+```python
+# uses traits from "An implementation sketch of Option 2"
+imageSpec = ImageSpecification()
+policy = manager.managementPolicy([imageSpec], context)[0]
```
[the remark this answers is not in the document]
foundrytom: The original code is correct, `managementPolicy` is a batched method.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,277 @@
...
+if policy == [kIgnored]:
+    return  # nothing with an ImageSpecification is managed
+
+# check if pngs and alembic caches are supported by the manager
+exrSpec = ImageSpecification()
+exrSpec.mimeType = "image/x-exr; version=\"2\""
+
+abcSpec = GeometryCacheSpecification()
+abcSpec.mimeType = "alembic"
+policy = manager.managementPolicy([exrSpec, abcSpec], context)[0]
```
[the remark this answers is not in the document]
foundrytom: The original code is correct, `managementPolicy` is a batched method.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,345 @@
...
+                             resultSpec=[ImageSpecification(),
+                                         GeometryCacheSpecification()])
+
+# expected output: [["asset://fxPointsCache"], ["asset://precompShadows"]]
+```
+
+#### Option 2
+
+```python
+manager.getRelatedReferences(shots, [], context, session,
```
foundrytom, line 312: [fix] This should be `manager.getRelatedReferences(shots, [CacheTrait], context, session)`

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,345 @@
...
+          "asset://precompShadows"]  # rendered image of shadows
+
+shots = ["shot://ABC/01"]
+
+```
+
+#### Option 1
+
+```python
+manager.getRelatedReferences(shots, [], context, session,
```
foundrytom, line 302: [fix] This should be `manager.getRelatedReferences(shots, [ImageSpecification(),
                                          GeometryCacheSpecification()], context, session)`

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,345 @@
...
+                             resultSpec=[ImageSpecification(),
```
foundrytom, line 303: [fix] If the Specifications are as described above, then for Option 1 the equivalent would be this, rather than using the two individual specifications:

`manager.getRelatedReferences(shots, [BlobCacheSpecification()], context, session)`

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,345 @@
...
+
+# expected output: [["asset://fxPointsCache", "asset://precompShadows"]]
+```
+
+### List all the frame ranges in the host application
+
+#### Option 1
+
+```python
+specifications = [ImageSpecification(), GeometryCacheSpecification(),
```
foundrytom, line 323: [fix] Same comment re: `specifications = [BlobCacheSpecification(), ShotSpecification()]`

foundrytom (changes requested):
@carmenpinto this is really taking shape, nice one! Feels much easier to follow now. I wonder if there is a way we can make the second example really highlight the difficulties of the "emergent" frame range behaviour from Option 1 that is solved with Option 2.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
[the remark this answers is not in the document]
carmenpinto: Should it be if `policy == [[kIgnored]]` for clarity?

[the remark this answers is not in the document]
foundrytom: No, the management policy is a bit mask, so the value returned is an array of ints the same length as the input array of specs. So `policy = manager.managementPolicy([imageSpec], context)[0]` gives you the bitfield of the policy fo `imageSpec`.  So if you stick with the version with `[0]` then it'd just be `[kIgnored]`.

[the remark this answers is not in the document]
foundrytom: Sorry if there was some confusion - I couldn't see the intermediate state as the diff was old. But go whichever is the most readable

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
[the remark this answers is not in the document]
foundrytom: Sorry, I'd missed this in relation to the above, the `[0]` would give you the policy for `exrSpec`, so you wouldn't want the subscript here.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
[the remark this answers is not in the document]
foundrytom: David was right in the context of the way it was then presented in the subsequent lines.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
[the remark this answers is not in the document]
foundrytom: Perhaps call it `policies` in this case to avoid confusion? And use the `[0]` in the one above if its a single spec?

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Addressed in [78c87c0d7038ae6c2f948e0d7ed5bc39023d108f](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/238/commits/78c87c0d7038ae6c2f948e0d7ed5bc39023d108f)

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,600 @@
...
+  Image <-- Sequence_1
+  Image <-- Panoramic
+  Sequence_2 <-- Shot
+  Container <-- Sequence_2
+  Sequence_3 <-- AnimationCache
+  
+  class Blob {
+    str mimeType
+  }
+  class Sequence_1 {
```
foundrytom, line 348: [consdier][Strong] In Option 1, they don't have to have unique names. So it can still be `Sequence` in all branches.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,600 @@
...
+    bool isEmpty
+  }
+```
+
+#### Option 2
+
+This lets us specify only the properties that apply to the
+specification, without duplication.
+
+```mermaid
```
foundrytom, line 392: [consider][strong] Is it possible to change the presentation of the `Trait` classes so its more obvious they are conceptually different to the resulting `Specification`s, and to Option 1 - where there are only `Specification`s

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,600 @@
...
+
+```python
+# uses traits from "An implementation sketch of Option 2"
+policy = manager.managementPolicy([ImageSpecification()], context)
+
+if policy == [kIgnored]:
+    pass  # nothing with an ImageSpecification is managed
+
+# check if pngs and alembic caches are supported by the manager
+exrSpec = BlobSpecification()
```
foundrytom, line 500: [question] Is there a reason for moving away from using `Blob.Image` here, and in the rest of the examples? The idea of specifications is that they should be used as precisely as possible, in general the policy would be more subject to the use of a specific type of data (eg: renders vs shader textures), over the format itself (eg: png).

foundrytom (commented):
Nice work @carmenpinto , the new diagrams are great! Make it much easier to understand the limitations of Option 1.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Even though they can have the same names, they are different specifications because the hierarchy is involved. This could lead to confusion as `blob.geometry.sequence` is not the same as `blob.image.sequence` and not the same as `container.sequence` .  Calling it different names highlights that they are different copies of the same specification. Calling them all Sequence without the number makes it falsely seem like these are shared attributes from the same class. Mermaid will show this as causing a contradiction of each class having a single parent.

```mermaid
classDiagram
  direction TB
  Blob <-- Geometry
  Geometry <-- Sequence
  Geometry <-- Rig
  Blob <-- Image
  Sequence <-- RenderCache
  Image <-- Sequence
  Image <-- Panoramic
  Sequence <-- Shot
  Container <-- Sequence
  Sequence <-- AnimationCache
  
  class Blob {
    str mimeType
  }
  class Sequence {
    int startFrame
    int endFrame
  }

  
  class Geometry {
    str upAxis
  }
  class AnimationCache {
    int size
  }
  class Rig {
    str department
  }
  class Image {
    int channels
  }
  class RenderCache {
    int size
  }
  class Panoramic {
    float angle
  }
  class Shot {

  }
  class Container {
    bool isEmpty
  }
```

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Added comments to the classes for the first diagrams of Option 1 and Option 2 in 78c87c0

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: It's to show that they can be done similarly in some cases. BlobSpecification is pretty similar to BlobTrait in the case that the hierarchy is the same as the shared traits.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
+# DR007 Hierarchical or composition based traits for specifications
+
+- **Status:** Proposed
+- **Impact:** High
+- **Driver:** @carmenpinto
+- **Approver:** @feltech
+- **Contributors:** @foundrytom @feltech @fnRaihanKibria
+- **Outcome:** A choice will be made for the specifications to be hierarchical or composition based
```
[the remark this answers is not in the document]
carmenpinto: Added summary of outcome in e7717b77781fa9927a5dd0e23863ca706d113557

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+introspection/help) Definable out of core (by host integrations/managers)
+
+## Options considered
+
+### Option 1
+
+Hierarchical specifications with a single parent per child. We assume that every child fulfills the
+contract of its parent. For example, that all alembic files are a geometry cache, and that all
+geometry caches might have coordinate system based data, and that all geometry caches are a type of
+binary blob. This could be described in a specification: entity. blob.geometry._cache.alembic.
```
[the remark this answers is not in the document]
carmenpinto: added code quotes in e7717b77781fa9927a5dd0e23863ca706d113557

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+Another example could be an exr render, which could be a type of image, which could also be a type
+of blob binary data, represented as: entity.blob.image.exr .
```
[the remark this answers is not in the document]
carmenpinto: added code quotes in e7717b77781fa9927a5dd0e23863ca706d113557

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,361 @@
...
+
+#### Pros
+
+- only one way to represent the specification
```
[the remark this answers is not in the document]
carmenpinto: Added sentence case in da3bdfe

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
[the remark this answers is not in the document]
carmenpinto: No problem, updated in e7717b77781fa9927a5dd0e23863ca706d113557

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Yes, sounds good. Added in e7717b77781fa9927a5dd0e23863ca706d113557

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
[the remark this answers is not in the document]
carmenpinto: No problem, changed to text in : 78c87c0d7038ae6c2f948e0d7ed5bc39023d108f . Added diagrams since then as well.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Replaced with diagrams in 78c87c0d7038ae6c2f948e0d7ed5bc39023d108f . These diagrams have changed the traits available to have a sequence trait.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Removed unused traits and replaced with diagrams: 78c87c0d7038ae6c2f948e0d7ed5bc39023d108f

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Removed bundle spec and replaced with traits here: 78c87c0d7038ae6c2f948e0d7ed5bc39023d108f

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Reorganised the declarations at the top 78c87c0d7038ae6c2f948e0d7ed5bc39023d108f

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Used the places where the schema was persisted wording: e7717b77781fa9927a5dd0e23863ca706d113557

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: added code quotes in e7717b77781fa9927a5dd0e23863ca706d113557

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: added code quotes in e7717b77781fa9927a5dd0e23863ca706d113557

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: added code quotes in  e7717b77781fa9927a5dd0e23863ca706d113557

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Added cost estimation: e7717b77781fa9927a5dd0e23863ca706d113557

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
```diff
@@ -0,0 +1,277 @@
...
+
+- difficult to insert new specifications in between types, requires
+  updating the child type identifier of the new insert
+- requires knowledge of as many possible entities as possible to be
+  efficiently taxonomied into an efficient and useful hierarchy
+- sharing behaviours between taxonomically unrelated types - for example
+  a frame range applied to both a geometry cache and a shot - is
+  difficult, it relies on duplicated property definitions on each of
+  specifications, meaning that the trait of “having a frame range
+  becomes” implied though well-known property names, rather than an
```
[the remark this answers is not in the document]
carmenpinto: moved quote here: a2c158cb3155b209aaa595dd5345b402dcbb687d

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
foundrytom: I think there might be some crossed wires. In the Option 1, the "Sequence" class as its depicted in your diagram does not exist at all, and there are no copies of the specification. The  _only_ things that exist are concrete Specifications with complete schema strings such as `blob.geometry.sequence`,  `container.sequence` etc. There is no manifestation of the individual partial intermediate schema name components, _only_ parent schemas that have the same exact prefix as some part of the name. So there is no need to disambiguate - the resulting `Specification` is already uniquely distinguished by the full schema name. So when you say "calling it sequence", "it" is referring to something that doesn't actually exist in the model. 

In the class hierarchy, there is only `BlobSpecification("blob")`, `GeometrySpecification("blob.geometry")`, `GeometrySequenceSpecificaiton("blob.geomety.sequence")`, `ImageSequenceSpecification("blob.image.sequence")`. Where `GeometrySpecification` is derived from `BlobSpecification`, etc. Consequently, there is never any ambiguity as the hierarchical namespace ensures this.

This is the main reason the trait approach is a) different and b) better suited for purpose - the explicit lack of any concrete, shared `sequence` concept _in_ the hierarchical model is why it is bad, and why we want option 2. In Option 2 it only exists as an emergent phenomena though commonly named sub-strings. If you do end up uniquing the component name, it actually ends up evaporating that already tenuous link further still.

I think this might be a side effect of using a node graph diagram to represent the hierarchical model in this way. It depicts (and consequently implies) the existence of intermediate specifications as distinct entities within the model. In reality, it is only _visualisation_ of the implicit hierarchy in the resulting scheme strings themselves, not a reflection of the actual model structure. The intermediate entities in the path don't actually exist in Option 1 at all. This is what results in the diagram noted above then being misleading - to represent the actual model conceptually, it would need three `Sequence` blocks all with the same name - the schema component names would never be suffixed in the specification itself. Ie.the schema name for `GeometrySequence` it would never be `blob.geometry.sequence_2` - that implies some rule that name components are globally unique, defeating the purpose of the namespace hierarchy in the first place.

foundrytom: Using nested blocks for Option 1 would probably be a more accurate representation, where an outer block represent a parent name component?

carmenpinto: The point of the diagram is to show that `Sequence` can't be shared between branches of the hierarchy and be the same specification, not that it's the best case representation of Option 1. The first diagram is the representation of Option 1, this diagram is to show that it's not possible to share the same specification without joining the hierarchies together. I could make a diagram that illustrates the naming rules, but that wouldn't illustrate the problem of having to join the branches together as well.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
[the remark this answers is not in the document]
foundrytom: Ahh, sorry, It's meant to be a summary of what the change will be, so you don't need to read on. So something like

`OpenAssetIO will use a trait-composition based typing system instead of classical hierarchy.`

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
foundrytom: Ace thanks! I think the template needs up dating, as if we have on option 2 - we'd need one on option 1 for comparison. If effort isn't a consideration, it can be removed from both.

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
foundrytom: Capturing an off-line conversation:

> Properties of specs can be a slightly grey area. The general rule is that the schema/traits should be enough to concretely define something. Ie: `rig.hair` and `rig.mesh` rather than `rig` with a `department` property, so it's best to use the most concrete definition at all times.

carmenpinto:
I added an example of USD having to rewrite one of their hierarchical classes (typed schema) to be a trait based API one 48fc44fe68ce01d14ec583fc3d1ba62385170607

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: No problem, here it is: 70bea88

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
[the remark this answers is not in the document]
carmenpinto: Used this wording here: 70bea88

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: Removed the effort ad33b2a

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
carmenpinto: How would it be updated? I can't find the line.

foundrytom: `L464` says:

> When using management interface functions the code using specifications is the same for Option 1 and Option 2.

The code for option 2 at `L496` and `L499` still uses a base `BlobTrait`, by the "most concrete definition" principal it should also be using an image or geometry specification so that it is "the same" as the other option (at that point do we just need a single code example?).

if you want this to be a realistic example (and avoid the property debate) I'd almost suggest not setting a `mimeType` on at least the image example. In all the situations we've implemented in the past, its rare that a specific type would be queried. Taking Nuke for example, to query if image read nodes should even bother the AMS (a concrete example of where `managementPolicy` should be used), would not specify a mime type as the type of the image could me multitude, and most likely determined by the pipeline, not the host.

meshula:
I think this aspect of the specification needs some clarification about how it fits into the big picture. It talks about EXR and Alembic binary blobs, and it talks about traits for very specific things like whether something is a rig or a light.

What's not clear is what happens when a referenced asset is not a single-purpose blob. For example, a USD file containing an set with light, cameras, meshes and animated components, or that same USD file, accomplishing the same set via referenced USD files.

Similarly, an OTIO file, with media references. It's also not excluded that an OTIO file can reference another OTIO file, even though OTIO doesn't currently recursively compose such a structure itself if encountered.

The notion of frame range is also ambiguous. A USD file, or an OTIO file, may have a time warp on a referenced clip. What frame range is reported? OTIO has a notion of an availableRange, which is the time range for which requesting data about a particular asset can be answered. This has the potential to get ambiguous if layers can be predicated, for example, this composition [ a ][ b ]. What is the frame range if [ b ] is muted, or otherwise supressed? If [ b ] is a referenced file, and itself has a predicated elements, what's the frame range? In many ways the frame range is a Heisen-attribute, that isn't known unless the scene is evaluated for a specific set of predications. At least, this specification could qualify the term with something like "frame range: the frame range within which the asset can report data. The temporal coordinate system of the frame range is the inherent coordinate system of the asset itself. There may be no data returned within that range; 'no data' is a valid result. Requests outside the range are invalid."

decisions/DR007-Hierarchical-or-compositional-traits-for-specifications.md
foundrytom: The point was more that they _should_ be the same code. The use of `BlobTrait` as it is presented for Option 2 is not in the spirit of how specifications should be used.

foundrytom:
> I think this aspect of the specification needs some clarification about how it fits into the big picture. It talks about EXR and Alembic binary blobs, and it talks about traits for very specific things like whether something is a rig or a light....

@meshula an excellent question indeed. We're going to take the opportunity to answer this through better documentation. Will have a PR shortly that hopefully sheds a bit more light on this.

foundrytom:
> > I think this aspect of the specification needs some clarification about how it fits into the big picture. It talks about EXR and Alembic binary blobs, and it talks about traits for very specific things like whether something is a rig or a light....
> 
> @meshula an excellent question indeed. We're going to take the opportunity to answer this through better documentation. Will have a PR shortly that hopefully sheds a bit more light on this.

@meshula If you have a sec, have a look at #249. It projects a future based on this and #247, and a few other changes that it enabled. Does that help? There is a build version of the docs [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/suites/5449043718/artifacts/173123945) - see `entities_traits_and_specifications.html`.

foundrytom:
@carmenpinto do you want to squash these commits ready for merge?

feltech:
For my part at least, this can be squashed and merged as soon as convenient.

foundrytom:
> For my part at least, this can be squashed and merged as soon as convenient.

Plan is to:
 - Circulate this more widely in advance of the next working group.
 - Assuming no objections from the community, merge after that meeting.

foundrytom (approved):

foundrytom:
@carmenpinto is on holiday, so have rebased this to `main` so we can merge.

PR #241, opened 2022-02-21
[API] Rename logging module to log

### LINKED ISSUE #58: Rename `logging` module
`logging` is a built-in Python module that we're shadowing, which could cause confusion and hard-to-debug bugs.  

A contrived example:
```
$ python python/openassetio/SpecificationFactory.py 
...
Error in sys.excepthook:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/apport_python_hook.py", line 72, in apport_excepthook
    from apport.fileutils import likely_packaged, get_recent_crashes
...
  File "/usr/lib/python3/dist-packages/urllib3/util/retry.py", line 21, in <module>
    log = logging.getLogger(__name__)
AttributeError: module 'logging' has no attribute 'getLogger'
```
Confirmed this is caused because we shadow the `logging` namespace, by adding a `print` of `logging.__file__` inside `urllib3/util/retry.py`.

## Acceptance Criteria

- The logging module is renamed to `openassetio.log`

fnRaihanKibria:
Rename the module from logging to log and fix all references in tests and documentation.

Closes #58

### DIFF
--- a/python/openassetio/hostAPI/Session.py
+++ b/python/openassetio/hostAPI/Session.py
@@ -66,11 +66,11 @@ class Session(Debuggable):
         HostInterface is supported, so if multiple sessions are created,
         they should all use the same HostInterface instance).
 
-        @param logger openassetio.logging.LoggerInterface The target for
+        @param logger openassetio.log.LoggerInterface The target for
         all logging output from the session API-level or Manager level
         messages will all be routed through this object. No severity
         filtering is performed. Hosts wishing to filter the messages can
-        use the @ref openassetio.logging.SeverityFilter wrapper if
+        use the @ref openassetio.log.SeverityFilter wrapper if
         desired.
 
         @param managerFactory
@@ -80,7 +80,7 @@ class Session(Debuggable):
 
           @see @ref useManager
           @see @ref currentManager
-          @see @ref openassetio.logging "logging"
+          @see @ref openassetio.log "log"
           @see @ref openassetio.hostAPI.ManagerFactoryInterface
           "ManagerFactoryInterface"
           @see @ref openassetio.pluginSystem.PluginSystemManagerFactory
--- a/tests/openassetio/test_imports.py
+++ b/tests/openassetio/test_imports.py
@@ -50,8 +50,8 @@ class Test_package_imports:
     def test_importing_exceptions_succeeds(self):
         from openassetio import exceptions
 
-    def test_importing_logging_succeeds(self):
-        from openassetio import logging
+    def test_importing_log_succeeds(self):
+        from openassetio import log
 
     def test_importing_Specification_succeeds(self):
         from openassetio import Specification
--- a/doc/src/Examples.dox
+++ b/doc/src/Examples.dox
@@ -23,7 +23,7 @@
  * openassetio.hostAPI.HostInterface "HostInterface" implementation.
  *
  * @code{.py}
- * from openassetio.logging import ConsoleLogger, SeverityFilter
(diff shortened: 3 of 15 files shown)


tests/openassetio/test_imports.py
```diff
@@ -51,7 +51,7 @@ def test_importing_exceptions_succeeds(self):
         from openassetio import exceptions
 
     def test_importing_logging_succeeds(self):
```
foundrytom, line 53: [fix]
```suggestion
    def test_importing_log_succeeds(self):
```

tests/openassetio/test_logging.py
```diff
@@ -14,7 +14,7 @@
 #   limitations under the License.
 #
 """
-Tests that cover the openassetio.logging module.
+Tests that cover the openassetio.log module.
```
foundrytom, line 17: [fix] This file needs renaming to `test_log.py` to match the module-under-test.

foundrytom (changes requested):
Great job @fnRaihanKibria, just a few obscure spots that have slipped through the net. 

On a minor note, the commit component for this should be `[Core]` (see the list [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/main/contributing/COMMITS.md#appendix---code-components)).

Can we add some more context to the commit message to? We treat commit messages as light-weight decision records, so they need to describe the rationale for the change, and any other options considered etc... Some guidelines [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/main/contributing/COMMITS.md#commit-messages). 

This ensures the commit message is self contained, and doesn't rely on GitHub - we may not have access to it in the future, and it is not available off-line.  It is definitely still worth including a GitHub issue reference in the commit message body itself though, to make it easier to link back to though if we do need to.

python/openassetio/hostAPI/Session.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, hostInterface, logger, managerFactory):
 
           @see @ref useManager
           @see @ref currentManager
-          @see @ref openassetio.logging "logging"
+          @see @ref openassetio.log "logging"
```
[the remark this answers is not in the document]
fnRaihanKibria: https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/241/commits/0c4bf0e7c64285f9d31de2c8f01e711e8b2ca2bc Fixed in the two places I found.

tests/openassetio/test_imports.py
fnRaihanKibria: https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/241/commits/d511403d67ce77181ccc7be96082ca1f4382a77d Fixed

tests/openassetio/test_logging.py
fnRaihanKibria: https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/241/commits/329754435c71b5a609218b7cc5b96580c62b5c14 Fixed

fnRaihanKibria:
@foundrytom Answered all requests, could you take another look please. Regarding the final commit, I think it makes sense to squash it all into a single commit in this case? The commit message could be:

```
[Core] Rename logging module to log

The logging module was shadowing the built-in Python module of the same
name, which could cause confusion and hard-to-debug bugs.

Closes #58
```

feltech (approved):

foundrytom (approved):
Thanks @fnRaihanKibria, spot on. Do you want to squash and amend the commit message, read for merge? That updated one looks perfect. Thanks.

PR #242, opened 2022-02-22
[Docs] Update commit component tags in `COMMITS.md`

foundrytom:
Adds a few missing ones that have surfaced over recent months, and fixes the missing `s` in the `Tests` component.

### DIFF
--- a/contributing/COMMITS.md
+++ b/contributing/COMMITS.md
@@ -256,8 +256,13 @@ naming convention.
 
 These components should be used in commit message titles:
 
-- `Build` For `Makefile`, `CMake`, `setup.py`, etc.
-- `CI` For GitHub actions related work.
+- `Build` For `Makefile`, `CMake`, `setup.py`, etc. use by all builds.
+- `CI` For work solely on GitHub Actions or similar deployments.
 - `Core` For work within the core, non-interactive API.
-- `Docs` For work in the (distribution) Doxygen documentation.
-- `Test` For work on test cases and test infrastructure.
+- `Docs` For work in the (distribution) Doxygen documentation or
+  in-repository Markdown files.
+- `Examples` For work included under the `examples` directory.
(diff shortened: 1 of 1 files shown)


feltech (approved):
Looks reasonable to me.  For reference, what we've used so far is:
```
$ git log --oneline | sed -nE 's/.*\[(.+)\].*/\1/p' | sort -u
All
Build
CI
Core
Docs
git
Licensing
Lint
ManagerInterface
PluginSystem
Test
Tests
```

feltech:
I guess `Licensing` should really have been `Docs`?

> > `git`
> 
> Do you think we should keep this one in, or lump that in `Build`?

I think perhaps `CI` could cover git-related changes? Maybe `.gitignore` is more of a `Lint` though.

> 
> > `ManagerInterface`
> 
> I think this one may be too specific? Thinking that long-term there will be UI and other such things too so broad categories are useful? We can certainly look into more specific ones if it would benefit, and make the message itself shorter.

Agreed, it should have been put in a broad category.

feltech:
Also `Build` and `CI` are closely related.  I can see how they could be kept distinct, but I wouldn't mind if they were combined.

foundrytom:
> Also `Build` and `CI` are closely related. I can see how they could be kept distinct, but I wouldn't mind if they were combined.

I suppose I liked the split between "A particular application of the build process, that is technically unrelated to the source tree as used by anyone with a checkout" and "something that is fundamental to every use of the source tree".

foundrytom:
> Also `Build` and `CI` are closely related. I can see how they could be kept distinct, but I wouldn't mind if they were combined.

I added some clarification [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/242/commits/eda56cd488b8535b1ba681de32e1d8167e61bab7), how about that?

feltech (approved):

PR #244, opened 2022-02-24
[Build] Re-instate Windows tests

fnRaihanKibria:
- Installing the test Python environment works

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,15 +10,22 @@ on:
 
 jobs:
   test:
-    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.config.os }}
+    runs-on: ${{ matrix.config.os }}
     strategy:
       matrix:
         # We can't properly align to the VFX Reference Platform as this
         # requires glibc 2.17, which is older than any of the available
         # environments.
-        # TODO(DF): Enable other platforms.
-        # os: [ubuntu-20.04, windows-2019, macos-11]
-        os: [ubuntu-20.04, windows-2019]
+        # TODO(DF): Enable macos-11.
+        config:
+        - {
+            os: windows-2019,
+            vcvars: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"
+          }
+        - {
+            os: ubuntu-20.04
+          }
 
     steps:
     - uses: actions/checkout@v2
@@ -34,21 +41,21 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ~/conan
-        key: ${{ matrix.os }}-conan
+        key: ${{ matrix.config.os }}-conan
 
     - name: Install dependencies (Unix)
       if: runner.os != 'Windows'
       # Configure the system and install library dependencies via conan
       # packages.
       run: |
-        source resources/build/bootstrap-${{ matrix.os }}.sh
+        source resources/build/bootstrap-${{ matrix.config.os }}.sh
 
     - name: Install dependencies (Windows)
       if: runner.os == 'Windows'
       # Configure the system and install library dependencies via conan
       # packages.
       run: |
-        cmd /C resources\build\bootstrap-${{ matrix.os }}.bat
+        cmd /C resources\build\bootstrap-${{ matrix.config.os }}.bat
 
     - name: Configure CMake build (Unix)
       if: runner.os != 'Windows'
@@ -64,7 +71,7 @@ jobs:
     - name: Configure CMake build (Windows)
       if: runner.os == 'Windows'
       run: >
-        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        call "${{ matrix.config.vcvars }}"
         
         cmake -G Ninja -S . -B build --install-prefix %GITHUB_WORKSPACE%\dist
         --toolchain %USERPROFILE%\conan\conan_paths.cmake -DCMAKE_VERBOSE_MAKEFILE=ON
@@ -81,7 +88,7 @@ jobs:
       if: runner.os == 'Windows'
       run: |
         cd build
-        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        call "${{ matrix.config.vcvars }}"
         ctest -VV
       shell: cmd
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
         # environments.
         # TODO(DF): Enable other platforms.
         # os: [ubuntu-20.04, windows-2019, macos-11]
-        os: [ubuntu-20.04]
+        os: [ubuntu-20.04, windows-2019]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


feltech:
Looking good so far.  Would be great to update `.github/workflows/test.yml` to re-enable Windows in the `runs-on` matrix, then we should see Windows show up in the checks in this PR.  Getting a tick next to the Windows check will be a good signal that this PR is ready.

.github/workflows/test.yml
```diff
@@ -52,10 +61,26 @@ jobs:
         -DOPENASSETIO_ENABLE_SANITIZER_ADDRESS=ON 
         -DOPENASSETIO_ENABLE_SANITIZER_UNDEFINED_BEHAVIOR=ON
 
-    - name: Build, install and test
+    - name: Configure CMake build (Windows)
+      if: runner.os == 'Windows'
+      run: |
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        cmake -G Ninja -S . -B build --install-prefix %GITHUB_WORKSPACE%\dist --toolchain %USERPROFILE%\conan\conan_paths.cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DOPENASSETIO_ENABLE_TESTS=ON -DOPENASSETIO_ENABLE_SANITIZER_ADDRESS=ON -DOPENASSETIO_ENABLE_SANITIZER_UNDEFINED_BEHAVIOR=ON
```
feltech, line 68: [fix] 
* Remove `*_SANITIZER_*` variables, they're unix-only.
* Would be good to line-wrap if possible.

src/openassetio-core/CMakeLists.txt
```diff
@@ -36,6 +36,13 @@ install(TARGETS openassetio-core EXPORT ${PROJECT_NAME}_EXPORTED_TARGETS)
 
 # TODO(DF): Generate and install .pdb on Windows. See OCIO.
```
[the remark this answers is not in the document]
feltech: When I looked at the CI output, it looked like this may already be done (I saw some reference in the command line to a .pdb).  So perhaps this TODO can simply be removed (should be evident from a local Windows build)?

feltech (commented):
Good stuff! I think `USES_TERMINAL` is reasonable. Just needs a bit of tidying up and I think this PR is good to go.

.github/workflows/test.yml
fnRaihanKibria: Fixed with https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/244/commits/a7c407c8f47bad76a4174cb319c3d1d502208e90 and https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/244/commits/c0d664c202ca31e383ddf68784f0b6c48abed789

src/openassetio-core/CMakeLists.txt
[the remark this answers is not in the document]
fnRaihanKibria: Installing pdbs with https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/244/commits/318b11dd272f399cf3745c1c40ca8ee603e13478

resources/build/conanfile.txt
```diff
@@ -10,3 +10,6 @@ cmake_paths
 # public ConanCenter packages (e.g. pybind11) since they disallow
 # bundling of `<PackageName>Config.cmake`-like files in the package.
 cmake_find_package
+
+[options]
+cpython:shared=True # Doesn't build properly on Windows without it. See: https://github.com/conan-io/conan-center-index/issues/9333
```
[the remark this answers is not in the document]
fnRaihanKibria: Fixed with https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/244/commits/a07581895f9b763895121701c4f3a5bf3fe0441f

.github/workflows/test.yml
```diff
@@ -52,10 +61,29 @@ jobs:
         -DOPENASSETIO_ENABLE_SANITIZER_ADDRESS=ON 
         -DOPENASSETIO_ENABLE_SANITIZER_UNDEFINED_BEHAVIOR=ON
 
-    - name: Build, install and test
+    - name: Configure CMake build (Windows)
+      if: runner.os == 'Windows'
+      run: >
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
```
fnRaihanKibria, line 67: This actually isn't great since the path depends on the runner version. This will be different for a new Windows/compiler version. I saw some people managed to store it inside the matrix but I don't know YAML well enough right now to fix it straight away.

.github/workflows/test.yml
```diff
@@ -52,10 +61,29 @@ jobs:
...
+      if: runner.os != 'Windows'
+      run: |
+        cd build
+        ctest -VV
+
+    - name: Build, install and test (Windows)
+      if: runner.os == 'Windows'
       run: |
         cd build
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
```
fnRaihanKibria, line 84: Also the path string is repeated.

.github/workflows/test.yml
feltech: The matrix stuff and the reference to it, e.g. `${{ matrix.os }}`, isn't YAML syntax, it's Github Action specific, so I would refer you to their docs.

However, there is also a way in YAML to avoid repetition - anchors and aliases, e.g. see https://www.educative.io/blog/advanced-yaml-syntax-cheatsheet

feltech: Perhaps under the matrix we can add a new variable under `matrix:`, e.g.

```
console_preamble: 
  # Ubuntu
  - ""
  # Windows
  - call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
```
Then add a `${{ matrix.console_preamble }}` line to the `run` blocks.

Does seem a bit of a waste having to have a blank Ubuntu line.  Surely there's a way in Github Actions to set a common variable outside the matrix, but I haven't looked.

You could perhaps add even more matrix parameters to parametrise the cmake command-line differences, avoiding the need for `if: runner.os`.  But that might not be possible, or be more hassle than it's worth.

OTIO's Github Actions may be a good reference.

src/openassetio-core/CMakeLists.txt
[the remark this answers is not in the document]
feltech: Awesome! Obviously this isn't tested by CI but looks reasonable to me if you're happy it works OK.

.github/workflows/test.yml
feltech: Interesting usage of `>`, I didn't realise that you can break up the line wrapping by adding a blank line, really nice.

.github/workflows/test.yml
foundrytom: Btw - it's totally fine if the windows setup is a bit "hard coded" at this point, we can't sink too much time into it at this point. Just having a working version is good for now, even if it breaks in a month or so.

foundrytom:
[question] No need to address this now, but given the number of windows specific steps, would it perhaps make more sense just to have a separate windows `.yml` file, rather than trying to shoehorn it into the matrix?

foundrytom:
Nice work @fnRaihanKibria! Great to have this working now :) don't wait for me to merge this once @feltech is happy.

fnRaihanKibria:
> [question] No need to address this now, but given the number of windows specific steps in CI, would it perhaps make more sense just to have a separate windows `.yml` file, rather than trying to shoehorn it into the matrix?

We might be able to clean this up as we go along and learn more YAML/Github features.

fnRaihanKibria:
@feltech Commits have been squashed.

feltech (approved):

Issue #247, opened 2022-02-25
Is `resolveEntityReference` still relevant?


foundrytom:
## What

Decide if `manager.resolveEntityReference(ref)` is meaningful or it could simply be `manager.entityData(ref, BlobTrait)`, where `BlobTrait ` defines a `url` property.

## Why

#238 resolves one of the lingering messy parts of the current API design. The spec hierarchy precluded the ability to define concrete, well known behaviour that was shared across branches in the spec tree. This meant Specifications could never be used to reliably define any particular aspect of an entity in isolation. This predicated methods such as `resolveEntityReference` in order to formalise a common shared behaviour of "resolving to a string".

Being able to concretely ask for specific data from an entity simplifies many aspects of the conceptual model, and removes the baked-in concept of a string-to-string mapping as a core premise. 

This already caused us issues with "non file based entities" where there may be no specific primary string. The name "primary string" itself shows the problem - that is, it is unspecific and leaves a lot to interpretation. This all goes away.

This could have performance implications though for the primary (current) use case of resolving paths? There should be no reason not to intern all spec/trait magic strings though I think?

If we do this, then it most certainly means that #31 should be `manager.entityTraits(ref, context)`, that returns the collection of trait names that the entity supports.

This change has the additional benefit that it encourages a formal definition of any well-known traits that hosts or managers may wish to use within the system, something lost with the old attributes free-for-all bucket.

### Potential downsides

A potential issue with this approach of a host specifying it's intended usage by requesting a specific trait is that you would need a "default" trait that a generic resolver with no knowledge of usage could use to 'resolve a ref to any old string'. 

This raises the question though - is the idea of "blind resolving" compatible with the overarching principal in this API of providing contextual information about the use of the data?

If such a behaviour was required, one way to achieve this could be to simply define a `string` trait, with a `value` property, and use `entityData(ref, StringTrait)`, which effectively models the behaviour of `resolveEntityReference`:

```python
def resolveEntityReference(self, ref, context):
    return StringTrait(self.entityData(ref, StringTrait.name, context)).getValue()
```

## ACs

- A decision on whether this is a sensible change.
- Documentation of the proposed API changes.
- A qualitative performance comparison against some naive manager implementation, that has the data for a `BlobTrait`  in memory that shows, the comparative cost of resolving 1,000,000 unique references in terms of the differences in:
   - Total wall-clock time.
[...]

PR #253, opened 2022-03-01
[Docs] Add decision record to unify entity data model

foundrytom:
#247 raises an interesting question. This PR presents a decision record to move to a unified entity data model for peer review.

We still need to quantify any concerns around changes to the low-level performance of the old `resolveEntityReference` method that this could introduce. The plan is to solicit feedback from community members as to representative numbers around time spent in different aspects of a typical resolution lookup, and compare these to the API-level overhead involved in bridging the respective data types of each approach.

The pretty view of this DR is available [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/48e5857045d84b745d48a00b64cb23a48212432d/decisions/DR008-Unify-the-entity-data-model.md).

### DIFF
--- a/decisions/DR008-Unify-the-entity-data-model.md
+++ b/decisions/DR008-Unify-the-entity-data-model.md
@@ -1,0 +1,361 @@
+# DR008 Unify the entity data model
+
+-   **Status:** Proposed
+-   **Impact:** High
+-   **Driver:** @foundrytom
+-   **Approver:** @carmenpinto @feltech
+-   **Outcome:** The existing `getEntityAttributes` and
+    `resolveEntityReference` methods will be combined into a trait-centric
+    resolution mechanism.
+
+## Background
+
+When a host application (DCC tool, pipeline script, etc.) wishes to
+retrieve data from one of a manager's entities (assets), it has two
+options:
+
+1. The "primary string" can be resolved.
+2. Arbitrary key-value pairs of "attributes" can be queried.
+
+The primary string is thought of as "the string the host would have
+had if there were no asset management system", and attributes are
+"bonus data that may be useful".
+
+The reason for this is the origins of the API. It is based on the Katana
+Asset API, where it is predominantly used to resolve strings, the
+majority of which being file paths.
+
+OpenAssetIO was forked to solve a wider variety of workflows, including
+editorial processes. This broadened the scope of what kinds of entity
+needed to be handled by the API, and consequently necessitated the
+ability to store and recall more than just single strings.
+
+If we consider a "shot" (a series of frames covered by a single camera
+angle in a film or tv show), the concept of a primary string is somewhat
+meaningless. A shot may have a name _and_ a frame range, where the frame
+range is actually the most important part.
+
+To provide a working solution to this problem, OpenAssetIO gained the
+attributes mechanism as it stands today, and the specification
+classification hierarchy to help everyone work out what "kind of thing"
+any given entity was.
+
+Upon this, a somewhat fragile scaffolding of conventions grew. It was
+expected that certain specifications (or an element of their parent
+hierarchy) had certain attributes, and perhaps more comically, that for
+some specifications of entity, the primary string was "largely
+irrelevant".
+
+In short, the hard partitioning between the primary string and
+attributes, established at the APIs origin was becoming less suited to
+the breadth of scenarios encountered today. The hierarchical
+specifications mechanism was a reaction to this, but one that in itself
+was inherently flawed.
+
+The the recent switch of the specifications mechanism from a
+hierarchical taxonomy to one that composes several distinct traits (see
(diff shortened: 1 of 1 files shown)


decisions/DR008-Combine-getEntityAttributes-and-resovle.md
```diff
@@ -0,0 +1,329 @@
...
+### Generic resolution
+
+Some generic code wishes to 'resolve' an entity to its primary string
+without any understanding of its specific usage as simple wrapper around
+any generic string parameter.
+
+#### Option 1
+
+The manager is already expected to figure out what to return as a
+primary string based type of entity it is within it's domain (or, what
```
carmenpinto, line 311: [fix] minor typo
```suggestion
primary string based type of entity it is within its domain (or, what
```

decisions/DR008-Combine-getEntityAttributes-and-resovle.md
```diff
@@ -0,0 +1,329 @@
...
+Note the asymmetry to `register`, only the `ImageTrait` is needed for
+lookup as otherwise if `ImageSpecification` was used, it would also
+request the already accessed `url` from the blob, which we "know" we
+already have through the primary string. Maybe we should have omit
+resolve and just fetched it all via attributes?
+
+#### Option 2
+
+```python
+data = manager.resolve(img_ref, ImageSpec.traitNames, context)
```
fnRaihanKibria, line 291: I think the proposal looks good overall. I was only wondering about corner cases where someone might want to do a bulk call for resolving many different refs (and only cares about the returned path), which may be cheaper if the manager doesn't have to create the full set of attributes for all of them. I expect though this is rare and the amount of data involved is small, so it's unlikely to be a problem.

foundrytom: Cool, thanks.  That situation is definitely a concern. The closest you could get with this is the single trait use below, which then is limited to just the (available) properties of the `BlobTriat` which would be the path part. You could request this for any entity that has that trait.

```python
path = BlobTrait(manager.resolve(image_refs, [BlobTrait.name], context)).getUrl()
```

decisions/DR008-Unify-the-entity-data-model.md
```diff
@@ -0,0 +1,358 @@
...
+path = manager.resolveEntityReference(img_ref, context)
+attr = manager.getEntityAttributes(img_ref, ImageTrait.attrNames, context)
+
+color_space = ImageTrait(attr).getColorSpace()
+```
+
+Note the asymmetry to `register`, only the `ImageTrait` is needed for
+lookup as otherwise if `ImageSpecification` was used, it would also
+request the already accessed `url` from the blob, which we "know" we
+already have through the primary string. Maybe we should have omit
```
feltech, line 314: [fix]
```suggestion
already have through the primary string. Maybe we should have omitted
```

feltech (changes requested):

decisions/DR008-Combine-getEntityAttributes-and-resovle.md
fnRaihanKibria: Seems reasonable I think. The cost of communication inside the AMS will likely outweigh any of the OpenAssetIO code.

decisions/DR008-Unify-the-entity-data-model.md
```diff
@@ -0,0 +1,358 @@
...
+    API not only to determine where to put the images, but what
+    colorspace, file format, bit depth and compression settings to use
+
+-   A primary design goal for OpenAssetIO is clarity. There should be as
+    little ambiguity as its usage or the expected behavior of a host or
+    manager as possible.
+
+-   The current API as implemented suggests that the lifetime of
+    attributes and the primary string are different. The primary string
+    can only be set during `register`, where as attributes can be
```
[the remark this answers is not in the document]
foundrytom: GitHub being silly, this is fixed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/253/commits/79986bad2498784db1248aec41382bc69beb3b3e).

decisions/DR008-Unify-the-entity-data-model.md
```diff
@@ -0,0 +1,358 @@
...
+}
+```
+
+### Publishing an image
+
+This example illustrates a host wanting to publish an image with its
+associated color space, to ensure it is accurately tracked.
+
+```python
+img = ImageSpec()
```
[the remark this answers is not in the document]
foundrytom: Do you mind leaving this as-is? I quite liked the fact that it tied in more concretely with the declaration of the spec in the preamble, reenforcing the "robustness", even if it is at the expense of accuracy to the case we performance tested. (It would be better right, as the trait IDs would be in a static vector rather than on the stack?)

feltech (approved):

Issue #250, opened 2022-02-28
`Specification`/`Trait` `C++` implementation


foundrytom:
## What

A `C++` only implementation of #245.

## ACs

- `C++` headers + implementation for the data layer, that considers requirements of a `C` API in the next step:
   - The `Specification` mechanism
   - The `Trait` mechanism
- Manually written python bindings for the data layer
- Manually written concrete C++ traits and specifications for:
   - `entity`, `relationship` and `locale` traits (no properties)
   - `blob` trait: `url` and `mimeType` (str) properties
  - Manually written python bindings to concrete traits/specs
- Testing:
   - `BlobSpecificaiton` specification that composes `entity` and `blob` traits for testing purposes.
   - `pytest` coverage of the implementation + bindings including python-defined test spec/traits, that incorporate a C++ defined trait.
- The API would support the workflows sketched in #226 [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/issues/226#issuecomment-1047993384), and in #247, and whatever the outcome of #246 is.

## Out of scope

- Autogen. This is for later, this task is to design the code that will could be autogenerated.
- Replacing any existing use of `Specification` with in the code base.
- Updating documentation outside of the in-code Doxygen documentation for the new code.
- An implementation of the C API
- Map value type for trait data

PR #255, opened 2022-03-02
[Core] Add specification/trait C++ implementation

feltech:
For #250 we require the C++ infrastructure to support specifications following the new design of #238 et al. 

This PR introduces an initial draft of specifications as dictionary-like data structures and traits as views on a specification, coded in C++ with Python bindings, and tested exclusively through the Python bindings.

Initial performance tests using #270 and reported in #253 lead us to believe the performance of this dictionary-like design is unlikely to be a serious bottleneck.

The Doxygen config is updated to render docs for the new C++ classes.

### DIFF
--- a/src/openassetio-core/include/openassetio/specification/Specification.hpp
+++ b/src/openassetio-core/include/openassetio/specification/Specification.hpp
@@ -1,0 +1,127 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2022 The Foundry Visionmongers Ltd
+/**
+ * Provide the base dynamic specification class.
+ */
+#pragma once
+
+#include <memory>
+#include <vector>
+
+#include <openassetio/export.h>
+
+#include "../trait/property.hpp"
+
+namespace openassetio {
+inline namespace OPENASSETIO_VERSION {
+/**
+ * Comprises the base Specification class and concrete classes of core
+ * specifications derived from it.
+ */
+namespace specification {
+
+/**
+ * Structure for data exchange between a @ref host and a @ref
+ * manager.
+ *
+ * A specification is logically a set of supported @needsref trait
+ * "traits", each identified by a unique string, plus optional
+ * key-value properties associated with each of those traits.
+ *
+ * Trait @ref trait::property::Key "property keys" are always strings.
+ * Property values are strings, integers, floating point, or booleans.
+ * Any of a trait's properties can be legitimately left unset - it is up
+ * to the consumer (host or manager, depending on the API method) to
+ * decide how this should be handled.
+ *
+ * @todo Add SimpleMap trait property value type.
+ *
+ * @see trait::property
+ *
+ * Various API methods require a populated specification to be provided
+ * by the host, which the manager can interrogate in order to determine
+ * the correct response.
+ *
+ * Conversely, various API methods, in particular @needsref
+ * ManagerInterface::resolve, require the manager to return a
+ * populated specification to the host. The traits (and hence their
+ * properties) contained within the returned specification are
+ * determined by the intersection of the traits that were requested by
+ * the host and the traits that the manager supports.
+ *
+ * Since specifications are generic dictionary-like data structures,
+ * accurate data access/mutation relies on well-known trait IDs and
+ * property names. This introduces a possible avenue for user error due
+ * to misspelling, as well as difficulty in discovering what properties
+ * may be available for a given trait.
+ *
+ * Therefore, it is strongly advised that accessing and mutating trait
+ * properties is performed using trait view wrapper classes wherever
+ * possible, rather than directly using the accessor/mutator functions
+ * on the specification.
+ *
+ * @see trait::TraitBase
+ *
+ */
+class OPENASSETIO_CORE_EXPORT Specification {
+ public:
+  /// List of supported trait IDs.
+  using TraitIds = std::vector<trait::TraitId>;
+
+  /**
+   * Construct such that this specification supports the given list of
+   * trait IDs.
+   *
+   * @param traitIds List of IDs of traits that this specification
+   * supports.
+   */
+  explicit Specification(const TraitIds& traitIds);
+
+  /**
+   * Defaulted virtual destructor establishing this as a base class
+   * for more concrete specifications.
+   */
+  virtual ~Specification();
+
+  /**
+   * Return whether this specification supports the given trait.
+   *
+   * @param traitId ID of trait to check for support.
+   * @return `true` if trait is supported, `false` otherwise.
+   */
+  [[nodiscard]] bool hasTrait(const trait::TraitId& traitId) const;
+
+  /**
+   * Get the value of a given trait property, if the property has
+   * been set.
+   *
+   * @param[out] out Storage fo result, only written to if the property
+   * is set.
+   * @param traitId ID of trait to query.
+   * @param propertyKey Key of trait's property to query.
+   * @return `true` if value was found, `false` if it is unset.
+   * @exception `std::out_of_range` if the trait is not supported by
+   * this specification.
+   */
+  [[nodiscard]] bool getTraitProperty(trait::property::Value* out, const trait::TraitId& traitId,
+                                      const trait::property::Key& propertyKey) const;
+
+  /**
+   * Set the value of given trait property.
+   *
+   * @param traitId ID of trait to update.
+   * @param propertyKey Key of property to set.
+   * @param propertyValue Value to set.
+   * @exception `std::out_of_range` if the trait is not supported by
+   * this specification.
+   */
+  void setTraitProperty(const trait::TraitId& traitId, const trait::property::Key& propertyKey,
+                        trait::property::Value propertyValue);
+
+ private:
+  class Impl;
+  std::unique_ptr<Impl> impl_;
+};
+}  // namespace specification
+}  // namespace OPENASSETIO_VERSION
+}  // namespace openassetio
--- a/src/openassetio-core/include/openassetio/trait/BlobTrait.hpp
+++ b/src/openassetio-core/include/openassetio/trait/BlobTrait.hpp
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2022 The Foundry Visionmongers Ltd
+/**
+ * Define the core `BlobTrait` trait class.
+ */
+#pragma once
+
+#include <openassetio/export.h>
+
+#include "TraitBase.hpp"
+
+#include "property.hpp"
+
+namespace openassetio {
+inline namespace OPENASSETIO_VERSION {
+namespace trait {
+
+/**
+ * Core trait class representing a locatable blob of data of a specific
+ * type.
+ *
+ * Has the ID of "blob" and defines "url" and "mimeType" properties.
+ *
+ * This core trait defines one of the most common set of properties in
+ * the specification::Specification of an @ref entity - specifically, a
+ * URL and data (MIME) type.
+ *
+ * For example, the URL can be a local file system path (i.e.
+ * `file:///`). Thus the common workflow of @ref resolve "resolving" an
+ * @ref entity_reference to a file path is likely to involve the host
+ * requesting, and the manager returning, a @ref
+ * specification::Specification "Specification" supporting this
+ * `BlobTrait`.
+ *
+ * @see TraitBase
+ * @see specification::Specification
+ */
+struct OPENASSETIO_CORE_EXPORT BlobTrait : TraitBase<BlobTrait> {
+  /// ID of this trait.
+  static inline const TraitId kId{"blob"};
+
+  /// Hoist base class constructor.
+  using TraitBase<BlobTrait>::TraitBase;
+
+  /**
+   * Get the URL property for this trait from the wrapped specification.
+   *
+   * @param[out] out Storage for value, if property is set.
+   * @return `true` if property is set, `false` otherwise.
+   * @exception `std::out_of_range` if this trait is not supported by
+   * the wrapped specification.
+   */
+  [[nodiscard]] TraitPropertyStatus getUrl(property::Str* out) const;
+
+  /**
+   * Set the URL property for this trait in the wrapped specification.
+   *
+   * @param url URL value to set.
+   * @exception `std::out_of_range` if this trait is not supported by
+   * the wrapped specification.
+   */
+  void setUrl(property::Str url);
+
+  /**
+   * Get the mime type property for this trait from the wrapped
+   * specification.
+   *
+   * @param[out] out Storage for value, if property is set.
+   * @return `true` if property is set, `false` otherwise.
+   * @exception `std::out_of_range` if this trait is not supported by
+   * the wrapped specification.
+   */
+  [[nodiscard]] TraitPropertyStatus getMimeType(property::Str* out) const;
+
+  /**
+   * Set the mime type property for this trait in the wrapped
+   * specification.
+   *
+   * @param mimeType Mime type value to set.
+   * @exception `std::out_of_range` if this trait is not supported by
+   * the wrapped specification.
+   */
+  void setMimeType(property::Str mimeType);
+};
+}  // namespace trait
+}  // namespace OPENASSETIO_VERSION
+}  // namespace openassetio
--- a/src/openassetio-core/include/openassetio/trait/TraitBase.hpp
+++ b/src/openassetio-core/include/openassetio/trait/TraitBase.hpp
@@ -1,0 +1,143 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2022 The Foundry Visionmongers Ltd
+/**
+ * Base class for all specification traits.
+ */
+#pragma once
+#include <algorithm>
+#include <memory>
+#include <utility>
+
+#include "../specification/Specification.hpp"
+
+namespace openassetio {
+inline namespace OPENASSETIO_VERSION {
+namespace trait {
+
+/**
+ * Abstract CRTP base class for specification traits.
+ *
+ * A trait view provides a way to hide the underlying dictionary-like
+ * data access from hosts and managers. Trait view classes wrap a
+ * @ref specification::Specification "Specification" and provide member
+ * functions that query/mutate properties on the specification.
+ *
+ * @see BlobTrait
+ *
+ * This non-virtual templated base class provides the common interface
+ * for a concrete trait view.
+ *
+ * As an example, assume a trait view called `MyTrait` and an arbitrary
+ * specification. Before we can extract `MyTrait` property values from
+ * the specification we must check that it supports `MyTrait`. We can
+ * then use the trait's concrete accessors to retrieve data from the
+ * underlying dictionary-like specification. Usage may thus look
+ * something like
+ *
+ *     property::Int myValue = 123;  // Default if property not found.
+ *
+ *     MyTrait myTrait{specification};
+ *
+ *     if (myTrait.isValid()) {
+ *
+ *       if (myTrait.getMyValue(&myValue) != TraitPropertyStatus::kFound) {
+ *
+ *         std::cerr << "Warning: myValue not found\n";
+ *       }
+ *     }
+ *
+ * A trait class inheriting from this base must have a
+ * `static inline TraitId kId` member giving the unique string ID of
+ * that trait.
+ *
+ * @see TraitId
+ *
+ * @warning The `kId` member must be `inline` to support @ref
+ * specification::Specification "specifications" that statically
+ * construct their list of supported trait IDs. This is to avoid the
+ * so-called _Static Initialization Order Fiasco_, when a
+ * specification's static trait list is constructed using static trait
+ * IDs.
+ *
+ * In addition, the derived class should implement appropriate typed
+ * accessor / mutator methods that internally call the wrapped
+ * specification's @ref specification::Specification::getTraitProperty
+ * "getTraitProperty" / @ref
+ * specification::Specification::setTraitProperty "setTraitProperty".
+ *
+ * Such accessor/mutator functions then provide developers with
+ * compile-time checks and IDE code-completion, which would not be
+ * available with arbitrary string-based lookups.
+ *
+ * @note Attempting to access a trait's properties without first
+ * ensuring the specification supports that trait via `isValid`, or
+ * otherwise, may trigger a `std::out_of_range` exception if the trait
+ * is not supported by the specification.
+ *
+ * @tparam Derived Concrete subclass.
+ */
+template <class Derived>
+struct TraitBase {
+  /// Ref-counted smart pointer to underlying Specification.
+  using SpecificationPtr = std::shared_ptr<specification::Specification>;
+
+  /**
+   * Construct this trait view, wrapping the given specification.
+   *
+   * @param specification Underlying specification to wrap.
+   */
+  explicit TraitBase(SpecificationPtr specification) : specification_{std::move(specification)} {}
+
+  /**
+   * Check whether the specification this trait has been applied to
+   * actually supports this trait.
+   *
+   * @return `true` if the underlying specification supports this trait,
+   * `false` otherwise.
+   **/
+  [[nodiscard]] bool isValid() const { return specification_->hasTrait(Derived::kId); }
+
+ protected:
+  /**
+   * Get the underlying specification that this trait is wrapping.
+   *
+   * @return Wrapped Specification.
+   */
+  SpecificationPtr& specification() { return specification_; }
+
+  /**
+   * Get the underlying specification that this trait is wrapping.
+   *
+   * @return Wrapped Specification.
+   */
+  [[nodiscard]] const SpecificationPtr& specification() const { return specification_; }
+
+  /**
+   * Convenience typed accessor to properties in the underlying
+   * specification.
+   *
+   * @tparam T Type to extract from variant property.
+   * @param[out] out Storage for value, if property is set.
+   * @param traitId ID of trait to query.
+   * @param propertyKey Key of property to query.
+   * @return Status of property in the specification.
+   */
+  template <class T>
+  [[nodiscard]] TraitPropertyStatus getTraitProperty(T* out, const TraitId& traitId,
+                                                     const property::Key& propertyKey) const {
+    if (property::Value value; specification()->getTraitProperty(&value, traitId, propertyKey)) {
+      if (T* maybeOut = std::get_if<T>(&value)) {
+        *out = *maybeOut;
+        return TraitPropertyStatus::kFound;
+      }
+      return TraitPropertyStatus::kInvalidValue;
+    }
+    return TraitPropertyStatus::kMissing;
+  }
+
+ private:
+  SpecificationPtr specification_;
+};
+}  // namespace trait
+}  // namespace OPENASSETIO_VERSION
+}  // namespace openassetio
--- a/tests/openassetio/test_traits.py
+++ b/tests/openassetio/test_traits.py
@@ -1,0 +1,190 @@
+"""
+Tests for the traits type system.
+"""
+# pylint: disable=invalid-name,missing-class-docstring
+# pylint: disable=redefined-outer-name,no-self-use
+# pylint: disable=missing-function-docstring
+import pytest
+# TODO(DF): @pylint - re-enable once Python dev vs. install mess sorted.
+# pylint: disable=no-name-in-module
+from openassetio import specification, trait
+
+
+class Test_Specification_hasTrait:
+    def test_when_has_trait_then_returns_true(self, a_specification):
+        assert a_specification.hasTrait("second_trait")
+
+    def test_when_not_has_trait_then_returns_false(self, a_specification):
+        assert not a_specification.hasTrait("unknown_trait")
+
+
+class Test_Specification_getsetTraitProperty:
+    def test_valid_values(self, a_specification):
+        a_specification.setTraitProperty("first_trait", "a string", "string")
+        a_specification.setTraitProperty("second_trait", "an int", 1)
+        a_specification.setTraitProperty("first_trait", "a float", 1.0)
+        a_specification.setTraitProperty("second_trait", "a bool", True)
+
+        assert a_specification.getTraitProperty("first_trait", "a string") == "string"
+        assert isinstance(a_specification.getTraitProperty("first_trait", "a string"), str)
+        assert a_specification.getTraitProperty("second_trait", "an int") == 1
+        assert isinstance(a_specification.getTraitProperty("second_trait", "an int"), int)
+        assert a_specification.getTraitProperty("first_trait", "a float") == 1.0
+        assert isinstance(a_specification.getTraitProperty("first_trait", "a float"), float)
+        assert a_specification.getTraitProperty("second_trait", "a bool") is True
+
+    def test_when_key_is_not_found_then_get_returns_None(self, a_specification):
+        assert a_specification.getTraitProperty("first_trait", "a string") is None
+
+    def test_when_trait_is_not_found_then_IndexError_raised(self, a_specification):
+        with pytest.raises(IndexError):
+            a_specification.setTraitProperty("unknown_trait", "a string", "string")
+
+        with pytest.raises(IndexError):
+            _ = a_specification.getTraitProperty("unknown_trait", "a string")
+
+    def test_when_value_is_not_supported_then_set_raises_TypeError(self, a_specification):
+        with pytest.raises(TypeError):
+            a_specification.setTraitProperty("first_trait", "unknown type", object())
+
+        with pytest.raises(TypeError):
+            a_specification.setTraitProperty("first_trait", "unknown type", None)
+
+
+def test_BlobTrait_traitId():
+    assert trait.BlobTrait.kId == "blob"
+
+
+class Test_BlobTrait_isValid:
+    def test_when_wrapping_BlobSpecification_then_returns_true(self, a_blob_specification):
+        assert trait.BlobTrait(a_blob_specification).isValid()
+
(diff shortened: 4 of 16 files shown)


src/openassetio-core/include/openassetio/specifications.hpp
```diff
@@ -0,0 +1,137 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2022 The Foundry Visionmongers Ltd
+#pragma once
+
+#include <memory>
+#include <string>
+#include <unordered_map>
```
foundrytom, line 7: Capturing off-line discussion:

Work out strategy to make sure we don't leak the storage implementation.

src/openassetio-core/include/openassetio/specifications.hpp
```diff
@@ -0,0 +1,137 @@
...
+#include <utility>
+#include <variant>
+#include <vector>
+// API export generated header
+#include <openassetio/export.h>
+
```
foundrytom, line 13: [fix] Add High-level architectural manifesto of what is in this file before too much implementation happens 📜

src/openassetio-core/include/openassetio/specification/Specification.hpp
```diff
@@ -0,0 +1,42 @@
...
+class OPENASSETIO_CORE_EXPORT Specification {
+ public:
+  using TraitIds = std::vector<trait::TraitId>;
+
+  explicit Specification(const TraitIds& traitIds);
+  virtual ~Specification();
+
+  [[nodiscard]] bool hasTrait(const trait::TraitId& traitId) const;
+
+  [[nodiscard]] bool getTraitProperty(trait::property::Value* out, const trait::TraitId& traitId,
```
foundrytom, line 30: [question] Do we want to consider typed accessors here?

src/openassetio-core/include/openassetio/specification/trait/TraitBase.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+ protected:
+  SpecificationPtr& spec() { return spec_; }
+  [[nodiscard]] const SpecificationPtr& spec() const { return spec_; }
+
+  template <class T>
+  [[nodiscard]] bool getTraitProperty(T* out, const TraitId& traitId,
+                                      const property::Key& propertyKey) const {
+    if (property::Value value; spec()->getTraitProperty(&value, traitId, propertyKey)) {
+      // TODO(DF): Make exception message more friendly.
```
foundrytom, line 39: [question] Do we want to `return false` instead of excepting, if the goal of the bool/out param was to avoid exceptions in this API? Otherwise, you need both wrappers for most typical get scenarios, as missing or mistyped are potentially equally likely?

foundrytom (commented):
WIP, but some initial Qs

python/openassetio/specification/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+"""
+Entity type system and property transport mechanism.
+"""
+__all__ = ["BlobSpecification", "Specification"]
+
+from openassetio._openassetio.specification import BlobSpecification, Specification
```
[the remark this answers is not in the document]
feltech: This came from an annoyance with my IDE, which is also now causing problems with `pylint` for a similar reason.  

The `python/openassetio` source directory doesn't contain `_openassetio.so` (since it's installed into the venv), so relative imports fail, tripping up my IDE code completion.  I attempted to work around this by making it an absolute import.  So I made it absolute just for selfish reasons and forgot about it.  

However, now with `pylint` the same problem occurs. Linting fails because of unrecognized imports, even if it's executed from the venv where `_openassetio.so` is installed.  This is apparently because the source vs. installed `openassetio` package names conflict (as evidenced by the fact that `pylint` on the `test` directory succeeds).

Placing `_openassetio.so` alongside, rather than within, the `openassetio` package would make dev life much easier on several fronts.  E.g. we may be able to re-introduce `--editable` installs for running tests, speeding up the TDD cycle.

I understand that we don't want to structure our package based on what's convenient for dev, though. 

We could run `pylint` on the installed package, working around that problem, but IDE code completion would still suffer.

We could "install" the Python extension module to the source directory, so that relative imports work in IDE/pylint, then add it as an installable artefact in `setup.py`, which may be the correct thing to do anyway (I don't know much about `setup.py` and extension modules).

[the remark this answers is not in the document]
feltech: In the meantime we can switch to absolute imports and disable the pylint checks as a quick fix rather than worrying about it in this PR.  But I would really like to solve this problem sooner rather than later - having good code completion (and thorough linting) is extremely useful.

src/openassetio-core/include/openassetio/specification/trait/TraitBase.hpp
feltech: I'm not sure about this. I didn't view the `bool` as a way to avoid all exceptions, but rather that "unset" is not an exceptional case, and is certainly not an error.

Conversely, I think a bad `T` is an exceptional error case, especially as users should be using well-behaved pre-agreed trait wrappers rather than the low-level arbitrary-dict API.  

I'm also not sure that it's a good idea to conflate "not set" with "set, but with an unexpected value type", with no way to know which.

But maybe there is a counterexample where such a case is valid and not an error?

src/openassetio-core/include/openassetio/specification/Specification.hpp
feltech: We can move the typed accessor from `TraitBase` to here.  I was in two minds where it belongs. 

I guess I viewed `getTraitProperty` as an advanced method that normal users would never call, since they should be using trait views, so didn't want to be too friendly.

The wrapper typed accessor in `TraitBase` is also `protected`, reducing the public API surface area, whereas if we move it here it'll have to be `public`.

Do you think it's worth adding as part of the public interface?

python/openassetio/specification/trait/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+"""
+Trait classes providing a concrete view on specification properties.
+"""
+__all__ = ["BlobTrait"]
+
+from openassetio._openassetio.specification.trait import BlobTrait
```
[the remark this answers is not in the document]
feltech: I don't have a strong opinion on this either way.  I thought in terms of specifications composing traits, so a parent-child kind-of thing.  But as you say, that's not really true in practice.  Maybe a peer relationship does make more sense.

python/openassetio/specification/__init__.py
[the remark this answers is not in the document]
foundrytom: Oh man, that sounds like a _load_ of fun. Lets scheme on this sooner rather than later.

python/openassetio/specification/trait/__init__.py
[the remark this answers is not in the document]
foundrytom: I think I leant towards peer on the basis of `managementPolicy` and `entityTraits`

src/openassetio-core/include/openassetio/specification/trait/TraitBase.hpp
foundrytom: Jotting down an off-line discussion:

The main thing we need to figure out is that is incorrect type an "exceptional" case, or does its very possibility force consumers to always defensively code around it, mitigating some of the semantic value brought by moving `InvalidKey` to the bool-outptr pattern.

src/openassetio-core/include/openassetio/specification/Specification.hpp
foundrytom: Capturing off-line discussion:

For now, the main choice is: do we leak `std::variant` in the public API?

src/openassetio-core/include/openassetio/specification/trait/TraitBase.hpp
feltech: Related to this: the current implementation will also throw if `traitId` doesn't exist in the dict (i.e. the specification doesn't support the requested trait).

src/openassetio-core/include/openassetio/specifications.hpp
feltech: Many things changed, but somewhat related, added some docstrings in 6a50c5f9

src/openassetio-core/include/openassetio/specifications.hpp
feltech: Used pimpl in the rewrite in 01d6a8ea

python/openassetio/specification/trait/__init__.py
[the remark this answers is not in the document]
feltech: Moved `trait` outside of `specification` in 831a9168.

foundrytom (commented):
A few more open questions

src/openassetio-core/include/openassetio/specification/trait/TraitBase.hpp
```diff
@@ -0,0 +1,51 @@
...
+struct TraitBase {
+  using SpecificationPtr = std::shared_ptr<Specification>;
+
+  explicit TraitBase(SpecificationPtr spec) : spec_{std::move(spec)} {}
+
+  [[nodiscard]] static const TraitId& traitId() { return Derived::kId; }
+  [[nodiscard]] bool isValid() const { return spec_->hasTrait(traitId()); }
+
+ protected:
+  SpecificationPtr& spec() { return spec_; }
```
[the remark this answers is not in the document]
feltech: In general I'd err on the side of being verbose.  In this case I thought it was pretty unambiguous.  But I'll happily expand it.

tests/openassetio/test_traits.py
```diff
@@ -0,0 +1,158 @@
...
+    def test_when_trait_is_not_found_then_IndexError_raised(self, a_specification):
+        with pytest.raises(IndexError):
+            a_specification.setTraitProperty("unknown_trait", "a string", "string")
+
+        with pytest.raises(IndexError):
+            _ = a_specification.getTraitProperty("unknown_trait", "a string")
+
+    def test_when_value_is_not_supported_then_set_raises_TypeError(self, a_specification):
+        with pytest.raises(TypeError):
+            a_specification.setTraitProperty("first_trait", "unknown type", object())
```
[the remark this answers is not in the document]
feltech: Yeah, I just didn't want to splat any existing code (yet), whilst keeping the new stuff separate.

src/openassetio-core/include/openassetio/specification/trait/TraitBase.hpp
foundrytom: Capturing offline discussion:

The choice comes from the fact that we had previously made an API-level decision that "missing" is less important than "invalid type". Upon reflection, we can't say that for sure. Without doubt, invalid type is always bad, but in many cases missing is equally so.  Having two mechanism to categorise the failure reason is confusing for everyone, and forces an opinion on what is more or less important.

Proposal:

- `C`/`C++` Core API:  Return an `enum` instead of a `bool` to cover `ok`, `not_set` and `invalid_type`.
- Future `C++` extension: return `optional<variant<>>` to allow the host to use more standard techniques to determine the status.
- `Python`: Return `None` or value by default, add `strict` mode that throws exceptions if missing/invalid type

src/openassetio-core/include/openassetio/specification/trait/TraitBase.hpp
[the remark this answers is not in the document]
foundrytom: Verbose is good from my p.o.v too as an OSS project, if you're happy.

src/openassetio-core/include/openassetio/specifications.hpp
foundrytom: These are a good start. I guess the main request was for something more along the lines of what even is a spec, when is used, the whole view/data separation and relation to traits.

tests/openassetio/test_traits.py
```diff
@@ -0,0 +1,158 @@
...
+
+        actual = trait.BlobTrait(a_blob_specification).getUrl()
+
+        assert actual == expected
+
+
+class Test_BlobTrait_setURL:
+    def test_when_spec_doesnt_have_blob_trait_then_raises_IndexError(self, a_specification):
+        with pytest.raises(IndexError):
+            trait.BlobTrait(a_specification).setUrl("some://url")
```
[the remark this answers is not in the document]
feltech: Does it then make sense to have concrete specification classes, i.e. `BlobSpecification`?  Since you could transform it into a different specification by setting new traits on it.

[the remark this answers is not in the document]
foundrytom: It felt like the concrete specs are more generally a template for consistency. The reason for an `ImageSpecification` being to ensure that all hosts.managers dealing with an image agree it combines `blob`, `image`, `raster` for example, and make it easy for a programmer to know what should/shouldnt be available (as there aren't trait accessors on that spec for traits an "image" doesn't have).

I guess it throws up the Q as to whether:
 - Direct trait use is allowed (feels like it should be) take `register` for example, wanting to just use a `BlobTrait`, and we dont need `BlobSpecification` specifically unless its composes multiple traits.
 - Or, should only `Specifications` be used, and we consequently always have a spec for each trait.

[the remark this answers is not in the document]
foundrytom: I guess, coupled with the other comment above, is this desirable for a manager in resolve:

```python

result = Specification()

if myAsset.filePath:
  blob = BlobTrait(result)
  blob.setUrl(f"file://{myAsset.filePath}")

if myAsset.isSequence:
  range = FrameRangeTrait(result)
  range.setStartFrame(myAsset.startFrame)
...
```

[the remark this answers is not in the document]
foundrytom: .. though it does pose the question of "does just wrapping add the trait ID", maybe we should look at how applied schemas work in USD.

src/openassetio-core/include/openassetio/specification/Specification.hpp
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * @param traitId ID of trait to update.
+   * @param propertyKey Key of property to set.
+   * @param propertyValue Value to set.
+   * @exception `std::out_of_range` if the trait is not supported by
+   * this specification.
+   */
+  void setTraitProperty(const trait::TraitId& traitId, const trait::property::Key& propertyKey,
+                        trait::property::Value propertyValue);
+
```
foundrytom, line 81: [question] Are we considering introspection one for a follow up PR? Ie, the ability to iterate traits and their properties as for serialisation.

feltech: I think it's probably a good idea to consider serialisation at a later date.  

Serialisation is not really needed for the the minimal "manager that refuses to manage anything" vertical slice.  We could define a new vertical slice that requires serialisation, with properly designed workflows and end-to-end tests, as a new unit of work?

tests/openassetio/test_traits.py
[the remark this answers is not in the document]
feltech: From what I read, USD's applied API schemas do indeed add the attributes. Non-applied schemas (which are similar to how I have viewed traits so far) do not.  So, as in all things, USD has it both ways.  I'm not 100% that's how applied vs. non-applied schemas work, but that's my understanding.

src/openassetio-core/include/openassetio/specification/trait/TraitBase.hpp
feltech: Updated in b1169a1a

tests/openassetio/test_traits.py
```diff
@@ -0,0 +1,158 @@
...
+
+        assert a_specification.getTraitProperty("first_trait", "a string") == "string"
+        assert isinstance(a_specification.getTraitProperty("first_trait", "a string"), str)
+        assert a_specification.getTraitProperty("second_trait", "an int") == 1
+        assert isinstance(a_specification.getTraitProperty("second_trait", "an int"), int)
+        assert a_specification.getTraitProperty("first_trait", "a float") == 1.0
+        assert isinstance(a_specification.getTraitProperty("first_trait", "a float"), float)
+        assert a_specification.getTraitProperty("second_trait", "a bool") is True
+
+    def test_when_key_is_not_found_then_get_returns_None(self, a_specification):
```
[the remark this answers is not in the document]
feltech: Ah, hmm, yeah I have used "missing" vs. "found", e.g. in the enum added in b1169a1a.  Would you prefer "set" vs "unset"?  My slight reservation is that these are verbs as well as adjectives.

tests/openassetio/test_traits.py
```diff
@@ -0,0 +1,158 @@
...
+        assert a_specification.getTraitProperty("first_trait", "a string") is None
+
+    def test_when_trait_is_not_found_then_IndexError_raised(self, a_specification):
+        with pytest.raises(IndexError):
+            a_specification.setTraitProperty("unknown_trait", "a string", "string")
+
+        with pytest.raises(IndexError):
+            _ = a_specification.getTraitProperty("unknown_trait", "a string")
+
+    def test_when_value_is_not_supported_then_set_raises_TypeError(self, a_specification):
```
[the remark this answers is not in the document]
feltech: Really good spot.  Adding a test for `None` revealed rather unintentional behaviour (was silently converted to `false`). Fixed in 6e68a638

src/openassetio-core/include/openassetio/specification/trait/TraitBase.hpp
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @tparam Derived Concrete subclass.
+ */
+template <class Derived>
+struct TraitBase {
+  using SpecificationPtr = std::shared_ptr<Specification>;
+
+  explicit TraitBase(SpecificationPtr spec) : spec_{std::move(spec)} {}
+
+  [[nodiscard]] static const TraitId& traitId() { return Derived::kId; }
```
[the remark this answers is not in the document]
feltech: Renamed in 996495a3

src/openassetio-core/include/openassetio/specification/trait/TraitBase.hpp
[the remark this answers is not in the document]
feltech: Renamed in 8e17d811

src/openassetio-core/include/openassetio/specification/Specification.hpp
foundrytom: Sounds good to me 👍

src/openassetio-core/include/openassetio/specification/Specification.hpp
foundrytom: We can certainly save this for a later PR if we're happy that we're leaking `std::variant` in the public API.

tests/openassetio/test_traits.py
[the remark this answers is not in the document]
foundrytom: I think the verb/adjective ambiguity is a great point.

tests/openassetio/test_traits.py
```diff
@@ -0,0 +1,158 @@
...
+        assert a_specification.getTraitProperty("second_trait", "an int") == 1
+        assert isinstance(a_specification.getTraitProperty("second_trait", "an int"), int)
+        assert a_specification.getTraitProperty("first_trait", "a float") == 1.0
+        assert isinstance(a_specification.getTraitProperty("first_trait", "a float"), float)
+        assert a_specification.getTraitProperty("second_trait", "a bool") is True
+
+    def test_when_key_is_not_found_then_get_returns_None(self, a_specification):
+        assert a_specification.getTraitProperty("first_trait", "a string") is None
+
+    def test_when_trait_is_not_found_then_IndexError_raised(self, a_specification):
```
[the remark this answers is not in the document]
foundrytom: Yeah, not required for this PR

tests/openassetio/test_traits.py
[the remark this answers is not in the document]
foundrytom: We can re-visit adding traits later, but definitely need to make sure that we highlight the use of `isValid` in the docs.

src/openassetio-core/include/openassetio/trait/TraitBase.hpp
```diff
@@ -0,0 +1,93 @@
...
+   * @param specification Underlying specification to wrap.
+   */
+  explicit TraitBase(SpecificationPtr specification) : specification_{std::move(specification)} {}
+
+  /**
+   * Get the ID of this trait.
+   *
+   * @return ID of this trait.
+   */
+  [[nodiscard]] static const TraitId& id() { return Derived::kId; }
```
foundrytom, line 39: [fix] Capturing an off-line conversation:

If we don't need to pass around base class `TraitPtr`s, then do we need this method? Is `kId` sufficient?

src/openassetio-core/include/openassetio/trait/BlobTrait.hpp
```diff
@@ -0,0 +1,77 @@
...
+ * Core trait class representing a locatable blob of data of a specific
+ * type.
+ *
+ * Has the ID of "blob" and defines "url" and "mimeType" properties.
+ */
+struct OPENASSETIO_CORE_EXPORT BlobTrait : TraitBase<BlobTrait> {
+  /// ID of this trait.
+  static inline const TraitId kId{"blob"};
+  /// Key for URL property.
+  static inline const property::Key kUrl{"url"};
```
foundrytom, line 28: [consider] Should we prefix property key names to avoid ambiguity? eg `kProperty_url` or similar, so that it is clear that `kId` isn't a property called `id`?

foundrytom (changes requested):
Adding a few off-line chats.

src/openassetio-core/include/openassetio/specification/BlobSpecification.hpp
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
feltech: Removed in dfb36cd7

src/openassetio-core/include/openassetio/trait/BlobTrait.hpp
feltech: I went a slightly different direction in 49c7861c, where I made `kUrl` and `kMimeType` private implementation details.  My argument is that if we do keep them as public members, then we should also expose them to Python.  So, since YAGNI I went with just making them private.

src/openassetio-python/trait/BlobTrait.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <optional>
+#include <type_traits>
+
+#include <pybind11/stl.h>
+
+// Public.
+#include <openassetio/export.h>
+#include <openassetio/trait/BlobTrait.hpp>
+// Private.
+#include <trait/keys.hpp>
```
feltech, line 12: [question] Here we're `#include`ing a header that is not part of the public API, i.e. not in the `include` directory.  

To make this work, the build target's include directories must overlap, i.e. both the `openassetio-core` directory and the `openassetio-core/include` directory are on the search path.  

I think the directory structure should be rejigged to remove the overlap, but I can't decide a good split.

Currently we have
```
openassetio-core
├── include
│   └── openassetio
│       ├── managerAPI
│       ├── specification
│       └── trait
├── managerAPI
├── specification
└── trait
```

We could e.g. restructure to something like
```
openassetio-core
├── include
│   └── openassetio
│       ├── managerAPI
│       ├── specification
│       └── trait
└── src
    └── private
        ├── managerAPI
        ├── specification
        └── trait
```
which would allow `#include <private/trait/keys.hpp>` (assuming `src` was on the include search path).  

I don't particularly like this option, but it's surprisingly hard to come up with a a scheme that doesn't overlap, isn't overly verbose, gives clear at-a-glance `#include`s, whilst (ideally) dog-fooding our use of public includes.

We can look at other projects to get some inspiration.

foundrytom: How come its not part of the public API? If it's needed for a custom trait, then it needs to be public - as traits have to be runtime extensible?

src/openassetio-core/include/openassetio/trait/BlobTrait.hpp
foundrytom: Should these be private to the implementation of `BlobTrait`, not to the API? _Any_ trait must be considered out-of-core so we can't have anything shared/etc.. that can't be done in a totally random, unconnected repo with just a copy of the source.

foundrytom: Also, was the removal of these to force the use of a traits as a view on a `Specification` instance, instead of using their key constants directly in `Specification::getTraitProperty`? Though higher overhead, that's probably a 'good thing' as there have been comments before that more than one way to do the same thing causes confusion.

src/openassetio-python/trait/BlobTrait.cpp
foundrytom: I think I just answered that Q when looking into the comment above (sorry was visiting these in reverse order), either way, we should probably have a huddle on src layout?

src/openassetio-core/include/openassetio/trait/BlobTrait.hpp
feltech: The only reason they were public was so the Python bindings could use them in crafting an exception message, which seemed like a bad excuse for making something public and, like you say, unintentionally encouraging their use as keys in a raw `getTraitProperty` query.

Good point that we want to keep the traits self-contained, though.  I can rename the `keys.hpp` to `BlobTrait.hpp` so it's clear.

feltech: Updated in 40cf7353

src/openassetio-core/include/openassetio/specifications.hpp
feltech: Did a second pass in 55fbc07a. Also enabled Doxygen for public C++ headers in c905c47d.

Perhaps still need a little fleshing out - more example snippets, perhaps.

That reminds me, one cool thing I've done with Doxygen in the long lost past is [snippets](https://www.doxygen.nl/manual/commands.html#cmdsnippet), where you can mark a section of source code with special comments, then refer to that section in the docs, which will reproduce it.  So you have a living example.  Would be great to eventually have some "workflow test" snippets that we can litter through the docs.

feltech: Restructured a bit in d5a7251b - how's it looking re. this review comment?

python/openassetio/specification/__init__.py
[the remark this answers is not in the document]
feltech: I disabled the pylint checks back in 4348ffb0 - though none of the Github checks are running at the moment for some reason :thinking:

src/openassetio-core/include/openassetio/specification/Specification.hpp
feltech: Do we need to chat about this some more?  It sounds like we do want to expose `std::variant`, along with (eventually) a `std::optional` API, to make the modern C++ folks happy?

But if you think we should hide it for now we can do that.  Hiding `std::variant` makes the Python bindings a bit trickier - pybind is clever enough to work with `std::variant` out of the box, so if we hide it's use behind a typed accessor then we'll need a way to "unhide" it (or equivalent) in the bindings.

feltech: One argument perhaps is that it's just coincidental that we return a `std::variant` as well as use them internally.  We can have an API that returns `std::variant` and swap out the underlying storage for something else, they're not necessarily coupled.

src/openassetio-core/include/openassetio/trait/BlobTrait.hpp
feltech: ... and then removed in 077c393f after off-line discussion, where we decided that simplifying the exceptions is a small price to pay for dodging this question entirely ;)

src/openassetio-core/include/openassetio/specification/Specification.hpp
```diff
@@ -0,0 +1,127 @@
...
+   * @param[out] out Storage fo result, only written to if the property
+   * is set.
+   * @param traitId ID of trait to query.
+   * @param propertyKey Key of trait's property to query.
+   * @return `true` if value was found, `false` if it is unset.
+   * @exception `std::out_of_range` if the trait is not supported by
+   * this specification.
+   */
+  [[nodiscard]] bool getTraitProperty(trait::property::Value* out, const trait::TraitId& traitId,
+                                      const trait::property::Key& propertyKey) const;
```
foundrytom, line 107: [fix] We should probably also document what constitutes a valid trait name/property key

src/openassetio-core/include/openassetio/specification/Specification.hpp
foundrytom: Shall we leave it as a variant for now, and pick this up in another ticket in the not too distant?

foundrytom (commented):
Sorry I was sure I posted these yesterday, maybe it got lost in the GitHub glitch

src/openassetio-core/include/openassetio/specifications.hpp
foundrytom: That is spot on!

src/openassetio-core/include/openassetio/specification/Specification.hpp
feltech: I'm not sure what you mean, as in, it's any arbitrary string, as can be found by looking at the type.  Do we have further restrictions to add?  I guess the trait ID needs to be globally unique (I think I've called that out somewhere).

foundrytom: Yeah, was wondering if its fine for them to contain any valid UTF-8 character, and how that then maps to method names, etc.

src/openassetio-core/include/openassetio/trait/BlobTrait.hpp
```diff
@@ -20,6 +20,20 @@ namespace trait {
  * type.
  *
  * Has the ID of "blob" and defines "url" and "mimeType" properties.
+ *
+ * This core trait defines one of the most common set of properties in
```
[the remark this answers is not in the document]
feltech: We discussed off-line and decided to leave it in for now, but likely remove in favour of moving it to the [OpenAssetIO-MediaCreation](https://github.com/TheFoundryVisionmongers/OpenAssetIO-MediaCreation) repo

src/openassetio-core/include/openassetio/trait/TraitBase.hpp
```diff
@@ -17,6 +17,53 @@ namespace trait {
 /**
  * Abstract CRTP base class for specification traits.
  *
+ * This non-virtual templated base class provides the common interface
+ * for a concrete trait view on a specification::Specification.
+ *
+ * A trait view provides a way to hide the underlying dictionary-like
+ * data access from hosts and managers.
```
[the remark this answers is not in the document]
feltech: Restructured slightly in bc850855 and fixed the example.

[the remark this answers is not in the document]
feltech: At the risk of dragging this out further, the example in the docs is rather boilerplatey, reproducing it here:
```c++
property::Int myValue = 123;  // Default if property not found.

MyTrait myTrait{specification};

if (myTrait.isValid()) {

  if (myTrait.getMyValue(&myValue) != TraitPropertyStatus::kFound) {

    std::cerr << "Warning: myValue not found\n";
  }
}
```
We could simplify to remove `isValid` and not raise an exception if the trait is unsupported (can be checked via `Specification::hasTrait` so the functionality exists elsewhere), and instead add another `TraitPropertyStatus::kUnsupportedTrait`, then the workflow simplifies to
```c++
property::Int myValue = 123;  // Default if property not found.

if (MyTrait{specification}.getMyValue(&myValue) != TraitPropertyStatus::kFound) {

  std::cerr << "Warning: myValue not found\n";
}
```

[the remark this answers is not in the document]
feltech: I guess this relates to the comment re. allowing mutators to set properties and implicitly make the trait supported, rather than requiring specifications to be pre-constructed with the supported traits.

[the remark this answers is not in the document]
foundrytom: Like the proposition, shall we do it in a follow up PR?

src/openassetio-core/include/openassetio/specification/Specification.hpp
feltech: Added some notes in c959d09f - I added a note to strongly recommend ASCII, which I'm not sure is really necessary these days, but would take some research to find out.

src/openassetio-core/include/openassetio/trait/TraitBase.hpp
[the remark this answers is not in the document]
foundrytom: > Restructured slightly in [bc85085](https://github.com/TheFoundryVisionmongers/OpenAssetIO/commit/bc8508552b00ee92ace36906d9a647aeabe0b012) and fixed the example.

This is def on the right track, and the example is great. What do you think about moving it above all the requirement for a derived class?

src/openassetio-core/include/openassetio/specification/Specification.hpp
foundrytom: GitHub won't let me resolve this comment 🤷‍♂️

src/openassetio-core/include/openassetio/trait/TraitBase.hpp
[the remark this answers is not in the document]
feltech: Move up in bdb6c289 - seems to read well like that.

PR #256, opened 2022-03-02
[Examples] Fix BAL JSON Schema URLs

foundrytom:
Now it is actually in GitHub, it illustrated the previous schema URL was incorrect.

Can anyone test this with their IDE to see if it is working ok? Editing the BAL test library json should validate correctly now.

I'm not sure this is set up right yet.

### DIFF
--- a/resources/examples/manager/BasicAssetLibrary/tests/resources/library_apiComplianceSuite.json
+++ b/resources/examples/manager/BasicAssetLibrary/tests/resources/library_apiComplianceSuite.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://github.com/TheFoundryVisionmongers/OpenAssetIO/resources/examples/manager/BasicAssetLibrary/schema.json",
+  "$schema": "https://raw.githubusercontent.com/TheFoundryVisionmongers/OpenAssetIO/main/resources/examples/manager/BasicAssetLibrary/schema.json",
   "entities": {
     "anAsset⭐︎": {
       "versions": [{ "primary_string": "resolved from 'anAsset⭐︎' using 📠" }]
--- a/resources/examples/manager/BasicAssetLibrary/schema.json
+++ b/resources/examples/manager/BasicAssetLibrary/schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://github.com/TheFoundryVisionmongers/OpenAssetIO/resources/examples/manager/BasicAssetLibrary/schema.json",
+  "$id": "https://raw.githubusercontent.com/TheFoundryVisionmongers/OpenAssetIO/main/resources/examples/manager/BasicAssetLibrary/schema.json",
   "title": "BAL Library File",
   "description": "The data store that backs an instance of the BAL manager",
   "type": "object",


feltech:
Works for me, for the most part.  PyCharm picks up the remote schema and validates the JSON.

Interestingly however, in PyCharm I hit [this problem](https://youtrack.jetbrains.com/issue/WEB-51501) whereby the ([apparently non-standard and unsupported](https://github.com/json-schema-org/json-schema-spec/issues/828)) `$schema` identifier in `library_apiComplianceSuite.json` triggers a warning, because `$schema` is not in the schema!

We can either leave this in as an oddity, or we can fix by adding `$schema` to the schema.  In my IDE  the following works (tested by replacing the URL with a relative file path on my box)
```json
    "$schema": {
      "type": "string",
      "pattern": "\\Qhttps://raw.githubusercontent.com/TheFoundryVisionmongers/OpenAssetIO/main/resources/examples/manager/BasicAssetLibrary/schema.json\\E"
    },
```
i.e. using `\\Q`/`\\E` to bookend a literal character sequence, which may not work universally (the [docs](https://cswr.github.io/JsonSchema/spec/basic_types/) I checked have nothing to say on the regex flavour used).

It's also a bit frustrating that you'd have to duplicate the `$id`.  So maybe just the presence of a `$schema` string is sufficient and we don't need to validate the exact URL (in fact, maybe we don't even want to, it may be legitimate to point `$schema` at a mirror). 

Or like I say, we can just leave that as an expected error in some IDEs due to the fact that `$schema` is non-standard.

feltech (approved):
OK as it is, but we may want to add `$schema` to the schema, as mentioned in my other comment.

foundrytom:
@feltech thanks for looking into the details. Based on `$schema` being unsupported, I just removed it [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/256/commits/51b96ee42fd72a51deb50b385463d0f44d8eea24)

resources/examples/manager/BasicAssetLibrary/tests/resources/library_apiComplianceSuite.json
```diff
@@ -1,5 +1,4 @@
 {
-  "$schema": "https://raw.githubusercontent.com/TheFoundryVisionmongers/OpenAssetIO/main/resources/examples/manager/BasicAssetLibrary/schema.json",
```
feltech: [consider][minor] It's unfortunate to remove it, since it does work (at least in JetBrains IDEs) - it's just that `$schema` gets flagged as not being in the schema. 

Selfishly, I can also configure the schema for a JSON file through the settings in JetBrains IDEs, so removing it from the .json itself doesn't preclude nice schema validation for me, it's just less convenient.

Is it just that we don't want to go down the route of using non-standard extensions?

foundrytom: Happy to leave it in here if you like, I guess I was just less keen on adding it to the schema itself as that seems very atypical?

feltech: So just let it flag a warning in IDEs that don't treat `$schema` as a special case?  I think I prefer that, rather than omit it.

foundrytom: I've dropped the commit that removed it from the library `json`.

PR #257, opened 2022-03-04
[CI] Avoid a couple of false-positives in C++ linters

feltech:
Clang-Tidy 10 flags a `performance-unnecessary-value-param` error with the following
```
  explicit BaseTrait(SpecificationDataPtr specificationData)
      : HasSpecificationData(std::move(specificationData)) {}
```
saying that `specificationData` should be a const reference. However, this snippet isn't a performance concern since `specificationData` is (immediately) moved. This idiom is beneficial in the case that `BaseTrait` was constructed with an rvalue `SpecificationDataPtr`, where a const reference would involve a pointless copy rather than a move.

Clang-Tidy 12 doesn't complain about this.  

Unfortunately Clang-Tidy 12 also doesn't complain if I remove the `std::move`.  Neither does Clang-Tidy 14.  However, if I assign to a member variable rather than pass to the base class constructor then I get the expected 

> parameter 'specificationData' is passed by value and only copied once; consider moving it to avoid unnecessary copies

I'm not sure if allowing pass-by-value to the base class without `move ` is a bug or by design. I can't find anything in the Clang-Tidy changelog.  I'm pretty sure the `move` is still valuable, as can be seen in [this compiler explorer demo](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEIADMXKQADqgKhE4MHt6%2B/ilpGQIhYZEsMXGJtpj2jgJCBEzEBNk%2BfgF2mA6Z9Y0ExRHRsQlJCg1NLbntY32hA2VDiQCUtqhexMjsHObxocjeWADUJvFuTqPEmKzH2CYaAIK3d%2BdeDgfKjVf3JgDsVvcHALeHxYEEWR3iABEDlgqEwvLQCMc/g9/oD3sRWBB0VczAA2cy4g6oAgIWKLR6Ao6/CmUwGjdAgFCrAjgtzHNxHMxmACyqAAbph0OYzKz2Qd6YzXLQkTSAT8IY9ZUCMSC0AxRsqcYTiaTiOTUXLqQbaRKmV4WeyxcKPMk8ILhaKTuKCAyQFKZcb5Yrvgr4sjHs9XhYmEpvcjKcGlFjgQdksCwSBY8CAPo/Cymlj8zAQOMq8k%2BqlWH2K43YlhJlWpv3e33%2B%2B6BlkQ2J4AXoA6JyOYMNKpvEFuC6MqiusBMHTs5%2BOFr1fYvVr73UIslhMUKg7vG3v9oW/Mtpr2z/0%2BjjLWicACsvD8HC0pFQnDZlms4tW60wnMSvAImmPywA1iBJC4AA6SQzCkM9vlxb5EjPM8zG%2BaRTw4SRL2/W9OF4BQQA0Ugv2vY9SDgWAYEQJkWFtegyAoCA0HIughj2QxgG%2BDQcJoBFYiwiAojQqJQkaABPTgeFIWi2EEAB5BhaCE/DSCwZcjHEOT8AuLoBSwuTMFUTpzU2G9F2qNDaDwKIMWIASPCwYTPz7FgbOWKgDGABQADU7QAdwk5JGBsmRBBEMR2Ckfz5CUNQ0N0JIDCMFBrGsfRTKw2BmDYEAolQNICFIAU4m4Fif1vZJanVTgAFp6WOCFTEfSwuG%2BTDqk6EqXAYdxPFaPRglmUpyj0VJ0hKiY/CSAbCgYfpeqGEYmq6OppmGvQOjmhgeiaSbBjiEYFo63Jtt6Db5i25YFBfDY9AIC5Nh4E9z1QuS7w4VQAA5cTK3FJAORijAOb4gI0f6DggB8rEsUgDlwQgSHfJIDg8OjKJhxZP2/RY/wA%2BIgLxDRvjAiCAE54g0M9cTPfROBQ0h7LPHCrxvR7MOw3DUcIkiICQZlknNchKFoijpsCfAiC20LAvEEL%2BEERQVHUOSPIxZIHPJjgL1IOneEeiTzS5llUCoA4Xrej6vpi4Bfv%2BwHgdQBHYiRlH8LR0hSSYLA4lBZXKep2m0IZ2wmbwrRHf/M9nqAiDnrPUD4jxZ7JEkXHlfie76Yw5mHdujgzGTjXU4DwrcvSZxJCAA).

### DIFF
--- a/cmake/StaticAnalyzers.cmake
+++ b/cmake/StaticAnalyzers.cmake
@@ -2,7 +2,10 @@
 # Copyright 2013-2022 The Foundry Visionmongers Ltd
 
 macro(enable_clang_tidy)
-    find_program(CLANGTIDY clang-tidy)
+    # Clang-Tidy warnings vary quite a bit between different versions.
+    # So pin to clang-tidy-12 for now, which happens to be the latest
+    # available in Ubuntu 20.04's (i.e. CI) apt repos.
+    find_program(CLANGTIDY clang-tidy-12)
 
     if (CLANGTIDY)
         # Construct the clang-tidy command line
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -36,7 +36,6 @@ Checks: >
     performance-*,
     portability-*,
     readability-*,
-    -google-runtime-references,
     -modernize-return-braced-init-list,
     -modernize-use-trailing-return-type,
     -modernize-avoid-c-arrays,
--- a/contributing/CODING_STANDARDS.md
+++ b/contributing/CODING_STANDARDS.md
@@ -18,7 +18,7 @@ with a few exceptions, as detailed in the configuration of the linters
 described below.
 
 CI is configured to enforce [Clang-Format](https://clang.llvm.org/docs/ClangFormat.html),
-[Clang-Tidy](https://releases.llvm.org/10.0.0/tools/clang/tools/extra/docs/clang-tidy)
+[Clang-Tidy 12](https://releases.llvm.org/12.0.0/tools/clang/tools/extra/docs/clang-tidy)
 and [Cpplint](https://github.com/cpplint/cpplint) checks.
 
 Clang-Tidy and Cpplint can be enabled as part of the build process using
--- a/resources/build/bootstrap-ubuntu-20.04.sh
+++ b/resources/build/bootstrap-ubuntu-20.04.sh
@@ -5,7 +5,7 @@
 set -xeo pipefail
 sudo apt-get update
 # Install gcc, clang-tidy and Python 3.
-sudo apt-get install -y build-essential clang-tidy python3-pip
+sudo apt-get install -y build-essential clang-tidy-12 python3-pip
 # Install system packages required for the Python 3.9 conan package.
 # These would be installed as part of the conan package install, but
 # we're caching the conan directory via the `actions/cache` Github
--- a/src/CPPLINT.cfg
+++ b/src/CPPLINT.cfg
@@ -1,4 +1,8 @@
 set noparent
 linelength=99
 root=.
-filter=-runtime/references
+# -runtime/references: Allow non-const references as function params.
+# -readability/nolint: Don't complain if unrecognized checks are
+#   disabled by a NOLINT comment. This is because both cpplint
+#   and clang-tidy use NOLINT, annoyingly.
+filter=-runtime/references,-readability/nolint


fnRaihanKibria (approved):
Nothing further from me to add.

cmake/StaticAnalyzers.cmake
```diff
@@ -2,7 +2,12 @@
 # Copyright 2013-2022 The Foundry Visionmongers Ltd
 
 macro(enable_clang_tidy)
-    find_program(CLANGTIDY clang-tidy)
+    # clang-tidy-10 (default on Ubuntu 20.04, i.e. Github CI) has false
+    # positives for `performance-unnecessary-value-param` when using the
+    # idiom of immediately `std::move`ing parameters. Luckily
+    # clang-tidy-12 is available in Ubuntu 20.04 and doesn't have this
+    # problem.
+    find_program(CLANGTIDY clang-tidy-12)
```
foundrytom, line 10: [question] This comment is very CI focused,  but this toolchain is generic for all builds right?

Given we do definitely need to be able to target reference platform CY2022 as well as the newer distros, is this going to cause any issues using the ASWF docker image, etc?

feltech: It's generic in that if you enable the `OPENASSETIO_ENABLE_CLANG_TIDY` CMake option then it'll affect all builds. 

For CI we only need to enable it for one build.  So if we're (also) building in CY22 Docker, we can simply disable the clang-tidy check.

You can also override it by setting CMake variable `-DCLANGTIDY=/path/to/clang-tidy` if you really want to use a different version (I did exactly that to test different versions).

Another option is that `find_program` can take a list of possible names, in order of preference, so we could allow older versions of clang-tidy to be used as a fallback.  Obviously that means possible disagreement on the clang-tidy warnings between local and CI builds, but perhaps that's preferable over not having clang-tidy checks at all?

foundrytom: I guess then, the main Q is - are we happy to say that clang tidy 12 is a base requirement for standard builds? (I know we need to enable it, but its part of our core validation suite).

The majority of real world uses of the API will be much closer to CY22 so we don't want to increase the barrier to entry of the base build checks if we can help it. I think we can't use the rational that it being available for Ubuntu 12 as a particular justification. That only helps our CI out...

If this was pylint wed just have  `# pylint: disable=performance-unnecessary-value-param` at the call site right? Is there any way we can do this here?

foundrytom: Capturing off-line comment:

[fix] Phrase comment in the context of pinning on a stable `clang-tidy` version and reasons for 12 specifically. Update `BUILDING.md`

feltech: Updated the code comment in 1eace612 and updated `CODING_STANDARDS.md` in 097d3d05 - there was no mention of `clang-tidy` in `BUILDING.md` previously, do you think we should add a note in there as well?

feltech: > If this was pylint wed just have `# pylint: disable=performance-unnecessary-value-param` at the call site right? Is there any way we can do this here?

Indeed, `clang-tidy` supports `// NOLINT(performance-unnecessary-value-param)`-like comments.  My reason for upgrading clang-tidy rather than just suppressing the warning was that it's clearly a false positive, and is likely to crop up again and again.

foundrytom: Thanks @feltech up to you re `BUILDING.md`, but I think we're good for now. Sorry I'd forgotten that we didn't mention it there.

foundrytom (approved):

foundrytom:
Feel free to squash and merge at your leisure @feltech

PR #261, opened 2022-03-09
[Core] Remove the host document API

### LINKED ISSUE #6: Remove `Host` document API
## What

Remove methods `Host. documentReference`, `Host.getKnownEntityReferences` and `Host.knownEntityReferences` until we can properly model specific scenarios in terms of multi-document support.

## Why

We have previously determined through research, that as a developer of a manager, you may want to query entity usage within a `Host` document, I need to be able to specify which open document the queries relate to. Many hosts support multiple open documents within the same process, and the current API stipulates `getKnownEntityReferences` and `getDocumentReference` should [reference the 'frontmost' document](https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/main/python/openassetio/hostAPI/HostInterface.py#L175), which is severely limiting.

We need to re-validate this assumption, along with know use cases to make sure that 1) this is still true 2) can any scenarios be generalised enough to make their inclusion in `HostInterface` legit, and 3) If not, do we even need `HostInterface` and is the host `identifier` just part of `Context.locale`?

That said, wherever we see a manager needing to interact with the host data model, we need to make sure that the process of getting "back to the host native SDK object" from an OpenAssetIO call is not onerous (this was one of the original motivations for the HostInterface being there in the first place).

foundrytom:
We still fundamentally believe that the ability for a manager to query the current host for information about entity usage and its documents is critical. However, the current API was not well thought out (as it
assumed a 'current' document, which may be a little tenuous).

This PR closes #6 by removing the current API, so that we can add a more fit-for-purpose API in over time as we have more concrete use cases from manager authors.

### DIFF
--- a/doc/src/Examples.dox
+++ b/doc/src/Examples.dox
@@ -29,7 +29,7 @@
  *
  * class ExampleHost(HostInterface):
  *     """
- *     A minimal host implementation with no document model.
+ *     A minimal host implementation.
  *     """
  *     @staticmethod
  *     def identifier():
--- a/doc/src/ForHosts.dox
+++ b/doc/src/ForHosts.dox
@@ -22,9 +22,7 @@
  *   implementation of the @ref HostInterface class. This represents the
  *   application or tool in a generic fashion <i>to</i> the asset
  *   management system. This allows the manager to adapt its behavior
- *   if necessary, or query information about entity usage within the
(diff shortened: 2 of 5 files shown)


feltech (approved):

feltech:
LGTM, rebase and merge when ready.

PR #262, opened 2022-03-09
[Core] Improve context thawing API

### LINKED ISSUE #79: Create a `Context` from frozen token
# What

Add the ability to create a context from a frozen state token.

# Why

In a distributed rendering scenario, each worker machine is required to create a `Context` then use `TransactionCoordinator.thawManagerState` to restore a previous state based on its frozen token.

Though this works, it means that the `ManagerInterface` has to first make a redundant state for the initial context creation, only to have it immediately discarded when the previous state is thawed. We should have a way for a host to create a context directly from the frozen state token to reduce wasted effort by the `ManagerInterface`.

foundrytom:
Changes TransactionCoordinator.thawManagerState to be the reverse of freezeManagerState.

The main motivation is to remove a redundant state creation when the host creates a context, then immediately thaws a serialized state back into it, overwriting the one just made by the manager.

This also makes the API more consistent with the actual behaviour, in terms of "context lifetime associates API calls".

Closes #79.

### DIFF
--- a/python/openassetio/hostAPI/transactions.py
+++ b/python/openassetio/hostAPI/transactions.py
@@ -19,6 +19,7 @@ This module provides convenience functionality for a @ref host to aid
 working managing transactions when interacting with a @ref manager.
 """
 
+from .. import Context
 from .._core.audit import auditApiCall
 from .._core.debug import debugApiCall, Debuggable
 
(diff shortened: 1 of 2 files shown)


python/openassetio/hostAPI/transactions.py
```diff
@@ -173,16 +174,15 @@ def freezeManagerState(self, context):
...
 
     @auditApiCall("Transactions")
-    def thawManagerState(self, token, context):
+    def thawManagerState(self, token):
         """
-        Restores the @ref manager_state in the supplied Context so that
-        it represents the context as previously frozen.
+        Restores the @ref manager_state returning a Context that is
+        associated with the same state as the context that was
+        previously frozen using freezeManagerState.
```
feltech, line 181: [consider][minor]
```suggestion
        previously frozen using `freezeManagerState`.
```
or
```suggestion
        previously frozen using @ref freezeManagerState.
```

feltech (approved):
LGTM. 

I know we discussed this previously, but perhaps for posterity (and because I can't quite remember :wink: ) it's worth adding a (Github) comment about why the host doesn't need the ability to imbue a previously created `Context` with a different state token?

foundrytom:
> LGTM.
> 
> I know we discussed this previously, but perhaps for posterity (and because I can't quite remember 😉 ) it's worth adding a (Github) comment about why the host doesn't need the ability to imbue a previously created `Context` with a different state token?

Does the updated commit message [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/262/commits/941e77979924b13fb0c4b2215908b0c38504578c) help?

feltech:
> > LGTM.
> > I know we discussed this previously, but perhaps for posterity (and because I can't quite remember wink ) it's worth adding a (Github) comment about why the host doesn't need the ability to imbue a previously created `Context` with a different state token?
> 
> Does the updated commit message [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/262/commits/941e77979924b13fb0c4b2215908b0c38504578c) help?

Yeah that really clarifies things, thanks.

feltech (approved):

PR #270, opened 2022-03-16
[Core] Core API performance test harness

### LINKED ISSUE #260: Core API performance test harness - phase 1
## What

Produce the starting point for a light-weight test harness that can be used to exercise the `cpp` - `cpp` happy path to measure performance.

**Time box to 2 days**

## Why

The `cpp` - `cpp` happy path must be as performant as possible. We need a simple harness that we can use to assess the impact of choices made as part of #247, #255, etc.

The long term goal is to simulate high-volume/multi-threaded access to the API.

## Scope

The initial goals of this harness is to ensure we can assess specific implementation choices in #247 and #255 _only_. It does not need to be fully generic or re-usable, but should be a starting point that we can grow as we build out the API and language bindings.

## Acceptance criteria

- Bare minimum relative difference of wall clock time for blob trait resolution vs resolveEntityReference for a given (large, >100,000) number of unique entity references due to the changes in API middleware/methods not "host" or "manager" implementation.
- Total wall clock time for any given run.
- Ability to repeat the test run without signifiant manual user steps. to account for instability in test environment over time (ie thermal limits, etc... meaning we can reuse historical data with any accuarcy)
- The test needs to ensure that the implementation of the host/manager code is not allowing the compiler to short circuit the cost of each call.
- Fine to live on a branch in a prototype state
- Ideally, this would be cross-platform code, but if this is prohibitive in terms of effort, we can pick an agreed platform for v1.

## Out of scope

- Multithreading is not needed for the scenarios in this story.

fnRaihanKibria:
Standalone 'perftest' executable for testing performance.

Closes #260

### DIFF
--- a/resources/perftest/CMakeLists.txt
+++ b/resources/perftest/CMakeLists.txt
@@ -1,0 +1,9 @@
+add_executable(openassetio-perftest
+  perftest.cpp)
+
+set_default_target_properties(openassetio-perftest)
+install(TARGETS openassetio-perftest EXPORT ${PROJECT_NAME}_EXPORTED_TARGETS)
+
+target_link_libraries(openassetio-perftest
+  PRIVATE
+  openassetio-core)
--- a/resources/perftest/perftest.cpp
+++ b/resources/perftest/perftest.cpp
@@ -1,0 +1,149 @@
+#include <iostream>
+#include <chrono>
+#include <functional>
+#include <random>
+#include <string>
+#include <algorithm>
+
+#include <openassetio/specification/Specification.hpp>
+#include <openassetio/trait/BlobTrait.hpp>
+using namespace openassetio;
+
+constexpr size_t nIterations = 100000;
+const size_t dbSize = 1000000;
+
+std::chrono::nanoseconds MeasureRuntime(std::function<void(void)> func)
+{
+  const auto start = std::chrono::high_resolution_clock::now();
+  func();
+  const auto stop = std::chrono::high_resolution_clock::now();
+  return stop - start;
+}
+
+auto MeasureRuntimeMs(std::function<void(void)> func)
+{
+  return std::chrono::duration_cast<std::chrono::milliseconds>(MeasureRuntime(func)).count();
+}
+
+///////////////////////////////////////////////////////////
+/// Enlarge the path so we likely exceed the small string optimization length
+const std::string kSubDirectory = "0123456789_0123456789_0123456789/";
+
+std::string getEntityReference(size_t i) 
+{
+  return std::string("ams://" + kSubDirectory + "myasset_") + std::to_string(i);
+}
+
+std::string getPrimaryString(size_t i) 
+{
+  return std::string("file://" + kSubDirectory + "myasset_") + std::to_string(i);
+}
+
+///////////////////////////////////////////////////////////
+class OldManagerInterface
+{
+public:
+  OldManagerInterface(const std::unordered_map<std::string, std::string>& refToPrimStrs) : refToPrimStrs{refToPrimStrs} {}
+  std::string resolveEntityReference(const std::string& ref);
+
+private:
+  const std::unordered_map<std::string, std::string>& refToPrimStrs;
+};
+
+std::string OldManagerInterface::resolveEntityReference(const std::string& ref)
+{
+  return refToPrimStrs.at(ref);
+}
+
+///////////////////////////////////////////////////////////
+auto benchmarkOldManager(const std::unordered_map<std::string, std::string>& refToPrimStrs, const std::vector<std::string>& refs)
+{
+  OldManagerInterface oldMgr(refToPrimStrs);
+
+  const auto doLookups = [&oldMgr, &refs]() {
+    for(size_t i = 0; i < nIterations; ++i) {
+      const auto& ref = refs[i];
+      const auto& primStr = oldMgr.resolveEntityReference(ref);
+    }
+  };
+
+  return MeasureRuntimeMs(doLookups);
+}
+
+///////////////////////////////////////////////////////////
+class NewManagerInterface
+{
+public:
+  NewManagerInterface(const std::unordered_map<std::string, std::string>& refToPrimStrs) : refToPrimStrs{refToPrimStrs} {}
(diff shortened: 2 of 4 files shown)


foundrytom:
Looks good @fnRaihanKibria, def worth adding a little more detail to the commit message about the design/motivation.

feltech:
Regarding linters: 
* `cpplint` is currently not configured to scan the `resources` directory. Adding that directory (in `StaticAnalyzers.cmake`) reveals a bunch of errors.
* `clang-tidy` complains about a bunch of things, but only if the CMake `BUILD_PERFTEST` option is enabled, which it isn't in Github CI.
* I get several compiler warnings on Linux `gcc`, again obviously they don't show up unless the CMake `BUILD_PERFTEST` option enabled, which it isn't in Github CI. Compiler warnings are failures in CI.

So I would expect this to pass CI, once Github gets it's act together and actually runs it (presumably on the next push), but only because we're not actually checking this code.

It would be nice, though perhaps out of scope (@foundrytom ?) to build this in the Vagrant VM with the linters enabled and sort that out, and enable `BUILD_PERFTEST` (hopefully renamed - see other review comments) as part of the linter CI build.  

Bonus points to configure warnings in `CompilerWarnings.cmake` for VC, but that's almost definitely out of scope.

resources/perftest/perftest.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+auto MeasureRuntimeMs(std::function<void(void)> func)
+{
+  return std::chrono::duration_cast<std::chrono::milliseconds>(MeasureRuntime(func)).count();
+}
+
+///////////////////////////////////////////////////////////
+std::string getEntityReference(size_t i) 
+{
+  return std::string("ams://myasset_") + std::to_string(i);
```
feltech, line 31: [consider][minor] It looks like we're right on the borderline of [small string optimisation](https://shaharmike.com/cpp/std-string/) here.  Perhaps for reproducibility we should increase the minimum length of the string to eliminate that as a factor.  To be fair, it probably doesn't make a difference.

resources/perftest/perftest.cpp
```diff
@@ -0,0 +1,150 @@
...
+  const auto doLookups = [&newMgr, &refs]() {
+    for(size_t i = 0; i < nIterations; ++i) {
+      const auto& ref = refs[i];
+      const auto spec = newMgr.resolve(ref, {trait::BlobTrait::kId});
+      trait::BlobTrait btrait(spec);
+      trait::property::Str url;
+      const auto success = btrait.getUrl(&url);
+      if(success != trait::TraitPropertyStatus::kFound) {
+        std::cerr << "getUrl() failed!" << std::endl;
+        exit(-1);
```
feltech, line 115: [consider] It's a bit odd to exit with a negative status, at least on Linux where I believe the program exit code will be `255` if you use `-1`.  A more portable way would be to use `EXIT_FAILURE`.

feltech: Similar comment for `benchmarkOldManager`.

feltech: Though do we even need to check the error case, which should never happen (or rather if it does we'll get an exception from `refToPrimStrs.at(` which will terminate the program anyway).

feltech (commented):
Looks good, just a few minor optional niggles.

foundrytom:
> It would be nice, though perhaps out of scope (@foundrytom ?) to build this in the Vagrant VM with the linters enabled and sort that out, and enable `BUILD_PERFTEST` (hopefully renamed - see other review comments) as part of the linter CI build.

Can we keep this as absolutely minimal as we can right now, and we can start to pick those things off once we have a better take on our C++/C test infrastructure, as it may well be that much of this moves there.

foundrytom:
@fnRaihanKibria we've merged #255 - would you mind rebasing this against there when you have a sec please?

src/openassetio-core/CMakeLists.txt
```diff
@@ -85,3 +85,8 @@ install(
     FILES ${PROJECT_BINARY_DIR}/include/openassetio/export.h
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/openassetio/
 )
+
+option(BUILD_PERFTEST "Build the performance test tool" OFF)
```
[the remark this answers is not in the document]
fnRaihanKibria: Addressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/270/commits/0517b333dbded81d9423a4baeed599aa8232d4ab 
- Added prefix and moved the option & add_subdirectory to the top CMake file

resources/perftest/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+add_executable(perftest
```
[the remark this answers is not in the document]
fnRaihanKibria: Addressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/270/commits/95f49b04419f6e711908722b704c8bb2a786de09

resources/perftest/perftest.cpp
fnRaihanKibria: Addressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/270/commits/cd4528dc2f32d3c6f86fd0294f3e169d17fcc170

This changes the measurements slightly:
```
resolveEntityReference: 51 ms // was 45 ms before
resolve: 138 ms // was 125 ms before
```

resources/perftest/perftest.cpp
fnRaihanKibria: Removed error check in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/270/commits/a1c5a4cbb5b29080ff35a32873be0290a921c406
I think we were worried the compiler might optimize away the code somehow but that's probably not an issue. Quick check shows same timing.

fnRaihanKibria:
@feltech Addressed remaining concerns

feltech (commented):
Looks good. Just need to rebase and squash ready for merge.  

For reference, when squashing and rewording the commit(s), double-check the commit message format in [our docs](https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/main/contributing/COMMITS.md#commit-messages).  It's good to include the Github issue # somewhere in the commit body too.

feltech (approved):
Nice, thanks.

PR #279, opened 2022-03-18
[Tests] Ensures `test.manager` checks use correct interpreter

foundrytom:
Ensures the tests run the CLI using the same python as the tests themselves. Fixes one of two issues surfaced by #276 (#278 sorts the other).

### DIFF
--- a/tests/openassetio/test/manager/test_main.py
+++ b/tests/openassetio/test/manager/test_main.py
@@ -61,10 +61,7 @@ class Test_CLI_output:
 class Test_CLI_arguments:
 
     def test_when_called_without_fixtures_arg_then_exits_with_usage_and_exit_code_is_two(self):
-        args = ["python", "-m", "openassetio.test.manager"]
-        # We explicitly don't want an exception to be raised.
-        # pylint: disable=subprocess-run-check
-        result = subprocess.run(args, capture_output=True)
+        result = execute_cli(None)
         assert result.returncode == 2
         assert "usage:" in str(result.stderr)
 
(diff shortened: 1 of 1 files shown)


tests/openassetio/test/manager/test_main.py
```diff
@@ -98,8 +96,15 @@ def execute_cli(fixtures_path, *extra_args):
...
     @return `subprocess.CompletedProcess` The results of the invocation.
     """
-    all_args = ["python", "-m", "openassetio.test.manager"]
-    all_args.extend(["-f", fixtures_path])
+    # We may well be running inside a venv interpreter, but without
+    # the venv being active in the current environment. The effect
+    # of this is that the python on $PATH may not be the one that is
+    # correctly configured with the appropriate dependencies, so we
+    # attempt to use the same executable as the tests.
+    prog = sys.executable or "python"
```
feltech, line 104: [consider] I'm not sure if the `or "python"` is a good idea.  Presumably this is only in case these tests are being executed in an embedded Python interpreter or something?  The test might work in that case, but likely it would fail and be harder to debug than allowing `None` to be passed to `subprocess` and fail earlier?

feltech (approved):

tests/openassetio/test/manager/test_main.py
foundrytom: Good call, removed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/279/commits/bf78f09daa08c9a1a2d15eae9f0b1885c4a8ca75)

feltech (approved):

feltech:
Looks good, squash and merge when ready.

Issue #251, opened 2022-02-28
[Build] Move `clang-format` checks into `CMake`


foundrytom:
## What

Move the `clang-format` setup in to `CMake` such that is run (optionally) as part of a standard build, as `clang-tidy` is.

## Why

Being a separate step means that `clang-format` may not be checked locally before pushing, leading to PR CI churn as formatting is rejected.

PR #295, opened 2022-03-21
[Core] Migrate `ManagerInterface` `identifier` and `displayName` to C++

feltech:
For #292

### DIFF
--- a/doc/src/Testing.dox
+++ b/doc/src/Testing.dox
@@ -28,7 +28,7 @@
  * values to pass to methods such as
  * @ref openassetio.managerAPI.ManagerInterface.ManagerInterface.isEntityReference
  * "isEntityReference", and the expected values from methods such as
- * @ref openassetio.managerAPI.ManagerInterface.ManagerInterface.identifier
+ * @ref openassetio::v0_0::managerAPI::ManagerInterface::identifier
  * "identifier"
  *
  * The file should set a top level variable called `fixtures`, holding
--- a/doc/src/Testing.dox
+++ b/doc/src/Testing.dox
@@ -28,8 +28,7 @@
  * values to pass to methods such as
  * @ref openassetio.managerAPI.ManagerInterface.ManagerInterface.isEntityReference
  * "isEntityReference", and the expected values from methods such as
- * @ref openassetio::v0_0::managerAPI::ManagerInterface::identifier
- * "identifier"
+ * @fqref{managerAPI::ManagerInterface::identifier} "identifier"
  *
  * The file should set a top level variable called `fixtures`, holding
  * a dict structured as-per the `fixtures` parameter documented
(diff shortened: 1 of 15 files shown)


feltech:
Force-pushes up to here would have been avoided if we had #251 done :slightly_frowning_face:

doc/src/Testing.dox
```diff
@@ -28,7 +28,7 @@
  * values to pass to methods such as
  * @ref openassetio.managerAPI.ManagerInterface.ManagerInterface.isEntityReference
  * "isEntityReference", and the expected values from methods such as
- * @ref openassetio.managerAPI.ManagerInterface.ManagerInterface.identifier
+ * @ref openassetio::v0_0::managerAPI::ManagerInterface::identifier
```
foundrytom, line 31: [consider] Do we need to create some Doxygen constant/macro/something so we dont have to keep `s/vX_X/vY_Y/g` every time we change the version?

foundrytom (commented):
Sorry only had time for a quick skim, but looks pretty good to me.

fnRaihanKibria (approved):
LGTM

doc/src/Testing.dox
feltech: The only solution I've found for string substitution is `$(...)` environment variable substitution.  

Annoyingly, although env vars are substituted in `@ref`s, the substitution is not performed before attempting to cross-reference, so the `@ref` fails, but the rendered output has the correct namespace. I.e.
```
 @ref openassetio::$(OPENASSETIO_VERSION)::managerAPI::ManagerInterface::identifier
```
emits an "unable to resolve reference to `openassetio:'" warning and renders as
```
openassetio::v0_0::managerAPI::ManagerInterface::identifier "identifier"
```

An option that works, though may be controversial, is to add another `ALIAS`, which I'll call `fqref` (fully-qualified ref) and looks like
```
@fqref{managerAPI::ManagerInterface::identifier} "identifier"
```
It seems the `{...}` is then required, unfortunately.  However, I kinda like that it removes the prefix cruft.

foundrytom: Good hunting. `@fref` looks good to me.

feltech: Added/used `@fqref` in 171ca60c

foundrytom: https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/295/commits/171ca60cc1a69aaffefe035e634c9b0714163939 = 🔥

foundrytom (approved):

foundrytom:
@feltech feel free to squash and merge at your leisure.

doc/src/Testing.dox
feltech: I feel sorry of @ref (though their repo is empty, probably for good reason)

PR #297, opened 2022-03-23
[Core] Migrate `defaultEntityReference` to use trait sets

foundrytom:
Updates the signature for `defaultEntityReference` to take a set of trait IDs rather than a `Specification`. This removes the ambiguity around whether or not the method should consider the properties of the traits vs their identifiers.

This method is meant as a high-level starting point for some "kind" of entity, before they exist, so it should not be affected by the data of any particular entity itself.

### DIFF
--- a/tests/openassetio/hostAPI/test_manager.py
+++ b/tests/openassetio/hostAPI/test_manager.py
@@ -66,8 +66,8 @@ class ValidatingMockManagerInterface(ManagerInterface):
     def flushCaches(self, hostSession):
         return mock.DEFAULT
 
-    def defaultEntityReference(self, specifications, context, hostSession):
-        self.__assertIsIterableOf(specifications, EntitySpecification)
+    def defaultEntityReference(self, traitSets, context, hostSession):
+        self.__assertIsIterableOf(traitSets, (list, tuple))
         self.__assertCallingContext(context, hostSession)
         return mock.DEFAULT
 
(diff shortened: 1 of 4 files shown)


fnRaihanKibria (approved):
LGTM

tests/openassetio/hostAPI/test_manager.py
```diff
@@ -262,6 +262,9 @@ def an_entity_spec():
 def some_entity_specs():
     return [EntitySpecification(), EntitySpecification()]
 
+@pytest.fixture
+def some_entity_trait_sets():
+     return [("blob",), ("blob", "image")]
```
feltech, line 267: [fix] I note Github hasn't run checks :angry:  Running pylint locally finds a couple of issues.
```
tests/openassetio/hostAPI/test_manager.py:267:0: W0311: Bad indentation. Found 5 spaces, expected 4 (bad-indentation)
tests/openassetio/hostAPI/test_manager.py:395:0: C0301: Line too long (103/99) (line-too-long)
```

feltech (changes requested):
Looks good apart from linter issues. Happy for you to squash now to trigger checks, if it helps.

tests/openassetio/hostAPI/test_manager.py
foundrytom: Doh! thanks. Fixed in rebase, but GitHub still doesn't seem to want to run any tests...

foundrytom: It ran [_some_](https://github.com/TheFoundryVisionmongers/OpenAssetIO/actions/runs/2035699014) tests 🤷 but not the all the ones listed 🙃

foundrytom: Actions have finally worked up, `pylint` passes [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/runs/5695151493?check_suite_focus=true).

PR #301, opened 2022-03-25
[Core] Add `Specifications::traitIDs` method

foundrytom:
Retrieves the list of traits the Specification has been populated with. This is the first step to allowing introspection of a Specification's data by exposing which traits have been set.

Be kind, it's been a long time since I've been writing `C++` 🙃.

### DIFF
--- a/src/openassetio-core/specification/Specification.cpp
+++ b/src/openassetio-core/specification/Specification.cpp
@@ -18,6 +18,15 @@ class Specification::Impl {
   }
   ~Impl() = default;
 
+  TraitIds traitIds() const {
+    TraitIds ids;
+    ids.reserve(data_.size());
+    for (const auto& item : data_) {
+      ids.push_back(item.first);
+    }
+    return ids;
+  }
+
   bool hasTrait(const trait::TraitId& traitId) const {
     return static_cast<bool>(data_.count(traitId));
   }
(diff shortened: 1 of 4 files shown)


src/openassetio-core/specification/Specification.cpp
```diff
@@ -51,6 +60,8 @@ Specification::Specification(const TraitIds& traitIds) : impl_{std::make_unique<
 
 Specification::~Specification() = default;
 
+Specification::TraitIds Specification::traitIds() const { return impl_->traitIds(); }
```
foundrytom, line 63: [question] Out of date `cpp` brain, is the compiler clever here and "do the right things and not copy the vector umpteen times"?

feltech: Yeah, there will likely be one copy then one move.  

In one of my earlier sketches I imagined a specification's trait set as being statically defined, so you could safely return `const TraitIds&` and avoid any copies or moves.  But that is a bit more restrictive.

I'm happy with your more flexible solution and we can tweak later if performance is a problem.

foundrytom: Cool, thanks. I think we definitely can't assume they will be static, eg [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/301/files#diff-29adf34bb4665be39936d3cbfcf6d37b4659d84d8fcb4a218f49fe75151079ffR20), and "the photoshop example".

feltech (approved):

feltech:

PR #302, opened 2022-03-25
[Lint] Add other build artifacts to `.gitignore`

foundrytom:
As these are created by README instructed steps, it helps make the default dev experience less chaotic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ __pycache__
 /dist/
 /resources/build/.vagrant/
 /.pybuild/
+/build


fnRaihanKibria (approved):
LGTM

.gitignore
```diff
@@ -6,3 +6,5 @@ __pycache__
 /dist/
 /resources/build/.vagrant/
 /.pybuild/
+/Testing
```
feltech, line 9: [consider] Not sure about this one. I assume the `Testing` directory is the one created by `ctest`?  The `README` instructions `cd build` before running `ctest`, so I don't think we need to ignore this independent of `build`?

For reference/posterity, there is also `.git/info/exclude` for ignoring local-specific things.

feltech (approved):

Issue #281, opened 2022-03-18
Update `defaultEntityReference` to use Trait IDs


foundrytom:
## What

`defaultEntityReference` should take a list of trait IDs instead of a specification.

## Why

The new specification and traits system introduced in #255 is replacing the current one. Existing methods that take a specification, that should really use trait IDs need updating. As the manager is not supposed to consider trait properties when determining its policy (it is a high-level question), the passing a specification is ambiguous, and is open to misinterpretation or inconsistent behaviour between manager implementations.

## Acceptance Criteria

 - `defaultEntityReference` updated to take a list of traitIDs
 - Documentation updated
 - Add tests to `apiComplienceSuite`

PR #305, opened 2022-03-29
[Core] Switch `TraitIDs` to `unordered_set`

foundrytom:
For any given set of traits, composed to form a specification, logically:

 - Any given trait ID should not be present more than once.
 - The order the traits are considered has no meaning.

We previously used `std::vector` as the container for a specifications's trait IDs. This commit switches to a `std::unordered_set` to make the above logic explicit in the implementation.

### DIFF
--- a/src/openassetio-core/specification/Specification.cpp
+++ b/src/openassetio-core/specification/Specification.cpp
@@ -22,7 +22,7 @@ class Specification::Impl {
     TraitIds ids;
     ids.reserve(data_.size());
     for (const auto& item : data_) {
-      ids.push_back(item.first);
+      ids.insert(item.first);
     }
     return ids;
   }
--- a/tests/openassetio/test_traits.py
+++ b/tests/openassetio/test_traits.py
@@ -12,15 +12,13 @@ from openassetio import specification, trait
 
 class Test_Specification_traitIds:
     def test_when_has_no_traits_returns_empty_list(self):
-        empty_specification = specification.Specification([])
-        assert empty_specification.traitIds() == []
+        empty_specification = specification.Specification(set())
+        assert empty_specification.traitIds() == set()
 
     def test_when_has_traits_returns_expected_trait_ids(self):
-        expected_ids = ["a", "b", "🐠🐟🐠🐟"]
+        expected_ids = {"a", "b", "🐠🐟🐠🐟"}
         populated_specification = specification.Specification(expected_ids)
-        ## TODO: Should we use set instead of vector for TraitIds, as
-        ## the order is meaningless (and not preserved).
-        assert  set(populated_specification.traitIds()) == set(expected_ids)
+        assert populated_specification.traitIds() == expected_ids
 
 
 class Test_Specification_hasTrait:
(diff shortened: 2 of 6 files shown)


src/openassetio-core/specification/Specification.cpp
```diff
@@ -22,7 +22,7 @@ class Specification::Impl {
     TraitIds ids;
     ids.reserve(data_.size());
     for (const auto& item : data_) {
-      ids.push_back(item.first);
+      ids.insert(item.first);
```
fnRaihanKibria, line 25: Consider using structured bindings here which I think is a little more readable:

      for (const auto& [key, value] : data_) {
        ids.insert(key);
      }

foundrytom: @feltech and had discussed this, and went for this one, as otherwise you have an unused `value` binding that triggers other warnings...

fnRaihanKibria: I see, makes sense.

foundrytom: Do you think it's worth a comment?

fnRaihanKibria: I didn't realize that it would lead to warnings, which is a valid point. I don't think it's worth littering the code with comments about this, or at least I can't come up with something that's short and to the point. 

Maybe this might work? I've not tried it in the sanitizer yet.

    for ([[maybe_unused]] const auto& [key, value] : data) {

foundrytom: Cool, we can pick this up as another PR maybe, as its out of scope for the change to `unordered_map`?

fnRaihanKibria: Yes, definitely not a deal breaker for this PR.

fnRaihanKibria (approved):
LGTM

foundrytom:
@fnRaihanKibria just realised I'd missed updated the previously merged #281

tests/openassetio/test_traits.py
```diff
@@ -12,15 +12,15 @@
...
-        empty_specification = specification.Specification([])
-        assert empty_specification.traitIds() == []
+        empty_specification = specification.Specification(set())
+        assert empty_specification.traitIds() == set()
 
     def test_when_has_traits_returns_expected_trait_ids(self):
-        expected_ids = ["a", "b", "🐠🐟🐠🐟"]
+        expected_ids = {"a", "b", "🐠🐟🐠🐟"}
         populated_specification = specification.Specification(expected_ids)
         ## TODO: Should we use set instead of vector for TraitIds, as
```
feltech, line 21: [fix] Can remove this TODO now :)

feltech (approved):
LGTM other than removing the defunct TODO. Once fixed I'm happy for squash/merge without me seeing it again, if you like.

tests/openassetio/test_traits.py
foundrytom: [Gone](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/305/commits/8ed138c9e80b376c6c74f7c89e6095226af094a9#diff-29adf34bb4665be39936d3cbfcf6d37b4659d84d8fcb4a218f49fe75151079ffL21) :) well remembered, thanks!

PR #249, opened 2022-02-25
[Docs] Update documentation to cover the use of traits

foundrytom:
Updates the main documentation to cover the switch to traits, and the unified entity data model.  See: #247, #246.

For anyone coming to this without the inclination to build the docs, or squint through Doxygen markup, there is a build artefact from the docs test [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/suites/6374903809/artifacts/231024559), that contains the build documentation - see `index.html` (NB. this takes a while to start downloading as it builds the archive on the fly).

@feltech, I've re-based this against HEAD on `feature/181` and fixed up a bunch of `@needsref`s to traits that had been added in the interim.  Giving it a good read through, I've (hopefully) improved various sections.  The first commit is the one you last approved. I've kept the rest as fixups so its hopefully easier to see where things have changed. A commit-by-commit approach may be easier on the brain.

PR #306, opened 2022-03-30
[Core] Migrate `getRelatedReferences` to use trait sets

foundrytom:
Updates the signature for `getRelatedReferences` to take a set of trait IDs rather than a `Specification`. This removes the ambiguity around whether or not the method should consider the properties of the traits vs their identifiers.

This argument is meant to indicate the "type" of entities returned, trait properties form the data of a specific entity.

Closes https://github.com/TheFoundryVisionmongers/OpenAssetIO/issues/282.

NB: This assumes #305 is merged imminently, and defines trait sets as a `set` not `list`/`tuple`.

### DIFF
--- a/tests/openassetio/hostAPI/test_manager.py
+++ b/tests/openassetio/hostAPI/test_manager.py
@@ -26,7 +26,7 @@ from unittest import mock
 
 import pytest
 
-from openassetio import Context
+from openassetio import Context, Specification
 from openassetio.specifications import EntitySpecification
 from openassetio.hostAPI import Manager
 from openassetio.managerAPI import HostSession, ManagerInterface
@@ -193,7 +193,15 @@ class ValidatingMockManagerInterface(ManagerInterface):
         return mock.DEFAULT
 
     def getRelatedReferences(
-            self, entityRefs, relationshipSpecs, context, hostSession, resultSpec=None):
+            self, entityRefs, relationshipSpecs, context, hostSession, resultTraitSet=None):
+        self.__assertIsIterableOf(entityRefs, str)
+        self.__assertIsIterableOf(relationshipSpecs, Specification)
+        self.__assertCallingContext(context, hostSession)
+        if resultTraitSet is not None:
+            assert isinstance(resultTraitSet, set)
+            self.__assertIsIterableOf(resultTraitSet, str)
+
+
         return mock.DEFAULT
 
     def register(self, primaryStrings, targetEntityRefs, entitySpecs, context, hostSession):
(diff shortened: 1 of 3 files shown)


tests/openassetio/hostAPI/test_manager.py
```diff
@@ -570,14 +585,16 @@ def test_wraps_the_corresponding_method_of_the_held_interface(
             assert manager.getRelatedReferences(
                 refs_arg, specs_arg, a_context) == method.return_value
             method.assert_called_once_with(
-                expected_refs_arg, expected_specs_arg, a_context, host_session, resultSpec=None)
+                expected_refs_arg, expected_specs_arg, a_context, host_session,
+                resultTraitSet=None)
             method.reset_mock()
 
-        # Check optional resultSpec
+        # Check optional resultTraitSet
```
carmenpinto, line 592: [minor] Do we need to include the type of the collection in the name? maybe it can just be resultTraits to make it shorter to type/read.

foundrytom: @feltech and I debated this when we worked on #301. We wanted to avoid the ambiguity of just saying "traits" for two reasons:

 - In many cases, this argument is present in a batched call, where an argument is a list of trait sets. As traits is already plural, it causes confusion (eg: [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/297/files#diff-0e6e2f71c0e91d80fc6b3ffa29832f79a53f4372224bafc9336b03859a7648e7R459). It felt like saying 'traitSets' was clear that it was multiple trait sets, rather than "traits" which seemed more readily interpreted as multiple trait IDs rather than multiple sets of ids. We need to be consistent in naming across all parameters that take this type, and so it becomes particularly relevant in this specific case - as it makes it clear that this argument is constant across the batch. If it said "resultTraits" it could be construed as a one per batch element (as `entityRefs` and `relationshipSpecs` are).

 - Given that traits can be instantiated, and this only refers to a set of IDs. Colloqually, for #249 and other documentation updates around the unified data model we wanted to work on a coherent use of language, we regularly use "the set of traits" in that documentation, to the point where a "trait set" almost becomes a proper noun for a collection of trait IDs.  As such, we tried to use that in all the places where a method takes a set of IDs (rather than a specification with data for these traits) for consistency , to try make it not sound like "some multiple trait instances". It's more tenuous than the first reason, but hopefully reinforceable by the documentation.

carmenpinto: Sounds like fair reasoning.

feltech (approved):
LGTM, for my part ready to squash and merge.

Issue #286, opened 2022-03-18
Migrate API to use traits and/or new specifications


foundrytom:
## What

The methods on `ManagerInterface` should be updated to use the new specifiation/trait API.

## Why

See: #253

## Tasks

- [x] #280 
- [x] #281 
- [x] #282 
- [x] #283 
- [x] #269 
- [x] #287
- [x] #342
- [x] #249

Issue #284, opened 2022-03-18
Update `resolveEntityReference`/`register` to use unified entity data model


foundrytom:
## What

`resolveEntityReference` and `register` should be updated such that the `primiaryString` is removed and the specification is used as the source of data.

## Why

The primary string model is outmoded as of #253.

## Acceptance Critiera

- `register` updated to remove the primary string, and take all data from the supplied specification.
- `resolveEntityReference` updated to take a list of requested Trait IDs
- `resolveEntityReference` renamed to `resolve`
- Documentation updated
- Example manager plugins updated

foundrytom:
Duplicate of #269

Issue #269, opened 2022-03-16
[Core] Switch to unified entity data model


foundrytom:
## What

Implement the change outlined #247, #253 in the Python `ManagerInterface` class and `Manager` wrapper.

## Why

Performance concerns have been mitigated, and the benefits of the new API recognised. We are seeing third parties keen to start implementing a manager plugin, and so affecting this change means the won't need to re-write resolve as soon as they've written it.

## Out of scope

- C++ `ManagerInterface` work

## Acceptance criteria

- [x] #307
- [x] Remove `resolveEntityReference`, add `resolve` w/list of requested Trait IDs.
- [x] `register` updated to remove the primary string, and take all data from the supplied specification.

PR #308, opened 2022-03-30
[Core] Remove attributes API

### LINKED ISSUE #307: Remove `get/setAttributes`
Part of #269, removes the old attributes mechanism.

foundrytom:
As part of #286, the attributes mechanism goes away in favour of registering/resolving traits and their properties. This change is discussed in more detail in DR008.

This interim commit removes the attributes API, and updates the surrounding documentation.

The only documentation left un-altered is the "Entities, Specifications and Attributes" page, which is tackled independently under #249.

Closes #307


doc/src/Thumbnails.dox
```diff
@@ -41,8 +43,8 @@
  *
  * @see @ref example_generating_a_thumbnail
  *
- * @todo Document the attributes used to determine width/hight/format,
- * these should be the same as the fields of the
+ * @todo Document the @needsref traits used to determine
+ * width/hight/format, these should be the same as the fields of the
```
carmenpinto, line 47: [fixup] height
```suggestion
 * width/height/format, these should be the same as the fields of the
```

carmenpinto:
[question] How will the host communicate the values of the entity trait properties that they're registering with the removal of setting the attributes?

doc/src/Examples.dox
```diff
@@ -236,7 +236,6 @@
  * # See if the manager can tell us where to put it, and what it should be like
  * if policy & constants.kWillManagePath:
  *    thumbnail_path = manager.resolveEntityReference([thumbnail_ref], context)[0]
- *    thumbnail_attr.update(manager.getEntityAttributes([thumbnail_ref], context)[0])
  *
  * # Generate a thumbnail using the supplied criteria
  * mk_thumbnail(thumbnail_path, thumbnail_attr["width"], thumbnail_attr["height"])
```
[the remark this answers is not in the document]
foundrytom: Was saving the bulk of the update of this till we do #284 - as the whole code will change then, and that'll all get updated from the traits and refactored. Happy to wait till then or can look at it now?

carmenpinto:
> > [question] How will the host communicate the values of the entity trait properties that they're registering with the removal of setting the attributes?
> 
> #269 makes `register`/`resolve` symmetrical (and also solves #4 for attributes ne metadata).

Great, that answers my question.

doc/src/MainPage.dox
```diff
@@ -235,13 +235,12 @@
  * rules for the manager:
  *
  * - Store and recall a @ref primary_string
- * - Store and recall @ref attributes
  * - Allow filtering by @ref Specification
```
fnRaihanKibria, line 238: - Just above it says "three simple rules", which is just two now?
- We mention primary strings here only, should this be something more generic? "Data"?

foundrytom: Good spot, thanks, will fix the count. Primary string will be removed in #269, as it still exists as of this PR.

foundrytom: Updated text [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/308/commits/939765c1b398a9facad409592dddf906752bcec8).

doc/src/EntitySpecMeta.dox
```diff
@@ -60,8 +60,8 @@
  * mapped to a manager's internal classification mechanisms. Additional
  * data from the specification's key/value pairs is not persisted beyond
  * the lifetime of the related API calls in any literal form. If a host
- * wishes to persist any other information about an entity, use the @ref
- * attributes mechanism instead.
+ * wishes to persist any other information about an entity, use the
+ * @needsref attributes mechanism instead.
```
feltech, line 64: [question] Should this be
```suggestion
 * @needsref traits mechanism instead.
```
(similarly below)

doc/src/Transactions.dox
```diff
@@ -35,8 +41,8 @@
  *    manager.startTransaction(context)
  *    try:
  *        # Create the shot first, so we can then publish the clips to it
- *        shot_ref = manager.register([shot.name], [shot.attributes],
- *                                    [shot.spec], [entity_ref], context)[0]
+ *        shot_ref = manager.register([shot.name], [shot.spec],
+ *                                    [entity_ref], context)[0]
```
feltech, line 45: [plan] What do we think about destructuring assignment?  I.e.
```suggestion
 *        [shot_ref] = manager.register([shot.name], [shot.spec],
 *                                    [entity_ref], context)
```

doc/src/EntitySpecMeta.dox
feltech: Indeed, there is a whole `@section` for Attributes below.

Ah, I should read the description
> The only documentation left un-altered is the "Entities, Specifications and Attributes" page, which is tackled independently under https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/249.

So I take it the `@needsref`s in this file are just so we can't forget to update later?

feltech (approved):
LGTM, for my part ready to squash and merge..

doc/src/EntitySpecMeta.dox
foundrytom: Yeah - and to stop the docs build failing

doc/src/Transactions.dox
foundrytom: Ahh yeah, that's much better isn't it, does that then also error if there is more than one with "too may values to unpack"

foundrytom: Added a note to https://github.com/TheFoundryVisionmongers/OpenAssetIO/issues/287

PR #309, opened 2022-03-30
[Build] Cache test requirements locally

### LINKED ISSUE #300: Can't build offline
## Observed behaviour

Running `ctest` on an existing and previously built and run codebase checkout fails if there is no internet connection:

```
1:   Installing build dependencies: finished with status 'error'
1:   ERROR: Command errored out with exit status 1:
1:    command: /src/build/src/openassetio-python/../../../dist/bin/python /tmp/pip-standalone-pip-gbgcwc8x/__env_pip__.zip/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-190x2gxq/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- 'setuptools>=40.8.0' wheel
1:        cwd: None
1:   Complete output (9 lines):
1:   WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7f421f964580>: Failed to establish a new connection: [Errno -2] Name or service not known')': /simple/setuptools/
```

## Expected behaviour

Once the dependencies have been retrieved/installed (or otherwise made available), the dev cycle should not require an active internet connection.

## Repro steps

1. Check out the repository
2. Run `cmake -S . -B build -DOPENASSETIO_ENABLE_TESTS=ON && cd build && ctest`
3. Disable internet connectivity
4. Run `ctest`, the error above will be displayed and the run aborted.

fnRaihanKibria:
Running ctest on an existing and previously built and run codebase
checkout fails if there is no internet connection because pip will
attempt to look online for the packages. The pip documentation
recommends caching the archives locally and then referring to them in
the install step:
https://pip.pypa.io/en/stable/user_guide/#installing-from-local-packages
For unknown reasons setuptools and wheel are not archived
automatically, so I had to manually store them in an additional step.
The archives are stored in BUILD/downloads.

Closes #300

Signed-off-by: Raihan Kibria <raihan.kibria@foundry.com>

### DIFF
--- a/src/openassetio-python/CMakeLists.txt
+++ b/src/openassetio-python/CMakeLists.txt
@@ -107,20 +107,55 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
         DEPENDS "${installed_pip}"
     )
 
+    # One-time step to download test-specific dependencies
+    # so we don't always need an online connection for running ctest
+    set(deps_dir "${CMAKE_BINARY_DIR}/dependencies")
+    set(python_deps_dir "${deps_dir}/python")
+    set(test_deps_dir "${python_deps_dir}/tests")
+    set(test_deps_downloaded "${deps_dir}/testDepsDwnld.out")
+    add_custom_command(
+        OUTPUT "${test_deps_downloaded}"
+        COMMAND
+        "${installed_pip}" "download" "--destination-directory" "${test_deps_dir}"
+            "-r" "${PROJECT_SOURCE_DIR}/tests/requirements.txt"
+        COMMAND
+        cmake -E touch "${test_deps_downloaded}"
+        DEPENDS "${installed_pip}" "${PROJECT_SOURCE_DIR}/tests/requirements.txt"
+    )
+
+    # One-time step to download project-specific dependencies
+    set(project_deps_dir "${python_deps_dir}/project")
+    set(project_deps_downloaded "${deps_dir}/projDepsDwnld.out")
+    add_custom_command(
+        OUTPUT "${project_deps_downloaded}"
+        # Need to cache the build tools explicitely due to build isolation, see
+        # https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#build-process
+        COMMAND
+        "${installed_pip}" wheel --wheel-dir ${project_deps_dir} setuptools wheel
+        COMMAND
+        cmake -E touch "${project_deps_downloaded}"
+        DEPENDS "${installed_pip}"
+    )
+
     # Command to install test-specific dependencies (e.g. pytest).
     add_custom_command(
         OUTPUT
         "${venv_python_dir}/pytest"
         COMMAND
-        "${installed_pip}" install -r "${PROJECT_SOURCE_DIR}/tests/requirements.txt"
-        DEPENDS "${installed_pip}"
+        "${installed_pip}" "install" "--no-index"
+            "--find-links" "${test_deps_dir}"
+            "-r" "${PROJECT_SOURCE_DIR}/tests/requirements.txt"
+        DEPENDS "${test_deps_downloaded}"
     )
 
     # Target to install the pure Python component of the project.
     add_custom_target(
         openassetio-python-py-install
-        COMMAND "${installed_pip}" install "${PROJECT_SOURCE_DIR}"
-        DEPENDS "${installed_pip}"
+        COMMAND
+        "${installed_pip}" "install" "--no-index"
+            "--find-links" "${project_deps_dir}"
+            "${PROJECT_SOURCE_DIR}"
+        DEPENDS "${project_deps_downloaded}"
     )
     # Ensure pre-requisite C++ component has been installed first.
     add_dependencies(openassetio-python-py-install openassetio-install)


src/openassetio-python/CMakeLists.txt
```diff
@@ -112,15 +113,19 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
         OUTPUT
         "${venv_python_dir}/pytest"
         COMMAND
-        "${installed_pip}" install -r "${PROJECT_SOURCE_DIR}/tests/requirements.txt"
+        "${installed_pip}" download --destination-directory ${download_dir} -r "${PROJECT_SOURCE_DIR}/tests/requirements.txt"
```
foundrytom, line 116: [fix] Can we add some notes on why we're downloading these? (ie, as the wheel build insists on installing things).

src/openassetio-python/CMakeLists.txt
```diff
@@ -112,15 +113,19 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
...
         DEPENDS "${installed_pip}"
     )
 
     # Target to install the pure Python component of the project.
     add_custom_target(
         openassetio-python-py-install
-        COMMAND "${installed_pip}" install "${PROJECT_SOURCE_DIR}"
-        DEPENDS "${installed_pip}"
+        COMMAND "${installed_pip}" install  --no-index --find-links ${download_dir} "${PROJECT_SOURCE_DIR}"
+        DEPENDS "${installed_pip}" "${venv_python_dir}/pytest"
```
foundrytom, line 128: [question] Is this additional dependency required?

foundrytom (changes requested):
Great to have this fixed @fnRaihanKibria, thanks!

src/openassetio-python/CMakeLists.txt
fnRaihanKibria: It was required, the build can fail because the install requires those downloaded files and without the dependency this step may run before it, and failing obviously (I tried this out).

fnRaihanKibria: It would be cleaner to put the package download into a separate step, which could also depend on the requirements.txt, and then let both the pytest- and openassetio-installer step depend on that one. Not hard to do and more maintainable. I will change it.

src/openassetio-python/CMakeLists.txt
fnRaihanKibria: Will add a note as I refactor the dependencies as mentioned below.

src/openassetio-python/CMakeLists.txt
foundrytom: Thanks - sounds like a good plan - then it's really clear about the coupling between the stages.

src/openassetio-python/CMakeLists.txt
fnRaihanKibria: Addressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309/commits/d5e0db78ef4e78635c7d3343eb825a936d2ccaa6

src/openassetio-python/CMakeLists.txt
fnRaihanKibria: Addressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309/commits/d5e0db78ef4e78635c7d3343eb825a936d2ccaa6

foundrytom (approved):
Thanks for those changes Raihan, LGTM if @feltech is happy with his larger CMake brain.

feltech:
For info, good explanation here: https://stackoverflow.com/a/67595330/535103

TL;DR it's because of pip's relatively new "build isolation" default, where build dependencies (i.e. `setuptools` and `wheel`) are installed in some temporary build directory.  

This can be disabled (`--no-build-isolation`) but doing that has it's own problems (e.g. requires `wheel` to be separately installed in the virtualenv, adding unnecessary pollution).

I think on balance the approach taken in this PR seems like the best option. Though I'm happy if you want to experiment with that flag in case it simplifies things @fnRaihanKibria

feltech (changes requested):

src/openassetio-python/CMakeLists.txt
```diff
@@ -107,20 +107,47 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
...
+    # Download dependencies into this directory
+    # so we don't always need an online connection for running ctest
+    set(download_dir "${CMAKE_BINARY_DIR}/downloads")
+
+    # One-time step to download test-specific dependencies
+    set(test_deps_downloaded "${download_dir}/test-deps-downloaded.txt")
+    add_custom_command(
+        OUTPUT "${test_deps_downloaded}"
+        COMMAND
+        "${installed_pip}" download --destination-directory ${download_dir} -r "${PROJECT_SOURCE_DIR}/tests/requirements.txt"
```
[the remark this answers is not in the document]
fnRaihanKibria: Addressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309/commits/b73fd650a02596cf37a2eca916ee578b7839323b

src/openassetio-python/CMakeLists.txt
```diff
@@ -107,20 +107,47 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
...
     )
 
+    # Download dependencies into this directory
+    # so we don't always need an online connection for running ctest
+    set(download_dir "${CMAKE_BINARY_DIR}/downloads")
+
+    # One-time step to download test-specific dependencies
+    set(test_deps_downloaded "${download_dir}/test-deps-downloaded.txt")
+    add_custom_command(
+        OUTPUT "${test_deps_downloaded}"
```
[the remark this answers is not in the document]
fnRaihanKibria: Addressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309/commits/b73fd650a02596cf37a2eca916ee578b7839323b

src/openassetio-python/CMakeLists.txt
```diff
@@ -107,20 +107,47 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
...
+        COMMAND
+        "${installed_pip}" "download" "--destination-directory" "${test_deps_dir}"
+            "-r" "${PROJECT_SOURCE_DIR}/tests/requirements.txt"
+        DEPENDS "${installed_pip}" "${PROJECT_SOURCE_DIR}/tests/requirements.txt"
+    )
+
+    # One-time step to download project-specific dependencies
+    set(project_deps_dir "${CMAKE_BINARY_DIR}/projDeps")
+    add_custom_command(
+        OUTPUT "${project_deps_dir}"
```
feltech, line 124: [fix] My bad, the downside of using a directory is that the `clean` target fails (at least on Linux+Ninja)

From my IDE
```
/usr/local/bin/cmake --build /home/dave/workspace/cloud/assetapi/OpenAssetIO/cmake-build-debug --target clean
[1/1] /snap/clion/184/bin/ninja/linux/ninja  -t clean 
FAILED: clean 
/snap/clion/184/bin/ninja/linux/ninja  -t clean 
Cleaning... ninja: error: remove(testDeps): Directory not empty
ninja: error: remove(/home/dave/workspace/cloud/assetapi/OpenAssetIO/cmake-build-debug/testDeps): Directory not empty
ninja: error: remove(projDeps): Directory not empty
ninja: error: remove(/home/dave/workspace/cloud/assetapi/OpenAssetIO/cmake-build-debug/projDeps): Directory not empty
13 files.
ninja: build stopped: subcommand failed.
```

One way around it that works is to add `ADDITIONAL_CLEAN_FILES` to a _target_ (can't be added to the custom command), e.g.
```
set_target_properties(openassetio-python-py-install PROPERTIES ADDITIONAL_CLEAN_FILES ${project_deps_dir})
```

To do this neatly I think we would have to have "dummy" targets that wrap the custom command, e.g.
```
add_custom_target(
    openassetio-python-download-project-deps
    DEPENDS "${project_deps_dir}"
)
set_target_properties(
    openassetio-python-download-project-deps 
    PROPERTIES ADDITIONAL_CLEAN_FILES ${project_deps_dir})
```
then depend on that target where `${project_deps_dir}` is required. This actually matches what [the docs](https://cmake.org/cmake/help/v3.22/command/add_custom_command.html) say

> Do not list the output in more than one independent target that may build in parallel or the two instances of the rule may conflict (instead use the [add_custom_target()](https://cmake.org/cmake/help/latest/command/add_custom_target.html#command:add_custom_target) command to drive the command and make the other targets depend on that one)

At this point, maybe having a signal file isn't so bad - it would mean `clean` would only clean the signal file not the downloaded content, but perhaps that would be an advantage...?

I'll leave the approach up to you.

feltech (changes requested):

src/openassetio-python/CMakeLists.txt
fnRaihanKibria: A general thought: I think it would be better that running clean should not delete those cached dependencies, or otherwise you destroy the ability to test once you run it and you happen to be without Internet at the time. What do you think?

fnRaihanKibria: Note: make clean already was problematic before this change I think. If you run it, it will delete the pip executable in dist/bin and the test fails.

fnRaihanKibria: The pip install's need a --force-reinstall option or the output files are not regenerated correctly, I just found.

feltech: Eek, good spot!

fnRaihanKibria: Addressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309/commits/c169f8fec784011e7c838c9c618ba36867a817f4

src/openassetio-python/CMakeLists.txt
```diff
@@ -99,28 +99,63 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
         COMMAND
         "${Python_EXECUTABLE}" -m venv "${CMAKE_INSTALL_PREFIX}"
         COMMAND
-        ${venv_python_dir}/python -m pip install --upgrade pip
+        ${venv_python_dir}/python -m pip install --upgrade --force-reinstall pip
```
feltech, line 102: [plan] This fails if I `clean` then disconnect from the internet then attempt to build. 

Arguably that's to be expected, and regardless is probably out of scope for this PR, so we should make a separate issue.

I think CMake `clean` deleting (only) the pip binary is just wrong and we should ultimately find a better solution. When I originally wrote this CMake I didn't consider the effect of `clean` on the `OUTPUT` of an `add_custom_command`, so my bad.

I'll make an issue.

feltech (approved):
Works well, as long as I don't `clean` before attempting to build offline, which is worth a separate issue.

src/openassetio-python/CMakeLists.txt
```diff
@@ -107,20 +107,53 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
         DEPENDS "${installed_pip}"
     )
 
+    # One-time step to download test-specific dependencies
+    # so we don't always need an online connection for running ctest
+    set(test_deps_dir "${CMAKE_BINARY_DIR}/testDeps")
+    set(test_deps_downloaded "${CMAKE_BINARY_DIR}/testDepsDwnld.txt")
```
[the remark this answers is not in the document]
fnRaihanKibria: It was arbitrary here, I've seen this extension used elsewhere and stuck with it. We could go with no extension, or ".out" possibly?

src/openassetio-python/CMakeLists.txt
feltech: Added #311

src/openassetio-python/CMakeLists.txt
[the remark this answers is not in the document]
fnRaihanKibria: Adressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309/commits/4e7909c651b42db005f6e9b190bd288570d9ad19 (renamed to .out in previous commit)

src/openassetio-python/CMakeLists.txt
```diff
@@ -107,20 +107,53 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
         DEPENDS "${installed_pip}"
     )
 
+    # One-time step to download test-specific dependencies
+    # so we don't always need an online connection for running ctest
+    set(test_deps_dir "${CMAKE_BINARY_DIR}/testDeps")
```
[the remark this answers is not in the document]
fnRaihanKibria: Adressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309/commits/3f9403abf95a74c7cfbbdaf327aa919b93058fe5

src/openassetio-python/CMakeLists.txt
```diff
@@ -99,28 +99,68 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
...
         DEPENDS "${installed_pip}"
     )
 
+    # One-time step to download test-specific dependencies
+    # so we don't always need an online connection for running ctest
+    set(deps_dir "${CMAKE_BINARY_DIR}/dependencies")
+    set(python_deps_dir "${deps_dir}/python")
+    set(test_deps_dir "${python_deps_dir}/tests")
+    # Store signal files under a common directory
+    set(signals_dir "${CMAKE_BINARY_DIR}/signals")
```
feltech, line 116: [consider][minor] Apologies for the churn, but would it be neater to put the signal files in the `dependencies` directory, rather than a separate directory?

src/openassetio-python/CMakeLists.txt
```diff
@@ -99,28 +99,68 @@ if (OPENASSETIO_ENABLE_TESTS AND Python_Interpreter_FOUND)
...
 
     # Target to install the pure Python component of the project.
     add_custom_target(
         openassetio-python-py-install
-        COMMAND "${installed_pip}" install "${PROJECT_SOURCE_DIR}"
-        DEPENDS "${installed_pip}"
+        COMMAND
+        "${installed_pip}" "install" "--no-index"
+            "--find-links" "${project_deps_dir}"
+            "--force-reinstall"
```
feltech, line 161: [question] Do we need `--force-reinstall` here?  My understanding was that we need to force in case a `clean` had destroyed an `OUTPUT`, but there is no `OUTPUT` here.

feltech (approved):

src/openassetio-python/CMakeLists.txt
fnRaihanKibria: Addressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309/commits/77d7dd99809c24f57d8bf11163d99a1e710ff983

src/openassetio-python/CMakeLists.txt
fnRaihanKibria: Addressed in https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309/commits/77d7dd99809c24f57d8bf11163d99a1e710ff983

Removed all force installs since they don't completely fix the issues with using clean.

feltech: :+1: to reiterate https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309#discussion_r842652391- we added #311 to deal with this separately.

fnRaihanKibria:
@feltech Squashed and ready to merge.

Issue #293, opened 2022-03-21
C API structure spike


feltech:

feltech:
I have posted PR #310, which includes a markdown doc and proposed implementation for discussion. I figured this would be better to iterate on in case of review comments, rather than having a lot of churn in this issue.  The markdown doc can be copied to this issue and the PR abandoned if/when we're happy with it.

Rendered output of the initial markdown doc [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/80dba9bcdf78af7a4819d0b1a69db7b6bb232966/decisions/C-ManagerInterface.md)

feltech:
The PR has had some good feedback, thankfully nothing fundamentally wrong, so I'll post the summary below.

feltech:
# Initial C API Proposal

This proposal is scoped to the C API bindings for the minimal 
`ManagerInterface` methods that have been ported to C++ to date, 
specifically `identifier()` and `displayName()`. Despite this tight
restriction in scope there are several design decisions to be addressed.

## Namespacing

In C, exported symbols must be namespaced by prefixing the symbol name
with some string.

Since our top-level namespace includes a version tag, which is likely to
change, we must use a macro to ease maintenance, e.g.

```c++
#define OPENASSETIO_NS(symbol) ...
```
such that `OPENASSETIO_NS(MyStruct)` expands to 
`openassetio_v0_0_MyStruct`.

## Strings

Taking inspiration from [this conversation](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/194#discussion_r794047468)
we can bundle a string buffer and length in a single C struct for error
messages and parameters, e.g.

```c++
typedef struct {
  const size_t maxSize;
  char* buffer;
  size_t usedSize;
} OPENASSETIO_NS(SimpleString);
```
`usedSize` can technically be removed if we're happy to assume
null-termination.

However, having a `usedSize` member is more optimal, avoiding the need
to re-measure the string's length as it is passed around.

In addition, other string representations (e.g. in other languages using
the C API through an FFI) may not use null-termination as standard, so
including a `usedSize` member improves language support.

Note that with UTF-8 the number of printed characters might not match
the number of characters required for encoding, so we mustn't naively
assume that `usedSize` reflects the rendered string's length.

## Opaque handles

The opaque handle pattern used commonly in C APIs involves an arbitrary
pointer to some user data, often coded simply as a `void*`. To add some
[...]

PR #310, opened 2022-04-01
[Temp] C API sketch

feltech:
For #293 this PR presents an initial implementation of the C API for the basic `ManagerInterface` consisting of only `identifier()` and `displayName()` methods. A markdown doc is included detailing the design choices made. 

This PR is just to provoke discussion and will not be merged.

See [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/c131aa9e5a0d8b7d0c470428b849cabec3cb5d69/decisions/C-ManagerInterface.md) for rendered markdown output (EDIT: updated to latest version).


decisions/C-ManagerInterface.md
```diff
@@ -0,0 +1,205 @@
...
+typedef struct {
+  const size_t maxSize;
+  char* buffer;
+  size_t usedSize;
+} OPENASSETIO_NS(SimpleString);
+```
+`usedSize` is for optimisation and safety and can technically be removed
+if we're happy to assume null-termination and accept the performance
+implications.
+
```
meshula, line 38: This form, with a length, means that language bindings don't need to measure the string upon receiving it, nice for non-zero-terminated-string languages. If the buffer has an unmeasured zero at the end of the string, it means that it's C/C++/Python friendly as well.

decisions/C-ManagerInterface.md
```diff
@@ -0,0 +1,205 @@
...
+pointer to some user data, often coded simply as a `void*`. To add some
+meagre type safety, we can define an opaque handle as a pointer to an
+incomplete and ultimately unused class. We then `reinterpret_cast` it 
+to/from our actual pointer type on the C++ side. E.g.
+
+```c++
+typedef struct OPENASSETIO_NS(managerAPI_ManagerInterface_t) *
+    OPENASSETIO_NS(managerAPI_ManagerInterface_h);
+```
+
```
meshula, line 51: I imagine this boiler-plater pattern with an intermediate type would be further wrapped in a macro for ease of use? Might it be better to just use _p instead of _h? Handle has historically often meant **, whereas this is a *.

decisions/C-ManagerInterface.md
```diff
@@ -0,0 +1,205 @@
...
+`usedSize` is for optimisation and safety and can technically be removed
+if we're happy to assume null-termination and accept the performance
+implications.
+
+## Opaque handles
+
+The opaque handle pattern used commonly in C APIs involves an arbitrary
+pointer to some user data, often coded simply as a `void*`. To add some
+meagre type safety, we can define an opaque handle as a pointer to an
+incomplete and ultimately unused class. We then `reinterpret_cast` it
```
meshula, line 44: You can get strong type safety with a wrapping pattern. i.e. put the type-erased pointer inside a strongly typed struct.

```
struct FooOpaque { void* p; };
```

decisions/C-ManagerInterface.md
```diff
@@ -0,0 +1,205 @@
...
+
+C++ functions may or may not return a value and may or may not throw an
+exception. C doesn't have exceptions, so to distinguish different error
+types we must use another mechanism. Commonly (and used here) this is
+via (integer) error codes. 
+
+Given this, a C API function wrapping a C++ class method has between 0
+and 3 possible output values, depending on whether a return value is
+expected and whether an error code and message (i.e. exception) is
+possible.
```
meshula, line 94: I wonder if it makes sense to say "(i.e. an exceptional condition)" here, to avoid the C++ term "exception". The reason is I had to read this section a couple of times to convince myself that it doesn't allow a C++ routine to throw across the C interface.

decisions/C-ManagerInterface.md
```diff
@@ -0,0 +1,205 @@
...
+```
+src/openassetio-core-c
+├── CMakeLists.txt
+├── include
+│   └── openassetio
+│       └── c
+│           ├── errors.h
+│           ├── managerAPI
+│           │   └── ManagerInterface.h
+│           ├── ns.h
```
meshula, line 193: maybe call this one namespace.h? ns.h makes me wonder if it is full of ObjC stuff, since every Apple Foundation class is named NSSomething.

meshula (commented):
I like the looks of it. Some random comments in line.

decisions/C-ManagerInterface.md
feltech: Interesting point on `_p` vs `_h`, we'll re-evaluate that. `_h` has the advantage of matching the terminology of "opaque handle", maybe we should switch to "opaque pointer"?

I wondered about wrapping in a short-form macro, a couple of considerations though
* The short-form macro surely requires some level of namespacing regardless? I'm not sure what the appropriate short-form name should be, whilst still avoiding conflicts. I guess we could remove the namespace entirely for a short-form, then its up to the user to be grown up and ensure they don't mix headers that have conflicting symbol/macro names?
* We envision the C API being primarily a binding layer - we don't really expect many people to develop in pure C - so looking overly verbose is perhaps not a big problem, anyway.

decisions/C-ManagerInterface.md
feltech: Ah yes, is there a compelling advantage of one approach over the other?  I guess `static_cast` can be used in your example, which is nice.

decisions/C-ManagerInterface.md
feltech: I've re-worded slightly in 46a0a3c0 - hopefully that removes any initial confusion.

decisions/C-ManagerInterface.md
feltech: Renamed as suggested in 4d4b589a

decisions/C-ManagerInterface.md
feltech: Good point about language support, where they may not use null-termination. Updated in a9a2a74c, where I also added a brief note re. UTF-8 and string lengths.

foundrytom:
Superseded by #320 and #312.

decisions/C-ManagerInterface.md
meshula: Yes, there's a compelling advantage :)

```cpp
namespace clarity {
    // clarity in legibility:
    typedef void* EntityPtr;
    typedef void* ContainerPtr;

    // unsafe in that compiler doesn't care:
    ContainerPtr a = CreateContainer();
    EntityPtr b = CreateEntity();
    ContainerPtr c = a;   // compiler ignores error
    ContainerAddEntity(b, b); // compiler ignores error
}

namespace clarity_and_safety {
    struct EntityPtr { void* p };
    struct ContainerPtr { void* p };

    // safe in that compiler will disallow assignment:
    ContainerPtr a = CreateContainer();
    EntityPtr b = CreateEntity();
    ContainerPtr c = a;   // compiler reports error
    ContainerAddEntity(b, b); // compiler reports error
}
```
In both cases, the types are `sizeof(void*)` and can be passed with the same expense, and the pattern works for C and C++. The C++ side can have template and operator shenanigans to add utility to fetching out p such that the C++ code can look idiomatic without casting.

feltech: Thanks for getting back to me. 

I guess by "compelling advantage" I meant versus the pointer-to-incomplete-type solution, i.e.
```c++
namespace clarity_and_safety_alternative {
    typedef struct UnusedIncompleteEntityType * EntityPtr;
    typedef struct UnusedIncompleteContainerType * ContainerPtr;

    ContainerPtr a = CreateContainer(); // Will `reinterpret_cast` real ptr to `ContainerPtr`
    EntityPtr b = CreateEntity(); // Will `reinterpret_cast` real ptr to `EntityPtr`
    ContainerPtr c = a;   // compiler reports error
    ContainerAddEntity(b, b); // compiler reports error
}
```

meshula: Ah :) `struct Foo { void* p };` is handled as a type, whereas `typedef struct Foo* FooPtr` is handled as a pointer to a type. To clarify a bit, what I've been doing is `struct Foo { uintptr_t p; };` so that no magic in any ffi or other binding mechanism attempts to do ANYTHING with the pointer ~ it's completely type erased and binding languages tend to understand how to copy structs with POD guts.

feltech: Thanks a lot for the clarification.  I really like the `uintptr_t` version, especially your argument about removing any ambiguity for FFI magic.  

We circulated the options internally, which of course just threw up yet more variations we hadn't considered :roll_eyes: 

But the TL;DR is that we're sticking with the classic opaque handle type, rather than wrapping a pointer in a struct, at least for now.

At least this is one of those decisions that doesn't require a major code overhaul if we change our mind.

PR #313, opened 2022-04-07
[Build] Move clang-format checks into CMake

### LINKED ISSUE #251: [Build] Move `clang-format` checks into `CMake`

feltech:
Closes #251


CMakeLists.txt
```diff
@@ -172,6 +172,9 @@ endif ()
 # Enable unit tests.
 option(OPENASSETIO_ENABLE_TESTS "Create test targets" OFF)
 
+# Enable clang-format formatting check.
+option(OPENASSETIO_ENABLE_CLANG_FORMAT "Enable clang-format check during build" OFF)
+
 # Enable clang-tidy static analysis tool.
 option(OPENASSETIO_ENABLE_CLANG_TIDY "Enable clang-tidy analysis during build" OFF)
```
foundrytom, line 179: [question] If these are off by default, should we have a 'dev' CMake preset that turns all this on, along with tests? Or just turn them on? It'd be great if the default experience was overzealous, as we should never be merging anything to a release branch where these two checks being `ON` would fail.

.github/workflows/code-quality.yml
```diff
@@ -71,6 +61,7 @@ jobs:
           --toolchain $HOME/conan/conan_paths.cmake
           -DCMAKE_CXX_STANDARD=17
           -DOPENASSETIO_WARNINGS_AS_ERRORS=ON
+          -DOPENASSETIO_ENABLE_CLANG_FORMAT=ON 
           -DOPENASSETIO_ENABLE_CLANG_TIDY=ON
           -DOPENASSETIO_ENABLE_CPPLINT=ON
```
foundrytom, line 66: [fix] Seems like we definitely need a CMake preset for this now, so we aren't relying on CI magic, and its easy to recreate locally.

foundrytom (changes requested):
Great to see this built into CMake now @feltech, think we just need to make the invocation more uniform so local dev "does the right thing" by default as much as it can.

CMakeLists.txt
feltech: The default really needs to be what we would expect a parent project to want if OpenAssetIO was used as a subproject.

Otherwise yeah I agree it's about time we had some presets. Is it worth doing as part of this PR or another issue?

.github/workflows/code-quality.yml
feltech: OK to convert this to a [plan], and I'll create an issue?

foundrytom: Perfect, thanks @feltech

CMakeLists.txt
foundrytom: Ahh yes, perfect, `OFF` + dev preset sounds ideal. Thanks.

.github/workflows/code-quality.yml
feltech: Added #315

CMakeLists.txt
feltech: Added #315

foundrytom (approved):
Thanks @feltech feel free to squash and merge at your leisure (incl the whitespace changes if you want too).

PR #314, opened 2022-04-07
[CI] Add check to ensure ASWF docker builds work

foundrytom:
Ensures the steps listed in README.md work. Ideally we'd be grabbing this from the file itself, but this is progress...

This does add some time to the CI check suite though 😞 but given we list these instructions in the top-level readme, and compliance with the reference platform is critical, we need to make sure local dev doesn't diverge.

A failing run is available [here](https://github.com/foundrytom/OpenAssetIO/runs/5867268090?check_suite_focus=true) to show errors in the command are reported.

It could be said that this should be part of the `Test` check, but ideally this would be a "check the steps in the documentation work" PR...

### DIFF
--- a/.github/workflows/build-instructions.yml
+++ b/.github/workflows/build-instructions.yml
@@ -1,0 +1,26 @@
+# A simple workflow that runs the documented build steps work.
+
+name: Build Instructions
+on:
+  pull_request:
+    branches-ignore:
+      - docs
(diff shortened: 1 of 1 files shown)


feltech (approved):

feltech:
LGTM rebase and merge when ready

PR #316, opened 2022-04-08
[Core] Migrate `preflight` to use trait sets

foundrytom:
Updates the signature for `preflight` to take a set of trait IDs rather than a `Specification`. This removes the ambiguity around whether or not the method should consider the properties of the traits vs their identifiers.

This argument is meant to indicate the "type" of entities that are about to have their data generated. Supplying the properties for a trait here makes no sense, as that data _is_ the data of the entity, and so the host can't provide it yet, as its not been made!

I think I got all the documentation updates here, but am some what sleepy so let me know if anything has slipped through the cracks!

Closes https://github.com/TheFoundryVisionmongers/OpenAssetIO/issues/283.

### DIFF
--- a/doc/src/Examples.dox
+++ b/doc/src/Examples.dox
@@ -183,7 +183,7 @@
  * # Whenever we make new data, we always tell the manager first,
  * # This allows it to create a placeholder version or similar.
  * # NOTE: It is critical to always use the working_ref from now on.
- * working_ref = manager.preflight([entity_ref], [file_spec], context)[0]
+ * working_ref = manager.preflight([entity_ref], file_spec.traitIds(), context)[0]
  *
  * # We then check if the manager can tell us where to save the file
  * if policy & constants.kWillManagePath:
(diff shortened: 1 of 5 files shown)


doc/src/Examples.dox
```diff
@@ -228,7 +228,7 @@
  * # Preflight the thumbnail spec with the target entity's reference,
  * # this gives us a reference we can now use for all interactions
  * # relating to the thumbnail itself.
- * thumbnail_ref = manager.preflight([final_ref], [thumbnail_spec], context)[0]
+ * thumbnail_ref = manager.preflight([final_ref], thumbnail_spec.traitIds, context)[0]
```
feltech, line 231: [fix] `traitIds` is a method
```suggestion
 * thumbnail_ref = manager.preflight([final_ref], thumbnail_spec.traitIds(), context)[0]
```

python/openassetio/hostAPI/Manager.py
```diff
@@ -966,9 +966,8 @@ def preflight(self, targetEntityRefs, entitySpecs, context):
         @param targetEntityRefs `List[str]` The entity references to
         preflight prior to registration.
 
-        @param entitySpecs `List[`
-            specifications.EntitySpecification `]`
-        A description of each entity that is being published.
+        @param traitSet `Set(str)` The @needsref traitSet of the
```
feltech, line 969: [consider][minor] Here and in `ManagerInterface`, for Python 3 type hint style
```suggestion
        @param traitSet `Set[str]` The @needsref traitSet of the
```

feltech (approved):
LGTM, I can't find anything missed. Just minor typos, happy for you to squash+merge once fixed.

PR #329, opened 2022-04-13
[Core] Validate specification trait sets passed to `register`

### LINKED ISSUE #328: [Core] Validate constant trait sets in `register` batch
## What

Add validation in `Manager.register` that the supplied Specifications all have the same trait set.

## Why

The API contract requires that the specifications passed to `ManagerInterface.register` have constant trait sets (to allow the manager to optimise the batch submission). We should take advantage of the API middleware to validate this, so the `ManagerInterface` implementation doesn't need to check. 

This will add some overhead but:
 - We consider `register` to be a far lower call volume than `resolve` and so consider this overhead negligible for now.
 - Removing it later is safer than adding the check in later, once lax code is already in use.

foundrytom:
Closes #328.

This is a follow up PR to #327, so ignore anything other than [this commit](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/329/commits/a7f8ba6b78f453b68315e96f24e53d01cbf50415) until it is merged.

### DIFF
--- a/python/openassetio/hostAPI/Manager.py
+++ b/python/openassetio/hostAPI/Manager.py
@@ -1079,6 +1079,15 @@ class Manager(Debuggable):
         if len(targetEntityRefs) != len(entitySpecs):
             raise IndexError("Parameter lists must be of the same length")
 
+        if entitySpecs:
+            # Check supplied specifcations share a trait set
+            expectedTraits = entitySpecs[0].traitIds()
+            for i, spec in enumerate(entitySpecs[1:]):
+                specTraits = spec.traitIds()
+                if specTraits != expectedTraits:
+                    raise ValueError(
+                            f"Mismatched traits at index {i+1}: {specTraits} != {expectedTraits}")
+
         return self.__impl.register(targetEntityRefs, entitySpecs, context, self.__hostSession)
 
     ## @}
--- a/tests/python/openassetio/hostAPI/test_manager.py
+++ b/tests/python/openassetio/hostAPI/test_manager.py
@@ -577,3 +577,13 @@ class Test_Manager_register:
 
         with pytest.raises(IndexError):
             manager.register(some_refs, specifications[1:], a_context)
+
+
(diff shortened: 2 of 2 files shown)


python/openassetio/hostAPI/Manager.py
```diff
@@ -1077,6 +1077,15 @@ def register(self, targetEntityRefs, entitySpecs, context):
         if len(targetEntityRefs) != len(entitySpecs):
             raise IndexError("Parameter lists must be of the same length")
 
+        if entitySpecs:
+            # Check supplied specifcations share a trait set
+            expectedTraits = entitySpecs[0].traitIds()
```
foundrytom, line 1082: [question] is `next(iter(entitySpecs))` preferred here, and manually tracking an index, on the off chance that this isn't a `list` or `tuple`?

feltech: An intriguing problem. A Pythonic one-liner, but doesn't short-circuit

```python
len(set(frozenset(s.traitIds()) for s in entitySpecs)) == 1
```

Ideally `traitIds()` would return a `frozenset` (needed to be used in a set of sets).

A quick google reveals solutions using `itertools.groupby`, though I suspect some `frozenset` shenanigans would be required for us, [e.g. from StackOverflow](https://stackoverflow.com/a/3844832/535103)

```python
from itertools import groupby

def all_equal(iterable):
    g = groupby(iterable)
    return next(g, True) and not next(g, False)
````

feltech: Ah but you're trying to make a nice error message, which limits the solution.

feltech: I think it's reasonable to assume, maybe even require, a sequential container that is indexable, so elements can correspond to elements in `targetEntityRefs`.  The only counterexample that springs to mind is a `dict` whose keys correspond, but that would require different treatment in the manager plugin, complicating matters.

feltech (approved):

feltech:
LGTM, rebase and merge when ready.

Issue #290, opened 2022-03-21
Migrate manager `info` (`SimpleMap` - also needed later for trait properties)


feltech:
# What

Add `info` method to the C++ ManagerInterface and inherit the default implementation in Python. Hence craft and bind a simple C++ dictionary class to hold the return value. (Incl C API)

# Why

This is a necessary step toward a C++ manager plugin and a C++->Python bridge.

# ACs

- Use a simple `std::<some>map` type and rely on `pybind` autogen bindings.
- No need for custom exceptions at this point
- C API (with ability to create a new map)
- ~Close~ Update #142.
   - `int64_t`, `double`, `bool`, `std::string`
   - documentation
   - ~solution for dict author to validate if any given data type they have would be accepted by the conversion to C++ (eg when generically converting arbitrary asset system data to a simple map)~

PR #330, opened 2022-04-14
[Core] Add InfoDictionary type(def) with initial C API

feltech:
Part of #290. 

~We may want to think of a better name than `InfoDictionary`.~

Future PRs will expand the C API to include:
* Creating a new `InfoDictionary`from the C API (currently only pre-existing C++ `InfoDictionary`s can be converted).
* Querying the value type of an entry.
* Iteration through entries.
* Writing (insert/update) entries.

### DIFF
--- a/src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
+++ b/src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
@@ -1,0 +1,145 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2022 The Foundry Visionmongers Ltd
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+#include <openassetio/c/export.h>
+
+#include "./StringView.h"
+#include "./errors.h"
+#include "./namespace.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+/**
+ * @addtogroup CAPI C API
+ * @{
+ */
+
+/**
+ * Opaque handle type representing a @fqref{InfoDictionary} "InfoDictionary"
+ * instance.
+ *
+ * The associated @fqcref{InfoDictionary_s} "InfoDictionary suite" functions
+ * operate on this handle type.
+ */
+// NOLINTNEXTLINE(modernize-use-using)
+typedef struct OPENASSETIO_NS(InfoDictionary_t) * OPENASSETIO_NS(InfoDictionary_h);
+
+/**
+ * Function pointer suite for a @fqref{InfoDictionary} "InfoDictionary" C API.
+ *
+ * Instances of this suite are provided by the `InfoDictionary_suite`
+ * function and operate on a @fqcref{InfoDictionary_h} "InfoDictionary handle".
+ */
+// NOLINTNEXTLINE(modernize-use-using)
+typedef struct {
+  /**
+   * Destructor function.
+   *
+   * This should be called by the owner of the handle when the handle is
+   * no longer in use. The underlying object will be destroyed and its
+   * memory freed.
+   *
+   * @param handle Opaque handle to InfoDictionary.
+   */
+  void (*dtor)(OPENASSETIO_NS(InfoDictionary_h) handle);
+
+  /**
+   * Retrieve a boolean value from the map.
+   *
+   * Missing values will result in a @fqcref{ErrorCode_kOutOfRange} error
+   * code.
+   *
+   * Values with the wrong data type will result in a
+   * @fqcref{ErrorCode_kBadVariantAccess} error code.
+   *
+   * @param[out] error Storage for error message, if any
+   * @param[out] out Storage for retrieved value.
+   * @param handle Opaque handle to InfoDictionary.
+   * @param key Key of entry to query.
+   * @return Error code.
+   */
+  OPENASSETIO_NS(ErrorCode)
+  (*getBool)(OPENASSETIO_NS(StringView) * error, bool* out,
+             OPENASSETIO_NS(InfoDictionary_h) handle, const OPENASSETIO_NS(ConstStringView) key);
+
+  /**
+   * Retrieve an integer value from the map.
+   *
+   * Missing values will result in a @fqcref{ErrorCode_kOutOfRange} error
+   * code.
+   *
+   * Values with the wrong data type will result in a
+   * @fqcref{ErrorCode_kBadVariantAccess} error code.
+   *
+   * @param[out] error Storage for error message, if any
+   * @param[out] out Storage for retrieved value.
+   * @param handle Opaque handle to InfoDictionary.
+   * @param key Key of entry to query.
+   * @return Error code.
+   */
+  OPENASSETIO_NS(ErrorCode)
+  (*getInt)(OPENASSETIO_NS(StringView) * error, int64_t* out,
+            OPENASSETIO_NS(InfoDictionary_h) handle, const OPENASSETIO_NS(ConstStringView) key);
+
+  /**
+   * Retrieve a floating point value from the map.
+   *
+   * Missing values will result in a @fqcref{ErrorCode_kOutOfRange} error
+   * code.
+   *
+   * Values with the wrong data type will result in a
+   * @fqcref{ErrorCode_kBadVariantAccess} error code.
+   *
+   * @param[out] error Storage for error message, if any
+   * @param[out] out Storage for retrieved value.
+   * @param handle Opaque handle to InfoDictionary.
+   * @param key Key of entry to query.
+   * @return Error code.
+   */
+  OPENASSETIO_NS(ErrorCode)
+  (*getFloat)(OPENASSETIO_NS(StringView) * error, double* out,
+              OPENASSETIO_NS(InfoDictionary_h) handle, const OPENASSETIO_NS(ConstStringView) key);
+
+  /**
+   * Retrieve a string value from the map.
+   *
+   * Missing values will result in a @fqcref{ErrorCode_kOutOfRange} error
+   * code.
+   *
+   * Values with the wrong data type will result in a
+   * @fqcref{ErrorCode_kBadVariantAccess} error code.
+   *
+   * An `out` parameter with insufficient capacity for the string value
+   * will result in truncation of the string as well as a
+   * @fqcref{ErrorCode_kLengthError} error code.
+   *
+   * @param[out] error Storage for error message, if any
+   * @param[out] out Storage for retrieved value.
+   * @param handle Opaque handle to InfoDictionary.
+   * @param key Key of entry to query.
+   * @return Error code.
+   */
+  OPENASSETIO_NS(ErrorCode)
+  (*getStr)(OPENASSETIO_NS(StringView) * error, OPENASSETIO_NS(StringView) * out,
+            OPENASSETIO_NS(InfoDictionary_h) handle, const OPENASSETIO_NS(ConstStringView) key);
+} OPENASSETIO_NS(InfoDictionary_s);
+
+/**
+ * Get an instance of a @fqcref{InfoDictionary_s} "InfoDictionary suite" of C API
+ * function pointers.
+ *
+ * @return Suite of function pointers.
+ */
+OPENASSETIO_CORE_C_EXPORT OPENASSETIO_NS(InfoDictionary_s) OPENASSETIO_NS(InfoDictionary_suite)();
+
+/**
+ * @}
+ */
+#ifdef __cplusplus
+}
+#endif
--- a/src/openassetio-core-c/include/openassetio/c/errors.h
+++ b/src/openassetio-core-c/include/openassetio/c/errors.h
@@ -17,7 +17,13 @@ typedef enum {
   /// Error code indicating an OK result from a C API function.
   OPENASSETIO_NS(ErrorCode_kOK) = 0,
   /// Error code representing a generic C++ exception.
-  OPENASSETIO_NS(ErrorCode_kUnknown)
+  OPENASSETIO_NS(ErrorCode_kUnknown),
+  /// Error code representing a C++ std::bad_variant_access exception.
+  OPENASSETIO_NS(ErrorCode_kBadVariantAccess),
+  /// Error code representing a C++ std::out_of_range exception.
+  OPENASSETIO_NS(ErrorCode_kOutOfRange),
+  /// Error code representing a C++ std::length_error exception.
+  OPENASSETIO_NS(ErrorCode_kLengthError)
 } OPENASSETIO_NS(ErrorCode);
 /**
  * @}
--- a/src/openassetio-core-c/include/openassetio/c/errors.h
+++ b/src/openassetio-core-c/include/openassetio/c/errors.h
@@ -12,9 +12,13 @@ extern "C" {
  * @{
  */
 
-/// Error code indicating an OK result from a C API function.
-const int OPENASSETIO_NS(kOK) = 0;
-
+// NOLINTNEXTLINE(modernize-use-using)
+typedef enum {
+  /// Error code indicating an OK result from a C API function.
+  OPENASSETIO_NS(ErrorCode_kOK) = 0,
+  /// Error code representing a generic C++ exception.
+  OPENASSETIO_NS(ErrorCode_kUnknown)
+} OPENASSETIO_NS(ErrorCode);
 /**
  * @}
  */
--- a/src/openassetio-core-c/include/openassetio/c/errors.h
+++ b/src/openassetio-core-c/include/openassetio/c/errors.h
@@ -16,8 +16,10 @@ extern "C" {
 typedef enum {
   /// Error code indicating an OK result from a C API function.
   OPENASSETIO_NS(ErrorCode_kOK) = 0,
-  /// Error code representing a generic C++ exception.
+  /// Error code representing a generic non-exception type thrown.
   OPENASSETIO_NS(ErrorCode_kUnknown),
+  /// Error code representing a generic C++ exception.
+  OPENASSETIO_NS(ErrorCode_kException),
   /// Error code representing a C++ std::bad_variant_access exception.
   OPENASSETIO_NS(ErrorCode_kBadVariantAccess),
   /// Error code representing a C++ std::out_of_range exception.
--- a/src/openassetio-core-c/private/errors.hpp
+++ b/src/openassetio-core-c/private/errors.hpp
@@ -47,6 +47,30 @@ template <class Exception>
 void extractExceptionMessage(OPENASSETIO_NS(StringView) * err, const Exception &exc) {
   openassetio::assignStringView(err, exc.what());
 }
+
+/**
+ * Wrap a callable such that all exceptions are caught and converted to
+ * an error code.
+ *
+ * This is intended as a fallback for unhandled exceptions.
+ *
+ * @tparam Fn Type of callable to wrap.
+ * @param err Storage for error message, if any.
+ * @param fn Callable to wrap.
+ * @return Error code.
+ */
+template <typename Fn>
+auto catchUnknownExceptionAsCode(OPENASSETIO_NS(StringView) * err, Fn &&fn) {
+  try {
+    return fn();
+  } catch (std::exception &exc) {
+    extractExceptionMessage(err, exc);
+    return OPENASSETIO_NS(ErrorCode_kException);
+  } catch (...) {
+    assignStringView(err, "Unknown non-exception object thrown");
+    return OPENASSETIO_NS(ErrorCode_kUnknown);
+  }
+}
 }  // namespace errors
 }  // namespace OPENASSETIO_VERSION
 }  // namespace openassetio
--- a/src/openassetio-core-c/private/errors.hpp
+++ b/src/openassetio-core-c/private/errors.hpp
@@ -21,8 +21,9 @@ namespace errors {
  * @param code Error code.
  * @param msg Error message to bundle in exception.
  */
-inline void throwIfError(const int code, [[maybe_unused]] const OPENASSETIO_NS(StringView) & msg) {
-  if (code != 0) {
+inline void throwIfError(const OPENASSETIO_NS(ErrorCode) code,
+                         [[maybe_unused]] const OPENASSETIO_NS(StringView) & msg) {
+  if (code != OPENASSETIO_NS(ErrorCode_kOK)) {
     Str errorMessageWithCode = std::to_string(code);
     errorMessageWithCode += ": ";
     errorMessageWithCode += std::string_view{msg.data, msg.size};
--- a/tests/openassetio-core-c/private/errorsTest.cpp
+++ b/tests/openassetio-core-c/private/errorsTest.cpp
@@ -67,3 +67,60 @@ SCENARIO("Using extractExceptionMessage to copy a C++ exception message to a C S
     }
   }
 }
+
+SCENARIO("Decorating an exception-throwing C++ function to instead return an error code") {
+  using openassetio::errors::catchUnknownExceptionAsCode;
+
+  // Error message storage.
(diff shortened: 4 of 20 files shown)


tests/openassetio-core-c/private/StringViewTest.cpp
```diff
@@ -38,6 +41,49 @@ SCENARIO("Creating, modifying and querying a C API mutable StringView") {
...
+    AND_GIVEN("a StringView wrapping a buffer with sufficient capacity for C++ string") {
+      openassetio::Str storage;
+      storage.reserve(expectedStr.size());
+
+      OPENASSETIO_NS(StringView) actualStringView{storage.capacity(), storage.data(), 0};
+
+      WHEN("assignStringView is used to copy the C++ string to the StringView") {
+        openassetio::assignStringView(&actualStringView, expectedStr);
+
+        THEN("StringView matches source string") { CHECK(actualStringView == expectedStr); }
```
foundrytom, line 50: [fix] Should we check also in these cases that it has copied and detached from the src string?

tests/openassetio-core-c/private/errorsTest.cpp
```diff
@@ -33,9 +35,27 @@ SCENARIO("throwIfError error code/message handling") {
...
+    const openassetio::Str expectedMessage = "some error";
+    const std::runtime_error runtimeError{expectedMessage};
+
+    openassetio::Str storage(expectedMessage.size(), '\0');
+    OPENASSETIO_NS(StringView) actualMessage{storage.capacity(), storage.data(), 0};
+
+    WHEN("extractExceptionMessage copies the message from the exception to the StringView") {
+      extractExceptionMessage(&actualMessage, runtimeError);
+
+      THEN("StringView matches expected message") { CHECK(actualMessage == expectedMessage); }
```
foundrytom, line 58: [fix] Should we also check that the copy is detached, so when the exception goes out of scope...

foundrytom (commented):
Need to give the main API and exception handling commits a proper look through, but some higher-level comments in the mean time.

tests/openassetio-core-c/private/StringViewTest.cpp
```diff
@@ -0,0 +1,42 @@
...
+        STATIC_REQUIRE(std::is_const_v<decltype(actualStringView.maxSize)>);
+        actualStringView.buffer[1] = '0';
+        actualStringView.usedSize = 4;
+
+        THEN("storage has been updated") { CHECK(expectedStr == "s0me string"); }
+
+        THEN("view has been updated") {
+          CHECK(std::string_view{actualStringView.buffer, actualStringView.usedSize} == "s0me");
+        }
+      }
```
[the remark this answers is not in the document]
feltech: Updated tests for `StringView` in 37e01eb4 and docs in dd02b246, then docs and tests for `ConstStringView` in 26f408be. Split into multiple commits to try to make eventual squash easier.

src/openassetio-core-c/private/errors.hpp
```diff
@@ -25,5 +29,18 @@ inline void throwIfError(const int code, [[maybe_unused]] const OPENASSETIO_NS(S
     throw std::runtime_error(errorMessageWithCode);
   }
 }
+
+/**
+ * Extract message from an exception and copy into a C StringView out-param.
+ *
```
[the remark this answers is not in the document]
feltech: Added a note on truncation in f88b8a91

tests/openassetio-core-c/private/StringViewTest.cpp
feltech: Updated tests to ensure data was copied not just pointed to in 61fcc94f

tests/openassetio-core-c/private/errorsTest.cpp
feltech: Updated tests to ensure data was copied not just pointed to in d4609176

tests/openassetio-core-c/private/StringViewTest.cpp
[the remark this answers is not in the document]
foundrytom: > Split into multiple commits to try to make eventual squash easier, sorry

Don't apologise, its a nightmare otherwise!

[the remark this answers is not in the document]
foundrytom: [question] What are your thoughts on dealing with the length of the underlying buffer changing? Should we document that one should really generally be careful not to update the underlying buffer when it is exposed through a string view?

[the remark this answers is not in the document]
feltech: I'll expand on the notes re. being careful with the underlying buffer, though it is C so we can hope people are grown-ups. 

Re. the length of the `data` buffer, I don't think that once set it can change, at least in "normal" C/C++, since this would require a reallocation, invalidating the `data` pointer (which is equally bad, if not worse).  

We could make the `capacity` non-const, so that it can be updated in case the user suddenly wants to use some of the bytes at the end of the `data` buffer for some other purpose, but that's getting pretty advanced.

One thing that just occurred to me to be aware of is that truncation in UTF-8 can mean a decoding failure.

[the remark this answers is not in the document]
feltech: Added some more notes in daf0aaf6 and 0bd6b807

Also, somewhat tangentially but related, updated `StringView::data` to be a const pointer, so it can't be re-assigned once set, in 85f8ebf2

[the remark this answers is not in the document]
foundrytom: > Re. the length of the data buffer, I don't think that once set it can change, at least in "normal" C/C++, since this would require a reallocation, invalidating the data pointer (which is equally bad, if not worse).

Sorry, was thinking of the `size` part, where you could change `"cat"` to `"ca"` in the underlying buffer, but `size` wouldn't be updated in the `StringView`.

foundrytom:
@feltech Do you want to squash these fixes in when you have a sec to make it a bit easier to review?

feltech:
> @feltech Do you want to squash these fixes in when you have a sec to make it a bit easier to review?

Done. Haven't yet addressed https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/330#discussion_r854349076 though, so a fixup commit incoming re. that.

tests/openassetio-core-c/private/StringViewTest.cpp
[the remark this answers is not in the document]
feltech: Simplified docstrings to simply say "don't use the buffer" in 0bac7408 and 8607eaf6, and removed the test that modifies the underlying buffer in 448599c2

src/openassetio-core-c/include/openassetio/c/SimpleMap.h
```diff
@@ -0,0 +1,138 @@
...
+#include <cstdint>
+
+#include <openassetio/c/export.h>
+
+#include "./StringView.h"
+#include "./namespace.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
feltech, line 15: [question] @meshula - we're curious to get your opinion on whether we should try to support pure C linking (i.e. headers) or whether having a C-compatible interface in the binary is sufficient?

We're raising this specifically now since this file (and others) has a contradiction, in that we're apparently supporting pure C via `#ifdef __cplusplus`, but then we're also `#include`ing C++-only headers (e.g. `cstddef`). 

We need to decide whether to continue supporting pure C headers or not.  

Off the top of my head the only reasons to have C-compatible headers are
* C projects linking against our library. Is that really likely?
* FFIs/bindings generators that are able to automate some boilerplate by parsing C headers (but such systems may well be able to parse C++ headers too).

meshula: The pattern I use is

```cpp
#ifdef __cplusplus
// individually adorn functions for complete control
// rather than block externing everything
#define MYAPI_EXTERNC extern "C"
#else
#define MYAPI_EXTERNC
#endif
#ifdef _MSC_VER
... extra junk for dll export/import and gcc visibility
#define MYAPI_EXPORT
#endif
// include the c headers not the c++ headers, the fussiness around
// std:: in front of everything when it comes from cmath seems like
// unhelpful complication and compiler to compiler variation in defaults

#include <math.h>

MYAPI_EXPORT MYAPI_EXTERN
int my_func();
```

as far as binding generators go, you get C for free pretty much across the board in languages that natively parse C headers, and autobind, e.g. Rust, Zig, Swift, Objective C, all do this. C++ always implies extra work. Rust, Zig, Swift, Objective C, will all require manual C binding of the C++ interfaces, and the results tend to be highly unidiomatic to the target language.

Languages with well developed c++ generators usually also require one to go back in and create a "porcelain" layer so that the generated goobledygook is consumable by humans. This tends not to be an issue with C autogens, and it all stems from the fact that C++ has extra semantics around ctor/dtor and memory management and the resultant bridging tends to be opinionated and clumsy IMO.

src/openassetio-core-c/include/openassetio/c/StringView.h
```diff
@@ -0,0 +1,112 @@
...
+ *   myDestString->size =
+ *     min(myDestString->capacity, mySrcStringSize);
+ *
+ *   strncpy(myDestString->data, mySrcStringData, myDestString->size);
+ * }
+ * @endcode
+ */
+// NOLINTNEXTLINE(modernize-use-using)
+typedef struct {
+  /// Number of available elements in buffer.
```
meshula, line 99:
```suggestion
  /// Memory capacity of the buffer.
```
I recommend noting as byte size not element size. If elements are unicode, "available elements" become ambiguous between memory capacity and code point count.

src/openassetio-core-c/include/openassetio/c/StringView.h
```diff
@@ -0,0 +1,112 @@
...
+  const size_t capacity;
+  /// Writeable buffer storing the string data.
+  char* const data;
+  /// Number of elements in the buffer used for string storage.
```
meshula, line 103:
```suggestion
  /// Occupied memory capacity
```
same note about codepoints being a confounding thing

src/openassetio-core-c/private/SimpleMap.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    return OPENASSETIO_NS(kOK);
+  });
+}
+}  // namespace
+
+extern "C" {
+
+OPENASSETIO_NS(SimpleMap_s) OPENASSETIO_NS(SimpleMap_suite)() {
```
meshula, line 61: There must be some axiom about things declared "Simple" and computer science oxymorons :) Instead of calling it Simple, why not point directly at its use, which I think might be better served by something like "AttirbuteDictionary" or some such.

src/openassetio-core-c/include/openassetio/c/SimpleMap.h
wuggy-ianw: I like @meshula 's pattern. Macroising the symbol visibility is common and familiar and if there's going to a by a dynamic library with this stuff in, something similar will be needed! Adding the extern-"C"-ness optionally when using the header as C++ is a nice addition.

We should definitely be using the C headers for things in all the C-compatible headers:

```
#include <math.h>
```

can be compiled as C or C++ (as long as the compilers math.h works as it should do! I had to read the C++ spec to check... :smile: ).

```
#include <cmath>
```

This just automatically requires all users of the header to be C++ compilers... which kind of defeats the point of making the headers valid C. We could instead use C++ and then control the ABI (if necessary) with compiler specific attributes in the function prefix macros.

src/openassetio-core-c/include/openassetio/c/SimpleMap.h
```diff
@@ -0,0 +1,138 @@
...
+typedef struct OPENASSETIO_NS(SimpleMap_t) * OPENASSETIO_NS(SimpleMap_h);
+
+/**
+ * Function pointer suite for a @fqref{SimpleMap} "SimpleMap" C API.
+ *
+ * Instances of this suite are provided by the `SimpleMap_suite`
+ * function and operate on a @fqcref{SimpleMap_h} "SimpleMap handle".
+ */
+// NOLINTNEXTLINE(modernize-use-using)
+typedef struct {
```
wuggy-ianw, line 38: Using the 'suite' of function pointers seems a bit overkill. Is this strictly necessary, or could we use simple linkage?

The times when using a fptr-suite is most useful is for runtime loadable plugin systems.
1. The factory method can hide the discovery mechanism and loading of plugins
2. The suite object provides a convenient way for plugins to 'self describe' their entry-points after loading (e.g. through a factory function of their own)
3. A single plugin can implement several different public APIs by providing different 'levels' of suites with different (but usually overlapping) behaviours.

Do we need any of that?

My preference would be just to have plain extern-"C" functions:
- It would be up to the user at link time to decide to use static or dynamic linking, and both would work
- In the static linking case, unused code would be strippable.
- Doesn't need any boilerplate-code on the library side or the user's side for accessing or setting up suite objects.

src/openassetio-core-c/include/openassetio/c/SimpleMap.h
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Opaque handle type representing a @fqref{SimpleMap} "SimpleMap"
+ * instance.
+ *
+ * The associated @fqcref{SimpleMap_s} "SimpleMap suite" functions
+ * operate on this handle type.
+ */
+// NOLINTNEXTLINE(modernize-use-using)
+typedef struct OPENASSETIO_NS(SimpleMap_t) * OPENASSETIO_NS(SimpleMap_h);
```
wuggy-ianw, line 29: Do we want the _h from SimpleMap_h, if it's the type that should be most used?
Feels a bit odd to have an handle type that we don't much care what the type is meant to be, but then have everyone type '_h' everywhere so we're reminded it's a handle! :smile:

src/openassetio-core-c/include/openassetio/c/errors.h
```diff
@@ -14,6 +14,16 @@ extern "C" {
 
 /// Error code indicating an OK result from a C API function.
 const int OPENASSETIO_NS(kOK) = 0;
```
wuggy-ianw, line 16: Any reason this is not just a simple enum?
(I imagine automated tools for language bindings may be able to gain more information from an enum than just a list of constants.)

src/openassetio-core-c/private/SimpleMap.cpp
```diff
@@ -0,0 +1,99 @@
...
+  return {// dtor
+          [](OPENASSETIO_NS(SimpleMap_h) handle) { delete HandleConverter::toInstance(handle); },
```
wuggy-ianw, line 63: Ouch. My brain hurts.

> This [user-defined conversion function](https://en.cppreference.com/w/cpp/language/cast_operator) is only defined if the capture list of the lambda-expression is empty. It is a public, constexpr, (since C++17) non-virtual, non-explicit, const noexcept member function of the closure object.

> The value returned by this conversion function is a pointer to a function with C++ [language linkage](https://en.cppreference.com/w/cpp/language/language_linkage) that, when invoked, has the same effect as invoking the closure object's function call operator directly.

This potentially means the actual function pointer for the lambda may have a different linkage (and therefor ABI) than a plain-C function pointer of the same parameters.

I think the safe way to do this would have to be declaring the function extern "C" with a C++ definition and taking a pointer to that.

src/openassetio-core-c/private/chandle.hpp
```diff
@@ -0,0 +1,39 @@
...
+ */
+template <class Type, class Handle>
+struct Converter {
+  /**
+   * Convert a pointer to a C++ instance to a handle.
+   *
+   * @param ptr Pointer to C++ instance.
+   * @return Opaque C handle.
+   */
+  static Handle toHandle(Type* ptr) { return reinterpret_cast<Handle>(ptr); }
```
wuggy-ianw, line 27: Is this necessary? If the handle is a pointer to an opaque struct, and we have the definition of the struct in C++, then no conversion should be necessary?

> // foo.h
>   extern "C" {
>   typedef struct Foo* FooHandle;
>
>   FooHandle makeFoo();
>   void freeFoo(FooHandle h);
> }

> // foo.cpp
> #include "foo.h"
> #include <strings>
>
> // we can do all kinds of C++ things!
> class FooBase { virtual ~FooBase() {} };
> struct Foo : public FooBase {
> private:
>  std::string wibble = "Wobble";
> public:
>  Foo() {};
>  override ~Foo();
> }
>
> FooHandle makeFoo() { return new Foo; }
> void freeFoo(FooHandle h) { delete h; }

That should be fine, right?

src/openassetio-core-c/include/openassetio/c/SimpleMap.h
feltech: Really great feedback, thanks!  I've created #337 to fix up the headers (since it's not just the PR where the mistake has been made :roll_eyes: ) and #338 to ensure we have a consistent and tested strategy for our C header structure.

src/openassetio-core-c/include/openassetio/c/StringView.h
feltech: Good point. We've now split the `StringView` commits into a separate PR #336, where I've re-worded to explicitly mention "bytes" rather than "elements".

src/openassetio-core-c/private/chandle.hpp
feltech: I think this is a misunderstanding.  In our approach, `struct Foo` would only be declared and never defined, it's just a dummy struct declaration, required so we can declare an opaque pointer type.

The actual data the `FooHandle` pointer points to would (likely) be an instance of a C++ namespaced and mangled class, which we `reinterpret_cast` to the C opaque pointer type for passing around in the C API.

src/openassetio-core-c/include/openassetio/c/SimpleMap.h
feltech: One concern with removing the `_h` is that it's less differentiated from the actual C++ class that it's associated with.  Plus I wonder if people would assume at a glance that it's a struct rather than an opaque handle.

src/openassetio-core-c/private/SimpleMap.cpp
feltech: It's been on our list to rename this and your comment is the straw that broke the camel's back.  So we've decided on `InfoDictionary`.  This sounds rather generic, but the primary purpose for this type is as a return value from an `info` member function. It will also be used as a possible value type received when `resolve`ing an entity reference, but we think the name works for that too, more-or-less.

src/openassetio-core-c/include/openassetio/c/SimpleMap.h
feltech: Good points well made and we do tend to agree so will refactor this.  As you mention, function suites are a valuable pattern for plugins, but not so much for the host (Katana notwithstanding).

For reference, we noted some advantages of function pointer suites in [C_API.md](https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/feature/181-coreCppBuild/contributing/C_API.md), which I'll reproduce below:

Such suites have a few advantages
* Aesthetically, a struct of function pointers mirrors a class with
  methods, and if the suite is providing a C API to a C++ class then
  mentally mapping one to the other is straightforward.
* Only the suite type needs namespacing, the "methods" are simply
  struct members, which reduces code verbosity.
* Since a suite is an instance of a struct, it must be provided by
  a function call, allowing the provider a great deal of flexibility
  in its construction, compared to link-time defined symbols.
* The suite is an object that can be passed around, including as input
  to a wrapper C++ class (see below).
* In C++ the suite's functions can be implemented as a series of 
  (non-capturing) lambdas, keeping related code grouped together.

src/openassetio-core-c/private/SimpleMap.cpp
feltech: This is a super interesting point, which I've still not fully absorbed :brain:,  though when we deal with your other good point in  https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/330#discussion_r855645900 then the lambdas will likely disappear.

src/openassetio-core-c/include/openassetio/c/errors.h
feltech: Good point, it's also better start out restrictive.  Using an enum is a bit more verbose in the C API function signatures (i.e. return type is an enum type rather than simply `int`), but that's a minor point.

src/openassetio-core-c/private/SimpleMap.cpp
feltech: Reading a little more on this, I believe the concern would be if the lambda function pointer's C++ language linkage implied a different calling convention than would be expected by a C caller.  From what I understand, this would never be the case for us, and would be difficult for anyone to cause.

Though I admit this is right at the edge of my level of understanding.  Always more to learn with C++!

src/openassetio-core-c/private/SimpleMap.cpp
feltech: I've gone and rewrote history so it's now (and always was ;) ) called `InfoDictionary`.

src/openassetio-core-c/include/openassetio/c/errors.h
feltech: As part of addressing https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/330#discussion_r855512524 I rewrote the git history, where I also updated the error codes to use an `enum`.

src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * Instances of this suite are provided by the `InfoDictionary_suite`
+ * function and operate on a @fqcref{InfoDictionary_h} "InfoDictionary handle".
+ */
+// NOLINTNEXTLINE(modernize-use-using)
+typedef struct {
+  /**
+   * Destructor function.
+   *
+   * This should be called when the handle is no longer in use.
```
foundrytom, line 43: [question] Do we need to describe the fact that this will immediately delete the object and so understanding ownership is critical?

src/openassetio-core-c/private/chandle.hpp
```diff
@@ -0,0 +1,39 @@
...
+#pragma once
+
+#include <openassetio/export.h>  // For OPENASSETIO_VERSION
+
+namespace openassetio {
+inline namespace OPENASSETIO_VERSION {
+/**
+ * Provides utilities for converting between C++ types and C handles.
+ */
+namespace chandle {
```
foundrytom, line 12: [question] Whats the convention here for the case style for word delimination, not having kebab, snake, or any camel variant seems like it might be hard to grok?

foundrytom (approved):
Good work @feltech, LGTM aside from a few minor questions, and splitting the error code -> enum change into its own commit. Feel free to rebase/squash/merge etc... when you're done.

feltech:
> splitting the error code -> enum change into its own commit

Split into 97e447ff and ad61c8ac - much cleaner now.

src/openassetio-core-c/private/chandle.hpp
feltech: Renamed to `handles` in d53e3389

foundrytom:
> > splitting the error code -> enum change into its own commit
> 
> Split into [97e447f](https://github.com/TheFoundryVisionmongers/OpenAssetIO/commit/97e447ff88c549965061b81bc7b05ae8347d490f) and [ad61c8a](https://github.com/TheFoundryVisionmongers/OpenAssetIO/commit/ad61c8ac75a8ecf60cf0bed21c27686fd754f68d) - much cleaner now.

Thanks @feltech, much easier to grok. This PR LGTM, and I believe we have addressed/plan to address the majority of the other reviewer comments. Feel free to squash and merge so we can get the dependent PRs in too.

PR #334, opened 2022-04-20
[Core] C API InfoDictionary basic introspection

feltech:
Part of https://github.com/TheFoundryVisionmongers/OpenAssetIO/issues/290. 

Add introspection functions allowing 
* The number of entries currently in the container.
* The type of an existing entry to be queried, so that the correct accessor function can be chosen.

### DIFF
--- a/src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
+++ b/src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
@@ -89,6 +89,13 @@ typedef struct {
   void (*dtor)(OPENASSETIO_NS(InfoDictionary_h) handle);
 
   /**
+   * Retrieve the number of entries currently in the map.
+   *
+   * @param handle Opaque handle to InfoDictionary.
+   */
+  size_t (*size)(OPENASSETIO_NS(InfoDictionary_h) handle);  // noexcept
+
+  /**
    * Get the type of value stored in an entry.
    *
    * @param[out] error Storage for error message, if any.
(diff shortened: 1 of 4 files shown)


src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
```diff
@@ -33,6 +33,14 @@ extern "C" {
 // NOLINTNEXTLINE(modernize-use-using)
 typedef struct OPENASSETIO_NS(InfoDictionary_t) * OPENASSETIO_NS(InfoDictionary_h);
 
+// NOLINTNEXTLINE(modernize-use-using)
+typedef enum {
```
foundrytom, line 37: [fix] Be good to record the rationale for the choice of an `enum` here over the other options.

foundrytom (changes requested):

src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
feltech: Added a docstring for this enum in eadd4de7.  Hopefully it answers the question why we're using an enum implicitly, but I can be more explicit if you think its necessary.

foundrytom: Thats fine, it might be worth calling out the other options considered in the commit message itself? As a mini DR.

foundrytom (approved):
Thanks @feltech, feel free to squash and merge at your leisure once we have #330 merged.

src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
feltech: Added a note in the commit message in 42b3988c

Issue #342, opened 2022-04-25
Remove old `Specifications` definitions/code


foundrytom:
- Remove `specifications.py`, `SpecificationFactory.py` and `Specification.py` along with corresponding tests.
- Remove doc references to `(Entity|Locale|Relationship)Specification`.
- Migrate test harness to manually implemented Python Specification/Traits

PR #349, opened 2022-04-26
[Core] Add `Specification` equality operator

foundrytom:
Allows Specifications to be compared "by value", based on their traits and property data. This is useful to compare the result of their manipulation.

Required as part of test migration in #342.


src/openassetio-core/include/openassetio/specification/Specification.hpp
```diff
@@ -128,6 +128,13 @@ class OPENASSETIO_CORE_EXPORT Specification {
   void setTraitProperty(const trait::TraitId& traitId, const trait::property::Key& propertyKey,
                         trait::property::Value propertyValue);
 
+  /**
+   * Compares specifications based on their trait and propertly values.
```
feltech, line 132: [fix]
```suggestion
   * Compares specifications based on their trait and property values.
```

tests/python/openassetio/test_traits.py
```diff
@@ -62,6 +62,40 @@ def test_when_value_is_not_supported_then_set_raises_TypeError(self, a_specifica
...
 
+class Test_Specification_equality:
+    def test_when_comparing_with_same_data_then_are_equal(self):
+        spec_a = specification.Specification({"a_trait"})
+        spec_a.setTraitProperty("a_trait", "a_property", 1)
+        spec_b = specification.Specification({"a_trait"})
+        spec_b.setTraitProperty("a_trait", "a_property", 1)
+        assert spec_a == spec_b
+
+    def test_when_comapring_with_different_property_value_then_are_not_equal(self):
```
feltech, line 73: [consider][minor] Here and below "comapring".

src/openassetio-core/specification/Specification.cpp
```diff
@@ -76,6 +78,8 @@ void Specification::setTraitProperty(const trait::TraitId& traitId,
                                      trait::property::Value propertyValue) {
   impl_->setTraitProperty(traitId, propertyKey, std::move(propertyValue));
 }
+
+bool Specification::operator==(const Specification& other) const { return *impl_ == *other.impl_; }
```
feltech, line 82: [note] I did wonder if we should check memory address equality as a short-circuit optimisation.

It occurred to me that `unordered_map` probably does an identity check to short-circuit its equality operator. So I stepped through the code and, at least in gcc, it doesn't.

Maybe they thought it's such a rare case as to not be worth the cost of adding a pointer comparison to every equality check.  

I suspect that argument applies to us too.

feltech (approved):
Other than minor typo, looks good. Happy for you to fix + merge without me seeing it again.

src/openassetio-core/specification/Specification.cpp
foundrytom: I was wondering the very same and was going to chat to you about that! Like your reasoning. I'm sure if it was valuable, then it would be in `unordered_map` too.

PR #353, opened 2022-04-26
[Core] Migrate `ManagerInterface` `info` method to C++

### LINKED ISSUE #290: Migrate manager `info` (`SimpleMap` - also needed later for trait properties)

feltech:
Closes #290. Migrate `info` method from Python to C++ with Python binding, and add its C API.

### DIFF
--- a/doc/src/Examples.dox
+++ b/doc/src/Examples.dox
@@ -109,9 +109,8 @@
  *
  * The API middleware provides assorted short-circuit validation
  * optimisations that can reduce the number of inter-language hops
- * required. See @ref
- * openassetio.managerAPI.ManagerInterface.ManagerInterface.info
- * "ManagerInterface.info" and the `kField_EntityReferencesMatchPrefix`
+ * required. See @fqref{managerAPI::ManagerInterface::info}
+ * "ManagerInterface::info" and the `kField_EntityReferencesMatchPrefix`
  * key.
  *
  * @code{.py}
--- a/python/openassetio/managerAPI/ManagerInterface.py
+++ b/python/openassetio/managerAPI/ManagerInterface.py
@@ -148,7 +148,7 @@ class ManagerInterface(_openassetio.managerAPI.ManagerInterface):
 
        @li @needsref identifier()
        @li @needsref displayName()
-       @li @ref info()
+       @li @needsref info()
        @li @ref updateTerminology()
        @li @ref getSettings()
        @li @ref setSettings()
@@ -199,40 +199,6 @@ class ManagerInterface(_openassetio.managerAPI.ManagerInterface):
     # asset_management_system itself.
     #
     # @{
-
-    def info(self):
-        """
-        Returns other information that may be useful about this @ref
-        asset_management_system. This can contain arbitrary key/value
-        pairs. For example:
-
-            { 'version' : '1.1v3', 'server' : 'assets.openassetio.org' }
-
-        @note Keys should always be strings, and values must be
-        plain-old-data types (ie: str, int, float, bool).
-
-        There are certain optional keys that may be used by a host or
-        the API:
-
-          @li openassetio.constants.kField_SmallIcon (upto 32x32)
-          @li openassetio.constants.kField_Icon (any size)
-
-        Because it can often be expensive to bridge between languages,
-        info can also contain an additional field - a prefix that
-        identifies a string as a valid entity reference. If supplied,
-        this will be used by the API to optimize calls to
-        isEntityReference when bridging between C/Python etc.
-        If this isn't supplied, then isEntityReference will always be
-        called to determine if a string is an @ref entity_reference or
-        not. Note, not all invocations require this optimization, so
-        @ref isEntityReference should be implemented regardless.
-
-          @li openassetio.constants.kField_EntityReferencesMatchPrefix
-
-        @note Keys should always be strings, and values must be
-        plain-old-data types (ie: str, int, float, bool).
-        """
-        return {}
 
     def updateTerminology(self, stringDict, hostSession):
         """
@@ -305,7 +271,7 @@ class ManagerInterface(_openassetio.managerAPI.ManagerInterface):
 
          @li @needsref identifier()
          @li @needsref displayName()
-         @li @ref info()
+         @li @needsref info()
          @li @ref updateTerminology()
          @li @ref getSettings()
          @li @ref setSettings()
--- a/src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
+++ b/src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
@@ -28,10 +28,6 @@ extern "C" {
 /**
  * Opaque handle type representing a @fqref{InfoDictionary}
  * "InfoDictionary" instance.
- *
- * The ownership semantics of this handle are "owned by client", that
- * is, the caller of the C API is responsible for deallocating using
- * the C API `dtor` function once the InfoDictionary is no longer in use.
  */
 // NOLINTNEXTLINE(modernize-use-using)
 typedef struct OPENASSETIO_NS(InfoDictionary_t) * OPENASSETIO_NS(InfoDictionary_h);
--- a/src/openassetio-core-c/private/managerAPI/CManagerInterface.cpp
+++ b/src/openassetio-core-c/private/managerAPI/CManagerInterface.cpp
@@ -6,6 +6,7 @@
 #include "CManagerInterface.hpp"
 
 #include "../errors.hpp"
+#include "../handles.hpp"
 
 namespace openassetio {
 inline namespace OPENASSETIO_VERSION {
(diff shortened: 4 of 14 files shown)


doc/src/Examples.dox
```diff
@@ -109,9 +109,8 @@
  *
  * The API middleware provides assorted short-circuit validation
  * optimisations that can reduce the number of inter-language hops
- * required. See @ref
- * openassetio.managerAPI.ManagerInterface.ManagerInterface.info
- * "ManagerInterface.info" and the `kField_EntityReferencesMatchPrefix`
+ * required. See @fqref{managerAPI::ManagerInterface::info}
+ * "ManagerInterface::info" and the `kField_EntityReferencesMatchPrefix`
```
feltech, line 113: [question] Should this PR also migrate the relevant constants to C++?  Or can we leave that to a separate issue?

foundrytom: We can do that as a separate issue I think, but should do it now.

foundrytom: Added https://github.com/TheFoundryVisionmongers/OpenAssetIO/issues/354

python/openassetio/managerAPI/ManagerInterface.py
```diff
@@ -148,7 +148,7 @@ class ManagerInterface(_openassetio.managerAPI.ManagerInterface):
 
        @li @needsref identifier()
        @li @needsref displayName()
-       @li @ref info()
+       @li @needsref info()
```
foundrytom, line 151: [fix] Can this be `@fqref{managerAPI::ManagerInterface::info} "info`"? (Sorry, for some reason I can't add suggestions).

python/openassetio/managerAPI/ManagerInterface.py
```diff
@@ -305,7 +271,7 @@ def initialize(self, hostSession):
 
          @li @needsref identifier()
          @li @needsref displayName()
-         @li @ref info()
+         @li @needsref info()
```
foundrytom, line 274: [fix] `@fqref` also?

src/openassetio-core/include/openassetio/managerAPI/ManagerInterface.hpp
```diff
@@ -194,6 +195,41 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+   * isEntityReference when bridging between C/Python etc.
+   * If this isn't supplied, then isEntityReference will always be
+   * called to determine if a string is an @ref entity_reference or
+   * not. Note, not all invocations require this optimization, so
+   * @needsref isEntityReference should be implemented regardless.
+   *
+   *   @li openassetio.constants.kField_EntityReferencesMatchPrefix
+   *
+   * @note Keys should always be strings, and values must be
+   * plain-old-data types (ie: str, int, float, bool).
```
foundrytom, line 227: [fix] Duplicate of `L205` and rephrase "should" to "are" (now its enforced) or remove?

tests/python/openassetio/managerAPI/test_managerinterface.py
```diff
@@ -50,6 +50,20 @@ def test_when_overridden_then_returns_value(self, manager_interface):
...
+
+        assert isinstance(info, dict)
+        assert info == {}
+
+    def test_when_overridden_then_returns_expected_dict(self, manager_interface):
+        info = manager_interface.info()
+
+        assert isinstance(info, dict)
+        assert info == {"stub": "info"}
+
```
foundrytom, line 65: [fix] Are we relying on automagic binding of `InfoDictionary` to Python via pybind, where it is always converted to a native python `dict`? If so, we need checks for the expected behaviour around:
- Non string keys
- Invalid value types (including None)

src/openassetio-core-c/private/managerAPI/CManagerInterface.cpp
```diff
@@ -13,15 +13,17 @@ namespace managerAPI {
 
 constexpr size_t kStringBufferSize = 500;
 
+// Buffer for error message. Statically allocated per-thread to avoid
+// the need to reallocate on ever function call.
+static thread_local char errorMessageBuffer[kStringBufferSize];
```
foundrytom, line 19: [question] Is this really a significant cost at the moment? To make it worth the risk of accidentally coupling test data?

src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
```diff
@@ -28,10 +28,6 @@ extern "C" {
 /**
  * Opaque handle type representing a @fqref{InfoDictionary}
  * "InfoDictionary" instance.
- *
- * The ownership semantics of this handle are "owned by client", that
- * is, the caller of the C API is responsible for deallocating using
- * the suite's `dtor` function once the InfoDictionary is no longer in use.
```
foundrytom: [plan] Longer-term, we should have a page on "memory management between languages" or similar, were we outline the strategy/best practices.

tests/openassetio-core-c/private/managerAPI/CManagerInterfaceTest.cpp
```diff
@@ -207,3 +218,71 @@ SCENARIO("A host calls CManagerInterface::displayName") {
...
+    }
+
+    AND_GIVEN("the C suite's info() call fails") {
+      const std::string_view expectedErrorMsg = "some error happened";
+      const auto expectedErrorCode = OPENASSETIO_NS(ErrorCode_kUnknown);
+      const openassetio::Str expectedErrorCodeAndMsg = "1: some error happened";
+
+      using trompeloeil::_;
+
+      // Check that `displayName` is called properly and update error
```
foundrytom, line 269: [Fix] `displayName` -> `info`

foundrytom (changes requested):

tests/python/openassetio/managerAPI/test_managerinterface.py
foundrytom: Don't worry about this for now - as we realised that when we have a C++ `Manager` implementation, we'll be able to return bogus values in the `ValidatingManagerInterface` implementation in this file. To try and do it now needs a lot of scaffolding as currently, a python override totally bypasses C++.

python/openassetio/managerAPI/ManagerInterface.py
feltech: I just copied what was done for `identifier` and `displayName`, which I think were `@needsref`ed in the expectation that, once all these methods are migrated, stripping the `needs` bit should "just work".

I guess since we're now saying that there will be a relatively long-lived intermediate state, where only some methods are migrated, then it's worth making the cross-references work.

src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
feltech: How about an issue to update the `C_API.md` doc? Perhaps once we've proved out the semi-end-to-end workflow?

python/openassetio/managerAPI/ManagerInterface.py
foundrytom: You make an excellent point! lets just leave as `@needsref` then it is at least consistent...

src/openassetio-core-c/include/openassetio/c/InfoDictionary.h
foundrytom: Sounds good to me!

foundrytom: Added https://github.com/TheFoundryVisionmongers/OpenAssetIO/issues/359

src/openassetio-core-c/private/managerAPI/CManagerInterface.cpp
feltech: Reverted in ffac5bd2 ready to be squashed out of existence, and updated `info` to use a stack-allocated buffer in cf73708f

src/openassetio-core/include/openassetio/managerAPI/ManagerInterface.hpp
feltech: Just removed in 26c7f80e

tests/openassetio-core-c/private/managerAPI/CManagerInterfaceTest.cpp
feltech: Fixed in 7ba186b9

python/openassetio/managerAPI/ManagerInterface.py
feltech: OK left as `@needsref` for now (though did fix a related issue in 4875942d)

foundrytom (approved):
👍 SAMAYL

feltech:
> +1 SAMAYL

I think we need an `ACRONYMS.md` in `contributing` :grin:

PR #356, opened 2022-04-27
[Docs] Add DR for locatable content trait design

### LINKED ISSUE #332: [Investigation] `file` vs `blob` trait

foundrytom:
Closes #332.

Pretty view: [available here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/356/files?short_path=f967dcd#diff-f967dcd0760a5812bf92149546082da9b3010893cf1e97929bb3a3d03ab62714)

Covers the recent discussions about how best to represent content location in terms of URLs or paths, depending on the capabilities of the host.

Note to @mmazerolle and @feltech - sadly I was unable to recreate the table in markdown.


decisions/DR009-Locatable-content-traits.md
```diff
@@ -0,0 +1,207 @@
...
+place of the URL-holding `locatableContent` when filesystem-only access
+is required.
+
+#### Pros
+
+-   Explicitly determines when a host requires data to be available
+    through the file system.
+-   More explicit representation of the filtering required during
+    browsing/etc.
+-   No sting manipulation overhead in path-based hosts.
```
feltech, line 103: [fix] :bee: :scorpion: :spider: 
```suggestion
-   No string manipulation overhead in path-based hosts.
```

decisions/DR009-Locatable-content-traits.md
```diff
@@ -0,0 +1,207 @@
...
+    gained from traits vs hierarchical specifications.
+-   Requires a manager to potentially re-write data depending on the
+    exact traits used for registration and resolution, converting
+    between paths/URLs.
+
+### Option 3
+
+A single `locatableContent` trait that holds a `location` property,
+whose content is determined by the options set in a
+`locatableContentReader` trait in the call's Context's locale.
```
feltech, line 118: [consider] 
```suggestion
`locatableContentReader` trait in the caller Context's locale.
```

decisions/DR009-Locatable-content-traits.md
```diff
@@ -0,0 +1,207 @@
...
+
+#### Pros
+
+-   Allows the host to declare the locator strategies it supports to
+    access content (eg: `path`, `URL`, etc.), which can influence
+    `resolve` and browsing.
+-   No sting manipulation overhead in path-based hosts.
```
feltech, line 125: [fix] :scorpion: 
```suggestion
-   No string manipulation overhead in path-based hosts.
```

decisions/DR009-Locatable-content-traits.md
```diff
@@ -0,0 +1,207 @@
...
+-   No need for duplicate specifications.
+-   Uses the API mechanism for adapting responses at runtime (the
+    calling context's locale).
+
+#### Cons
```
feltech, line 130: [consider][minor] Precludes requesting both URL and file path in a single call (can't really think why you would want that though)

feltech (approved):
Content looks good, struggling to think of things to add, just a few minor typos.

decisions/DR009-Locatable-content-traits.md
```diff
@@ -0,0 +1,207 @@
...
+
+-   The meaning of the `location` property is less well defined, and
```
carmenpinto, line 132: This makes the trait modified by the context in a somewhat unexpected way, and the result of the location is actually dependent on another trait - belonging to another specification, which isn't atomic anyway. What about a location trait that has both a path and a url? The plugin writer can decide which one is reasonable to use. Otherwise, it might be worthwhile to make entities queryable by traits using `and` and `or` so you could specify image `and` path `or` url traits.

foundrytom: 👋  @carmenpinto!

The main goal was to try and make it clear to the manager what the host supports, similarly to the way you have the HTTP `accept` header.  

> The plugin writer can decide which one is reasonable to use. 

Inverting this was the main goal of this option. The plugin writer needs to choose the correct thing to pass back to the host, and without supporting information it can't make that decision as to which one is the best one to use. As it stands, or with any of the other options, there is no way to hint to the plugin writer what a host can even accept and URL, so it relies on out-of-band or prior knowledge. This didn't seem like a good thing. The API is about trying to make all this explicit. So this the idea of being able to formalise exactly what the host can support was very appealing, as it takes all the guess work away from the plugin writer as the host is the only thing that knows whether it can deal with an URL or not, and through this mechanism can describe this In as much detail as is appropriate.

> the result of the location is actually dependent on another trait - belonging to another specification, which isn't atomic anyway.

It is atomic in the sense that no trait in the result affects any other in the same _result_. It is only affected by some property of the calling context which is an _input_ to the method.

> This makes the trait modified by the context in a somewhat unexpected way,

I'm not sure it is particularly unexpected. The context and it's locale is defined as the mechanism by which a host imparts information to the manager about the calling environment so it can best formulate its response (similar to the request headers in HTTP). The manager is already obliged to consider this when deriving the data to return in `resolve`. So `resolve` is, by design, rarely static and is defined as "the most appropriate answer for the host". As such, this is just following the established patterns for resolve.

> queryable by traits using `and` and `or`

We did discuss this but it felt like a bit of a can of worms, as It significantly complicates trait sets, and made the logic much harder to implement for a manager.

Option 2 felt like it might be easier than introducing the extra logic, but the downsides of having to duplicate specifications etc.. again (undoing a lot of the niceties of traits in the first place), and  the more extensive subsequent data-rewriting that a manager would be expected to do wasn't worth it - given we already had the context as a mechanism for adapting response.

foundrytom: PS. If we can convince you on this, then will def update the DR with extended Rationale!

carmenpinto: Hello! I think you're right about the host needing to tell the plugin writer which format it would like, but maybe it's worth having that as either documentation or as a queryable function on the host's API - because otherwise it's difficult to tell what the host will request at all. The issue I have with the context is that the option isn't connected to the trait, so it's something that would have to be written as a special case for just this location trait, and something that be necessary to remember while writing the plugin. What would prevent every call to resolve from having to check if it was the location trait being resolved? You've convinced me on the traits using `and` and `or`, I think it would add too much complexity at this point.

foundrytom: re:

> but maybe it's worth having that as either documentation

One of the tenets of the API is that either side should hopefully not have to know who is on the other side, so there is definitely a preference for an unambiguous solution for this (ie, the `HostInterface` or `Context`)

carmenpinto: I think it makes sense to use the locale, but is there a way to avoid the manager's resolve implementation looking something like:
`value resolve(ref, traitid, context) {
   if(traitid == LocatableContentTrait.kID) {
      if(context.useUrl == true) {
         // return url path
     }
    else {
        // return file path
    }
   }
  // any other context specific traits
  
  // return generic trait data fetching instead
}`

foundrytom: I think you'd end up with some similar looking code which ever the solution is?  Just some slightly different conditionals? Maybe even worse if it was two traits:
```
if traitId in {FileTrait.kId, UrlTrait.kId}:
   location = my_asset_db.get_asset_location(ref)
   if traitId == UrlTrait.kId:
      location = f"file:///{location}"
```
At least this way you only have the one trait to worry about?

carmenpinto: I think if there's two traits, you could use the generic getter for both of them, and store either/both a url and a path. Is it worth adding the ability to convert between traits, like url to path, or timecode to frames, and the specification could match any one that has convertible traits - or is that way too complicated / not going to have good performance? That would help in the case where the host needs a file, but isn't sure if there's anything else that can supply it, like the url, and wouldn't require handling the url/path switch on both the manager side and the host's side.

For example, if I was writing a host plugin for Nuke, I sometimes need the file path, without file://. The manager plugin will have to implement the url to path conversion. If the manager plugin doesn't, then the host can either ignore potentially having a url, or handle the url to path conversion on the host's side - because the host doesn't know what the manager can supply either.

foundrytom: > I think if there's two traits, you could use the generic getter for both of them, and store either/both a url and a path.

I think you can use the generic getter either way for any implementaion, ie, to update your sketch

```
value resolve(ref, traitSet, context) {
   result = Specification()
   if(LocatableContentTrait.kID in traitSet) {
      location = my_asset_db.get_asset_path(ref)
      if(context.useUrl == true) {
         location f"file:///{location}"
     }
     LocatableContentTrait(result).setLocation(location)
   }
  // any other supported/understood/generic traits
}
```

> then the host can either ignore potentially having a url, or handle the url to path conversion on the host's side - because the host doesn't know what the manager can supply either.

I think it might be simpler than that. In the proposed design, if the host doesn't set the `usesUrls` trait in the context's locale, then the location must _always_ be a file path. If the host can support URLs, then it should set that trait and it will then it will always be a URL.

So, in the Nuke host, you just make sure the locale you use for any API calls that need a file path doesn't set `usesUrls`, and in cases where you can support an URL for a broader range of access schemes, then you set that trait. So there I no need for any switches on the host side, and it is explicit to the manager what is required.

> If the manager plugin doesn't...

If the manager doesn't then it is out of spec, and sending invalid data... the definition of `locatableContent.location` is a file path unless the `usesUrls` trait is set in the calling context, then it is an URL.  Not doing so is like a web server sending back XML to an `accept: application/json` request.

> Is it worth adding the ability to convert between traits ... or is that way too complicated / not going to have good performance?

I think this is all an implementation detail for the manager? Depending on their data model, there is probably always a conversion in one case, and not the other?

If a manager tracks locations internally via paths, then we get best performance for the file-only case (which is probably 90%+ of all use cases in DCCs today - aside from Houdini), and it is simply a case of prefixing when a host wants URLs. This was part of the motivation for this design. There was a performance concern around forcing URLs everywhere, as in a lot of cases, you'll be prefixing and then immediately stripping it on the other side, which is kind of pointless.

If a manager tracks locations internally via URLs, then in most current implementations, it's probably already making sure they're file URLs, and stripping the scheme off for the majority of tools that only work with file paths.

Either way, there is one conversion or another that has to happen, regardless of which solution we pick. I think we wanted to avoid setting both, as its redundant, forces the conversion in every case, and means that the API has failed to communicate the requirements of the host properly. Different traits for resolve are perhaps even more explicit, but have significant downsides during registration, browsing, etc. that tipped the balance in favour of the proposed solution.

We could certainly add conveniences on a specific trait, or provide utilities in the OpenAssetIO-MediaCreation repo for common conversions if they become onerous or error prone?

carmenpinto: I was thinking that having the conversion be something formalised is in OpenAssetIO it would help with the specification library, as it could be formalised which traits can be equivalent and distributed that way (like the utilities in the OpenAssetIO-MediaCreation). This would mean that you could store the file trait, and add a conversion for file -> url if the majority of uses was file, set both if you don't want to do the conversion, or set url -> file for if you mostly have urls, and sometimes want the file path. I agree with the performance concern - a lot of the time it's just going to be either urls or paths. The other potential issue with this is if you need two contexts because the host sometimes needs urls and sometimes needs paths. The registration and complexity of formalising the trait conversion could be not worth adding it to OpenAssetIO, though - in that case, using the locale seems like a good solution to me.

foundrytom: >  I was thinking that having the conversion be something formalised is in OpenAssetIO

Yeah, we can totally add those utils to the repo with the traits so they're all together in one place.

> The other potential issue with this is if you need two contexts 

That's OK, you can use the same context, just change the locale. If you need concurrent access, its easy to [make a child context](https://thefoundryvisionmongers.github.io/OpenAssetIO/classopenassetio_1_1host_a_p_i_1_1_session_1_1_session.html#a0cf0c66325b67abe2563c310c46e1565), so it's all still tied to the same state token for stability.

> The registration and complexity of formalising the trait conversion could be not worth adding it to OpenAssetIO, though - in that case, using the locale seems like a good solution to me.

I think we can have our cake and eat it - by having all the conversion stuff in the `-MediaCreation` repo, then its all in once place, centralised and stable, without any danger of contaminating the core API with something that is specific to our niche use case in M&E.

carmenpinto: That sounds good. That resolves the last of the questions I had. Thanks for your time @foundrytom !

foundrytom: Awesome, any time @carmenpinto - thanks for the massively valuable input. I will attempt to improve the rationale to capture more of this discussion.

foundrytom:
[fix] Add option 4 - URL only, and make undecided.

decisions/DR009-Locatable-content-traits.md
foundrytom: @carmenpinto ruminating upon this some more, it became clear that there was a significant problem with the proposed approach. Namely, that though it is fine for a manager to rewrite the _value_ of a property, based on the context/locale/etc. it should never change the format/interpretation of that value - as it breaks the "generic store and recall" tenet.

We pondered this some more, and came full circle, lol. See the revised outcome and additional options added for ways this could be tackled without violating this principal. What do you think?

foundrytom:
This needs moving to the `OpenAssetIO-MediaCreation` repository, so do not merge once approved.

decisions/DR009-Locatable-content-traits.md
carmenpinto: I like this solution the most. It doesn't overcomplicate and can be extended later if necessary, whereas changing from using the context to something else would make legacy code harder to support.

decisions/DR009-Locatable-content-traits.md
```diff
@@ -0,0 +1,207 @@
...
+reader_trait.setSupportedURLSchemes({"file": True, "https": True})
+
+[entity_data] = manager.resolve([entity_ref], {LocatableContentTrait.kID}, context)
+
+# The manager now returns a URL instead
+url = LocatableContentTrait(entity_data).getLocation()  # https://secure/access/to/data
+```
+
+
+[^1]: Forging UNC paths, for now - see [this saga](https://bugzilla.mozilla.org/show_bug.cgi?id=66194).
```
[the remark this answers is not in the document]
foundrytom: Mystery [option C](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/356/commits/6ee8d07a748fda8d4470fe162946b10a56116fc4)!

decisions/DR009-Locatable-content-traits.md
foundrytom: Added [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/356/commits/e4bac68cafdfa0114e57aa1966972b132d1851f6#diff-f967dcd0760a5812bf92149546082da9b3010893cf1e97929bb3a3d03ab62714R141).

feltech (approved):

foundrytom:
Merged under https://github.com/TheFoundryVisionmongers/OpenAssetIO-MediaCreation/pull/4

PR #358, opened 2022-04-28
[Core] Remove `BlobTrait`

foundrytom:
Closes #358. We can grab the removed from history tests in the future when we add traits to `OpenAssetIO-MediaCreation`.


tests/python/openassetio/test_traits.py
```diff
@@ -1,231 +0,0 @@
...
-        spec_b = Specification({"another_trait"})
-        assert spec_a != spec_b
-        assert spec_a != spec_b
-
-
-def test_BlobTrait_traitId():
-    assert trait.BlobTrait.kId == "blob"
-
-
-class Test_BlobTrait_isValid:
```
feltech, line 99: [plan] Removing `BlobTrait` tests means that `TraitBase` is now untested. 

We had avoided testing `TraitBase` in isolation because that would mean C++-specific tests, rather than testing through the Python interface.  

We now have most of the boilerplate required for testing C++ specifics (i.e. copy-paste the C API testing boilerplate), so we could add `TraitBase`-specific tests without too much fuss.

Alternatively, we can leave this until we start developing traits in C++... its conceivable `TraitBase` would be redesigned or disappear, and if it's not being used maybe there's no point investing the time to test it.

feltech (approved):
LGTM, just need to decide what to do about `TraitBase`, then ready to merge.

tests/python/openassetio/test_traits.py
foundrytom: Are you happy to push this back to when we work on our first `c++` trait, incase, as you say, `TraitBase` changes in any way?

foundrytom: Added [this note](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/358/commits/ff24bb7adf96b6a92ecb3242f50f2623ce4dc716).

feltech: That's fine, thanks.

feltech (approved):

PR #360, opened 2022-05-03
[Core] Minimal Manager C++ implementation

### LINKED ISSUE #345: Minimal `Manager` C++ implementation
## What

- A C++ implementation of the `Manager.py` stateful wrapper around a `ManagerInterface`, limited to:
  - `identifier`, `displayName`, `info`
  - No `HostSession`, auditing/etc...
- Bound to python as the base class for `Manager.py`
- The existing Python `test_manager.py` is used to cover the C++ API

## Notes

- Needs to use a shared pointer to hold the `ManagerInterface` instance as the `Manager` does not own the instance of the manager.

feltech:
Closes #345. Migrate `Manager`'s `identifier`, `displayName` and `info` methods from Python to C++ and add Python bindings. Wrangle tests to work with the enhanced type checking caused by this.

### DIFF
--- a/src/openassetio-core/include/openassetio/hostAPI/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostAPI/Manager.hpp
@@ -1,0 +1,106 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2022 The Foundry Visionmongers Ltd
+#pragma once
+
+#include <openassetio/export.h>
+#include <openassetio/managerAPI/ManagerInterface.hpp>
+#include <openassetio/typedefs.hpp>
+
+namespace openassetio {
+inline namespace OPENASSETIO_VERSION {
+/**
+ This namespace contains code relevant to anyone wanting to add support
+ for a host application.
+
+ If you are a asset management system developer, see @ref managerAPI.
+*/
+namespace hostAPI {
+/**
+ * The Manager is the Host facing representation of an @ref
+ * asset_management_system. The Manager class shouldn't be directly
+ * constructed by the host.  An instance of the class for any given
+ * asset management system can be retrieved from an API @needsref
+ * Session, using the @ref
+ * openassetio.hostAPI.Session.Session.currentManager
+ * "Session.currentManager" method, after configuring the session with
+ * the appropriate manager @ref identifier.
+ *
+ * @code
+ * session = openassetio.hostAPI.Session(
+ *     hostImpl, consoleLogger, pluginFactory)
+ * session.useManager("org.openassetio.test")
+ * manager = session.currentManager()
+ * @endcode
+ *
+ * A Manager instance is the single point of interaction with an asset
+ * management system. It provides methods to uniquely identify the
+ * underlying implementation, querying and resolving @ref
+ * entity_reference "entity references" and publishing new data.
+ *
+ * The Manager API is threadsafe and can be called from multiple
+ * threads concurrently.
+ */
+class OPENASSETIO_CORE_EXPORT Manager {
+ public:
+  explicit Manager(managerAPI::ManagerInterfacePtr managerInterface);
+
+  /**
+   * @name Asset Management System Information
+   *
+   * These functions provide general information about the @ref
+   * asset_management_system itself. These can all be called before
+   * @needsref initialize has been called.
+   *
+   * @{
+   */
+
+  /**
+   * Returns an identifier to uniquely identify the Manager.
+   *
+   * This identifier is used with the Session class to select which
+   * Manager to initialize, and so can be used as in preferences etc...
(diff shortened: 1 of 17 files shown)


src/openassetio-core/include/openassetio/hostAPI/Manager.hpp
```diff
@@ -0,0 +1,112 @@
...
+   * @{
+   */
+
+  /**
+   * Returns an identifier to uniquely identify the Manager.
+   *
+   * This identifier is used with the Session class to select which
+   * Manager to initialize, and so can be used as in preferences etc...
+   * to persist the chosen Manager. The identifier will use only
+   * alpha-numeric characters and '.', '_' or '-'. They generally follow
```
foundrytom, line 63: I've made https://github.com/TheFoundryVisionmongers/OpenAssetIO/issues/361 to track validating comments such as this.

src/openassetio-core/include/openassetio/hostAPI/Manager.hpp
```diff
@@ -0,0 +1,112 @@
...
+   *
+   * There is no requirement to use any of the information in the
+   * info dict, but it may be useful for optimisations or display
+   * customisation.
+   *
+   * There are certain well-known keys that may be set by the
+   * Manager. They include things such as:
+   *
+   *   @li openassetio.constants.kField_SmallIcon (upto 32x32)
+   *   @li openassetio.constants.kField_Icon (any size)
```
foundrytom, line 94: [Fix] Can we remove these two for now, as there not well defined (yet), and we can add them back in properly when we need them. We need MatchPrefix though.

src/openassetio-core/include/openassetio/hostAPI/Manager.hpp
```diff
@@ -0,0 +1,112 @@
...
+   *   @li openassetio.constants.kField_EntityReferencesMatchPrefix
+   *
+   * Keys will always be str, and Values will be int, bool, float or
```
foundrytom, line 97: [fix] I think we removed this elsewhere as it is now implicit in the `InfoDictionary` type?

tests/python/conftest.py
```diff
@@ -47,3 +50,186 @@ def unload_openassetio_modules():
...
+    `autospec`ed to `ManagerInterface` and accessible as a public `mock`
+    member variable, so that expectations can be configured.
+
+    @see mock_manager_interface
+    """
+
+    # pylint: disable=too-many-public-methods,too-many-arguments
+
+    def __init__(self):
+        super(ManagerInterface, self).__init__()
```
foundrytom, line 80: [fix] I _think_ this is now just `super().__init__()`

https://peps.python.org/pep-3135/

foundrytom (changes requested):
Great stuff @feltech - perilously close to and end-to-end test soon! Just a few minor things, and `pylint` is grumbling at you 🤣

src/openassetio-core/include/openassetio/hostAPI/Manager.hpp
feltech: What is the problem with this comment, is it that we should allow any UTF-8?

tests/python/conftest.py
feltech: Perhaps it's time to remove pylint disable config from `pyproject.toml` where we had
```
    # Disable Python 3 improvements:
    # - We've not definitely ruled 2 out
    # - Much of this code will move to C++
```

src/openassetio-core/include/openassetio/hostAPI/Manager.hpp
feltech: Removed in 4d324160

src/openassetio-core/include/openassetio/hostAPI/Manager.hpp
feltech: Good point, remove in 89049d80

tests/python/conftest.py
feltech: Fixed for this particular case in 1317b51b

src/openassetio-core/include/openassetio/hostAPI/Manager.hpp
foundrytom: Yeah, I think, back in the day we said that as it was a "simpler string to persist", but given we are utf-8 everywhere, do we need to be so ASCII/English centric?

tests/python/conftest.py
foundrytom: Ahh yes, sorry I forgot about those. I'd taken them out in the new `MediaCreation` repo.  Added https://github.com/TheFoundryVisionmongers/OpenAssetIO/issues/364 to track this.

foundrytom (approved):
🎉  GJD, LGTMSAMAYL 🙃

PR #366, opened 2022-05-09
[Build] Reinstate basic macOS build support

foundrytom:
This is incomplete, but gets basic tests running on macOS. The asan section still needs some work along with figuring out our deployment target.

Set up so RPATH uses `@loader_path` for libs, and `@executable_path` for the test binaries.

### DIFF
--- a/cmake/DefaultTargetProperties.cmake
+++ b/cmake/DefaultTargetProperties.cmake
@@ -57,7 +57,8 @@ function(set_default_target_properties target_name)
     )
 
     # Hide all symbols from external statically linked libraries.
-    if (IS_GCC_OR_CLANG)
+    if (IS_GCC_OR_CLANG AND NOT APPLE)
+        # TODO(TC): Find a way to hide symbols on macOS
         target_link_options(${target_name} PRIVATE -Wl,--exclude-libs,ALL)
     endif ()
 
@@ -84,8 +85,7 @@ function(set_default_target_properties target_name)
 
     if (UNIX)
         if (APPLE)
-            # TODO(DF): handle RPATH on OSX - e.g. @rpath vs. @loader_path
-            message(AUTHOR_WARNING "OSX RPATH not configured")
+            set(rpath "@loader_path")
         else ()
             set(rpath "$ORIGIN")
         endif ()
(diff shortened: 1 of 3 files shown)


cmake/DefaultTargetProperties.cmake
```diff
@@ -58,7 +58,10 @@ function(set_default_target_properties target_name)
 
     # Hide all symbols from external statically linked libraries.
     if (IS_GCC_OR_CLANG)
-        target_link_options(${target_name} PRIVATE -Wl,--exclude-libs,ALL)
+        if (NOT CMAKE_SYSTEM_NAME STREQUAL "Darwin")
```
feltech, line 61: [consider][minor] We discussed being consistent re `CMAKE_SYSTEM_NAME` vs. `APPLE`

feltech:
> thinking
> 
> ```
> 2: -- Installing: /Users/tomcowland/dev/openassetio.darwin/dist/lib/python3.9/site-packages/openassetio/_openassetio.so
> 2: error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: for: /Users/tomcowland/dev/openassetio.darwin/dist/bin/openassetio-core-c-test-exe (for architecture x86_64) option "-add_rpath @executable_path/../lib" would duplicate path, file already has LC_RPATH for: @executable_path/../lib
> ```
> 
> Seem to get this on rebuilds, so I'm guessing it's trying to update the same lib it installed before? It all works, but the error aspect (and the fact that then doesn't fail?) is more concerning?

Yeah that's annoying. Tried a quick google, not much useful info.

I think its innocuous enough.  It looks like CMake is clever enough to not re-install an unmodified binary, but not then clever enough to avoid attempting to re-set the rpath on that unmodified binary?

If that's the case, one solution I can think of is to set `BUILD_WITH_INSTALL_RPATH` to `TRUE`, which should set the rpath at build time rather than install time, avoiding the attempt to re-set the rpath when re-installing an unmodified binary.

Happy to leave fixing this for another time, if you prefer.

feltech (approved):

cmake/DefaultTargetProperties.cmake
foundrytom: Sorry missed that one, doh. Fixed [here](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/366/commits/b4afe8f0507fbf62299921ad18518c9aa87d69d7).

foundrytom:
> Happy to leave fixing this for another time, if you prefer.

Cool, thanks. Happy [middle-ground](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/366/commits/50ea32206caadd6d37868fb458b62813de994d97)?

feltech (approved):

Issue #383, opened 2022-05-18
Initial alpha release


foundrytom:
## What

Before #373 is merged, we should make an initial pre-alpha release to pin the MediaCreation and OTIO PoC repositories to so they don't break. At this point, it's just a tag, but it will mark the start of using the changlog.

## ACs

 - Tag with the format `v0.0.0-alpha.202205<date>` on the `feature/181-coreCppBuild` branch.

PR #386, opened 2022-05-19
Add `RELEASE_NOTES.md`

foundrytom:
Required for #383.


contributing/CHANGES.md
```diff
@@ -1,8 +1,8 @@
-# Recording user facing changes in CHANGES.md
```
feltech, line 1: [question] Is this file's name remaining the same?

feltech, line 1: [fix] The example under "Formatting" below still has "Changes" in the title (maybe the title is irrelevant to that section and can be removed)

feltech (changes requested):

contributing/CHANGES.md
foundrytom: If that seems ok? Was thinking it referred to the 'first' use of changes, rather than the second.

foundrytom: Good spot, thanks! Fixed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/386/commits/0a9f7bf51e7070e80142554ac20bd7f81d9c4515).

feltech (approved):

PR #393, opened 2022-05-20
Support un-versioned `clang-tidy`

### LINKED ISSUE #392: `clang-tidy`/`clang-format` fallback
## What

Support the use of non-version-suffixed `clang-tidy` and `clang-format` as part of cmake `OPENASSETIO_ENABLE_CLANG_*` options. Document our minimum target clang version for code compliance.

## Why

The sufficed executable (symlink) seems to be a quirk of the `apt` installed LLVM on Ubuntu. Release distributions do not suffix these tools.

A side effect is that the executed version of the tool is uncontrolled and may result in variation between validations.

## ACs

- Update cmake to allow fallback to non-versioned tools
- Document policy of when using a newer version, satisfy its requirements and hope for backwards compatibility.

foundrytom:
Closes #392

### DIFF
--- a/cmake/StaticAnalyzers.cmake
+++ b/cmake/StaticAnalyzers.cmake
@@ -2,10 +2,7 @@
 # Copyright 2013-2022 The Foundry Visionmongers Ltd
 
 macro(enable_clang_tidy)
-    # Clang-Tidy warnings vary quite a bit between different versions.
-    # So pin to clang-tidy-12 for now, which happens to be the latest
-    # available in Ubuntu 20.04's (i.e. CI) apt repos.
-    find_program(OPENASSETIO_CLANGTIDY_EXE clang-tidy-12)
+    find_program(OPENASSETIO_CLANGTIDY_EXE NAMES clang-tidy clang-tidy-12)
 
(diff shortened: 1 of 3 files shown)


cmake/StaticAnalyzers.cmake
```diff
@@ -57,7 +58,8 @@ macro(enable_cpplint)
 endmacro()
 
 macro(enable_clang_format)
-    find_program(OPENASSETIO_CLANGFORMAT_EXE clang-format clang-format-10)
+    find_program(OPENASSETIO_CLANGFORMAT_EXE
+        NAMES clang-format clang-format-10)
```
themissingcow, line 62: @feltech Did we mean to have different versions for format vs tidy?

foundrytom: Capturing the off-line discussions, we decided to simply use un-versioned bin names for both `clang-tidy` and `clang-format` as described in the commit comment.

feltech (approved):

cmake/StaticAnalyzers.cmake
foundrytom: We re-added v12 support as some distros seem to only install with a suffix.............. 🙃

PR #394, opened 2022-05-20
[Build] Better `clean`ing

### LINKED ISSUE #311: [Build] Better handling of `clean` CMake target
# What

Redesign CMake code, especially where `add_custom_command` is used, so that it behaves as expected when the `clean` target is executed. If feasible, ensure that `clean` followed by a build + test succeeds when offline.

# Why

In general `clean` has not been well considered in the CMake code up until this point, so there may be several issues.

In particular, the CMake code makes use of `add_custom_command` in several places, especially in the targets for running Python tests.

The `OUTPUT` of an `add_custom_command` is used to allow targets to depend on the presence of a command output (i.e. execute the command iff the `OUTPUT` file(s) is not found).

However, the `OUTPUT` file(s) are deleted when running the `clean` target. Our usage didn't consider that, so we can end up with e.g. a broken Python environment left around with no `pip` executable, after `clean`ing.

See #309 and https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/309#discussion_r842652391

# AC

- The `clean` target should revert the build tree only, leaving the installation untouched.
- ~Don't clean per-downloaded python _build_ deps in clean (eg: the `pip`, `setuptools`), do those a separate `clean-offline-cache` target.~ (Easy enough to do just removing the `build` dir)
- Move default `dist` to `build/dist`

feltech:
Closes #311

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,12 +76,11 @@ project(
 #-----------------------------------------------------------------------
 # Default install directory
 
-# If the user hasn't configured cmake with an explicit
-# CMAKE_INSTALL_PREFIX, then set it to safely install into ./dist, to
-# help prevent the user from accidentally writing over /usr/local or
-# whatever.
+# Default install to a `dist` directory under the build directory, ready
+# for use in tests and for packaging. But don't override if user has
+# explicitly set CMAKE_INSTALL_PREFIX.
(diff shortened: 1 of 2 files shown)


foundrytom (approved):

foundrytom:
@feltech Feel free to rebase and merge at your leisure.

PR #400, opened 2022-05-23
Add CMake presets

### LINKED ISSUE #315: [Build] Add CMake presets
## What

Add a [CMake presets](https://cmake.org/cmake/help/v3.22/manual/cmake-presets.7.html) `CMakePresets.json` file to the project, with presets that we expect developers to use and presets that we expect CI to use (if different), with platform-specific variants as required. 

## Why

We now have a slew of optional CMake options, many of which should be enabled during development, though some only work on specific platforms. For example, the current Ubuntu CI C++ linter build uses 
```
          cmake -S .
          --install-prefix $GITHUB_WORKSPACE/dist
          --toolchain $HOME/conan/conan_paths.cmake
          -DCMAKE_CXX_STANDARD=17
          -DOPENASSETIO_WARNINGS_AS_ERRORS=ON
          -DOPENASSETIO_ENABLE_CLANG_FORMAT=ON 
          -DOPENASSETIO_ENABLE_CLANG_TIDY=ON
          -DOPENASSETIO_ENABLE_CPPLINT=ON 
```
All of these options (including non-variable options such as `--toolchain`) can be bundled into a single CMake preset.  

Presets can be inherited, allowing us to share common base configurations.

Ultimately, having  "release" presets will also be useful for reproducible builds of our redistributable artefacts.

The straw that broke the camel's back: https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/313#discussion_r845137140 and https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/313#discussion_r845139666

## ACs

- Well-named, environment agnostic option centric high level presets, composed from functional sub-presets for:
   - Test modes (eg): lint (warn as error + clang-* + py*) test (ctest) sanitizers (asan, undef)
   - Standard dev profile 
   - No build/install/toolchain values specific
- CI updated to use a high-level preset, only supplying CI env specific options itself
- Generate issues that stop specific environments from using presets

## Out of scope

- Updating CMakeLists etc.. to make specific checks work on currently unsupported platforms
- Which targets build

foundrytom:
Closes #315.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -65,3 +65,34 @@ jobs:
 
       - name: Build and lint
         run: cmake --build build
+
+  sanitizers:
+    name: Sanitizers
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Cache conan packages
+        id: cache-conan
+        uses: actions/cache@v2
+        with:
+          path: ~/conan
+          key: ubuntu-20.04-conan
+
+      - name: Install dependencies
+        # Configure the system and install library dependencies via
+        # conan packages.
+        run: |
+          source resources/build/bootstrap-ubuntu-20.04.sh
+        env:
+          WORKSPACE: .
+
+      - name: Configure CMake build
+        run: >
+          cmake -S . -B build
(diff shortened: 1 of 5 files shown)


feltech (approved):

PR #408, opened 2022-05-25
[Lint] Default to trim whitespace in .editorconfig

feltech:
The reason for not doing this before is that some languages may have significant trailing whitespace. In particular, a double trailing whitespace in markdown forces a line break. However, there are better alternatives to achieving the same effect. The amount of trailing whitespace that slips through is far more of a problem.


foundrytom (approved):
LGTM FFT~SQMAYL~RBAMAYL

PR #425, opened 2022-05-31
Add HostSession to C++

### LINKED ISSUE #417: Add `HostSession` to C++
## What

A C++ `HostSession` class, limited to exposing the `Host` wrapper, no logging mechanism yet.

## Why

 This is required by the more interesting manager API methods. We omit logging at this point as it can easily be added later and it keeps the slice thinner - so we can get to the harder problems such as the vector + variant types for `resolve/register` et. al.

## ACs

- A `HostSession` class in the `managerAPI` namespace.
- Support for:
  -  `host`
- Issue for C implementation with any design notes.
- Python bindings as a base for the existing python `HostSession` class
- Test coverage via Python

## Out of scope

- logging
- C implementation

## Notes

The supplied host should be held by value vs pointer, as it is a simple stateless container

feltech:
Closes #417.

### DIFF
--- a/src/openassetio-core/include/openassetio/managerAPI/HostSession.hpp
+++ b/src/openassetio-core/include/openassetio/managerAPI/HostSession.hpp
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2022 The Foundry Visionmongers Ltd
+#pragma once
+
+#include <openassetio/export.h>
+#include <openassetio/managerAPI/Host.hpp>
+#include <openassetio/typedefs.hpp>
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace managerAPI {
+/**
+ * The HostSession is a manager-facing class that represents a discrete
+ * API session started by a @ref host in order to communicate with a
+ * manager.
+ *
+ * Any generalised API interactions a Manager may wish to make with a
+ * Host should be performed through the HostSession instance supplied
+ * with any ManagerInterface entrypoint. These objects should not be
+ * directly constructed, cached or otherwise persisted by a Manager.
+ *
+ * The HostSession provides access to:
+ *   - A concrete instance of the @fqref{managerAPI::Host} "Host",
+ *     implemented by the tool or application that initiated the API
+ *     session.
+ *
+ * @see @fqref{managerAPI::Host} "Host"
+ *
+ * @todo Expose logging mechanism through HostSession
+ */
+class OPENASSETIO_CORE_EXPORT HostSession {
+ public:
+  explicit HostSession(HostPtr host);
+
+  /**
+   * @return The host that initiated the API session.
+   */
+  [[nodiscard]] HostPtr host() const;
+
+ private:
+  HostPtr host_;
+};
+
+}  // namespace managerAPI
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
--- a/src/openassetio-core/managerAPI/HostSession.cpp
+++ b/src/openassetio-core/managerAPI/HostSession.cpp
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2022 The Foundry Visionmongers Ltd
+#include <openassetio/managerAPI/HostSession.hpp>
+
+#include <openassetio/managerAPI/Host.hpp>
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace managerAPI {
+
+HostSession::HostSession(HostPtr host) : host_{std::move(host)} {}
+
+HostPtr HostSession::host() const { return host_; }
+
+}  // namespace managerAPI
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
--- a/tests/python/openassetio/managerAPI/test_hostsession.py
+++ b/tests/python/openassetio/managerAPI/test_hostsession.py
@@ -30,8 +30,8 @@ from openassetio.managerAPI import Host, HostSession
 
 
 @pytest.fixture
-def mock_host():
-    return mock.create_autospec(spec=Host)
+def a_host(mock_host_interface):
+    return Host(mock_host_interface)
 
 
 @pytest.fixture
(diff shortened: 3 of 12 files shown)


src/openassetio-core/include/openassetio/managerAPI/HostSession.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * Any generalised API interactions a Manager may wish to make with a
+ * Host should be performed through the HostSession instance supplied
+ * with any ManagerInterface entrypoint. These objects should not be
+ * directly constructed, cached or otherwise persisted by a Manager.
+ *
+ * The HostSession provides access to:
+ *   - A concrete instance of the @fqref{managerAPI::Host} "Host",
+ *     implemented by the tool or application that initiated the API
+ *     session.
+ *
```
foundrytom, line 25: [fix]
```suggestion
 * @todo Expose logging mechanism through HostSession
```

src/openassetio-core/include/openassetio/managerAPI/HostSession.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * @see @fqref{managerAPI::Host} "Host"
+ */
+class OPENASSETIO_CORE_EXPORT HostSession {
+ public:
+  explicit HostSession(Host host);
+
+  /**
+   * @return A wrapped @fqref{hostAPI::HostInterface} "HostInterface".
```
foundrytom, line 33: [fix]
```suggestion
   * @return The host that initiated the API session".
```

src/openassetio-core/managerAPI/HostSession.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+#include <openassetio/managerAPI/Host.hpp>
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace managerAPI {
+
+HostSession::HostSession(Host host) : host_{std::move(host)} {}
+
+Host HostSession::host() const { return host_; }
```
foundrytom, line 13: [fix] Can we document why this is by-value

tests/python/openassetio/managerAPI/test_hostsession.py
```diff
@@ -40,14 +41,39 @@ def mock_logger():
...
+class Test_HostSession_init:
+    def test_when_invalid_host_then_raises_TypeError(self, mock_logger):
+        with pytest.raises(TypeError) as err:
+            HostSession(object(), mock_logger)
+
+        assert str(err.value).startswith("__init__(): incompatible constructor arguments")
+
+
+class Test_HostSession_host:
+    def test_returns_copy_of_host(self, host_session, a_host, mock_host_interface):
```
foundrytom, line 57: [fix] This is testing that we return by value in cpp. Is the title of this test the most descriptive? There isn't any particularly reason that it _should_ be a copy, we just choose to do so as it's a simple stateless container that can trivially be handled by value. From a TDD perspective it perhaps makes it sound like there is some ulterior motive behind it being a copy, and that that is required for proper functioning of the system.

tests/python/openassetio/managerAPI/test_hostsession.py
```diff
@@ -40,14 +41,39 @@ def mock_logger():
...
+        #   HostSession takes a Host by-value and is non-polymorphic,
+        #   so is no longer the Python subclass instance when queried.
+        #   This will be fixed when the pure-Python Host is removed, at
+        #   which point we should make Host a `final` class in both C++
+        #   and pybind.
+        # assert isinstance(actual_host, Host)
+        assert isinstance(actual_host, _openassetio.managerAPI.Host)
+
+        assert actual_host is not a_host
+        assert actual_host.identifier() == expected_identifier
```
foundrytom, line 73: [fix] Im not sure this is a helpful test - as the identifiers matching doesn't guarantee that the host returns is holding the same interface instance as the supplied host - which is what we really _must_ ensure.

foundrytom (changes requested):

tests/python/openassetio/managerAPI/test_hostsession.py
foundrytom: I guess this maybe raises the question that do we provide an equality operator for `Host` that compares the interface ptr?

feltech: The Python `Host` class has `_interface()`, which we could migrate and check that for equality?

tests/python/openassetio/managerAPI/test_hostsession.py
feltech: Could it be seen as documenting the behaviour?  I.e. users should expect the host retrieved from the session to be fresh, i.e. any changes they make to the host instance won't be persisted.

E.g. if the user does `host.my_variable = 123` then retrieves `.host()` again (or after passing `host` to the `HostSession` constructor), `my_variable` will be lost. 

IIRC hanging new variables off an instance is disallowed by default with pure-pybind instances, so removing the Python `Host` will mitigate that.

How about `test_returns_new_host_wrapper`?

foundrytom: Do we even need this test? It seems like its not testing system behaviour but instead some specifics of how we currently have the language bindings. All we really care about is the host supplied "the same" as the one that went into it.

tests/python/openassetio/managerAPI/test_hostsession.py
foundrytom: Yeah - that was def one of the things on the list to re-evaluate at this point.

src/openassetio-core/include/openassetio/managerAPI/HostSession.hpp
feltech: Added in ec57907b

src/openassetio-core/include/openassetio/managerAPI/HostSession.hpp
feltech: Updated in ec57907b

tests/python/openassetio/managerAPI/test_hostsession.py
feltech: Updated to use `SharedPtr` in 3ca11608 - certainly makes this test a lot simpler!

tests/python/openassetio/managerAPI/test_hostsession.py
feltech: Much simplified by the move to `SharedPtr` in 3ca11608

foundrytom (changes requested):
Thanks @feltech, just one minor thing to check now we're using Ptrs, do you want to squash ready for merge too?

foundrytom:
Thanks @feltech, a few Doxygen refs:

 - `ManagerInterface.hpp:58`
 - `ManagerInterface.hpp:70`
 - `ManagerInterface.hpp:72`
 - `ManagerInterface.py:77`
 -  `ManagerInterface.py:89`
 - `ManagerInterface.py:90`

feltech:
> Thanks @feltech, a few Doxygen refs:
> 
>     * `ManagerInterface.hpp:58`
> 
>     * ~`ManagerInterface.hpp:70`~
> 
>     * ~`ManagerInterface.hpp:72`~
> 
>     * `ManagerInterface.py:77`
> 
>     * ~`ManagerInterface.py:89`~
> 
>     * ~`ManagerInterface.py:90`~
> 
> 
> [edit] Sorry forgot we still have log in python

Those refs are to the `HostSession` class, which is currently still more complete in Python than it is in C++ (although frustratingly there's no hyperlink in the docs to the base class docs), are we sure we want to point the docs to the less complete C++?

foundrytom:
> > Thanks @feltech, a few Doxygen refs:
> > ```
> > * `ManagerInterface.hpp:58`
> > 
> > * ~`ManagerInterface.hpp:70`~
> > 
> > * ~`ManagerInterface.hpp:72`~
> > 
> > * `ManagerInterface.py:77`
> > 
> > * ~`ManagerInterface.py:89`~
> > 
> > * ~`ManagerInterface.py:90`~
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [edit] Sorry forgot we still have log in python
> 
> Those refs are to the `HostSession` class, which is currently still more complete in Python than it is in C++ (although frustratingly there's no hyperlink in the docs to the base class docs), are we sure we want to point the docs to the less complete C++?

Indeed we dont! Sorry! As you were.

feltech:
> [fix] Can we add a test when `None` is supplied? IIRC you can end up with a nullptr in these situations (or could with `boost::python`, and maybe only with raw ptrs).

Good catch! Fixed in 9699d823

PR #429, opened 2022-06-07
[Build] Update bootstrap script to use clang-tidy-12

foundrytom:
Signed-off-by: Tom Cowland <tom@foundry.com>

### DIFF
--- a/resources/build/bootstrap-ubuntu-20.04.sh
+++ b/resources/build/bootstrap-ubuntu-20.04.sh
@@ -4,8 +4,8 @@
 
 set -xeo pipefail
 sudo apt-get update
-# Install gcc, clang-tidy and Python 3.
-sudo apt-get install -y build-essential clang-tidy-12 python3-pip
+# Install gcc, linters, build tools used by conan and Python 3.
+sudo apt-get install -y build-essential pkgconf clang-format-12 clang-tidy-12 python3-pip
 # Install system packages required for the Python 3.9 conan package.
 # These would be installed as part of the conan package install, but
 # we're caching the conan directory via the `actions/cache` Github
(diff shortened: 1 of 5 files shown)


resources/build/bootstrap-ubuntu-20.04.sh
```diff
@@ -5,7 +5,7 @@
 set -xeo pipefail
 sudo apt-get update
 # Install gcc, clang-tidy and Python 3.
-sudo apt-get install -y build-essential clang-tidy-12 python3-pip
+sudo apt-get install -y build-essential pkg-config clang-format-12 clang-tidy-12 python3-pip
```
feltech, line 8: [consider][minor] Might be worth calling out why `pkg-config` is required. I would say put it with the big list below, but that doesn't quite match since "These would be installed as part of the conan package" clearly isn't true for `pkg-config`.

feltech, line 8: [consider][minor] Install `pkgconf` instead, since its apparently better and actively maintained, e.g. from [Fedora changelog](https://fedoraproject.org/wiki/Changes/pkgconf_as_system_pkg-config_implementation#Detailed_Description)

> pkgconf is a newer, actively maintained implementation of pkg-config that supports more aspects of the pkg-config file specification

I tested it and installing `pkgconf` does work in the Vagrant box.

feltech (approved):

resources/build/bootstrap-ubuntu-20.04.sh
foundrytom: Good call, thanks. Updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/429/commits/5be2d2cce999fa17d43c17eadb565120811a18db).

foundrytom: Wording updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/429/commits/d2a882f910b536ed9c480cdd8cdf7dfffc0dd930).

foundrytom:
@feltech Also updated the release notes [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/429/commits/99056c32749a6c377401aa67e3f4386505736eb0).

feltech (approved):

PR #435, opened 2022-06-08
[Llint] Add `flake8` config to `pyproject.toml`

foundrytom:


pyproject.toml
```diff
@@ -10,6 +10,10 @@ disable = [
     "consider-using-f-string"
 ]
 
+[tool.flake8]
```
feltech, line 13: [fix] According to the [docs](https://flake8.pycqa.org/en/latest/user/configuration.html), `pyproject.toml` isn't one of the supported config files for flake8. There is a loong [Github issue](https://github.com/PyCQA/flake8/issues/234) about it.  

I tested it and indeed passing this file as `--config` doesn't work (doesn't give an error, but doesn't obey the config).

There are alternative wrappers that add `pyproject.toml` support. I tried `pyproject-flake8`, which requires running `pflake8` rather than `flake8` as the executable, and that works.

So we could either 
* Add a comment detailing the package that must be used.
* Move the config to a supported file - we already have a `setup.cfg` that we could add to.

feltech (changes requested):

PR #436, opened 2022-06-08
Move Context management functions to `Manager`

### LINKED ISSUE #430: RFC: Move context management to `Manager`
## What

Move host API Context management methods from `Session` to `Manager`, and rename accordingly.

## Why

#428 moves `freezeManagerState` and `thawManagerState` to `Session` following the removal of the transactions API. This was a stop-gap as `Session::createContext` already existed so it kind of made sense.

Looking at the bigger picture, `Session` probably isn't the right place for them:

 - `Context` creation/freeze/thaw only actually involves the manager, the session is just coordinating.
 - A `Context` is tied to the manager whose state it holds, and so can't be shared/reused with multiple managers. This means that the current manager of the session may well change in-between calls, encouraging invalid context use.

Moving context management to `Manager` eliminates any ambiguity about the fact that any given context can only be used with the manager that created the state it holds.

foundrytom:
Closes #430

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -24,15 +24,22 @@ v1.0.0-alpha.X
   - `TraitsData::TraitIds` to `TraitSet`
   - `TraitsData::traitIds()` to `traitSet()`
 
-- Removed the Transactions API, including the `ManagerInterface`
-  methods, and `TransactionCoordinator` helpers. Added `freezeContext`
-  and `thawContext` methods to the `Session` class to allow
-  serialization of a manager's state for persistence or distribution.
-  [#421](https://github.com/OpenAssetIO/OpenAssetIO/issues/421)
-
 - Removed `Session.host()` as it is not useful for foreseeable workflows
   (beyond tests). Note that `HostSession.host()` remains.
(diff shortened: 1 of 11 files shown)


RELEASE_NOTES.md
```diff
@@ -25,11 +25,18 @@ v1.0.0-alpha.X
...
-  methods, and `TransactionCoordinator` helpers. Added `freezeContext`
-  and `thawContext` methods to the `Session` class to allow
-  serialization of a manager's state for persistence or distribution.
+  methods, and `TransactionCoordinator` helpers.
   [#421](https://github.com/OpenAssetIO/OpenAssetIO/issues/421)
 
+- Removed the `Session::createContext` method.
+  [#430](https://github.com/OpenAssetIO/OpenAssetIO/issues/430)
+
+- Added createContext, `freezeContext` and `thawContext` methods to the
```
feltech, line 34: [consider] 
```suggestion
- Added `createContext`, `freezeContext` and `thawContext` methods to the
```

feltech (approved):
LGTM, just a minor typo, can squash and merge when ready.

PR #438, opened 2022-06-08
[Python] Remove `Holder<T>` alias for `openassetio::shared_ptr`

### LINKED ISSUE #432: Use native shared ptr definition over `Holder<T>` in pybind
## What

Replace the use of:

```
py::class_<Host, Holder<Host>>(...)
```

With

```
py::class_<Host, HostPtrt>(...)
```

## Why

- Reduces the ambiguity that `Holder` might have additional functionality.
- Ensures that the python holding object is compatible with the ptr type were using else-where.

## AC

- Existing uses of `Holder<T>` updated (no new ones added)
- `Holder` alias removed

foundrytom:
We added this abstraction to facilitate easier refactoring later on. In use though, it hampers readability of binding code, and potentially allows the pybind holder to differ from what ever pointer we use internally, which could have dire side-effects.

Closes #432

### DIFF
--- a/src/openassetio-python/TraitsDataBinding.cpp
+++ b/src/openassetio-python/TraitsDataBinding.cpp
@@ -11,11 +11,12 @@
 
 void registerTraitsData(const py::module& mod) {
   using openassetio::TraitsData;
+  using openassetio::TraitsDataPtr;
   namespace trait = openassetio::trait;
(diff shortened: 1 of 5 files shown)


feltech (approved):
LGTM rebase and merge when ready

PR #444, opened 2022-06-10
Move `Context` to C++

### LINKED ISSUE #291: `Context` class

feltech:
Closes #291


tests/python/openassetio/test_context.py
```diff
@@ -80,31 +80,30 @@ class TestState(managerAPI.ManagerStateBase):
...
 
-        assert (str(err.value) ==
-                "123 (123) is not a valid Retention (ignored, transient, session, permanent)")
+        with pytest.raises(TypeError, match=expected_msg):
+            a_context.retention = 123
 
 
 class Test_Context_locale:
     def test_when_set_to_unknown_value_then_raises_ValueError(self, a_context):
-        with pytest.raises(ValueError) as err:
```
foundrytom: [fix] Can we add a test where we check that set and get of a valid `TraitsData` instance `is` the same, (to check we do accept that type, and that we're not copying by accident etc... re all the holder discussions)?

tests/python/openassetio/test_context.py
```diff
@@ -114,10 +113,11 @@ def test_when_set_to_None_then_returns_None(self, a_context):
 
 class Test_Context_managerState:
     def test_when_set_to_unknown_type_then_raises_TypeError(self, a_context):
-        with pytest.raises(ValueError) as err:
```
foundrytom: [fix] Same comment re: test as above for assignment with a valid value.

tests/python/openassetio/test_context.py
```diff
@@ -80,31 +80,30 @@ class TestState(managerAPI.ManagerStateBase):
...
-            a_context.access = "unrecognized"
 
-        assert (str(err.value) ==
-                "'unrecognized' is not a valid Access Pattern"
-                " (read, readMultiple, write, writeMultiple, unknown)")
+        expected_msg = (r"incompatible function arguments.*\n"
+                        r".*arg0: openassetio._openassetio.Context.Access")
+
+        with pytest.raises(TypeError, match=expected_msg):
+            a_context.access = "unrecognized"
```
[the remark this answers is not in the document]
foundrytom: Also, for both the enum types should we have a valid assignment test to make sure we're not relying on the coverage from the parent constructor test above? (Incase we ever removed/changed that mechanism).

foundrytom (changes requested):

RELEASE_NOTES.md
```diff
@@ -45,9 +45,21 @@ v1.0.0-alpha.X
   be subclassed.
   [#331](https://github.com/OpenAssetIO/OpenAssetIO/issues/331)
 
+- Removed `Context.managerOptions`.
+  [#291](https://github.com/OpenAssetIO/OpenAssetIO/issues/291)
+
+- Changed `Context.kOther` to `kUnknown`, and changed the default
+  context access to `kUnknown`. This better describes its use, and
+  encourages hosts to properly configure the context before use.
+
```
foundrytom, line 54: [fix] We also need a note here on the change of `Context.access` to an enum from a `string`.

tests/python/openassetio/test_context.py
```diff
@@ -112,6 +112,19 @@ def test_when_set_to_None_then_returns_None(self, a_context):
         assert a_context.locale is None
 
 
+class Test_Context_managerState:
+    def test_when_set_to_unknown_type_then_raises_TypeError(self, a_context):
+        with pytest.raises(ValueError) as err:
+            a_context.managerState = object()
+
+        assert str(err.value) == ""
```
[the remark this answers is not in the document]
feltech: This was updated in a subsequent commit, a8a97d2b

tests/python/openassetio/test_context.py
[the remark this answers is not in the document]
feltech: Yeah you can't assign a raw int, even if its within range of the enum, which is nice.  I'll update both the tests to attempt to set a raw int within range (i.e `a_context.access = 0`), which catches all three assertions in one go.

I'll add happy-path tests too.

tests/python/openassetio/test_context.py
feltech: Added test in aa39e3c1

tests/python/openassetio/test_context.py
feltech: Added test in cfc8bb8a

foundrytom (approved):
Thanks @feltech looks great - do you want to squash ready for merge?

foundrytom (approved):
FFTSAMAYLAHAGW

Issue #445, opened 2022-06-10
Migrate `Manager` context methods to C++ & rename


foundrytom:
## What

Migrate context management methods to C++ from Python, and updated names to be more meaningful.

## Why

Freeze/Thaw imply a behaviour that no longer exists. It used to be a requirement that a context wasn't used after freeze, but this is no longer necessary now it is de-coupled from transactions. It is perfectly legitimate for a host to get the token from a context to pass to a child process, etc... or hold this in a document whilst a process restarts.

## ACs

- Move:
  - `Manager.createContext` -> `Manager::createContext`
  - `Manager.freezeContext` -> `std::string Manager::persistenceTokenForContext(const Context& context)`
  - `Manager.thawContext` -> `Manager::contextFromPersistenceToken(const std::string& token)`
  - `ManagerInterface.freezeState` -> `ManagerInterface::persistenceTokenForState`
  - `ManagerInterface.thawState` -> `ManagerInterface::stateFromPersistenceToken`
- Update relevant documentation to introduce the persistence concept.

## Tasks

- [x] #450 
- [x] #451
- [x] #452
- [x] #467
- [ ] #471

foundrytom:
`persistence` was the best name we could come up with. It is not great, but it helps reinforce the concept that the token is to allow you to keep a context alive beyond the object's lifetime.

PR #450, opened 2022-06-13
Pass `HostSession` to C++ `Manager` during construction

### LINKED ISSUE #447: Add `HostSession` to C++ `Manager`
## What

Add a `HostSession` instance to the C++ `Manager` as per the Python implementation.

## Why

This is required to migrate any methods that require the host session to C++.

## ACs

- `Manager` constructor takes a `HostSessionPtr`, and forwards this to any relevant `ManagerInterface` methods.
- Update `C` bindings

foundrytom:
Closes #447, required for #445.

### DIFF
--- a/tests/openassetio-core-c/private/hostAPI/ManagerTest.cpp
+++ b/tests/openassetio-core-c/private/hostAPI/ManagerTest.cpp
@@ -2,19 +2,24 @@
 // Copyright 2013-2022 The Foundry Visionmongers Ltd
 #include <openassetio/c/errors.h>
 #include <openassetio/c/hostAPI/Manager.h>
+#include <openassetio/c/managerAPI/HostSession.h>
 #include <openassetio/c/managerAPI/ManagerInterface.h>
 #include <openassetio/c/namespace.h>
 
 #include <catch2/catch.hpp>
 #include <catch2/trompeloeil.hpp>
 
+#include <openassetio/hostAPI/HostInterface.hpp>
 #include <openassetio/hostAPI/Manager.hpp>
+#include <openassetio/managerAPI/Host.hpp>
+#include <openassetio/managerAPI/HostSession.hpp>
 #include <openassetio/managerAPI/ManagerInterface.hpp>
 #include <openassetio/typedefs.hpp>
 
 // Private headers.
 #include <handles/InfoDictionary.hpp>
 #include <handles/hostAPI/Manager.hpp>
+#include <handles/managerAPI/HostSession.hpp>
 #include <handles/managerAPI/ManagerInterface.hpp>
 
 #include "../StringViewReporting.hpp"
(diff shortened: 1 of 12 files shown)


tests/openassetio-core-c/private/hostAPI/ManagerTest.cpp
```diff
@@ -117,9 +157,14 @@ SCENARIO("A host calls Manager::identifier") {
...
     auto& mockManagerInterface = static_cast<MockManagerInterface&>(*mockManagerInterfacePtr);
+    // Create a HostSession with our mock HostInterface
+    managerAPI::HostSessionPtr hostSessionPtr =
+        openassetio::makeShared<openassetio::managerAPI::HostSession>(
+            openassetio::makeShared<managerAPI::Host>(
+                openassetio::makeShared<MockHostInterface>()));
 
     // Create the Manager under test.
-    hostAPI::Manager manager{mockManagerInterfacePtr};
+    hostAPI::Manager manager{mockManagerInterfacePtr, hostSessionPtr};
```
feltech, line 167: [consider][minor] I think this chunk of bootstrapping (from after `GIVEN` to `actualErrorMsg`) is long enough and duplicated enough times to warrant a fixture struct (a bit like `InfoDictionaryFixture`).

tests/python/openassetio/hostAPI/test_manager.py
```diff
@@ -85,24 +78,19 @@ def some_refs():
...
 
-        assert str(err.value) == (
-            '__init__(): incompatible constructor arguments. The following argument types '
-            'are supported:\n'
-            '    1. openassetio._openassetio.hostAPI.Manager(managerInterface: '
-            'openassetio._openassetio.managerAPI.ManagerInterface)\n'
-            '\n'
-            'Invoked with: None')
+        # Will contain a repr of the mock HostSession instance including address
+        assert str(err.value).startswith('__init__(): incompatible constructor arguments.')
```
feltech, line 93: [consider] Might be worth also asserting the expected types are reported, i.e. `openassetio._openassetio.hostAPI.Manager` and possibly `openassetio._openassetio.managerAPI.ManagerInterface`. 

This then tests we get the order of `register*` functions correct in the bindings.  If the order is wrong then these won't be reported as the Python types (IIRC it'll use `typeinfo` and be all messy C++ types - see [pybind docs](https://pybind11.readthedocs.io/en/stable/advanced/misc.html#avoiding-c-types-in-docstrings)).

There is a handy keyword argument to `pytest.raises`, `match="<regex>"`, which could help.

Used in `test_context.py` for a similar reason.

feltech (approved):

tests/python/openassetio/hostAPI/test_manager.py
foundrytom: Updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/450/commits/1797f6a9e0c98c524fff42fd912813cd661446ed).

feltech: Do you prefer the full error message rather than stripping the cruft via a regex using `match=`?

foundrytom: How do you find [this](https://github.com/OpenAssetIO/OpenAssetIO/pull/450/commits/4b6a5bdc3a6db4795567a424ddcb6e8fde083753) for readability?

foundrytom: Sorry I missed the focus on type presentation. I think its also important to check that the message includes some illustration that the types are wrong. How about [this](https://github.com/OpenAssetIO/OpenAssetIO/pull/450/commits/c4c16388bb839c5f486b672b70bb46010fae3ffe) (+fix [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/450/commits/34682b4fcc2b730971e14b568cdf426054abcb75)).

feltech (approved):
LGTM squash and merge when ready.

PR #465, opened 2022-06-14
[Core] Migrate Manager.initialize() to C++

### LINKED ISSUE #455: Migrate `initialize` to C++
## What

Move `initialize` to C++ 

## ACs

 - `initialize` can be called from C++ or Python
 - Issue logged for C API work

## Out of scope

- C API

feltech:
Closes #455


src/openassetio-core-c/private/managerAPI/CManagerInterfaceAdapter.hpp
```diff
@@ -41,6 +41,10 @@ class OPENASSETIO_CORE_C_EXPORT CManagerInterfaceAdapter : ManagerInterface {
   /// Wrap the C suite's `info` function.
   [[nodiscard]] InfoDictionary info() const override;
 
+  /// Wrap the C suite's `initialize` function.
+  /// @todo Unimplemented for now - simply a no-op.
```
foundrytom, line 45: [fix] `@todo`s get turned into [this page](https://openassetio.github.io/OpenAssetIO/todo.html), its good to write them in the imperative tense if we can so its easier to remember what is actually required.

Given how simple this method is, is it worth implementing this now to save having to go back to it, then we don't think its done already as it's in the header?

foundrytom (changes requested):
👍

foundrytom (approved):
Feel free to squash and merge once the TODO is updated @feltech

src/openassetio-core-c/private/managerAPI/CManagerInterfaceAdapter.hpp
feltech: Updated ans squashed in. Though note that this particular file isn't included in the Doxygen output, since it's under `private`.

foundrytom: Doh!

Issue #130, opened 2021-12-07
Use the `when` and `then` voice for unit tests


foundrytom:
## What

Update the existing code in [tests](tests) to use the revised test voice policy defined in [CONTRIBUTING.md](CONTRIBUTING.md).

## Why

It properly describes the expectations of the system under test.

## Notes

If tests make more sense as `then` only, also update `CONTRIBUTING.md`

foundrytom:
Whilst where here we should conform the naming of tests such that the target matches the case, rather than being all lowercase. See https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/133#discussion_r765721658

foundrytom:
And also move fixtures after the test cases to increase readability.

foundrytom:
See also https://github.com/OpenAssetIO/OpenAssetIO/pull/467#discussion_r900001740

PR #467, opened 2022-06-15
[Core] Rename context persistence methods

foundrytom:
Part of #445, renames the various `freeze` and `thaw` methods.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -32,16 +32,23 @@ v1.0.0-alpha.X
   methods, and `TransactionCoordinator` helpers.
   [#421](https://github.com/OpenAssetIO/OpenAssetIO/issues/421)
 
-- Removed `Session::createContext`.
+- Removed the `Session` methods `createContext`, `freezeContext` and
+  `thawContext`.
   [#430](https://github.com/OpenAssetIO/OpenAssetIO/issues/430)
+  [#445](https://github.com/OpenAssetIO/OpenAssetIO/issues/445)
 
-- Added `createContext`, `createChildContext`, `freezeContext` and
-  `thawContext` methods to the `Manager` class to facilitate context
-  creation and the serialization of a manager's state for persistence or
-  distribution.
+- Added `createContext`, `createChildContext`,
+  `persistenceTokenForContext` and `contextFromPersistenceToken` methods
+  to the `Manager` class to facilitate context creation and the
+  serialization of a manager's state for persistence or distribution.
   [#421](https://github.com/OpenAssetIO/OpenAssetIO/issues/421),
   [#430](https://github.com/OpenAssetIO/OpenAssetIO/issues/430)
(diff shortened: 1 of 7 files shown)


python/openassetio/hostAPI/Manager.py
```diff
@@ -1042,52 +1042,69 @@ def createChildContext(self, parentContext):
...
+        actions.
 
-        @param stateToken `str` A token previously returned from @ref
-        freezeContext by this manager.
+        @param persistenceToken `str` A token previously returned from
+        @ref persistenceTokenForContext by this manager.
 
         @return @fqref{Context} "Context" A context that will be
-        associated with the previously frozen one by the manager.
+        associated with the same logical group of action as the context
```
feltech, line 1093: [consider]/[plan] It was clear to me when we discussed it, but on re-reading I think it could be worth some disambiguation of context presistence tokens vs. child contexts, perhaps use-case examples of when one should be used vs. the other?  

E.g. under `createChildContext ` we say

> [...] in order to parallelise actions that are part of the same logical group [...]

and here we say

> [...] context that will be associated with the same logical group of action [...]

feltech, line 1093: [fix]
```suggestion
        associated with the same logical group of actions as the context
```

tests/python/openassetio/hostAPI/test_manager.py
```diff
@@ -514,37 +514,38 @@ def test_when_called_with_parent_then_props_copied_and_createState_called_with_p
         mock_manager_interface.mock.createState.assert_not_called()
 
 
-class Test_Manager_freezeContext:
+class Test_Manager_persistenceTokenForContext:
 
-    def test_when_called_then_the_managers_frozen_token_is_returned(
+    def test_when_called_then_the_managers_persistence_token_is_returned(
```
feltech, line 519: [consider][minor] Not introduced by this PR, but I think these tests would be a candidate for avoiding the `when`/`then` style, since there is no branching logic.

RELEASE_NOTES.md
```diff
@@ -32,17 +32,24 @@ v1.0.0-alpha.X
   methods, and `TransactionCoordinator` helpers.
   [#421](https://github.com/OpenAssetIO/OpenAssetIO/issues/421)
 
-- Removed `Session::createContext`.
+- Removed the `Session` methods `createContext`, `freezeContext` and
+  `thawContext.
```
feltech, line 36: [fix]
```suggestion
  `thawContext`.
```

feltech (changes requested):

python/openassetio/hostAPI/Manager.py
foundrytom: Added #470.

tests/python/openassetio/hostAPI/test_manager.py
foundrytom: Maybe we could pick this up when we get back to #130?

feltech (approved):

PR #469, opened 2022-06-17
[Core] Refine context creation mechanism

foundrytom:
A draft PR (dependent on #467) for review of a more refined approach to context creation, that means we can make the `ManagerInterface` methods easier to implement

Changes start at [this commit](https://github.com/OpenAssetIO/OpenAssetIO/pull/469/commits/a5c83f81ca46487b57f9b1490b9de85488440e55).

### DIFF
--- a/python/openassetio/hostAPI/Manager.py
+++ b/python/openassetio/hostAPI/Manager.py
@@ -1037,13 +1037,17 @@ class Manager(_openassetio.hostAPI.Manager, Debuggable):
         @see @ref createContext
         @see @fqref{Context} "Context"
         """
-        return Context(
+        context = Context(
             access=parentContext.access,
             retention=parentContext.retention,
-            locale=parentContext.locale,
-            managerState=self.__impl.createChildState(self.__hostSession,
-                                                        parentContext.managerState)
+            locale=parentContext.locale
         )
+
+        if parentContext.managerState:
+            context.managerState = \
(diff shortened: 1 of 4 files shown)


python/openassetio/hostAPI/Manager.py
```diff
@@ -1037,57 +1037,83 @@ def createChildContext(self, parentContext):
...
+        @warning This only encapsulates the logical identity of the
+        Context, such when restored, any API calls made using the
+        resulting Context will be logically associated with the one
+        supplied here. It does not encode the current access, retention
+        or locale.
 
         @see @ref stable_resolution
         """
-        token = self.__impl.freezeState(context.managerState, self.__hostSession)
+        if not context.managerState :
```
feltech, line 1080: [question] I think this makes sense from the perspective of a host that wants to work generically with all managers that may or may not support state tokens, without changing the host code to cover both cases. But I wonder if it is ever valid to use a state token based workflow with managers that don't support it?

feltech (approved):
OK, just a question to throw a spanner in the works.

python/openassetio/hostAPI/Manager.py
feltech: Capturing offline discussion: we assume that if the (well behaved) manager doesn't provide a `managerState`, then it simply doesn't require one for the workflows that `managerState` (and hence state tokens) is designed to simplify, so its fine for the host to be oblivious.

PR #477, opened 2022-06-21
[Docs] Conform use of `fqref`

### LINKED ISSUE #472: Conform namespace token in doc `@fqref`s
## What

Ensure all `@fqref{...}` usage uses `.` as a namespace token.

## Why

Both are supported by Python, and `.` is more universal.

foundrytom:
We had a mix of . and :: use when invoking the custom Doxygen command 'fqref.' Both work, but we elected to use . as the documentation is used for multiple languages.

Closes #472

### DIFF
--- a/src/openassetio-core/include/openassetio/InfoDictionary.hpp
+++ b/src/openassetio-core/include/openassetio/InfoDictionary.hpp
@@ -13,8 +13,8 @@ inline namespace OPENASSETIO_CORE_ABI_VERSION {
 /// Types available as values in a InfoDictionary.
 using InfoDictionaryValue = std::variant<Bool, Int, Float, Str>;
 /**
- * Dictionary type used for @fqref{managerAPI::ManagerInterface::info}
- * "ManagerInterface::info".
+ * Dictionary type used for @fqref{managerAPI.ManagerInterface.info}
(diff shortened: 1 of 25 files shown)


src/openassetio-core/include/openassetio/InfoDictionary.hpp
```diff
@@ -13,7 +13,7 @@ inline namespace OPENASSETIO_CORE_ABI_VERSION {
 /// Types available as values in a InfoDictionary.
 using InfoDictionaryValue = std::variant<Bool, Int, Float, Str>;
 /**
- * Dictionary type used for @fqref{managerAPI::ManagerInterface::info}
+ * Dictionary type used for @fqref{managerAPI.ManagerInterface.info}
  * "ManagerInterface::info".
```
feltech, line 17: [fix] Just this one missed

feltech (approved):
Neat, I could only find one missed case. Happy for this to be squashed and merged once fixed.

src/openassetio-core/include/openassetio/InfoDictionary.hpp
foundrytom: Good spot thanks! Fixed in squash.

PR #480, opened 2022-06-22
[Docs] Document TraitsData value retrieval API decision

foundrytom:
Finally!


feltech (approved):
Looks good re. covering the choice between `bool` vs. `optional`. Though bear in mind we ended up using an error code return for C++ `TraitBase`, so that the reason for the missing value could be communicated without resorting to exceptions.  Its a separate issue, but it's also an additional advantage of returning a status rather than relying on a binary `optional`.

Issue #481, opened 2022-06-22
Add `create` methods to replace `openassetio::makeShared`


foundrytom:
## What

Add static factory methods (`::create`, with constructor arguments/docs) to all shared ownership classes, and make the constructor private.

## Why

For consistent initialization of resources, and to encapsulate what we actually use for our shared object handles, enforcing a single way to construct these objects avoids consumer code inconsistencies or churn in the long term. This matches best practice as described [here](https://en.cppreference.com/w/cpp/memory/enable_shared_from_this).

## ACs

- Tests for the visibility of C++ constructor.
- Python to bind `__init__` to the `create` method.

PR #482, opened 2022-06-22
[Core] Remove `openassetio::SharedPtr` alias

### LINKED ISSUE #460: Remove `openassetio::SharedPtr` alias

foundrytom:
We plan to `enable_shared_from_this`, which makes less sense to alias as it is specific to std::shared_ptr. It was raised as odd to alias one and not the other.

We plan to properly isolate callers from the nature of the pointers in #481 using static factory methods vs bare constructors, to ensure we can still refactor their implementation in future if needed.

Closes #460.


src/openassetio-core/include/openassetio/Context.hpp
```diff
@@ -3,6 +3,7 @@
 #pragma once
 #include <array>
 #include <cstddef>
+#include <memory>
```
feltech, line 6: :heart:

Issue #456, opened 2022-06-14
Design batch method C/C++ API


foundrytom:
## What

Design the C++ API for methods such as `managementPolicy` and `resolve` that handles the following considerations in a fashion that can be coherently be bound through our C API. 

- Arrays of inputs
- Arrays of outputs
- Arrays of single types (values and ptrs)
- Arrays of mixed-types

## ACs

 - An approach that yields consistent, predictable API signatures across the project.
 - C++ API signature attached to issue for `managementPolicy`.
   - [x] #486
 - [x] #487
 - Justification for any specific choices made, especially regarding relationship between python, C and C++

PR #483, opened 2022-06-22
[Docs] Add batch result mechanism DR

feltech:
Part of #456. Add a decision record to detail our strategy choice for batch-first functions. 

Specifically, whether such functions should return an array of results, or should accept callbacks that are executed on each result individually. 

A sketch and some performance tests in C++ are included in the Appendix.


decisions/DR009-Batch-method-result-types.md
```diff
@@ -97,7 +97,176 @@ performant c++ to c++ path.
...
+
+Since we're concerned with the relative performance, the key output of
+the perftest is the ratio of timings between two solutions. The mean and
+standard deviation of these timing ratios are reported in the tables
+below. We round to 3 significant figures - the resolution of the fastest
+results are 3 significant figures (hundreds of nanoseconds).
+
+As shorthand for the different test cases, we define
+* _vector_ as the case that returns an array of results.
+* _callback_ as the case that executes a `std::function` callback on
```
foundrytom, line 190: [fix] Could we really spell it out and have `(Option 1)` and `(Option 2)` on these two?

decisions/DR009-Batch-method-result-types.md
```diff
@@ -0,0 +1,272 @@
...
+  each result.
+* _callbackToVector_ as the case that builds an array as callbacks
+  are executed and then iterates over it.
```
foundrytom, line 193: [fix]
```suggestion
  are executed and then iterates over it (simulating Option 1).
```

decisions/DR009-Batch-method-result-types.md
```diff
@@ -0,0 +1,272 @@
...
+batch size of 1 (i.e. only querying a single entity reference).
+
+Firstly, the primary results comparing the array return value solution
+versus the `std::function` callback solution is shown below.
+
+| numerator/denominator     | mean | std dev |
+|---------------------------|------|---------|
+| vector/callback           | 1.29 | 0.761   |
+| vector/callbackToVector   | 1.15 | 0.570   |
+| callbackToVector/callback | 1.24 | 0.745   |
```
foundrytom, line 212: [question] Do you think this is a useful comparison to have in the table (at the risk of shadowing the conclusion that the callback method is faster, even when building a vector?) Same applies to the larger batch results.

foundrytom (approved):
Great work @feltech. I'll save comment on the changes to `perftest`, and leave that to those more knowledgeable on the subtleties of performance testing. Just a few minor tweaks/q.

Can we squash all these commits together when we merge?

decisions/DR009-Batch-method-result-types.md
feltech: Good point, this comparison isn't discussed below here, so should be removed.

decisions/DR009-Batch-method-result-types.md
```diff
@@ -0,0 +1,272 @@
...
+- **Outcome:** Batch methods will adopt a per-result callback mechanism.
+
+## Background
+
+OpenAssetIO employs a batch-first design. In that, all data centric
+methods take multiple inputs, and produce multiple outputs. This allows
+a `ManagerPlugin` to optimize queries to potentially high-latency
+back-end services.
+
+Many methods may have per-value error states. Such that the output of
```
meshula, line 17:
```suggestion
Many methods may have per-value error states, such that the output of
```

decisions/DR009-Batch-method-result-types.md
```diff
@@ -0,0 +1,272 @@
...
+these methods is never of a uniform type. For example, `resolve` for any
+given entity reference may provide a `TraitsData` instance, or an error.
+An error for a specific value should not error the entire batch.
+
+This raises the question of how best express the resulting data across
```
meshula, line 22:
```suggestion
This raises the question of how to best express the resulting data across
```

decisions/DR009-Batch-method-result-types.md
```diff
@@ -0,0 +1,272 @@
...
+API serves as a binding point for other languages.
+
+A design mantra is that the implementation in any given language should
+be as idiomatic as possible. An exception to which is that Python and
+C++ are often used in the same breath within M&E DCC applications,
+keeping those two languages as close to each other as possible has many
+benefits.
+
+The performance of API methods is important, as many scenarios
+(particularly `resovle`) have a high query volume. Syntax alone is not
```
meshula, line 37:
```suggestion
(particularly `resolve`) have a high query volume. Syntax alone is not
```

decisions/DR009-Batch-method-result-types.md
```diff
@@ -0,0 +1,272 @@
...
+enough of a justification for any particular design choice.
+
+For the scope of this document, it is assumed that the input to any of
+these methods will be an array, regardless of the choices here.
+
+Input options are not explored here, as the sole purpose of the
+batch-centric nature is to optimise queries, which requires all input
```
meshula, line 44:
```suggestion
batch-centric pattern is to optimise queries, which requires all input
```

decisions/DR009-Batch-method-result-types.md
```diff
@@ -0,0 +1,272 @@
...
+#### Cons
+
+- Less idiomatic in some target languages, especially Python.
+- Precludes data-centric functional programming without caller-side
+  wrapping.
+- Function-centric result is harder to propagate across process/host
+  boundaries.
+- Potential for high per-result call cost depending on the language
+  stack.
+
```
meshula, line 90:
```suggestion
- has possible ramifications around concurrency ~ client side code will potentially have to be smart to avoid races and other faults
- at the same time that it has the possibility to simplify value handling, it also has the possibility of just the opposite, causing complexity, under async conditions
```

meshula: I am definitely buying the argument of performance as outlined in Appendix 2, but it isn't instantly clear to me that the performance characteristics hold up under a a situation such as 200 requests are dispatched to as many threads. I'd be curious to see something of the nature of a worst case

```
condvar<int> resultCount = 0;
mutex serialSorry;
void handler(int requestIndex) { guard(serialSorry); wait(rand()); }
dispatchgroup {
    for (0...199) { dispatchRequest( { wait(rand()); handler(i); ++resultCount; }); }
}
await(resultCount == 200);
```

okay, I guess I have convinced myself of Appendix 2, this worst case example would be no worse than the single threaded example. Move along, nothing to see here :)

decisions/DR009-Batch-method-result-types.md
```diff
@@ -0,0 +1,272 @@
...
+Interestingly, using callbacks and appending to a vector on the caller
+side (_callbackToVector_) appears slightly more performant than simply
+returning a vector directly (_vector_). This could be due to compiler
+intricacies (e.g. moving the vector across the DSO boundary), but more
+investigation is required.
+
+However, with larger batches we start to see the performance
+characteristics of the chosen option dominate over external factors.
+Here, the callback solution(s) is the clear winner.
+
```
meshula, line 262: Per my musings above, I'd find it compelling to have concurrent examples worked, in addition to single-threaded.

decisions/DR009-Batch-method-result-types.md
feltech: Hopefully clarified in 1f0f5ae5

decisions/DR009-Batch-method-result-types.md
feltech: Hopefully clarified in https://github.com/OpenAssetIO/OpenAssetIO/commit/1f0f5ae5bb2b1a3f534c4eabfe6d6f97ac15e70c

decisions/DR009-Batch-method-result-types.md
feltech: Removed in 73ccbccd

decisions/DR009-Batch-method-result-types.md
```diff
@@ -0,0 +1,272 @@
...
+taken, on the assumption that this represents the results that are least
+skewed by external factors.
+
+A database (simple map of entity reference to URL) of size 10000 is
+generated to draw from. Input entity references are randomly chosen from
+this database, with the probability of an erroneous entity reference set
+to 0.1. Timings are nanosecond precision.
+
+Since we're concerned with the relative performance, the key output of
+the perftest is the ratio of timings between two solutions. The mean and
```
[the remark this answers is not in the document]
feltech: Elaborated in 1e49700d

decisions/DR009-Batch-method-result-types.md
```diff
@@ -0,0 +1,296 @@
...
+
+- The result type is complex and requires significant boilerplate
+  on the host side to un-pack.
+- Requires additional allocations/copies of the result container.
+
+### Option 2
+
+Provide success/error callbacks to be invoked for each input value.
+
+The manager guarantees to invoke one of the callbacks for each input
```
feltech, line 80: [question] Is is worth a note on exceptions, or is that just muddying the waters?

foundrytom: We def need to address this (especially in terms of exceptions thrown during the callback), can we save this for the full design once the high-level choice of callback or not is agreed?

decisions/DR009-Batch-method-result-types.md
foundrytom: @meshula we've added some notes on threading [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/483/commits/04e25971225bb0c3ca0077074a39e36a2585cede). How does this fit into your thoughts more generally?

meshula: Yes, thanks for laying out the strategy explicitly.

antirotor:
Just quick comment from our point of view - we are mostly used to Option 1 (array of results), but that's mainly because we are Python-centric and it makes more sense there. This said, your decision is well researched and argumented and isn't breaking anything for us.

foundrytom:
@feltech Tidied the results of today's ruminations [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/483/commits/4da107f86a7beb99ba49f4a5d667cbaf7d036547?short_path=1b89b23#diff-1b89b2344444427a500f96c553304d9c137317ced2502be234c77229669e60ea).

feltech:
LGTM - I can't approve my own PR, so please approve on my behalf - then squash and merge when ready.

foundrytom (approved):

foundrytom:
> LGTM - I can't approve my own PR, so please approve on my behalf - then (I can) squash and merge when ready.

It's also odd that I can given that I'm also an assignee. I think _maybe_ I was a reviewer before I was assigned and that trumps it?

foundrytom:
@feltech I just force-pushed a change to mark it `Decided`.

PR #495, opened 2022-06-30
C++ Factory methods

### LINKED ISSUE #481: Add `create` methods to replace `openassetio::makeShared`

foundrytom:
This deliberately omits classes that are not constructed by API code, or callers of the API (from either side). `HostInterface`, `ManagerInterface` and `ManagerStateBase` are only ever constructed by their direct owners (the host, the manager's plugin and manager respectively).

Closes #481

### DIFF
--- a/src/openassetio-core/Context.cpp
+++ b/src/openassetio-core/Context.cpp
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 The Foundry Visionmongers Ltd
+#include <openassetio/Context.hpp>
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+ContextPtr Context::make(Access access, Retention retention, TraitsDataPtr locale,
+                         managerApi::ManagerStateBasePtr managerState) {
+  return std::shared_ptr<Context>(
+      new Context(access, retention, std::move(locale), std::move(managerState)));
+}
+
+Context::Context(Access access_, Retention retention_, TraitsDataPtr locale_,
+                 managerApi::ManagerStateBasePtr managerState_)
+    : access{access_},
+      retention{retention_},
+      locale{std::move(locale_)},
+      managerState{std::move(managerState_)} {}
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
--- a/src/openassetio-core/TraitsData.cpp
+++ b/src/openassetio-core/TraitsData.cpp
@@ -65,6 +65,16 @@ class TraitsData::Impl {
   PropertiesByTrait data_;
 };
 
+TraitsDataPtr TraitsData::make() { return std::shared_ptr<TraitsData>(new TraitsData()); }
+
+TraitsDataPtr TraitsData::make(const TraitSet& traitSet) {
+  return std::shared_ptr<TraitsData>(new TraitsData(traitSet));
+}
+
+TraitsDataPtr TraitsData::make(const TraitsDataConstPtr& other) {
+  return std::shared_ptr<TraitsData>(new TraitsData(*other));
+}
+
 TraitsData::TraitsData() : impl_{std::make_unique<Impl>()} {}
 
 TraitsData::TraitsData(const TraitSet& traitSet) : impl_{std::make_unique<Impl>(traitSet)} {}
--- a/src/openassetio-core/include/openassetio/Context.hpp
+++ b/src/openassetio-core/include/openassetio/Context.hpp
@@ -34,7 +34,7 @@ OPENASSETIO_DECLARE_PTR(Context)
  *  own, one will always be supplied through the ManagerInterface entry
  *  points.
  */
-class Context final {
+class OPENASSETIO_CORE_EXPORT Context final {
  public:
   /**
    * Storage for enum name lookup array.
@@ -80,7 +80,7 @@ class Context final {
    * then it will hint as to whether the Host is wanting to choose a
    * new file name to save, or open an existing one.
    */
-  Access access{kUnknown};
+  Access access;
 
   /**
    * A concession to the fact that it's not always possible to fully
@@ -102,7 +102,7 @@ class Context final {
    * to handle these situations correctly, Hosts are required to set
    * this property to reflect their ability to persist this information.
    */
-  Retention retention{kTransient};
+  Retention retention;
 
   /**
    * In many situations, the @ref trait_set of the desired @ref entity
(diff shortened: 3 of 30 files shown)


tests/openassetio-core-c/private/hostApi/ManagerTest.cpp
```diff
@@ -159,14 +159,11 @@ SCENARIO("A host calls Manager::identifier") {
         openassetio::makeShared<MockManagerInterface>();
     auto& mockManagerInterface = static_cast<MockManagerInterface&>(*mockManagerInterfacePtr);
     // Create a HostSession with our mock HostInterface
-    managerApi::HostSessionPtr hostSessionPtr =
-        openassetio::makeShared<openassetio::managerApi::HostSession>(
-            openassetio::makeShared<managerApi::Host>(
-                openassetio::makeShared<MockHostInterface>()));
+    managerApi::HostSessionPtr hostSessionPtr = managerApi::HostSession::make(
+        managerApi::Host::make(openassetio::makeShared<MockHostInterface>()));
```
feltech, line 163: [consider] We should probably remove the `makeShared` alias at this point.

src/openassetio-core/Context.cpp
```diff
@@ -0,0 +1,20 @@
...
+                         managerApi::ManagerStateBasePtr managerState) {
+  return std::shared_ptr<Context>(
+      new Context(access, retention, std::move(locale), std::move(managerState)));
+}
+
+Context::Context(Access access_, Retention retention_, TraitsDataPtr locale_,
+                 managerApi::ManagerStateBasePtr managerState_)
+    : access{access_},
+      retention{retention_},
+      locale{std::move(locale_)},
```
feltech, line 17: [question] Is it expected that `locale` is not a deep copy?

src/openassetio-core/include/openassetio/Context.hpp
```diff
@@ -129,6 +129,16 @@ class Context final {
...
   managerApi::ManagerStateBasePtr managerState{};
 
+  /**
+   * Constructs a new context.
+   *
+   * @warning This method should never be called directly by host code -
+   * @fqref{hostApi.Manager.createContext} "Manager.createContext"
+   * should always be used instead.
+   */
+  [[nodiscard]] static ContextPtr make(Access access = kUnknown, Retention retention = kTransient,
```
feltech, line 139: [question] Is it ever valid to construct a `Context` with a `nullptr` `locale`?  Just wondering about the order of arguments here, i.e. we could put `locale` first without a default argument, to enforce that it is set.

src/openassetio-core/include/openassetio/Context.hpp
```diff
@@ -129,6 +129,16 @@ class Context final {
    */
   managerApi::ManagerStateBasePtr managerState{};
```
feltech, line 130: [consider][minor] Here and for `locale` it would perhaps be clearer to a reader if we get rid of the default initialisation (i.e. the `{}`).

src/openassetio-core/TraitsData.cpp
```diff
@@ -65,6 +65,16 @@ class TraitsData::Impl {
   PropertiesByTrait data_;
 };
 
+TraitsDataPtr TraitsData::make() { return std::shared_ptr<TraitsData>(new TraitsData()); }
+
+TraitsDataPtr TraitsData::make(const TraitSet& traitSet) {
+  return std::shared_ptr<TraitsData>(new TraitsData(traitSet));
+}
+
+TraitsDataPtr TraitsData::make(const TraitsDataPtr& other) {
```
feltech, line 74: [consider] / [plan] This seems like a good place to use a `TraitsDataConstPtr` as the argument. Would also be nice to use in the tests where we take a copy of the data for comparisons.  

I'm not sure if taking a ConstPtr and passing a non-const Ptr would cause a ConstPtr temporary to be created and hence do a pointless refcount.  But it still feels like the right thing to do.

feltech (approved):

src/openassetio-core/Context.cpp
foundrytom: The current model is that is isn't - ie: its just holding a reference to whatever you supply. Think we should look into that?

src/openassetio-core/include/openassetio/Context.hpp
foundrytom: 🤔 or maybe the default should be a new `TraitsData`?

src/openassetio-core/TraitsData.cpp
foundrytom: Good call, Ill fix that.

I _think_ we use it in the tests already as we bind this method to python for the copy constructor?

src/openassetio-core/Context.cpp
foundrytom: Added [this Issue](https://github.com/OpenAssetIO/OpenAssetIO/issues/499) to determine the most desirable behavior once we have the minimal vertical slice working.

src/openassetio-core/include/openassetio/Context.hpp
foundrytom: Added [this Issue](https://github.com/OpenAssetIO/OpenAssetIO/issues/499) to determine the most desirable behavior once we have the minimal vertical slice working.

src/openassetio-core/include/openassetio/Context.hpp
foundrytom: Removed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/495/commits/9e028fe534c2d59bcc6dfd2ca1caa704ed2afce6).

tests/openassetio-core-c/private/hostApi/ManagerTest.cpp
foundrytom: Removed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/495/commits/eda7e8bb2c04f8c5f1a365f95a32c7a1fd4e71a5).

foundrytom:
@feltech fixed those various comments, but sorry had to rebase to pick up #500.

feltech (approved):

PR #501, opened 2022-07-04
[Core] Migrate managementPolicy to C++

### LINKED ISSUE #458: Migrate `managementPolicy` to C++
## What

Allow `managementPolicy` to be called from C++ or Python

## ACs

- `managementPolicy` migrated to C++ as per #456

feltech:
Closes #458

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -129,13 +129,17 @@ v1.0.0-alpha.X
   [#455](https://github.com/OpenAssetIO/OpenAssetIO/issues/455)
 
 - Migrated the following `ManagerInterface` methods to C++
-  `initialize`, `createState`, `createChildState`,
+  `initialize`, `managementPolicy`, `createState`, `createChildState`,
   `persistenceTokenForState`, `stateFromPersistenceToken`.
+  [#455](https://github.com/OpenAssetIO/OpenAssetIO/issues/455)
+  [#458](https://github.com/OpenAssetIO/OpenAssetIO/issues/458)
   [#445](https://github.com/OpenAssetIO/OpenAssetIO/issues/445)
 
-- Migrated the following `Manager` methods to C++ `createContext`,
-  `createChildContext`, `persistenceTokenForContext`,
-  `contextFromPersistenceToken`.
+- Migrated the following `Manager` methods to C++ `initialize`
+  `managementPolicy`, `createContext`, `createChildContext`,
+  `persistenceTokenForContext`, `contextFromPersistenceToken`.
+  [#455](https://github.com/OpenAssetIO/OpenAssetIO/issues/455)
+  [#458](https://github.com/OpenAssetIO/OpenAssetIO/issues/458)
   [#445](https://github.com/OpenAssetIO/OpenAssetIO/issues/445)
 
 - Switched to preferring un-versioned `clang-tidy` executables when
--- a/python/openassetio/test/manager/apiComplianceSuite.py
+++ b/python/openassetio/test/manager/apiComplianceSuite.py
@@ -20,8 +20,7 @@ manager plugin complies to the relevant core OpenAssetIO API contract.
 
 This suite is solely concerned with verifying that a plugin meets the
 requirements of the API, and can handle all documented calling patterns.
-For example, that when a
-@ref openassetio.managerApi.ManagerInterface.ManagerInterface.managementPolicy
+For example, that when a @fqref{managerApi.ManagerInterface.managementPolicy}
 "managementPolicy" query returns a non-ignored state, that there are no
 errors calling the other required methods for a managed entity with
 those @ref trait "traits".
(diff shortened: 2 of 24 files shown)


python/openassetio/test/manager/apiComplianceSuite.py
```diff
@@ -170,7 +170,7 @@ def __assertPolicyResults(self, numTraitSets, context, traitSet=("entity",)):
         @param traitSet `List[str]` The set of traits to pass to
         the call to managementPolicy.
         """
-        traitSets = [traitSet for _ in range(numTraitSets)]
+        traitSets = [set(traitSet) for _ in range(numTraitSets)]
```
foundrytom, line 173: [fix] Could we change `traitSet=("entity",)` to `traitSet={entity,}` instead of converting here? Then this conversion won't mask incorrect types fed into this method? Just we know people use tests as examples and so thinks like `L160` could be misleading (as they should be a set too).

RELEASE_NOTES.md
```diff
@@ -120,13 +120,17 @@ v1.0.0-alpha.X
   [#455](https://github.com/OpenAssetIO/OpenAssetIO/issues/455)
 
 - Migrated the following `ManagerInterface` methods to C++
-  `initialize`, `createState`, `createChildState`,
+  `initialize`, `managementPolicy`, `createState`, `createChildState`,
   `persistenceTokenForState`, `stateFromPersistenceToken`.
+  [#455](https://github.com/OpenAssetIO/OpenAssetIO/issues/455)
+  [#458](https://github.com/OpenAssetIO/OpenAssetIO/issues/458)
   [#445](https://github.com/OpenAssetIO/OpenAssetIO/issues/445)
```
foundrytom, line 127: [consider][minor] Might be nice to sort these, and below?

foundrytom (changes requested):
Great stuff @feltech thanks- good idea on the re-org too. Just a few minor things.

python/openassetio/test/manager/apiComplianceSuite.py
feltech: I did it this way because tuples are immutable, so are fine to use as default arguments, whereas if using a `set` then `pylint` says "Dangerous default value set() as argument (163:4) [dangerous-default-value]".

An alternative is to default to `None` then `if traitSet is None: traitSet = {"entity"}`, would that be preferred?

RELEASE_NOTES.md
feltech: I sorted these in roughly the same order as the list of methods in the text.  Happy to sort in issue order though, if you prefer? 

In general, how should these be sorted?  Issue number order, completion order, or as here attempt to match the order with the text?

foundrytom: You raise an excellent question! :) The OCD in me was probably just hanging after numeric, but totally happy to leave as-is

python/openassetio/test/manager/apiComplianceSuite.py
foundrytom: Happy for a pylint disable (as it's definitely not dangerous in this case as manager isn't ever going to modify traitSet), or the latter, which ever you would prefer. I think the main thing is not using anything other than `set` at the call site, so people reading tests will have the 'setness' cemented in their brains.

foundrytom: fwiw, I guess one thing I prefer about the default argument (given its safe here), is that its more readable what the behaviour is without having to look at the implementation.

src/openassetio-core-c/private/managerApi/CManagerInterfaceAdapter.cpp
```diff
@@ -81,6 +81,13 @@ InfoDictionary CManagerInterfaceAdapter::info() const {
 
 void CManagerInterfaceAdapter::initialize([[maybe_unused]] const HostSessionPtr& hostSession) {}
 
+trait::TraitsDatas CManagerInterfaceAdapter::managementPolicy(
+    [[maybe_unused]] const trait::TraitSets& traitSets,
+    [[maybe_unused]] const ContextConstPtr& context,
+    [[maybe_unused]] const HostSessionPtr& hostSession) const {
+  return openassetio::trait::TraitsDatas{};
```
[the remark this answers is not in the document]
feltech: Added a `runtime_error` here and in `initialize()` in 5f855e39

src/openassetio-core-c/private/managerApi/CManagerInterfaceAdapter.hpp
```diff
@@ -45,6 +47,12 @@ class OPENASSETIO_CORE_C_EXPORT CManagerInterfaceAdapter : ManagerInterface {
   /// @todo Implement C API. Currently a no-op.
   void initialize(const HostSessionPtr& hostSession) override;
 
+  /// Wrap the C suite's `managementPolicy` function.
+  /// @todo Implement C API. Currently a no-op.
```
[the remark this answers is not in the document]
feltech: Added https://github.com/OpenAssetIO/OpenAssetIO/issues/509

python/openassetio/test/manager/apiComplianceSuite.py
feltech: OK added the `set` default argument and suppressed pylint in 48b52532

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -280,6 +282,79 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+   * they are registered for an entity, and returning that data via
+   * resolve.
+   *
+   * @param traitSets The entity @ref trait "traits" to query.
+   *
+   * @param context The calling context.
+   *
+   * @param hostSession The API session.
+   *
+   * @return a `TraitsData` for each  element in `traitSets`.
```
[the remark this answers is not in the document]
feltech: Fixed in ff94adcf

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -150,6 +151,68 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
+   * write data to, prior to calling @ref register. If this trait is
+   * not imbued, then you should take care of writing data yourself
+   * (maybe prompting the user for a location on disk), and then only
+   * call @ref register to create the new entity.
+   *
+   * @param traitSets The entity @ref trait "traits" to query.
+   *
+   * @param context The calling context.
+   *
+   * @return a `TraitsData` for each  element in `traitSets`.
```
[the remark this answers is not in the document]
feltech: Fixed in 82f7354e

foundrytom (approved):

foundrytom:
Thanks @feltech feel free to SQMAYL

Issue #502, opened 2022-07-05
C++ Session intitalization using python manager plugins


foundrytom:
## What

Allow a C++ or Python host to load a python manager plugin and query it's `managementPolicy`.

## Work to be done

- [x] #503
- [x] #504
- [x] #506
- [x] #507
- [x] #510
- [x] #505
- [x] #508

PR #512, opened 2022-07-06
[Core] Rework ManagerInterface initialization

### LINKED ISSUE #503: Rework `ManagerInterface` initialization
## What

Refactor the `setSetting`/`initialize` behaviour to remove some of the current ambiguities. 

## Why

Settings and initialization are deliberately de-coupled to facilitate the following workflow:

1. A host queries all registered managers to obtain their display name (managers are required to be cheap to construct for this query).
2. A user determines which manager to use
3. A host instantiates the manager of choice with its identifier.
4. It queries the manager's default settings with `getSettings` so that the user can customize them before using the manager (eg: to determine which server should be used).
5. It updates the manager's settings based on a users changes.
6. It initializes the manager, where non-trivial setup operations happen.

This split of instantiation of a manager for general queries and initialization is critical as initialization may be non-trivial, and so initializing with default settings just to ask for a display name may perform significant redundant work, especially if many manager plugins are present. It may also be impossible to initialize a manager without certain user-supplied settings and so this needs to be a conscious step.

The current mechanism is functional, but certain usage requirements are less than obvious, specifically:
 
- It is not clear that a host must call `initialize` after calling `setSettings`, even if it has previously been called.
- The manager is required to buffer the settings, so that changes as a result of new values in `setSettings` don't take effect until after `initialize` has been (re)called. 

The revised design makes this clear by forcing a host to call `initialize(settings)` if it wishes to update them, and avoids the need for a manager to juggle not-yet-initialized settings updates.

The existing workflow is still supported:

1. A host queries all registered managers to obtain their display name
2. A user determines which manager to use
3. A host instantiates the manager of choice with its identifier.
4. It queries the manager's default settings with `settings`
5. It initializes the manager with the required settings, where non-trivial setup operations happen.

## ACs:

- Settings mechanism migrated to C++
- Remove `setSettings`.
- Rename `getSettings` -> `settings`.
- Change `initialize(hostSession)` to `initialize(managerSettings, hostSession)`.
- Relevant documentation updates (see/update #478).

feltech:
Closes #503

### DIFF
--- a/python/openassetio/test/manager/apiComplianceSuite.py
+++ b/python/openassetio/test/manager/apiComplianceSuite.py
@@ -82,40 +82,71 @@ class Test_info(FixtureAugmentedTestCase):
     def test_matches_fixture(self):
         self.assertEqual(self._fixtures["info"], self._manager.info())
 
-class Test_setSettings(FixtureAugmentedTestCase):
+
+class Test_settings(FixtureAugmentedTestCase):
+    """
+    Check plugin's implementation of managerApi.ManagerInterface.settings.
+    """
+    def test_when_retrieved_settings_modified_then_newly_queried_settings_unmodified(self):
+        original = self._manager.settings()
+        expected = original.copy()
+        original["update"] = 123
+        self.assertEqual(self._manager.settings(), expected)
+
+
+class Test_initialize(FixtureAugmentedTestCase):
     """
     Check plugin's implementation of
-    managerApi.ManagerInterface.setSettings.
+    managerApi.ManagerInterface.initialize.
     """
     def setUp(self):
-        self.collectRequiredFixture('some_settings_with_valid_keys', skipTestIfMissing=True)
+        self.__original_settings = self._manager.settings()
 
-    def test_valid_settings_succeeds(self):
-        self._manager.setSettings(self.some_settings_with_valid_keys)
+    def tearDown(self):
+        self._manager.initialize(self.__original_settings)
 
-    def test_unknown_settings_keys_raise_KeyError(self):
-        unknown_settings = self.requireFixture("some_settings_with_invalid_keys")
+    def test_when_settings_are_empty_then_all_settings_unchanged(self):
+        expected = self._manager.settings()
+
+        self._manager.initialize({})
+
+        self.assertEqual(self._manager.settings(), expected)
+
+    def test_when_settings_have_invalid_keys_then_raises_KeyError(self):
+        invalid_settings = self.requireFixture(
+            "some_settings_with_new_values_and_invalid_keys", skipTestIfMissing=True)
+
         with self.assertRaises(KeyError):
-            self._manager.setSettings(unknown_settings)
+            self._manager.initialize(invalid_settings)
 
+    def test_when_settings_have_invalid_keys_then_all_settings_unchanged(self):
+        expected = self._manager.settings()
+        invalid_settings = self.requireFixture(
+            "some_settings_with_new_values_and_invalid_keys", skipTestIfMissing=True)
 
-class Test_getSettings(FixtureAugmentedTestCase):
-    """
-    Check plugin's implementation of managerApi.ManagerInterface.getSettings.
-    """
+        try:
+            self._manager.initialize(invalid_settings)
+        except Exception:  # pylint: disable=broad-except
+            pass
 
-    def test_when_set_then_get_returns_updated_settings(self):
-        updated = self.requireFixture("some_new_settings_with_all_keys", skipTestIfMissing=True)
-        self._manager.setSettings(updated)
-        self.assertEqual(self._manager.getSettings(), updated)
+        self.assertEqual(self._manager.settings(), expected)
 
-    def test_when_set_with_subset_then_other_settings_unchanged(self):
+    def test_when_settings_have_all_keys_then_all_settings_updated(self):
+        updated = self.requireFixture("some_settings_with_all_keys", skipTestIfMissing=True)
+
+        self._manager.initialize(updated)
+
+        self.assertEqual(self._manager.settings(), updated)
+
+    def test_when_settings_have_subset_of_keys_then_other_settings_unchanged(self):
         partial = self.requireFixture(
-                "some_new_settings_with_a_subset_of_keys", skipTestIfMissing=True)
-        settings = self._manager.getSettings()
-        self._manager.setSettings(partial)
-        settings.update(partial)
-        self.assertEqual(self._manager.getSettings(), settings)
+            "some_settings_with_a_subset_of_keys", skipTestIfMissing=True)
+        expected = self._manager.settings()
+        expected.update(partial)
+
+        self._manager.initialize(partial)
+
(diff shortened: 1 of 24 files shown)


python/openassetio/test/manager/apiComplianceSuite.py
```diff
@@ -82,47 +82,65 @@ def test_is_correct_type(self):
...
+    def test_when_settings_expanded_then_manager_settings_updated(self):
         updated = self.requireFixture("some_new_settings_with_all_keys", skipTestIfMissing=True)
-        self._manager.setSettings(updated)
+
+        self._manager.initialize(updated)
+
         self.assertEqual(self._manager.settings(), updated)
 
-    def test_when_set_with_subset_then_other_settings_unchanged(self):
+    def test_when_subset_of_settings_modified_then_other_settings_unchanged(self):
```
foundrytom, line 134: [consider] How about `test_when_settings_have_subset_keys_then_other_settings_unchanged` to match the other names?

python/openassetio/test/manager/apiComplianceSuite.py
```diff
@@ -82,47 +82,65 @@ def test_is_correct_type(self):
...
         partial = self.requireFixture(
-                "some_new_settings_with_a_subset_of_keys", skipTestIfMissing=True)
+            "some_new_settings_with_a_subset_of_keys", skipTestIfMissing=True)
```
foundrytom, line 136: [consider] `"some_settings_with_a_subset_of_keys"` for consistency?

tests/python/openassetio/test/manager/resources/plugins/StubManager.py
```diff
@@ -50,9 +50,11 @@ def info(self):
         return self.__info
 
     def settings(self, hostSession):
+        # pylint: disable=unused-argument
         return self.__settings
 
     def initialize(self, managerSettings, hostSession):
+        # pylint: disable=unused-argument
```
foundrytom, line 57: [fix] Is this needed? We look to use both args now?

foundrytom (changes requested):
Thanks @feltech just a few minor tweaks

python/openassetio/test/manager/apiComplianceSuite.py
```diff
@@ -82,47 +82,65 @@ def test_is_correct_type(self):
...
+
+class Test_initialize(FixtureAugmentedTestCase):
     """
     Check plugin's implementation of
-    managerApi.ManagerInterface.setSettings.
+    managerApi.ManagerInterface.initialize.
     """
     def setUp(self):
-        self.collectRequiredFixture('some_settings_with_valid_keys', skipTestIfMissing=True)
+        super(Test_initialize, self).setUp()
```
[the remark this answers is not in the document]
foundrytom: Do we really need them? It says the default implementation is a no-op, and were not including them in other `setUp`/`tearDown` methods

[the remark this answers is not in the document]
feltech: I added the `super` since it's not directly derived from `unittest.TestCase`, so the base class _could_ also override.  It doesn't though, so I can remove it.  I guess its a YAGNI situation at the moment.

python/openassetio/test/manager/apiComplianceSuite.py
```diff
@@ -82,47 +82,65 @@ def test_is_correct_type(self):
...
 
-class Test_settings(FixtureAugmentedTestCase):
-    """
-    Check plugin's implementation of managerApi.ManagerInterface.settings.
-    """
+        with self.assertRaises(KeyError):
+            self._manager.initialize(unknown_settings)
 
-    def test_when_set_then_get_returns_updated_settings(self):
+    def test_when_settings_expanded_then_manager_settings_updated(self):
```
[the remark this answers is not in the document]
feltech: Good call, I made a silly assumption that the initial state was a reduced set of keys.

tests/python/openassetio/test/manager/resources/plugins/StubManager.py
feltech: :confused: good spot, wonder why I put that there.

python/openassetio/test/manager/apiComplianceSuite.py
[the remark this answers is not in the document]
feltech: Updated to your suggestion in 08b748cd

python/openassetio/test/manager/apiComplianceSuite.py
feltech: Much better, updated to your suggestion in 28b67cfa

python/openassetio/test/manager/apiComplianceSuite.py
feltech: Updated a couple of fixture names for consistency in 262730ae

tests/python/openassetio/test/manager/resources/plugins/StubManager.py
feltech: Removed in 4104ea14

python/openassetio/test/manager/apiComplianceSuite.py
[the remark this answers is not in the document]
feltech: Also removed from `tearDown` in 1f16027e :roll_eyes:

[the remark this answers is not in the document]
foundrytom: Ah yes, sorry I'd forgotten it wasn't just a direct derivation - good point. As long as you're happy with where were at now then 👍

foundrytom (approved):
Thanks for those fixes @feltech, could we squash ready for merge please? Can we also make sure that we include the "why" from #502 in the commit message, as I think this one will def be useful for archaeology at some point.

foundrytom (approved):
Changes look good David, do you want to squash ready for merge?

feltech:
> Thanks for those fixes @feltech, could we squash ready for merge please? Can we also make sure that we include the "why" from #503 in the commit message, as I think this one will def be useful for archaeology at some point.

I've included the "Why" in the final commit.  I thought it worth not squashing everything into a single commit, but happy to do so if you prefer.

foundrytom:
> > Thanks for those fixes @feltech, could we squash ready for merge please? Can we also make sure that we include the "why" from #503 in the commit message, as I think this one will def be useful for archaeology at some point.
> 
> I've included the "Why" in the final commit. I thought it worth not squashing everything into a single commit, but happy to do so if you prefer.

Cool, thanks. I guess the Q is, does the rename of `settings` make sense on its own?

feltech:
> > > Thanks for those fixes @feltech, could we squash ready for merge please? Can we also make sure that we include the "why" from #503 in the commit message, as I think this one will def be useful for archaeology at some point.
> > 
> > 
> > I've included the "Why" in the final commit. I thought it worth not squashing everything into a single commit, but happy to do so if you prefer.
> 
> Cool, thanks. I guess the Q is, does the rename of `settings` make sense on its own?

Taken the easy way out, squashed down into a single commit.

PR #513, opened 2022-07-07
[Core] Minimal C++ logging interface

### LINKED ISSUE #504: Minimal C++ logging interface.
## What

Port a skeleton logging interface to C++. This does not necessarily need full functionality, but enough to support the migration of other classes that require a logger. If the current design is adequate and simple to port, it should be migrated as-is.

## Why

Without the ability to inject a ptr to a specific logging implementation through C++, Python wrappers are required to avoid breaking existing functionality. Even an empty base class would avoid the need to un-`final` classes and wrap for python. 

## ACs

- Classes that trade in logger instances can be made `final` in C++ whilst maintaining logging functionality in python.

## Out-of-scope

- C++ logging ability
- Migrating existing python logger implementations.
- `@debugApi` functionality.

feltech:
Closes #504.

Migrate the `LoggerInterface` to C++ with Python bindings., removing `LoggerInterface.progress()` along the way.

This will then enable future work to fully migrate `HostSession` and `ManagerFactoryInterface` C++ without requiring Python wrappers.

Testing is a little minimal. We currently have no (Python->)C++->Python `log()` calls.  Once `HostSession` is migrated to C++, its Python tests will implicitly include a round-trip of `log()` through C++.

### DIFF
--- a/python/openassetio/log.py
+++ b/python/openassetio/log.py
@@ -18,68 +18,13 @@
 Provides the core classes that facilitate message and progress logging.
 """
 
-import abc
 import os
 import sys
 
+from . import _openassetio  # pylint: disable=no-name-in-module
 
-class LoggerInterface(object, metaclass=abc.ABCMeta):
-    """
-    An abstract base class that defines the receiving interface for log
-    messages generated a @ref manager or the API middleware.
-    """
 
-    ##
-    # @name Log Severity
-    # @{
-
-    kDebugApi = 6
-    kDebug = 5
-    kInfo = 4
-    kProgress = 3
-    kWarning = 2
-    kError = 1
-    kCritical = 0
-
-    ## A mapping of severity levels to readable labels
-    kSeverityNames = ['critical', 'error', 'warning', 'progress', 'info', 'debug', 'debugApi']
-
-    ## @}
-
-    @abc.abstractmethod
-    def log(self, message, severity):
-        """
-        Logs a message to the user.
-
-        This method must be implemented to present the supplied message
-        to the user in an appropriate fashion.
-
-        @param message str, The message string to be logged.
-
-        @param severity int, One of the severity constants defined in
-        @ref openassetio.log
-        """
-        raise NotImplementedError
-
-    def progress(self, decimalProgress, message=""):
-        """
-        Logs the progress of a task.
-
-        This method should be overridden if you wish to customize the
-        presentation of progress messages beyond the default
-        implementation that simply logs them as any other message.
-
-        @param decimalProgress float, Normalised progress between 0 and
-        1, if set to a value less than 0 it should be considered
-        cancelled, if greater than one, complete.
-
-        @param message str, A message to display with the progress. If
-        None is supplied, assume that there is no message and any
-        previous message may remain. Set to an empty string if it is
-        desired to always clear any previous message.
-        """
-        msg = "%3d%% %s" % (int(100 * decimalProgress), message if message is not None else "")
-        self.log(msg, self.kProgress)
+LoggerInterface = _openassetio.LoggerInterface
 
 
 class SeverityFilter(LoggerInterface):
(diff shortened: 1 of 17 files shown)


src/openassetio-core/include/openassetio/LoggerInterface.hpp
```diff
@@ -0,0 +1,44 @@
...
+class OPENASSETIO_CORE_EXPORT LoggerInterface {
+ public:
+  /**
+   * @name Log Severity
+   * @{
+   */
+  enum Severity : EnumIdx { kCritical = 0, kError, kWarning, kProgress, kInfo, kDebug, kDebugApi };
+
+  static constexpr EnumNames<7> kSeverityNames{"critical", "error", "warning", "progress",
+                                               "info",     "debug", "debugApi"};
```
foundrytom, line 25: [question] Was the extra white space intentional?

python/openassetio/log.py
```diff
@@ -93,14 +93,20 @@ def getSeverity(self):
     # LoggerInterface methods
 
     def log(self, message, severity):
-
+        """
+        Log only if severity is less than or equal to maximum allowed
```
foundrytom, line 97: [fix] Here (and progress, but that's going away...). The concept of min/max for severity is always a head-mess. I think we have been referring to the "logical level" ie "critical is high, debug is low", rather than the enum value. The docs at L33 suggest messages below the specified severity are dropped, so I think this needs to be "Log only if severity is greater or equal to the loggers configured severity level" (or similar).

python/openassetio/log.py
```diff
@@ -147,19 +93,15 @@ def getSeverity(self):
...
```
foundrytom, line 97:
```suggestion
        Log only if severity is greater or equal to the logger's configured
        severity level.
```

python/openassetio/log.py
```diff
@@ -192,7 +135,9 @@ def __init__(self, colorOutput=True, forceDefaultStreams=False):
                 self.__stderr = sys.__stderr__
 
     def log(self, message, severity):
-
+        """
+        Log to stderr for severity less than `kInfo`, stdout otherwise.
```
foundrytom, line 139: [fix]
```suggestion
        Log to stderr for severity greater than `kInfo`, stdout otherwise.
```

foundrytom (changes requested):
👍  Just a few small docs tweaks.

python/openassetio/log.py
foundrytom: If we want to change the enum around that's totally fine too!

src/openassetio-core/include/openassetio/LoggerInterface.hpp
feltech: That's `clang-format` aligning things.  I don't particularly like it either.

python/openassetio/log.py
feltech: Updated the docstrings in e0862b82.  

I also added tests for the severity names in c536ceb4, in preparation for reversing the severity enum order.  However, it occurred to me that we're more likely to add lower logical severities than higher (e.g. a `kTrace` severity), so having lower logical severities with higher enum values allows us to easily add more.  

It is a bit mind-bending though. I'm happy to reverse the enum values if you think its worthwhile.

python/openassetio/log.py
feltech: Updated docstring in e0862b82

python/openassetio/log.py
feltech: Updated docstring in e0862b82

python/openassetio/log.py
foundrytom: Can we [plan] an Issue for re-ordering? And we can pick it up once this slice is done. Good idea on the tests 👍

foundrytom (approved):
Thanks @feltech Feel free to squash and merge as you see fit in amongst all the other PRs.

python/openassetio/log.py
feltech: Re. leaving some room, a good idea, but kinda breaks the `kSeverityNames` lookup, so we'd have to rethink that.

feltech: Added https://github.com/OpenAssetIO/OpenAssetIO/issues/516 to re-order severity enums.

Issue #395, opened 2022-05-23
[Test] Improve speed of executing tests from CMake


feltech:
## What

Improve the speed of execution of pure Python tests during TDD loop.

## Why

Currently, executing the `openassetio-python` CMake target to run the `pytest` tests takes more than 10 secs even for a warm start (i.e. where everything is already installed and no changes have been made).  This noticeably impacts dev efficiency during the TDD loop.

## Notes

~We found that the `pip` flag `--no-build-isolation` can drastically improve our `setup.py` "build" times, at the expense of polluting the execution environment with the otherwise-unnececssary `wheel` package.  It's also discouraged as not the modern way to build packages. But perhaps it makes sense for our particular use case.~ , done, see #515 

~A development install via `--editable` would be great, but this has issues with the compiled `_openassetio.so` component.~ Not possible just yet, out of scope for this specific task.

~We could have a "fast build" CMake/CTest target/option that skips steps such as dependency installs, or even simply `rsync`s the Python modules.~ One for the future.

Ideally we would avoid polluting the source tree with `.egg-info`s.

Having multiple CTest targets, whose underlying CMake targets all have the same dependencies, wastes CPU time and precludes parallel execution.  

So we should remove all `add_dependencies` for the low-level CMake targets, and instead wrap these targets in `add_test`s, then use the `FIXTURES_SETUP` CTest property to mark appropriate tests as fixtures (e.g. creating a Python environment). 

This then allows us to strip redundant work and parallelise test execution. 

We probably want to continue to have build targets that create an installed environment for us, in which case we can add separate wrapper CMake targets with `add_dependencies` between them.

## ACs

- CMake/ctest targets updated to facilitate parallel execution of tests/build.
- Notes in `CODING_STANDARDS.md` (if appropriate), noting how new targets should be configured to support the above.
- Support for building the environment (ie a venv with `openassetio` and `openassetio-codegen` installed) independently of running tests (exposed as a build target)
- Set any flags in CI/et. al. to use parallel/multi-threaded builds/tests where possible.

feltech:
Testing #515 reveals that, assuming the build isolation overhead is gone, approx half the remaining time is due to checking dependencies and half is actually running the tests.

PR #515, opened 2022-07-07
[Build] Reduce Python install time for tests

feltech:
Part of #395. We run Python unit tests against a full install of OpenAssetIO, for better dogfooding. The time taken by this install
significantly slows the TDD feedback loop. Any time we can shave off this will improve dev efficiency.

So reduce the time taken to "build" our pure Python package by avoiding use of temporary build directories. This also removes the need to download/upgrade `setuptools` and `pip`, shaving a little more time. However, it does necessitate installing `wheel` into the venv.

### DIFF
--- a/src/openassetio-python/CMakeLists.txt
+++ b/src/openassetio-python/CMakeLists.txt
@@ -117,7 +117,7 @@ if (Python_Interpreter_FOUND)
         ${Python_EXECUTABLE} -m pip download
         --retries 0 --timeout ${OPENASSETIO_PYTHON_PIP_TIMEOUT} --find-links ${python_deps_dir}
         --destination-directory ${python_deps_dir}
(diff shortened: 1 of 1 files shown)


foundrytom (approved):
I just noticed, it does actually seem to create egg-info's in the source tree now :( We do ignore those though... probably worth it for now, we can always re-visit this later when we sort out packaging.

foundrytom:
@feltech Sorry meant to say, FFTSAMAYL if you're happy with 🥚🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚 🥚

Issue #510, opened 2022-07-06
Remove `Session`


foundrytom:
## What

Remove `Session` class, as since #507, it is no longer needed

## ACs

 - Session removed (updating existing usage to `ManagerFactory`)
 - Remove `_session` in `FixtureAugmentedTestCase`
 - Examples/Documentation updated

PR #522, opened 2022-07-12
[Core] Add ManagerFactory

### LINKED ISSUE #507: Add `ManagerFactory`
## What

Migrate business logic from `Session` into a simple `ManagerFactory`.

## Why

`Session` should be migrated to a host-convenience library. It is simply a state management helper designed to aid monolithic widget based cpp apps manage state. It is certainly not an interaction pattern required by the API, and complicates a number of simple host implementation scenarios.

## ACs

- Python/C++ support
- A static method that can be used to correctly instantiate a `Manager` wrapper given an identifier, host interface, factory and logger.
- A utility that can be used to inspect available managers based on the supplied factory:
  - `identifiers`
  - `availableManagers`
     - Map of `identifier`, `displayName`, `info`, `settings`, structs by `identifier`.
  - `create(identifier)`,

## Out of scope

- Updating docs that mention the use of `Session`

feltech:
Closes #507.

### DIFF
--- a/python/openassetio/hostApi/ManagerFactory.py
+++ b/python/openassetio/hostApi/ManagerFactory.py
@@ -1,0 +1,61 @@
+#
+#   Copyright 2022 The Foundry Visionmongers Ltd
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+"""
+@namespace openassetio.hostApi.Manager
+A single-class module, providing the ManagerFactory class.
+
+@todo This module is a temporary measure until we reach parity between
+C++ and Python Manager and HostSession classes.
+"""
+
+from .. import _openassetio  # pylint: disable=no-name-in-module
+
+from .Manager import Manager
+from ..managerApi.HostSession import HostSession
+
+
+__all__ = ['ManagerFactory']
+
+
+class ManagerFactory(_openassetio.hostApi.ManagerFactory):
+    """
+    @see @fqref{hostApi.ManagerFactory} "hostApi.ManagerFactory".
+    """
+    def __init__(self, hostInterface, managerInterfaceFactory, logger):
+        _openassetio.hostApi.ManagerFactory.__init__(
+            self, hostInterface, managerInterfaceFactory, logger)
+        self.__managerInterfaceFactory = managerInterfaceFactory
+        self.__hostInterface = hostInterface
+        self.__logger = logger
+
+    def createManager(self, identifier):
+        """
+        @see @fqref{hostApi.ManagerFactory.createManager}
+        "ManagerFactory.createManager".
+        """
+        return Manager(
+            self.__managerInterfaceFactory.instantiate(identifier),
+            HostSession(_openassetio.managerApi.Host(self.__hostInterface), self.__logger))
+
+    @staticmethod
+    def createManagerForInterface(identifier, hostInterface, managerInterfaceFactory, logger):
+        """
+        @see @fqref{hostApi.ManagerFactory.createManagerForInterface}
+        "ManagerFactory.createManagerForInterface".
+        """
+        return Manager(
+            managerInterfaceFactory.instantiate(identifier),
+            HostSession(_openassetio.managerApi.Host(hostInterface), logger))
--- a/src/openassetio-core/hostApi/ManagerFactory.cpp
+++ b/src/openassetio-core/hostApi/ManagerFactory.cpp
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 The Foundry Visionmongers Ltd
+#include <openassetio/LoggerInterface.hpp>
+#include <openassetio/hostApi/HostInterface.hpp>
+#include <openassetio/hostApi/Manager.hpp>
+#include <openassetio/hostApi/ManagerFactory.hpp>
+#include <openassetio/hostApi/ManagerInterfaceFactoryInterface.hpp>
+#include <openassetio/managerApi/Host.hpp>
+#include <openassetio/managerApi/HostSession.hpp>
+#include <openassetio/managerApi/ManagerInterface.hpp>
+#include <openassetio/typedefs.hpp>
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace hostApi {
+
+ManagerFactoryPtr ManagerFactory::make(HostInterfacePtr hostInterface,
+                                       ManagerInterfaceFactoryInterfacePtr managerInterfaceFactory,
+                                       LoggerInterfacePtr logger) {
+  return openassetio::hostApi::ManagerFactoryPtr{new ManagerFactory{
+      std::move(hostInterface), std::move(managerInterfaceFactory), std::move(logger)}};
+}
+
+ManagerFactory::ManagerFactory(HostInterfacePtr hostInterface,
+                               ManagerInterfaceFactoryInterfacePtr managerInterfaceFactory,
+                               LoggerInterfacePtr logger)
+    : hostInterface_{std::move(hostInterface)},
+      managerInterfaceFactory_{std::move(managerInterfaceFactory)},
+      logger_{std::move(logger)} {}
+
+Identifiers ManagerFactory::identifiers() const { return managerInterfaceFactory_->identifiers(); }
+
+ManagerFactory::ManagerDetails ManagerFactory::availableManagers() const {
+  const Identifiers& ids = identifiers();
+  if (ids.empty()) {
+    return {};
+  }
+
+  // `HostSession` required for `settings()` calls.
+  const managerApi::HostSessionPtr hostSession =
+      managerApi::HostSession::make(managerApi::Host::make(hostInterface_));
+
+  ManagerDetails managerDetails;
+
+  for (const Identifier& identifier : ids) {
+    const managerApi::ManagerInterfacePtr managerInterface =
+        managerInterfaceFactory_->instantiate(identifier);
+
+    managerDetails.insert({identifier,
+                           {managerInterface->identifier(), managerInterface->displayName(),
+                            managerInterface->info()}});
+  }
+  return managerDetails;
+}
+
+ManagerPtr ManagerFactory::createManager(const Identifier& identifier) const {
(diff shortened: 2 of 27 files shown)


python/openassetio/hostApi/ManagerFactory.py
```diff
@@ -0,0 +1,61 @@
...
+from .. import _openassetio  # pylint: disable=no-name-in-module
+
+from .Manager import Manager
+from ..managerApi.HostSession import HostSession
+
+
+__all__ = ['ManagerFactory']
+
+
+class ManagerFactory(_openassetio.hostApi.ManagerFactory):
```
feltech, line 33: [question] The need for this class is rather unfortunate. We're missing functionality in the C++ bound `HostSession` and `Manager` classes, so must override the `create*` methods to use the Python classes instead.

An alternative possiblity, which would keep the Python sources cleaner, may be to override `createManager` and `createManagerForInterface` in pybind, using `py::module::import()` to get the Python `Manager` (and `HostSession`) class and instantiate that instead.  

However for `createMananger` that would entail giving pybind access to `ManagerFactory`'s private members to gather the dependencies required for the `Manager` constructor.  There are probably other gotchas I haven't thought of too.

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
```diff
@@ -0,0 +1,74 @@
...
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace hostApi {
+
+OPENASSETIO_DECLARE_PTR(ManagerFactory)
+
+/**
+ * The ManagerFactory is the primary mechanism for querying for
+ * available @ref manager "managers" and constructing a
+ * @fqref{hostApi.Manager} "Manager", using the supplied manager
+ * interface factory.
```
foundrytom, line 29: [fix] Could we add a comment along the lines of "Hosts should never attempt to directly construct a `Manager` class or interact with the implementation factory directly" - just to really spell it out...

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
```diff
@@ -28,6 +30,54 @@ OPENASSETIO_DECLARE_PTR(ManagerFactory)
 class OPENASSETIO_CORE_EXPORT ManagerFactory final {
  public:
   using Identifiers = std::vector<Str>;
+
+  /**
+   * Simple struct containing the default configuration details of a
+   * potential @ref manager implementation.
+   */
+  struct ManagerDetail {
```
foundrytom, line 42: [question] We're not fussed about the MSVC struct issues here are we - as we only use it by value?

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
```diff
@@ -0,0 +1,194 @@
...
+   * @see @fqref{hostApi.Manager.identifier} "Manager.identifier".
+   */
+  [[nodiscard]] Identifiers identifiers() const;
+
+  /**
+   * Get the details for each available @ref manager as a map of
+   * manager identifier to manager details.
+   *
+   * This provides the default settings that can be taken and mutated
+   * before being used in the creation of a @fqref{hostApi.Manager}
```
foundrytom, line 121: [fix]
```suggestion
   * before being used in the initialization of a @fqref{hostApi.Manager}
```

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
```diff
@@ -29,7 +30,10 @@ OPENASSETIO_DECLARE_PTR(ManagerFactory)
  */
 class OPENASSETIO_CORE_EXPORT ManagerFactory final {
  public:
-  using Identifiers = std::vector<Str>;
+  /// Unique identifier of a manager.
+  using Identifier = Str;
```
foundrytom, line 34: [consider] As this is common to both interfaces, and the wrappers, for both host and manager, would lit make sense to add this to `typedefs.hpp` or similar?

tests/python/openassetio/hostApi/test_managerfactory.py
```diff
@@ -134,3 +135,53 @@ def assert_expected_host(hostSession):
...
+
+        # action
+
+        manager = a_manager_factory.createManager(expected_identifier)
+
+        # confirm
+
+        # TODO(DF): This can be changed to openassetio.hostApi.Manager
+        #  once C++ migration of Manager is complete.
+        assert isinstance(manager, _openassetio.hostApi.Manager)
```
foundrytom, line 177: [consider] Moving this to its own test in the spirit of "test cases describe expected behaviour"

src/openassetio-core/hostApi/ManagerFactory.cpp
```diff
@@ -56,8 +56,15 @@ ManagerFactory::ManagerDetails ManagerFactory::availableManagers() const {
...
 ManagerPtr ManagerFactory::createManager(const Identifier& identifier) const {
-  return Manager::make(managerInterfaceFactory_->instantiate(identifier),
-                       managerApi::HostSession::make(managerApi::Host::make(hostInterface_)));
+  return createManagerForInterface(hostInterface_, managerInterfaceFactory_, identifier);
+}
+
+ManagerPtr ManagerFactory::createManagerForInterface(
+    const HostInterfacePtr& hostInterface,
+    const ManagerInterfaceFactoryInterfacePtr& managerInterfaceFactory,
+    const ManagerFactory::Identifier& identifier) {
```
foundrytom, line 66: [fix] In our "order arguments as per the name" convention, would it make sense to order the arguments as:

`createManagerForInterface(identifier, hostInterface, factory, logger)`

It also reenforce that this is an alternate version of `createManager` where the other two arguments are held by the factory, and perhaps reduces that ambiguity around the two interfaces?

python/openassetio/hostApi/ManagerFactory.py
foundrytom: I think the solution you have is eloquent and spot on for now, its also very easy to remove.

foundrytom (changes requested):
Great job on the workaround for the C++/Python disparity @feltech - just a few tiny tweaks I think.

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
feltech: The `struct` vs. `class` issue only surfaces if we forward declare it, which I'm not sure is even possible for child classes. Do you think we should make it `class` regardless, for consistency?

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
```diff
@@ -129,6 +133,22 @@ class OPENASSETIO_CORE_EXPORT ManagerFactory final {
...
+   * The instance returned should then be used for all interaction with
+   * the manager.
+   *
+   * @param identifier Unique manager identifier.
+   *
+   * @return Newly instantiated manager.
+   */
+  // TODO(DF): This could probably return a ManagerConstPtr, but
+  //  currently not all Manager member functions are const, even though
+  //  they probably could be. See also #518.
```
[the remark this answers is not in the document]
feltech: Removed the TODO comment in f42bb482

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
feltech: Added your suggestion in 4a07d906

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
```diff
@@ -59,6 +109,26 @@ class OPENASSETIO_CORE_EXPORT ManagerFactory final {
...
+   * manager identifier to manager details.
+   *
+   * This provides the default settings that can be taken and mutated
+   * before being used in the creation of a @fqref{hostApi.Manager}
+   * "Manager".
+   *
+   * Additional metadata is also included that is useful for deciding
+   * between manager implementations to use, beyond simply their opaque
+   * identifiers. For example, this can be presented in a manager
+   * picker UI widget.
```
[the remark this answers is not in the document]
feltech: Reworded in 5917e49d

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
feltech: Added a global `Identifier` alias and used throughout the codebase in 88df07b1

tests/python/openassetio/hostApi/test_managerfactory.py
feltech: Split instance type check and config check in 227da8cb

src/openassetio-core/hostApi/ManagerFactory.cpp
feltech: Good point, updated in a961620b

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
foundrytom: [question] Are we happy to keep using `StringView` in C for these?

foundrytom (changes requested):
Looking good @feltech, just a few doc things. I think `RELEASE_NOTES.md` might need updating too?

foundrytom:
[question] What are you thinking for the final commit series for this - most looked to have squash/fixup - be nice to keep things like the addition of `Identifiers` typedefs in it's own commit if possible?

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
feltech: Well grepped. Swapped more places to use `Identifier` in 59aa6e86

I think `StringView` is fine for now, I kinda see `Identifier` as more of a documentation thing. If we did want to swap `Identifier` for some other type, then chances are it would still need to be convertible to a `char` array for C API purposes.

feltech:
> I think `RELEASE_NOTES.md` might need updating too?

Whoops!  Added release notes in 28b1ec54 (note that there's also mention of `ManagerFactory` in release notes for [#510](https://github.com/OpenAssetIO/OpenAssetIO/issues/510))

feltech:
> @feltech I think we need to add a check to
> 
> https://github.com/OpenAssetIO/OpenAssetIO/blob/0ade99b1167057ca2e4e3e723d8094d70c9f4aab/tests/python/openassetio/test_imports.py#L75

Added in 47cda3d9

foundrytom (approved):
@feltech - LGTM - feel free to rebase-rejig and merge at your leisure.

feltech:
As discussed offline re. https://github.com/OpenAssetIO/OpenAssetIO/pull/525#discussion_r926871004 - removed `settings` from `availableManagers` output in 66fcf945

foundrytom:
> As discussed offline re. [#525 (comment)](https://github.com/OpenAssetIO/OpenAssetIO/pull/525#discussion_r926871004) - removed `settings` from `availableManagers` output in [66fcf94](https://github.com/OpenAssetIO/OpenAssetIO/commit/66fcf945ff1ab1fe7511d6c58f296e67223d3023)

Thanks @feltech LGTM

Issue #524, opened 2022-07-19
Python instances held as `shared_ptr` in C++ can still be destroyed


feltech:
Python instances can be passed to and held by a C++ instance as a C++ base class pointer. The Python instance can then be destroyed (i.e. refcount can reach zero), leaving the base class pointer "dangling", resulting in e.g. `RuntimeError Tried to call pure virtual function` from the pybind 'trampoline'.

This is a known issue in Pybind. See https://github.com/pybind/pybind11/issues/1333.

PR #525, opened 2022-07-19
[Core] Remove `Session` class

### LINKED ISSUE #510: Remove `Session`

feltech:
Closes #510.

Session is simply a state management helper designed to aid monolithic widget based C++ apps manage state. It is certainly not an interaction pattern required by the API, and complicates a number of simple host implementation scenarios.

So remove all usages of `Session` from code and documentation in favour of the new, simpler, `ManagerFactory`, which is now responsible for querying available manager implementations and creation of `Manager` wrappers.


foundrytom (commented):
Nice work @feltech, I made a few tweaks to the docs as fixups, what do you think?

doc/src/Examples.dox
```diff
@@ -77,20 +77,26 @@
...
+ * # A manager's current (or in this case default) settings can be
+ * # queried if needed:
+ * settings = manager.settings()
+ * # ...and updated with new values as desired.
+ * settings["server"] = "my.server.com"
+ *
+ * # Finally, we can initialize the manager with the desired settings,
+ * # preparing it for use. Note that this may include non-trivial
+ * # amounts of work. Settings updates are sparse, so if you don't have
+ * # any custom settings, you can pass an empty dictionary here.
```
foundrytom, line 99: [question] Should we make it so the `Manager` wrapper makes settings optional?

python/openassetio/test/manager/_implementation.py
```diff
@@ -27,20 +27,25 @@
...
+
 
 def createHarness(managerIdentifier, settings=None):
     """
     Create the test harness used begin test case execution.
     @private
     """
+    if settings is None:
+        settings = {}
+    # global hostInterface
```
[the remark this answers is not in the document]
feltech: :roll_eyes: My bad, this is leftover debugging from when I originally got stuck on the problem that spawned #524.  I discovered the problem went away when I made `hostInterface` global, so it was not destroyed by Python.  Clearly I forgot to go back in and clean up after myself.

doc/src/Examples.dox
```diff
@@ -48,47 +46,58 @@
...
+ *
+ * # We now have an instance of the requested manager, but it is not
+ * # quite ready for use yet. The manager returned by the
+ * # ManagerFactory needs to be initialized before it can be used to
+ * # query or publish assets. Setup is split into two stages to allow
+ * # adjustments to its settings to be made prior to use if required.
+ *
+ * # A manager's current (or in this case default) settings can be
+ * # queried if needed:
+ * settings = manager.settings()
```
feltech, line 92: [question] I like your rewordings :+1: . 

I guess it makes more sense to get the settings from the manager instance once created, rather than extracting from the `availableManagers`.  

But then that begs the question, is there really any need for `settings` (for that matter, `info`) in the `availableManagers` output?

doc/src/Examples.dox
feltech: :thinking: Forcing the user to pass settings has a couple of advantages
* Passing an empty dict explicitly states that they want the default settings and must have considered the consequences.
* Throwing an error if they don't pass any settings acts as a form of documentation, in case the user forgets when and how to set settings.
* Other than the first `initialize` call, it doesn't make sense to call `initialize` without (a change of) settings(?).
* If we start strict we can make it optional later, but not vice versa.

Having said all that, I don't have a strong opinion if we want to make it optional.  Especially if it turns out that default settings are the most common case.

python/openassetio/test/manager/_implementation.py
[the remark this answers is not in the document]
feltech: Removed in 34710b25

doc/src/Examples.dox
foundrytom: Im sold :) stick as we are!

doc/src/Examples.dox
foundrytom: Good Q - looks like we added it in #522 - it wasn't in the current python implementation. Shall we bin if off?

feltech: Removed from Examples.dox in this PR in e9cefa2a and removed from the codebase in #522 (https://github.com/OpenAssetIO/OpenAssetIO/pull/522#issuecomment-1192444052)

foundrytom: @feltech when you do the squashy squash, we'll need to remove the inclusion of the settings in the output in the example.

foundrytom: Haha, sorry, great minds think alike - sorry I should have known you'd be on it already.

foundrytom (approved):

PR #534, opened 2022-07-25
[Tests] Add missing Python import checks

### LINKED ISSUE #526: Check we have cases in `test_imports.py` for recently added modules
## What

Ensure `test_imports.py` covers all public python.

## Why

We forgot to add these tests for some recent additions.

foundrytom:
Accidentally omitted during recent additions/changes.

Closes #526

### DIFF
--- a/tests/python/openassetio/test_imports.py
+++ b/tests/python/openassetio/test_imports.py
@@ -18,6 +18,10 @@
 These test ensure that all modules are importable. This can help catch
 dependencies between packages due to the hoisting required to hide
 duplicate namespaces to match the future C++ implementation appearance.
+
+Cases should be added for any of the following:
+ - Discreet Python modules.
+ - C++ implementations hoisted in an __init__.py file.
 """
 
 import pytest
@@ -38,7 +42,7 @@ def always_unload_openassetio_modules(unload_openassetio_modules): # pylint: dis
 
 class Test_package_imports:
 
-    def test_importing_openassetui_succeeds(self):
+    def test_importing_openassetio_succeeds(self):
         import openassetio
 
     def test_importing_constants_succeeds(self):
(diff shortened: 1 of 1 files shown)


tests/python/openassetio/test_imports.py
```diff
@@ -115,3 +121,18 @@ def test_importing_PluginSystemManagerInterfaceFactory_succeeds(self):
 
     def test_importing_PluginSystemPlugin_succeeds(self):
         from openassetio.pluginSystem import PluginSystemPlugin
+
+
+class Test_test_imports:
```
feltech, line 126: [question] Looks all good, I'm just wondering what the rule is for what should be included in this test module.  

At first I thought the intention was to test classes have been hoisted successfully in `__init__.py`s. But the tests here and below break that rule.

feltech (approved):

tests/python/openassetio/test_imports.py
foundrytom: Originally, the main intent of these tests was to catch cyclic imports etc... so I guess it should be that any individually importable module is importable, but it also doubles up testing C++ hoists now too. So I guess its:

- Any individual python defined module
- Any C++ hoisted module.

foundrytom: Can you think of anything else? Ill add a note to such an effect at the top

feltech: In which case we're missing `test.manager.harness`/`apiComplianceSuite`?  Or is it that these are somewhat "private"?

foundrytom: Added notes on coverage [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/534/commits/52d92df5eed34eb32374273fc9e21d5247213b8f), hows that? Added missing tests [https://github.com/OpenAssetIO/OpenAssetIO/pull/534/commits/be3a4e649a7c969296ec27b1515a8187e49e87d0] - somehow I'd totally spaced them.

feltech: LGTM, thanks.

Issue #508, opened 2022-07-05
Rename `PluginSystemManagerFactory` and expose to C++


foundrytom:
## What
 
Rename `PluginSystemManagerFactory` -> `PythonPluginSystemManagerImplementationFactory` and expose to C++.

## Why

In the future, we will need independent C++ and Python plugin systems. This disambiguates the two, so they can be combined for #191.

The convenience method is to allow a host to instantiate the python plugin system without needing access to a compatible version of pybind.

## ACs

- Factory renamed
- C++ function to create a `PythonPluginSystemManagerImplementationFactory ` and return a `ManagerImplementationFactoryInterfacePtr` (in `openassetio-core-python`), so that it can be passed to `ManagerFactory` by a host.
- Unit test that simply checks class of the returned object, it's fine for now for these tests to be in the python, calling contrived methods in `_openassetio_test`.

## Notes

NB: Names are illustrative

- Add `libopenassetio-core-python.so`. to `openassetio-core-python`, which provides the implementation of the C++ function that instantiates the python plugin system factory and retrieves a ptr to it. Along with headers for that method under a `python` namespace.

- No dependencies between `_openassetio.so` and `libopenassetio-core-python.so`

- ~`libopenassetio-core-python.so` links against `libpython` same as `_openassetio.so`.~ #528 

- The C++ helpers should be optional, so pure python builds can be made, and should only be able to be enabled if python is enabled.

- The factory function can assume python is already available and initialized.

PR #539, opened 2022-07-27
[Core] Add `createPluginSystemManagerImplementationFactory`

feltech:
Closes https://github.com/OpenAssetIO/OpenAssetIO/issues/508.

The convenience method is to allow a C++ host to instantiate the Python plugin system without needing access to a compatible version of Pybind.


src/openassetio-core-python/python/hostApi.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+  // Instantiate Python object.
+  py::object pyInstance = pyClass(std::move(logger));
+
+  // Keep Python object alive whilst shared_ptr chain is alive.
+  auto pyInstancePtr =
+      std::shared_ptr<py::object>{new py::object{pyInstance}, [](py::object* pyObjectPtr) {
+                                    py::gil_scoped_acquire gil;
+                                    delete pyObjectPtr;
+                                  }};
```
foundrytom, line 36: [consider] Is there any way we can move this to a helper in `PyRetaningSharedPtr.hpp` as it's the same as the inner code in the `type_caster`, and it's well documented what it's upto there.

foundrytom (changes requested):
Just a Q around re-use, plus the larger topic we spoke of yesterday around moving this into `openassetio-python` or otherwise avoiding `core` in the name.

src/openassetio-core-python/python/hostApi.cpp
feltech: Split out in 09f09b2a (note: juggled commits and force-pushed).

feltech:
I wasn't sure how far to go with the rename of `PluginSystem...` to `PythonPythonSystem...`, but it looked odd only renaming the one class/module.  The renames are in separate commits anyway, so can be easily dropped if desired.

tests/python/openassetio/pluginSystem/test_pythonpluginsystem.py
```diff
@@ -26,7 +26,7 @@
 import pytest
 
 from openassetio.log import ConsoleLogger
-from openassetio.pluginSystem import PluginSystem
+from openassetio.pluginSystem import PythonPluginSystem
 
 
 class Test_PluginSystem_scan:
```
foundrytom, line 32: [fix]
```suggestion
class Test_PythonPluginSystem_scan:
```

tests/python/openassetio/pluginSystem/resources/pathA/modulePlugin.py
```diff
@@ -1,16 +1,18 @@
...
+Provides a test PythonPluginSystemPlugin implemented within a single file
 module.
 """
 
-from openassetio.pluginSystem import PluginSystemPlugin
+from openassetio.pluginSystem import PythonPluginSystemPlugin
 
-class ModulePlugin(PluginSystemPlugin):
+
+class ModulePlugin(PythonPluginSystemPlugin):
```
foundrytom, line 9: [question] Do we need to consider moving these python plugins into a separate `python` space as part ofd this too?

foundrytom (changes requested):
Great job @feltech just a few small things.

src/openassetio-python/bridge/python/hostApi.cpp
```diff
@@ -24,8 +24,9 @@ ManagerImplementationFactoryInterfacePtr createPluginSystemManagerImplementation
     LoggerInterfacePtr logger) {  // NOLINT(performance-unnecessary-value-param)
   // Get Python class.
   py::object pyClass =
-      py::module_::import("openassetio.pluginSystem.PluginSystemManagerImplementationFactory")
-          .attr("PluginSystemManagerImplementationFactory");
+      py::module_::import(
+          "openassetio.pluginSystem.PythonPluginSystemManagerImplementationFactory")
```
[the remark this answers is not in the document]
feltech: Will do. I wasn't sure about this, since it's in the `python` namespace so prefixing with `Python` seemed redundant.  But better to be super explicit for the avoidance of any doubt.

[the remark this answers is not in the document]
foundrytom: Yeah - I was just imagining that when we have a unifying plugin system, we may well still want to expose that to python too, which then might cause ambiguity, so having it match the class name is probably quite valuable.

tests/python/openassetio/pluginSystem/resources/pathA/modulePlugin.py
foundrytom: ... So we have a python-only search path, at the point we have multiple plugin systems living in this folder, with their own search tests? That being said, we maybe want to leave it where it is, so we can double check coexistence between the two systems works too!

feltech: Does that open a can of worms, i.e. should then the `pluginSystem` package have a `python` subpackage?  Maybe we should just do that now, but I'm happy to defer until we work on the C++ plugin system, YAGNI and all that.

Another argument, specific to these tests, is perhaps that any plugin defined in Python must by definition be using the Python plugin system, so no need to disambiguate.

foundrytom: Lets save it till we know what it looks like. We will def need too support a shared plugin path for both plugin systems,

tests/openassetio-python/hostApiTest.cpp
```diff
@@ -0,0 +1,21 @@
...
+ */
+#include <pybind11/pybind11.h>
+#include <openassetio/LoggerInterface.hpp>
+#include <openassetio/hostApi/ManagerImplementationFactoryInterface.hpp>
+#include <openassetio/python/hostApi.hpp>
+
+namespace py = pybind11;
+
+void registerHostApiTestTypes(py::module_& mod) {
+  mod.def("createPluginSystemManagerImplementationFactory",
```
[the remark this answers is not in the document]
feltech: Renamed as suggested in 715df1e3

src/openassetio-python/bridge/python/hostApi.cpp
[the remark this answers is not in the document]
feltech: Renamed in dcd5528b

tests/python/openassetio/test_imports.py
```diff
@@ -114,17 +114,17 @@ def test_importing_ManagerStateBase_succeeds(self):
...
-        from openassetio.pluginSystem import ManagerPlugin
+    def test_importing_PythonPluginSystemManagerPlugin_succeeds(self):
+        from openassetio.pluginSystem import PythonPluginSystemManagerPlugin
 
-    def test_importing_PluginSystem_succeeds(self):
-        from openassetio.pluginSystem import PluginSystem
+    def test_importing_PythonPluginSystem_succeeds(self):
+        from openassetio.pluginSystem import PythonPluginSystem
 
     def test_importing_PluginSystemManagerImplementationFactory_succeeds(self):
```
foundrytom, line 123: [fix[
```suggestion
    def test_importing_PythonPluginSystemManagerImplementationFactory_succeeds(self):
```

foundrytom (changes requested):
Sorry had also missed that `tests/python/openassetio/pluginSystem/resources/README.md` references the old class names.

foundrytom:
@feltech Feel free to squash and merge when you've fixed those last two gremlins.

foundrytom (approved):

Issue #516, opened 2022-07-08
Re-order logger severity enum values


feltech:
# What

Re-order the logger severity enum values such that they match the logical order of severity

# Why

Having opposite order in the enum vs. logical severity makes the code more difficult to read and docstrings less obvious as to their meaning (e.g. "higher severity" could be interpreted the opposite of its intention).

PR #545, opened 2022-08-01
[Core] Invert log severity constant order

foundrytom:
Reverses the order of logging severity constants, such that a logically higher severity now has a higher associated numeric value.

Closes #516

### DIFF
--- a/python/openassetio/log.py
+++ b/python/openassetio/log.py
@@ -40,11 +40,11 @@ class SeverityFilter(LoggerInterface):
     def __init__(self, upstreamLogger):
         LoggerInterface.__init__(self)
 
-        self.__maxSeverity = self.kWarning
+        self.__minSeverity = self.kWarning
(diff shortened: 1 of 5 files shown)


python/openassetio/log.py
```diff
@@ -75,7 +75,7 @@ def setSeverity(self, severity):
```
feltech, line 75: [fix] Several lines above here (annoying github won't let me comment there) the `@name Filter Severity` docstrings can remove the "Confusingly" part now :partying_face:

feltech (approved):
Looks good, just a minor bit of doc tidying. Happy for this to be squashed+merged when fixed.

python/openassetio/log.py
foundrytom: Ahh yes! Good spot, thanks :) removed in squish.

Issue #546, opened 2022-08-03
Improve `SeverityFilter`


foundrytom:
## What

`log::SeverityFilter` is pretty lame. This needs some re-working to support:

 - Things other than `int` in the env var.
 - Proper messaging when the value is invalid, debug logging when correct, etc..

feltech:
> There's a question as to whether we should just merge SeverityFilter into LoggerInterface [#1014](https://github.com/OpenAssetIO/OpenAssetIO/issues/1014)

To follow-up - the `LoggerInterface` class now has a `isSeverityLogged()` function to query if a severity level _should_ be logged; however the `LoggerInterface` class itself doesn't filter anything, so we still rely on a wrapper or subclass (e.g. `SeverityFilter`) to implement log filtering.

To be clear, `SeverityFilter` is only really useful in fairly basic use-cases where the host application doesn't have its own logging implementation (complete with filtering) that is abstracted for use in OpenAssetIO by a `LoggerInterface` subclass implementation.

PR #552, opened 2022-08-05
Migrate Python loggers to C++

foundrytom:
Migrates the current (fairly minimal) logger implementations (`ConsoleLogger` and `SeverityFilter`) over to C++ so we can leave the GIL out of messaging.

It is a fairly literal port that takes the path of least resistance to get us going, and we have #546 to address some of the more questionable aspects of the current implementation.

### DIFF
--- a/src/openassetio-core/include/openassetio/log/ConsoleLogger.hpp
+++ b/src/openassetio-core/include/openassetio/log/ConsoleLogger.hpp
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 The Foundry Visionmongers Ltd
+
+#include <openassetio/export.h>
+#include <openassetio/log/LoggerInterface.hpp>
+
+#pragma once
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace log {
+OPENASSETIO_DECLARE_PTR(ConsoleLogger)
+/**
+ * A logger that sends messages to the console (stderr).
+ */
+class OPENASSETIO_CORE_EXPORT ConsoleLogger final : public LoggerInterface {
+ public:
+  /**
+   * Creates a new instance of the ConsoleLogger
+   *
+   * @param shouldColorOutput When true, messages will be colored based on
+   * their severity.
+   */
+  [[nodiscard]] ConsoleLoggerPtr static make(bool shouldColorOutput = true);
+
+  /**
+   */
+  void log(Severity severity, const Str& message) override;
+
+ private:
+  explicit ConsoleLogger(bool shouldColorOutput);
+
+  bool shouldColorOutput_;
+};
+}  // namespace log
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
--- a/src/openassetio-core/include/openassetio/log/LoggerInterface.hpp
+++ b/src/openassetio-core/include/openassetio/log/LoggerInterface.hpp
@@ -8,6 +8,16 @@
 #pragma once
 namespace openassetio {
 inline namespace OPENASSETIO_CORE_ABI_VERSION {
+/**
+ This namespace contains code relevant to message logging within the
+ API.
+
+ @ref host "Host" authors provide an implementation of @ref
+ LoggerInterface to a @ref manager to channel its messages. The API
+ middleware also makes use of this logger to provide debugging
+ information about use of the API at runtime.
+*/
+namespace log {
 OPENASSETIO_DECLARE_PTR(LoggerInterface)
 /**
  * An abstract base class that defines the receiving interface for
@@ -40,5 +50,6 @@ class OPENASSETIO_CORE_EXPORT LoggerInterface {
    */
   virtual void log(Severity severity, const Str& message) = 0;
 };
+}  // namespace log
 }  // namespace OPENASSETIO_CORE_ABI_VERSION
 }  // namespace openassetio
--- a/src/openassetio-core/log/ConsoleLogger.cpp
+++ b/src/openassetio-core/log/ConsoleLogger.cpp
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 The Foundry Visionmongers Ltd
+#include <iomanip>
+#include <iostream>
+
+#include <openassetio/log/ConsoleLogger.hpp>
+
+// Foreground ANSI color codes (combined with \033[<color>m)
+constexpr const int kSeverityColors[] = {36, 32, 32, 0, 33, 31, 31};
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace log {
+ConsoleLoggerPtr ConsoleLogger::make(bool shouldColorOutput) {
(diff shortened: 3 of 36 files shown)


src/openassetio-core/log/ConsoleLogger.cpp
```diff
@@ -0,0 +1,36 @@
...
+namespace log {
+ConsoleLoggerPtr ConsoleLogger::make(bool colorOutput) {
+  return std::shared_ptr<ConsoleLogger>(new ConsoleLogger(colorOutput));
+}
+
+ConsoleLogger::ConsoleLogger(bool colorOutput) : colorOutput_(colorOutput) {}
+
+void ConsoleLogger::log(Severity severity, const Str& message) {
+  if (colorOutput_) {
+    std::cerr << "\033[0;" << kSeverityColors[severity] << "m";
```
feltech, line 22: [consider] If we want to be grown-up, we should probably 
* Use `std::cerr` for `severity > kInfo` or possibly `> kWarning`.
* Use `std::clog` otherwise.

It's remarkably difficult to find easy to understand and detailed docs on this.  

`std::cerr` is unbuffered, so flushes with every `<<`, and it's also tied to `std:cout`, so forces a flush of `std::cout` before writing to stderr.

`std::clog`, on the other hand, also writes to stderr, but does not automatically flush and is not tied to `std::cout`.

So the general advice is, use `clog` for logging, `cerr` for errors - for performance when logging vs. immediate feedback when reporting errors.

Some optional references
* https://open-std.org/JTC1/SC22/WG21/docs/papers/2019/p1721r0.pdf (`29.4.3 Narrow stream objects`)
* https://en.cppreference.com/w/cpp/io/clog
* https://en.cppreference.com/w/cpp/io/cerr
* https://cplusplus.com/forum/general/284339/ - thread where people seem to know what they're talking about

src/openassetio-core/include/openassetio/log/ConsoleLogger.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <openassetio/export.h>
+#include <openassetio/log/LoggerInterface.hpp>
+
+#pragma once
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace log {
+OPENASSETIO_DECLARE_PTR(ConsoleLogger)
+/**
+ * A logger that sends messages to the console (std:cerr).
```
feltech, line 13: [fix] 
```suggestion
 * A logger that sends messages to the console (stderr).
```

src/openassetio-core/include/openassetio/log/ConsoleLogger.hpp
```diff
@@ -0,0 +1,36 @@
...
+   * their severity.
+   */
+  [[nodiscard]] ConsoleLoggerPtr static make(bool colorOutput = true);
+
+  /**
+   */
+  void log(Severity severity, const Str& message) override;
+
+ private:
+  explicit ConsoleLogger(bool colorOutput = true);
```
feltech, line 30: [consider] Unnecessary default argument.

src/openassetio-core/include/openassetio/log/ConsoleLogger.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+  bool colorOutput_;
```
feltech, line 32: [consider][minor] Super minor but I always try to use an `is` prefix for booleans, i.e. `isColorOutput`, so it's clear at a glance.  Maybe that's a bit old-school though.

src/openassetio-core/log/ConsoleLogger.cpp
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  // NOLINTNEXTLINE(readability-magic-numbers)
+  std::cerr << std::setw(11) << kSeverityNames[severity] << ": " << message;
+
+  if (colorOutput_) {
+    std::cerr << "\033[0m";
+  }
+
+  std::cerr << std::endl;
```
feltech, line 32: [fix] `std::endl` does a `\n` followed by a `std::flush`.  As mentioned in the other comment, this is unnecessary, especially for `cerr`, and has performance implications.  

Also frowned upon by [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rio-endl):
>  it simply slows down output by doing redundant flush()s.
```suggestion
  std::cerr << "\n";
```

feltech (changes requested):

src/openassetio-core/log/ConsoleLogger.cpp
foundrytom: Deferring until #555.

src/openassetio-core/include/openassetio/log/LoggerInterface.hpp
```diff
@@ -8,6 +8,7 @@
 #pragma once
 namespace openassetio {
 inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace log {
```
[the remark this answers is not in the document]
foundrytom: Added [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/552/commits/6fc9845fa5af0db5cad19fa4e12ddbba29c4c4ed).

src/openassetio-core/include/openassetio/log/ConsoleLogger.hpp
foundrytom: Removed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/552/commits/0f034a1c2816dc46d2607758705401e666973533).

src/openassetio-core/include/openassetio/log/ConsoleLogger.hpp
foundrytom: Renamed [here]()https://github.com/OpenAssetIO/OpenAssetIO/pull/552/commits/bffe11744668200772b6a459a4779b6eafc12c95).

src/openassetio-core/log/ConsoleLogger.cpp
foundrytom: Updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/552/commits/40e5b149e5a972fdb0596987c1a80e23f680c125).

foundrytom:
@feltech Also spotted a discrepancy between default arg values in C++ and Python, fixed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/552/commits/6251a609e8773ddacdcc005526a2c845b77bd134) with tests.

tests/python/openassetio/test_log.py
```diff
@@ -55,6 +55,55 @@ def severity_filter(mock_logger):
...
+        output = capfd.readouterr()
+
+        assert output.out == ""
+        assert output.err == "\n".join(
+            [f"{lg.LoggerInterface.kSeverityNames[s] : >11}: A message" for s in all_severities]
+        ) + "\n"
+
+    def test_when_shouldColorOutput_not_set_then_messages_are_colored(self, capfd):
+        _ = capfd.readouterr()
+        logger = lg.ConsoleLogger(shouldColorOutput=True)
```
feltech, line 84: [fix] 
```suggestion
        logger = lg.ConsoleLogger()
```

feltech (approved):
Just a minor fixup to a test, once that's done happy for this to be squashed and merged.

Issue #549, opened 2022-08-05
[Core] Strongly typed `isEntityReference` workflow


foundrytom:
# What

Convert uses of entity references to use a strongly-typed object rather than a string, as decided in #532. Implement this system in C++ with Python bindings.

# Why

In order for a C++ host to use an entity reference string as a parameter to manager API methods, the host is required to validate that it is relevant to the target manager.  In the current Python API, this is a trust-based system, whereby the host is expected to call `Manager.isEntityReference` on a string, before using that string as an entity reference parameter.  

The new design adds strong-typing semantics to wrap the entity reference string in a dedicated `EntityReference` object. This object should only ever be created by by the manager middleware, which will ensure validation of the entity reference string.

To support diverse workflows, we still require `isEntityReference` and further require both an exception-throwing `EntityReference`  factory function, as well as a `std::optional`-based `EntityReference`  factory function.

# ACs

* `isEntityReference` migrated to C++.
* All methods that take an entity reference string instead take a strongly-typed `EntityReference` object.
* `isEntityReference` renamed to `isEntityReferenceString` to disambiguate that it checks a string rather than `EntityReference` value-semantic object.
* `EntityReference` objects are convertible back to the underlying string entity identifier: i.e. `const string& ref.toString() const`
* A factory method `EntityReference Manager::createEntityReference(std::string)` (in a similar pattern to `createContext`), which throws a `std::domain_error` if the reference is not relevant to the manager.
* A factory method `std::optional<EntitiyReference> Manager::createEntityReferenceIfValid(std::string)`, which returns an empty `optional` if the reference is not relevant to the manager.
* For testing purposes, similar to `Context`, the `EntityReference` type can be trivially constructed (so no validation). However, this should be clearly marked as unsupported and private in any docs.
* Examples and related docs updated.
* Python bindings for the above.
* C API design sketch

# Out of scope

* C API.
* Middleware repeat validation (see #553)
* `info` based short-circuit middleware implementation
* Port of custom OpenAssetIO exceptions to C++

# Tasks
- [x] Rename `isEntityReference`
- [x] Migrate `isEntityReferenceString` to C++.
- [x] Add `EntityReference` strong type.
- [x] Add `Manager.createEntityReference` factory function
- [x] Add `Manager.createEntityReferenceIfValid`
- [x] Update all uses of entity references to use `EntityReference` strong type.
- [x] C API design for entity reference validation and usage workflow
[...]

PR #560, opened 2022-08-09
`isEntityReference` refactor

foundrytom:
Part of #549, refactors `isEntityReference` to the non-batch `isEntityReferenceString`, see the individual commit messages for more details.

### DIFF
--- a/python/openassetio/managerApi/ManagerInterface.py
+++ b/python/openassetio/managerApi/ManagerInterface.py
@@ -246,7 +246,7 @@ class ManagerInterface(_openassetio.managerApi.ManagerInterface):
     # @{
 
     @abc.abstractmethod
-    def isEntityReference(self, tokens, hostSession):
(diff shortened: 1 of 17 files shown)


python/openassetio/managerApi/ManagerInterface.py
```diff
@@ -246,29 +246,29 @@ def flushCaches(self, hostSession):
...
         @warning The result of this call should not depend on the
         context Locale.
 
-        @param tokens `List[str]` The strings to be inspected.
+        @param someString str The string to be inspected.
 
         @param hostSession HostSession The API session.
 
-        @return `List[bool]` `True` if the supplied token should be
+        @return `bool` `True` if the supplied stringn should be
```
feltech, line 271: [fix]
```suggestion
        @return `bool` `True` if the supplied string should be
```

feltech (approved):
Minor typo, otherwise looks good, I'm happy for this to be squashed+merged once fixed.

PR #565, opened 2022-08-10
[Core] Add `Manager.createEntityReference`

feltech:
Part of #549. Added the `createEntityReference` method to C++ and Python, which should always be used to create `EntityReference` objects.

The method will validate that the provided entity reference string is relevant to the manager and throw a `std::domain_error` if it isn't (exception type to be migrated to a custom type in future work).

### DIFF
--- a/src/openassetio-core/hostApi/Manager.cpp
+++ b/src/openassetio-core/hostApi/Manager.cpp
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2013-2022 The Foundry Visionmongers Ltd
+#include <stdexcept>
+
 #include <openassetio/Context.hpp>
 #include <openassetio/hostApi/Manager.hpp>
 #include <openassetio/managerApi/HostSession.hpp>
@@ -70,6 +72,15 @@ ContextPtr Manager::contextFromPersistenceToken(const std::string &token) {
 
 bool Manager::isEntityReferenceString(const std::string &someString) const {
   return managerInterface_->isEntityReferenceString(someString, hostSession_);
+}
+
+const Str kCreateEntityReferenceErrorMessage = "Invalid entity reference: ";
+
+EntityReference Manager::createEntityReference(Str entityReferenceString) const {
+  if (!isEntityReferenceString(entityReferenceString)) {
+    throw std::domain_error{kCreateEntityReferenceErrorMessage + entityReferenceString};
+  }
+  return EntityReference{std::move(entityReferenceString)};
 }
 
 }  // namespace hostApi
--- a/src/openassetio-core/hostApi/Manager.cpp
+++ b/src/openassetio-core/hostApi/Manager.cpp
@@ -83,6 +83,14 @@ EntityReference Manager::createEntityReference(Str entityReferenceString) const 
   return EntityReference{std::move(entityReferenceString)};
 }
 
+std::optional<EntityReference> Manager::createEntityReferenceIfValid(
+    Str entityReferenceString) const {
+  if (!isEntityReferenceString(entityReferenceString)) {
+    return {};
+  }
+  return EntityReference{std::move(entityReferenceString)};
+}
+
 }  // namespace hostApi
 }  // namespace OPENASSETIO_CORE_ABI_VERSION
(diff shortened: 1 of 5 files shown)


src/openassetio-core/hostApi/Manager.cpp
```diff
@@ -68,10 +70,19 @@ ContextPtr Manager::contextFromPersistenceToken(const std::string &token) {
   return context;
 }
 
-bool Manager::isEntityReferenceString(const std::string &someString) const {
+bool Manager::isEntityReferenceString(const Str &someString) const {
```
foundrytom, line 73: [fix] Be good to split this out into a separate commit, so we can add rationale (given our recent conversation). IIRC the idea of `std::string` for "raw input from the outside world" as with the persistence token stuff, was that it felt to make sense to use our own typedefs for "things that come _from_ the API", whereas every thing that goes into this method _has_ to come from some other code base - so using the typedef forces developers to have to go hunt what that type actually is to work out if they can get their string into that format. If its just "this is the type we accept here" then its easier to understand the conversion from whatever form the data is already in?

foundrytom (changes requested):
👍

src/openassetio-core/hostApi/Manager.cpp
foundrytom: Was thinking too, if we are going to do this, we need to add an issue to update the persistence token methods.

feltech: Conceptually, I had imagined that `Str` is just some string type, which is convertible/constructible to/from a `std::string`, and the fact that it's implemented as an alias to `std::string` is an implementation detail.

I thought of the manager state token as a reasonable exception, since `std::string` is acting more as a container for some arbitrary bytes, so is not semantically a string, e.g. that you would want to render to a UI.  Maybe that too should be its own "type", e.g. `using SerializedManangerState = std::string`.

However, point well taken about readability.  From experience I know it can be annoying to jump through code trying to find the underlying type of an alias (though that's mainly when trying to grok templated type traits).

On the broader point of have two string "types", `std::string` for external calls and `Str` for internal usage, I can get behind that, but I suspect the dividing line will be tricky.  We'll have to flesh out what this means for various parts of the API.

Maybe it would be simpler to just remove `Str` if it's adding confusion, especially since the alias is supposed to reduce confusion by abstracting the detail. If the opposite is true, and we're pretty certain we'll never drop-in replace it with a custom type, then it's doing more harm than good.

foundrytom: > Conceptually, I had imagined that `Str` is just some string type, which is convertible/constructible to/from a `std::string`, and the fact that it's implemented as an alias to `std::string` is an implementation detail.

Ahh, so maybe that's the reason for the ambiguity - I'm not sure that we surface that expectation (in terms of a public contract for the type) in an obvious-to-new-devs-way in the API (eg: docs in `typedefs.hpp`).

> I thought of the manager state token as a reasonable exception, since `std::string` is acting more as a container for some arbitrary bytes, so is not semantically a string, e.g. that you would want to render to a UI. Maybe that too should be its own "type", e.g. `using SerializedManangerState = std::string`.

Right, got you - I guess that would also need its public contract explicitly defining in the saw way as `Str`.
 
> On the broader point of have two string "types", `std::string` for external calls and `Str` for internal usage, I can get behind that, but I suspect the dividing line will be tricky. We'll have to flesh out what this means for various parts of the API.

Yeah, with you on that. It does feel like we can't actually make that differentiation, as even something like Identifier - which is only ever API derived, needs persisting, so maybe it really is more a question of "how visible is the public contract of these types" - and the scenarios we've looked at here are a specific enough to force the discussion.

> Maybe it would be simpler to just remove `Str` if it's adding confusion, especially since the alias is supposed to reduce confusion by abstracting the detail. If the opposite is true, and we're pretty certain we'll never drop-in replace it with a custom type, then it's doing more harm than good.

I'd imagine its unlikely we'd ever drop a type in that isn't 100% compatible with whatever went before, so is `std::string` the clearest way to explain that API contract?

foundrytom (approved):
LGTMSAMAYL

PR #569, opened 2022-08-11
[Core] Add BatchElementError

### LINKED ISSUE #557: `BatchElementError` type for callback-based batch methods
# What

Add a `BatchElementError` class to be used in error callbacks for batch-first API methods.

# Why

Our batch-first design means that errors that affect only a single element of the batch should not cause the whole batch to fail. Instead, such errors are communicated via an error callback, with a unique call per failed batch element.

We need some way for the manager to communicate the reason for a failure to the error callback. In order for this to be extensible we need to bundle the error details into a class.  At minimum these details will include an error code and message, giving the same level of information as a standard exception.

# ACs

* A header containing error codes in the core library that is importable in the C library, in order to prevent duplication of error codes. 
  - This should not have a dependency on the C library/headers.  
  - Initially supporting only "kUnknown".
  - Do not overlap these new error codes with the exception error codes.
* `BatchElementError` value semantic struct containing
  - `enum class ErrorCode` (initially) supporting only "kUnknown".
  - Error code value, `const ErrorCode code`.
  - Error message value `const std::string message`.
* C API sketch

# Out of scope

* New error codes.

feltech:
Closes #557.

Add a simple `BatchElementError` aggregate class, initially containing just an error code and string message. The error code is a strongly typed enumeration, currently only supporting a fallback code of `kUnknown`.

To prevent ambiguity of these error codes with the C API's exception error codes, and to allow reuse of the error codes between C++ and C, add a new C-compatible header the core library, with `#define`s computing the new error codes as offset from the C API's exception error codes.

### DIFF
--- a/src/openassetio-core/include/openassetio/BatchElementError.hpp
+++ b/src/openassetio-core/include/openassetio/BatchElementError.hpp
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 The Foundry Visionmongers Ltd
+#pragma once
+
+#include <string>
+#include <utility>
+
+#include <openassetio/errors.h>
+#include <openassetio/export.h>
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+/**
+ * Structure representing per-element batch operation errors.
+ *
+ * Many OpenAssetIO API functions take multiple inputs, i.e. a batch of
+ * elements, in order to allow the @ref manager backend to optimise bulk
+ * queries. The results of such queries are usually returned
+ * element-by-element via a callback.
+ *
+ * It is possible for the whole batch to fail due to some catastrophic
+ * error, in which case a standard exception workflow is expected. Using
+ * HTTP status codes as an analogy, a client error (4xx) would likely
+ * correspond to a `BatchElementError`, whereas a server error (5xx)
+ * would likely cause the whole batch to fail with an exception.
+ *
+ * However, it is also possible for a subset of elements in the batch to
+ * fail, whilst the remainder succeed. An exception workflow doesn't
+ * work so well here, and so every success callback is paired with an
+ * error callback, allowing per-element errors to be communicated back
+ * to the original caller (i.e. the @ref host application).
+ *
+ * The information for these per-element errors is bundled in instances
+ * of this simple BatchElementError structure for passing to error
+ * callbacks.
+ *
+ * This structure provides an error code, for control flow, and an
+ * error message, for more (human-readable) detail.
+ */
+class BatchElementError final {
+ public:
+  /// Possible classes of error.
+  enum class ErrorCode {
+    /// Fallback for uncommon errors.
+    kUnknown = OPENASSETIO_BatchErrorCode_kUnknown
+  };
+
+  /// Error code indicating the class of error.
+  const ErrorCode code;
+  /// Human-readable error message.
+  const std::string message;
+};
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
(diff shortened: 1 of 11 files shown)


src/openassetio-core/include/openassetio/BatchElementError.hpp
```diff
@@ -0,0 +1,51 @@
...
+  /// Possible classes of error.
+  enum class ErrorCode {
+    /// Fallback for uncommon errors.
+    kUnknown = OPENASSETIO_BatchErrorCode_kUnknown
+  };
+
+  /// Error code indicating the class of error.
+  const ErrorCode code;
+  /// Human-readable error message.
+  const Str message;
```
foundrytom, line 48: [fix] Can we change this to `std::string` based on the char around that whole world, save revisiting own the not too distant.

src/openassetio-core/include/openassetio/errors.h
```diff
@@ -0,0 +1,30 @@
...
+ * exception-like errors.
+ */
+#define OPENASSETIO_ErrorCode_BEGIN 1
+
+/**
+ * First error code used for representing a batch element error.
+ *
+ * Assumes a maximum of 100 exception-like error codes before these.
+ */
+#define OPENASSETIO_BatchErrorCode_BEGIN (OPENASSETIO_ErrorCode_BEGIN + 100)
```
foundrytom, line 27: [question] Is there any merit in making this some power of 2?

foundrytom (changes requested):

src/openassetio-core/include/openassetio/errors.h
feltech: :thinking: Interesting. Off the top of my head, we _could_ use bit-masking/shifting instead of inequality to test if a given error code is exception-like or batch element-like, but the additional boilerplate makes that not so useful. 

Anything else you can think of?

src/openassetio-core/include/openassetio/BatchElementError.hpp
feltech: Updated in 0876e51a

src/openassetio-core/include/openassetio/errors.h
foundrytom: Yeah I think that is what was lurking around in my mind - that it would might make it easier to work out what 'class' of code it was? But as they're never meant to be used in the same context - do we think we'd ever have a use case?

src/openassetio-core/include/openassetio/BatchElementError.hpp
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Structure representing per-element batch operation errors.
+ *
+ * Many OpenAssetIO API functions take multiple inputs, i.e. a batch of
+ * elements, in order to allow the @ref manager backend to optimise bulk
+ * queries. The results of such queries are usually returned
+ * element-by-element via a callback.
+ *
+ * It is possible for the whole batch to fail due to some catastrophic
+ * error, in which case a standard exception workflow is expected.
```
foundrytom, line 22: [consider] Based on our chat earlier about BAL, I was looking into HTTP status codes. They separate [Client Errors - 4xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) from [Server Errors - 5xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses). This feels like the axis we've been grasping for when trying to work out what is a `BatchElementError` and what is an exception. Might be good to add this as a concrete example here as to when to use `BatchElementError`?

src/openassetio-core/include/openassetio/errors.h
feltech: Doesn't do any harm, tweaked in 84f7f84a

src/openassetio-core/include/openassetio/BatchElementError.hpp
feltech: HTTP status codes as an analogy is a bit tenuous, but a good rule of thumb.  Added a note in c1373efd, how is that?

foundrytom (approved):
FFTSAMAYL

PR #571, opened 2022-08-12
[Core] Migrate `resolve` to C++ with callback-based API

### LINKED ISSUE #533: Migrate `resolve` to C++

feltech:
Closes #533.

The new design from #483 (DR009) for batch-first methods involves adding success and error callbacks, to be called per element
(entity reference, in this case) with the results. So switch over to this new scheme for `resolve`.

Support invalid entity references being passed to `resolve`, which are pre-validated with `isEntityReferenceString`, but nevertheless malformed for the particular entity. The docs and usage surrounding this was previously limited and ambiguous. So add an `apiComplianceSuite` check to make it concrete, with BAL extended to make use of this error code.

Finally, support implementing and calling `resolve` in C++, with Python bindings.


python/openassetio/test/manager/apiComplianceSuite.py
```diff
@@ -304,24 +303,47 @@ def test_when_resolving_read_only_reference_for_write_then_resolution_error_is_r
...
         context.access = access
-        [result] = self._manager.resolve([reference], self.a_set_of_valid_traits, context)
-        self.assertIsInstance(result, EntityResolutionError)
-        self.assertEqual(str(result), str(expected_error))
+
+        results = []
+        self._manager.resolve(
+            [reference], self.a_set_of_valid_traits, context,
+            lambda _idx, _traits_data: self.fail("Unexpected success callback"),
+            lambda idx, batch_element_error: results.append(batch_element_error)
```
foundrytom, line 347: [fix] `_idx`?

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
```diff
@@ -110,11 +110,18 @@ def entityExists(self, entityRefs, context, hostSession):
             results.append(result)
         return results
 
-    def resolve(self, entityRefs, traitSet, context, hostSession):
+    def resolve(self, entityRefs, traitSet, context, hostSession, successCb, errorCb):
```
foundrytom, line 113: [consider] Surprised `pylint` didn't moan about renaming the parameters here, worth matching the base class names?

python/openassetio/test/manager/apiComplianceSuite.py
foundrytom, line 347: [fix] `_idx`?

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
```diff
@@ -123,11 +130,13 @@ def resolve(self, entityRefs, traitSet, context, hostSession):
...
                     if trait_data is not None:
                         self.__add_trait_to_traits_data(trait, trait_data, result)
-            except Exception as exc:  # pylint: disable=broad-except
-                exc_name = exc.__class__.__name__
-                result = EntityResolutionError(f"{exc_name}: {exc}", ref)
-            results.append(result)
-        return results
+
+                successCb(idx, result)
+            except KeyError:
```
foundrytom, line 141: [fix] Are we sure that only `KeyErrors` are relevant here? There could be other errors from `TraitsData.setTraitProperty` or unpacking to `namedtuple` depending on what is in the JSON, which might be nice to keep as batch errors vs breaking the whole call? if we do want to make those exceptions rather than tie them to the ref (which is helpful), we should document the reason for only catching `KeyError` and leaving the rest.

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
```diff
@@ -110,11 +110,18 @@ def entityExists(self, entityRefs, context, hostSession):
...
         return results
 
-    def resolve(self, entityRefs, traitSet, context, hostSession):
+    def resolve(self, entityRefs, traitSet, context, hostSession, successCb, errorCb):
         if context.isForWrite():
-            return [EntityResolutionError("BAL entities are read-only", ref) for ref in entityRefs]
-        results = []
-        for ref in entityRefs:
+            # TODO(DF): Should this be a whole-batch error (i.e.
+            #  exception) or per-element error?
```
foundrytom, line 116: [fix] Remove TODO. This is the correct behavior I think, as BAL just happens to have this as a holistic response. For the majority of cases, this _is_ a per-element scenario (ie. in any real-world asset system that supports publishing), so we definitely need consistent behaviour, and exceptions should only really be for high-level exceptional happenings that affects the ability to derive _any_ data (eg: server dead).

foundrytom (commented):
WIP comments - ⛴️ has arrived!

python/openassetio/test/manager/apiComplianceSuite.py
feltech: Prefixing an underscore is fairly standard to flag an argument as unused.  It stops pylint complaining.  I could just change it to a lone `_` if you prefer?  Or remove the `_` prefix and suppress any pylint complaints?

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
feltech: Hmm. I think if there's any other error than a `KeyError` from `bal.entity(...)`, then it's a programmer error and should be exceptional?

How about a dedicated exception type for "unknown entity" to replace `KeyError`?

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
feltech: Turns out we don't run pylint on the examples. Trying to pylint the `resources` directory gives lots of issues, mostly about the lack of camelCase conforming, but lots of other errors too.

feltech: Fixed in a9ed554b then realised I'd forgot an argument and fixed in 0c8e7f08

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
feltech: Good spot. Removed in 74321bbc

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
feltech: Added a dedicated exception type in 847d58e8, then restructured a bit in 068d8645 and 7f554ad7

python/openassetio/test/manager/apiComplianceSuite.py
foundrytom: Sorry meant:
```suggestion
            lambda _idx, _traits_data: self.fail("Unexpected success callback"),
            lambda _idx, batch_element_error: results.append(batch_element_error)
```

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
foundrytom: Oh good spot! Sorry that other one has been there for ages 🙃

python/openassetio/test/manager/apiComplianceSuite.py
foundrytom:
```suggestion
            lambda _idx, batch_element_error: results.append(batch_element_error)
```

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
feltech: Based on offline conversation, did a little refactor in c4bf164d to allow unexpected `KeyError`s to propagate.

foundrytom: This is great @feltech a much better approach, capturing our off-line chat, as this is effectively calling out a subtle but important part of the resolve API contract, can we make sure that we have something in the documentation for `resolve` in `Manager`/`ManagerInterface` that makes the HTTP analogy - that `5xx` should always be an exception, and `4xx` batch result error?

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
foundrytom: I forgot about not linting too - it was as we wanted to use pep8 - and local pylint was harder!

python/openassetio/test/manager/apiComplianceSuite.py
feltech: :open_mouth: good spot, yet pylint didn't complain :thinking:

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
feltech: Is it sufficient for that analogy to be in the docs for `BatchElementError` as added in https://github.com/OpenAssetIO/OpenAssetIO/pull/569#discussion_r946580578 ?

python/openassetio/test/manager/apiComplianceSuite.py
foundrytom: I wonder if `pylint` doesn't check lambda args?

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
foundrytom: I think It'd be nice to have some spell-it-out-like-I'm-five examples in the documentation for `resolve` directly, so that someone implementing the method in a manager has a nice cheat sheet to hand, when they're learning about the obligations of that method. For the host - making it really obvious what the failure modes are, and what they mean in terms of business logic would be really useful.

python/openassetio/test/manager/apiComplianceSuite.py
feltech: Fixed (by your suggestion) in https://github.com/OpenAssetIO/OpenAssetIO/commit/90cafc7f2548b8e4ded291f25305545be8e374ae

src/openassetio-core/include/openassetio/BatchElementError.hpp
```diff
@@ -45,9 +45,8 @@ class BatchElementError final {
     kUnknown = OPENASSETIO_BatchErrorCode_kUnknown,
     /**
      * Error code used during @ref entity resolution if a supplied
-     * entity reference does not have a meaningful string
-     * representation, or it is a valid reference format that doesn't
-     * exist.
+     * entity reference is valid but the entity does not exist or
+     * does not have any meaningful representation.
```
feltech, line 49: [question] I updated the docs here from the ones copy-pasted from `exceptions.py` - prompted by "have a meaningful string representation", which I assume is from the dark ages before traits were invented. Is this OK?

resources/examples/manager/BasicAssetLibrary/plugin/BasicAssetLibrary/BasicAssetLibraryInterface.py
feltech: I've expanded the docs a little in 13e8925f, how is that?

foundrytom: Perfect, thanks!

src/openassetio-core/include/openassetio/BatchElementError.hpp
foundrytom: Good spot! Yes that is definitely out of date. As we say that missing traits data should result in the trait not being set, then it would most likely not result in an EntityResolutionError, how about something like this?

```
Error code used during @ref resolve "entity resolution" when the reference
itself is valid, but it is not possible to retrieve data for the referenced 
@ref entity. This could be because it does not exist, or a read-only entity
is being resolved for write. This should not be used if an entity does not
have a requested trait - simply do not set that trait in the resulting data.
```

feltech: Updated to mostly match your suggestion in 5578135b

resources/examples/manager/BasicAssetLibrary/tests/fixtures.py
```diff
@@ -97,9 +97,9 @@
             "a_reference_to_a_missing_entity": "bal:///missing_entity",
             "the_error_string_for_a_reference_to_a_missing_entity": (
                 "Entity 'bal:///missing_entity' not found"),
-            "a_malformed_reference_to_a_readable_entity": f"bal:///{an_existing_entity_name}?arg",
-            "the_error_string_for_a_malformed_reference_to_a_readable_entity": (
-                f"BAL does not support query parameters")
+            "a_malformed_entity_reference": f"bal:///",
```
feltech, line 100: [fix] We discussed offline about removing the assumption that BAL doesn't support query parameters, which was added just to exercise the apiComplianceSuite, and instead make use of the pre-existing check for an empty entity path.

This commit does that.

foundrytom: That looks spot on I think! Thanks

foundrytom (approved):
Great job @feltech LGTMFFTSAMAYLAHAGD

PR #576, opened 2022-08-16
[Core] Remove `Str`

### LINKED ISSUE #572: Use `str::string` for entity reference related methods.
## What

Use `std::string` for `createEntityRefrence` and friends, and in all methods that use `Str` that aren't container types such as `TraitsData` or `InfoDict`.

## Why

We currently have a mixture of `Str` and `std::string` in the methods on `Manager`/`ManagerInterface`.

foundrytom:
Closes #572 (and some).

Having solicited broader consensus around the use of typedefs for simple types, the conclusion was that aliasing `std::string` to `Str` adds more conceptual overhead than is justifiable by the potential future benefits of an abstraction.

Principally, casual reading of the code now calls a spade a `std::spade` which means the reader already understands the semantics of the type.

We have no intention to allow customization of this type at build time, so the main merits in future maintenance are hard to justify if its makes learning the code base harder for an integrator.

As there is no compiler checking of the miss-use of these types of aliases, we had also ended up with situations where the wrong one was accidentally used - potentially creating further ambiguity as to the correct approach.

### DIFF
--- a/resources/perftest/lib.cpp
+++ b/resources/perftest/lib.cpp
@@ -2,7 +2,7 @@
 // Copyright 2013-2022 The Foundry Visionmongers Ltd
 #include "lib.hpp"
 
-void openassetio::LocateableContentTrait::setUrl(openassetio::Str url) {
+void openassetio::LocateableContentTrait::setUrl(std::string url) {
   data()->setTraitProperty(kId, kUrl, std::move(url));
 }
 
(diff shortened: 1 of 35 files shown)


feltech (changes requested):

foundrytom:
https://github.com/OpenAssetIO/OpenAssetIO/pull/912#discussion_r1184724821

foundrytom:
Closing as we're going to go the other way for b1 - and keep the aliases

PR #578, opened 2022-08-18
[Core] Use scoped enum for Context and LoggerInterface

### LINKED ISSUE #568: Context enum restructuring
# What

Investigate whether `enum class` (rather than old-style `enum`) can/should be used for `Context::Access` and `Context::Retention`.

If old-stye `enum`s are chosen for `Context`, then decide whether `BatchElementError::ErrorCode` should be likewise defined.

# Why

Old-style enums do not create a namespace of their own, meaning that `Access` and `Retention` values are exported to the parent `Context` class, i.e. `Context::kIgnored` lives alongside `Context::kUnknown`, as opposed to `Context::Retention::kIgnored` and `Context::Access::kUnknown`, respectively. 

On the other hand, old-style enum constants are trivially convertible to their underlying numeric representation, whereas `enum class` constants require a `static_cast`.  This is relevant to the `Context::kRetentionNames` code-to-string mapping.

feltech:
Closes #568.

Use `enum class` in C++ and do not `export_values()` in Pybind, so that `Context` constants are namespaced under `Context.Retention` and `Context.Access`, and `LoggerInterface` severity constants are namespaced under `LoggerInterrface.Severity`.


contributing/CODING_STANDARDS.md
```diff
@@ -118,6 +118,16 @@ inline namespace OPENASSETIO_CORE_ABI_VERSION {
...
 
+### Enumerations
+
+All enumerations in C++ should use scoped enumerations, i.e.
+`enum class`, to ensure constants do not pollute the parent namespace,
+and to add strong typing, avoiding a class of programmer error. This
+matches the [ISO C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class).
+
+Python bindings should therefore not use [export_values()](https://pybind11.readthedocs.io/en/stable/classes.html#enumerations-and-internal-types)
+when binding a scoped enum type, for consistency with C++.
```
foundrytom, line 129: [fix] Could we add notes on what should be used instead?

foundrytom (changes requested):
Feel free to squash ready for merge when making doc change.

contributing/CODING_STANDARDS.md
feltech: Updated in 1483f316 (as part of a rebase) - a bit tricky since there is nothing to use "instead", it's just about omitting a call.  Does the updated text clarify that?

foundrytom: Spot on, thanks!

foundrytom (approved):
👍 I wish GitHub had a "Merge when checks pass" button like GitLab 😞

Issue #582, opened 2022-08-19
[Core] Fix `Manager.resolve` Python binding


feltech:
# What

During [#533](https://github.com/OpenAssetIO/OpenAssetIO/issues/533) we added an implementation of `Manager.resolve` to C++ but did not migrate Python to use that function via pybind11. Python would continue to use the implementation in `Manager.py`.

This hid a bug, in that we did not include `pybind11/functional.h` in the `Manager` and `ManagerInterface` bindings, which is required for pybind to convert a `std::function` from C++ into a Python callable.  Doing the reverse, i.e. converting a callable from Python into a C++ `std::function`, seems to be supported without that header, which is why our tests succeeded.

PR #590, opened 2022-08-23
[Tests] Add checks for source of `Manager` implementation

foundrytom:
Inspired by #582, where we inadvertently neglected to delete the Python implementation when migrating to C++.

Attempts to ensure that the method is implemented in the expected language during the Python -> C++ transition. The approach is slightly tenuous, but seem to work ok.

Methods like `getsource` could also be used, but ultimately, they raise a `TypeError` if the supplied object isn't a function... so this seems msimpler to invoke/manager.


fnRaihanKibria (approved):
LGTM

Issue #587, opened 2022-08-22
Migrate `preflight`/`register` to callbacks


foundrytom:
## What

Update the signature/documentation for `preflight`/`register` to use the batch callback mechanism as described in #483.

## Why

Pre-requisite for migration to C++

## ACs

- `preflight`/`register` Python implementation updated
- Documentation/~Examples updated~ (See #592)

PR #591, opened 2022-08-23
[Core] Add `kEntityAccessError` to `BatchElementError`

foundrytom:
As part of #587, we need to properly define the API behaviour when calling `preflight` or `register` for an un-versionable, immutable entity.  This required a new error code. Rather than tying this to the specific method, this adds `kEntityAccessError`, that can be used in all cases where the access is the sole cause of error.

As this also applies to `resolve`, updates the `apiComplianceSuite` to check for this too (and consequently updates BAL to be compliant).

### DIFF
--- a/src/openassetio-core/include/openassetio/errors.h
+++ b/src/openassetio-core/include/openassetio/errors.h
@@ -32,5 +32,8 @@
 /// Failure due to malformed entity reference.
 #define OPENASSETIO_BatchErrorCode_kInvalidEntityReference (OPENASSETIO_BatchErrorCode_BEGIN + 1)
 
+/// Failure due to an API method being called with an invalid @ref Context access.
+#define OPENASSETIO_BatchErrorCode_kEntityAccessError (OPENASSETIO_BatchErrorCode_BEGIN + 2)
+
 /// Entity resolution failure.
-#define OPENASSETIO_BatchErrorCode_kEntityResolutionError (OPENASSETIO_BatchErrorCode_BEGIN + 2)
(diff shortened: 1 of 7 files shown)


src/openassetio-core/include/openassetio/errors.h
```diff
@@ -29,8 +29,11 @@
...
-
 /// Failure due to malformed entity reference.
-#define OPENASSETIO_BatchErrorCode_kInvalidEntityReference (OPENASSETIO_BatchErrorCode_BEGIN + 2)
+#define OPENASSETIO_BatchErrorCode_kInvalidEntityReference (OPENASSETIO_BatchErrorCode_BEGIN + 1)
+
+/// Failure due to an API method being called with an invalid @ref Context access.
+#define OPENASSETIO_BatchErrorCode_kEntityAccessError (OPENASSETIO_BatchErrorCode_BEGIN + 2)
+
+/// Entity resolution failure.
+#define OPENASSETIO_BatchErrorCode_kEntityResolutionError (OPENASSETIO_BatchErrorCode_BEGIN + 3)
```
DuncanHopkinsFoundry, line 39: [question] I assume this is something todo with the C++ and Python interop, but why not have these as an enum instead of #defines?

foundrytom: A good q! We also have a [`C` API](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/src/openassetio-core-c/include/openassetio/c/errors.h), which precludes `enum class`, and so we wanted a single source of truth for the actual values, to avoid inadvertent drift.  We haven't added the equivalent of the `BatchElementError` to that API yet though.

PR #594, opened 2022-08-25
[Core] Add `ManagerInterface._createEntityReference`

foundrytom:
In #549, we added a 'strongly typed' entity reference workflow. This required hosts to ensure that strings are validated through the manager before being used as an entity reference. This ensures that once initial checks have been completed, manager implementations no longer need to worry about re-checking method inputs to see if they are one of their references.

We plan on adding assorted validation checks in the API middleware that will help avoid coding errors when multiple managers are in use.

This raises the question of how should a manager implementation create these strongly typed objects - where it doesn't need the validation?

We wanted to avoid direct construction of `EntityReference` objects on the stack by code so that we can evolve the manager validation aspect with minimal code churn at all the construction sites.

We previously debated an object oriented design (`EntityReference(string, manager)` vs the current function + data approach (`Manager.createEntityReference(string)`). We picked the latter as it matched the approach we have with `Context` and is more portable to other languages.

This seems to have lead to a simple solution to this question - where by, we add an alternate (protected) factory method to the ManagerInterface that bypasses the usual checks performed when the strings originate from outside the manager's implementation.

This should then be used by all manager implementation code that needs to create an entity reference to return to the host

### DIFF
--- a/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
+++ b/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
@@ -681,6 +681,21 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
                        const ResolveErrorCallback& errorCallback) = 0;
 
   /// @}
+ protected:
+  /**
+   * Create an @ref EntityReference object wrapping a given @ref
+   * entity_reference string. This should be used for all reference
+   * creation by a manager's implementation.
+   *
+   * No validation is performed as this method is only visible to the
+   * manager implementation, and so it is assumed that its internal
+   * business logic inherently ensures only valid strings are
+   * returned.
+   *
+   * @param entityReferenceString Raw string representation of the
+   * entity reference.
+   */
+  [[nodiscard]] EntityReference createEntityReference(Str entityReferenceString) const;
 };
 }  // namespace managerApi
 }  // namespace OPENASSETIO_CORE_ABI_VERSION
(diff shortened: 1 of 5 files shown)


src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -681,6 +681,21 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+   *
+   * No validation is performed as this method is only visible to the
+   * manager implementation, and so it is assumed that its internal
+   * business logic inherently ensures only valid strings are
+   * returned.
+   *
+   * @param entityReferenceString Raw string representation of the
+   * entity reference.
+   */
+  [[nodiscard]] EntityReference createEntityReference(Str entityReferenceString) const;
```
feltech, line 698: [consider] I think this should be public and used by the `Manager`'s `createEntityReference*` functions. DRY and whatnot.  

Assuming we eventually bundle a reference/ID to the manager implementation in the entity reference, for validation purposes, this then gives us a handy factory function to change.

feltech (commented):
Something to ponder

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
foundrytom: My concern here, is that this method deliberately _doesn't_ validate the string, as it is assumed that it is a valid ref as it is being supplied by the managers implementation (and can only be). If we make this public, that is no longer assured, which then makes it feel like this method should validate, which is the thing its deliberately avoiding.

feltech: Capturing offline discussion, we'll defer this conversation until we look into enhanced validation of entity references. Making this `protected` for now and potentially making `public` later is safer than the reverse.

feltech (approved):
LGTM

PR #595, opened 2022-08-25
[Core] Add `TraitsData.traitProperites`

foundrytom:
This allows introspection of a `TraitsData` instance by a manager in order to extract values for persistent storage.

My modern C++ is very rusty, so please let me know if there is a better way to extract the keys!


src/openassetio-core/include/openassetio/TraitsData.hpp
```diff
@@ -148,6 +148,14 @@ class OPENASSETIO_CORE_EXPORT TraitsData final {
   void setTraitProperty(const trait::TraitId& traitId, const trait::property::Key& propertyKey,
                         trait::property::Value propertyValue);
 
+  /**
+   * Returns the properties set for a given trait.
+   *
+   * If the trait has not been given to this instance, or the trait has
+   * no properties set, then it will return an empty set.
+   */
+  [[nodiscard]] trait::TraitProperties traitProperties(const trait::TraitId& traitId) const;
```
feltech, line 157: [consider] Naming things. I still feel that `traitProperties` is a bit misleading - my first instinct would be that this returns a dictionary of properties, rather than a set of keys. I think `traitPropertyKeys` / `traitPropertyNames` / `traitPropertiesKeys` or similar would be clearer.

src/openassetio-core/include/openassetio/trait/collection.hpp
```diff
@@ -26,6 +26,15 @@ namespace trait {
...
 using TraitSet = std::unordered_set<TraitId>;
 
+/**
+ * A collection of trait property keys.
+ *
+ * Trait property keys collections are a set, rather than a list. In
+ * that, no single key can appear more than once and the order of the
+ * keys has no meaning and is not preserved.
+ */
+using TraitProperties = std::unordered_set<property::Key>;
```
feltech, line 36: [consider] Perhaps this should be under the `property` namespace, e.g. `trait::property::Keys` / `KeySet`

src/openassetio-core/TraitsData.cpp
```diff
@@ -58,6 +58,18 @@ class TraitsData::Impl {
     data_[traitId][propertyKey] = std::move(propertyValue);
   }
 
+  [[nodiscard]] trait::TraitProperties traitProperties(const trait::TraitId& traitId) const {
+    const auto& traitEntry = data_.find(traitId);
+    if (traitEntry == data_.end()) {
+      return {};
+    }
+    trait::TraitProperties propertyKeys;
```
feltech, line 66: [consider][minor] Micro-optimisation
```suggestion
    trait::TraitProperties propertyKeys;
    propertyKeys.reserve(traitEntry->second.size());    
```

feltech (approved):

src/openassetio-core/include/openassetio/TraitsData.hpp
foundrytom: Yeah, good idea. I went for singular `Property` [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/595/commits/5392327b1fe9b2923b12e041c80625a8d78980eb), as otherwise it seemed like the keys for all the properties. 🤷‍♂️

src/openassetio-core/include/openassetio/trait/collection.hpp
foundrytom: Like it! [done](https://github.com/OpenAssetIO/OpenAssetIO/pull/595/commits/b85eb1de7b209d872689b80180fd39df1fc50147).

src/openassetio-core/TraitsData.cpp
foundrytom: Added [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/595/commits/c21a05f065bd214e0de97e4bafab4be79a5e4be1).

PR #599, opened 2022-08-30
[CI] Pin `sqlite3` conan recipe version

foundrytom:
The latest recipe requires Conan 1.51.3, which introduces this bug:
  https://github.com/conan-io/conan/issues/11822

Hopefully this is a band-aid until 1.52 is out and we can create a proper conan lockfile.

### DIFF
--- a/resources/build/conanfile.txt
+++ b/resources/build/conanfile.txt
@@ -10,6 +10,10 @@ catch2/2.13.8
 # Mocking library
 trompeloeil/42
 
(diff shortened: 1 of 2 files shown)


foundrytom:
@feltech Do you prefer the `Revert "<commit>"` approach for the commit title of the conan revert, or better to just retitle to `[CI] Revert Conan version to 1.50.2`?

foundrytom:
Closing, as we decide to postpone this till it becomes a problem (it's only an issue when rebuilding, and the pre-built packages work on CI platforms).

PR #601, opened 2022-09-01
[Core] Migrate `preflight` and `register` to C++

### LINKED ISSUE #588: Migrate `preflight`/`register` to C++
## What

Allow C++ hosts to call `preflight`/`register`.

## ACs

- Python implementation removed
- Issue for C API + NotImplementedError in C Adapter

feltech:
Closes #588

Note that `register` is migrated to `register_` in C++, since "register" is a reserved keyword in C++.


tests/python/openassetio/hostApi/test_manager.py
```diff
@@ -116,6 +116,7 @@ def test_when_constructed_with_ManagerInterface_as_None_then_raises_TypeError(
 class Test_Manager_identifier:
     def test_method_defined_in_cpp(self):
         assert not is_defined_in_python(Manager.identifier)
+        assert is_implemented_once(Manager, "identifier")
```
foundrytom, line 119: [question] Curious as to the motivation for this? `not is_defined_in_python` should fail if you leave the Python implementation around - so is this really just catching the situation where we forget to update the first check?

feltech: Yeah exactly, it forces us to update the test. We can drop this new check if you like.

src/openassetio-python/module/hostApi/ManagerBinding.cpp
```diff
@@ -38,5 +44,24 @@ void registerManager(const py::module& mod) {
...
+          [](Manager& self, const EntityReferences& entityReferences,
+             const trait::TraitsDatas& traitsDatas, const ContextConstPtr& context,
+             const Manager::PublishSuccessCallback& successCallback,
+             const Manager::BatchElementErrorCallback& errorCallback) {
+            // Pybind has no built-in way to assert that a collection
+            // does not contain any `None` elements, so we must add our
+            // own check here.
+            if (std::any_of(traitsDatas.begin(), traitsDatas.end(),
+                            std::logical_not<TraitsDataPtr>{})) {
+              throw pybind11::type_error{"Traits data cannot be None"};
```
feltech, line 61: [question] Based on WIP notes and previous offline chat, I've added this check for no `None` traits datas. Is this OK?

foundrytom: Looks spot on, thanks!

tests/python/openassetio/hostApi/test_manager.py
foundrytom: Cool, no its a good idea as we'll only forget to update the tests otherwise 🙃 as you mentioned too, we really should add it to `ManagerInterface` too - how about splitting this commit out into a new PR and adding the `ManagerInterface` bit there too?

feltech: Spun out a separate PR as https://github.com/OpenAssetIO/OpenAssetIO/pull/606

tests/python/openassetio/hostApi/test_manager.py
```diff
@@ -538,8 +558,9 @@ def test_wraps_the_corresponding_method_of_the_held_interface(
 
 
 class Test_Manager_preflight:
-    def test_method_defined_in_python(self):
+    def test_method_defined_in_cpp(self):
```
foundrytom, line 561: [fix] Think this snuck in to this commit - might need rescuing when dropping when rebasing :)

RELEASE_NOTES.md
```diff
@@ -31,6 +31,11 @@ v1.0.0-alpha.x
   keys for a given trait.
   [#498](https://github.com/OpenAssetIO/OpenAssetIO/issues/498)
 
+- Migrated the `ManagerInterface.preflight` method to C++.
+  [#588](https://github.com/OpenAssetIO/OpenAssetIO/issues/588)
+
+- Migrated the `Manager.preflight` method to C++.
```
foundrytom: [fix] Needs a breaking change for the rename of the callback. Can we split that into it's own commit as its not clear from the title/message that this has been done too.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -561,6 +561,19 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+   * element in a batch.
+   *
+   * This must be called for errors that are specific to a particular
+   * reference in a batch. Exceptions can be thrown to indicate a
+   * whole-batch error.
+   *
+   * The appropriate error code should be used for these errors. See
+   * @fqref{BatchElementError.ErrorCode} "ErrorCode".
+   */
+  using BatchElementErrorCallback = std::function<void(std::size_t, const BatchElementError&)>;
```
foundrytom, line 576: [question] Is there any merit in a common declaration for these? Do we ever imagine changing the callback signature between `Manager`/`ManagerInterface`?

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -665,7 +665,94 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
+   * not share the same trait set.
+   *
+   * Other exceptions may be raised for fatal runtime errors, for
+   * example server communication failure.
+   *
+   * @see @fqref{TraitsData} "TraitsData"
+   * @see @ref preflight
+   */
+  // NOLINTNEXTLINE(readability-identifier-naming)
+  void register_(const EntityReferences& entityReferences,
```
foundrytom, line 753: [question] I just realized, there is a slightly unfortunate collision here, this is our google clang-format convention for class data members. Can we get away with that for now until we think of a new name?

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -753,9 +754,9 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
    */
 
   /**
-   * Callback signature used for a successful publish preflight.
+   * Callback signature used for a successful publish operation.
    */
-  using PreflightSuccessCallback = std::function<void(std::size_t, EntityReference)>;
+  using PublishSuccessCallback = std::function<void(std::size_t, EntityReference)>;
```
foundrytom, line 759: [fix] Can we keep these as two specific callbacks (`PreflightSuccessCallback` and `RegisterSuccessCallback`). The entity hasn't been published at this point, so it feels a bit misleading to have a "publish success" before it has happened in `register`.

tests/python/openassetio/hostApi/test_manager.py
```diff
@@ -573,7 +594,8 @@ def call_callbacks(*_args):
 
 class Test_Manager_register:
     def test_method_defined_in_python(self):
```
foundrytom, line 596: [fix]
```suggestion
    def test_method_defined_in_cpp(self):
```

foundrytom (changes requested):
I need to give this a final pass when I'm back but wanted to get this feedback in for you this week.

tests/python/openassetio/hostApi/test_manager.py
feltech: If I understand correctly, I think that's done. See first commit, 97796b12

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Yeah this should really just be a temporary name.  The new name may be influenced by the conveniences we choose to add, e.g. `registerOne` vs. `registerMultiple` etc.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
foundrytom: Can we go with `PreflightSuccessCallback` and `RegisterSuccessCallback` for largely hand wavey/does what it says on the tine reasons.

RELEASE_NOTES.md
feltech: Split into (first) commit da4fc8f0

tests/python/openassetio/hostApi/test_manager.py
feltech: 682ac146 is now the first change of this line in the PR

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Split in d7346f5a

tests/python/openassetio/hostApi/test_manager.py
feltech: Fixed in b0da5f46

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
foundrytom: Cool, thanks, what do you think about the duplication of the other callbacks in `Manager` and `ManagerInterface` too?

feltech: It's a tricky one. From a technical perspective, if they were to share an alias, which header would it go in?  From a design perspective, are they only coincidentally the same signature and should really be seen as independent?

foundrytom: 👍 lets leave it as-is for now. Thanks.

foundrytom (approved):

PR #606, opened 2022-09-02
[Tests] Assert Manager methods implemented once

feltech:
Add test for robustness against forgetful developers migrating Python methods to C++ then leaving the old Python implementation around, masking the new C++ implementation.

I.e. ensure the `Manager` Python class hierarchy only has a single implementation of each method.

### DIFF
--- a/tests/python/openassetio/hostApi/test_manager.py
+++ b/tests/python/openassetio/hostApi/test_manager.py
@@ -20,7 +20,6 @@ Tests that cover the openassetio.hostApi.Manager wrapper class.
 # pylint: disable=no-self-use
 # pylint: disable=invalid-name,redefined-outer-name
 # pylint: disable=missing-class-docstring,missing-function-docstring
-import inspect
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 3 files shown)


tests/python/openassetio/hostApi/test_manager.py
```diff
@@ -761,3 +787,32 @@ def is_defined_in_python(method):
...
+
+
+def is_implemented_once(cls, method_name):
+    """
+    Check if a method is only implemented once in the class hierarchy.
+
+    In particular, this is useful to ensure we aren't hiding a (C++)
+    base class method with a (Python) subclass method. I.e. this checks
+    that we haven't migrated a method to C++ then forgot to remove the
+    Python implementation.
```
foundrytom, line 799: [fix] Can we add a note that it is a backup check incase we forget to update the `is_implemented_in_python` based tests?

Whilst were there, could we add a `@todo` to remember to remove all these tests once the `Manager.py` is deleted?

foundrytom (changes requested):
Small docs tweak - did we want to add these to `ManagerInterface` too - as you noted earlier?

feltech:
> did we want to add these to `ManagerInterface` too -

Ah yes, good point, will add.

feltech:
> > did we want to add these to `ManagerInterface` too -
> 
> Ah yes, good point, will add.

Refactored in bbf88560 then added tests to `ManagerInterface` in 09147850

tests/python/openassetio/hostApi/test_manager.py
feltech: Added notes in cb6b3684, but then migrated notes  to conftest.py in bbf88560

foundrytom (approved):
Great stuff @feltech like the `conftest` change.

FFTSAMAYL

PR #608, opened 2022-09-08
Default manager config mechanism

### LINKED ISSUE #494: `$OPENASSETIO_DEFAULT_SESSION_SETTINGS`
## What

A standardized mechanism for initializing a session with an externally pre-determined manager/settings.

## Why

Many common production deployment scenarios would have a standard manager plugin + settings to be used across a variety of hosts (See `Session.getSettings`). Requiring each host to independently determine these settings, and persist them individually does not scale well in such deployments.

Ideally there would be a standardized way for a host to construct a 'default session' (which could always be overridden, etc... on a per host/user/session/etc. basis), to minimize the Api initialization boilerplate, and facilitate centralized management of OpenAssetIO configuration in common use cases.

## Sketch

```bash
export OPENASSETIO_DEFAULT_SESSION_SETTINGS=/some/path/to/config.toml
```

### Config.toml
```toml
identifier = "org.openassetio.example.manage.bal"

[manager]
library_path = "/data/shared_library.json"
```

### A Host
```python
session = openassetio.Session.createDefaultSession()
manager = session.currentManager()
```

foundrytom:
Closes #494, with a more modern interpretation of the problem :) The commit notes hopefully explain the design choices.

~Draft PR as we need to work out how to get the TOML dependency into the ASWF container for the one-liner builds.~

### DIFF
--- a/doc/src/ForHosts.dox
+++ b/doc/src/ForHosts.dox
@@ -76,8 +76,9 @@
  *   @fqref{hostApi.HostInterface.identifier} "identifier" and
  *   @fqref{hostApi.HostInterface.displayName} "displayName".
  *
- * - Create a @fqref{hostApi.ManagerFactory} to bootstrap the API,
- *   provide the following objects:
+ * - Use @fqref{hostApi.ManagerFactory.defaultManagerForInterface}
+ *   "ManagerFactory.defaultManagerForInterface" to bootstrap the
+ *   API, provide the following objects:
  *    - An instance of the custom class derived from
  *      @fqref{hostApi.HostInterface} "HostInterface".
  *    - A `logger` (derived from
@@ -91,6 +92,11 @@
  *      openassetio.pluginSystem.PythonPluginSystemManagerImplementationFactory
  *      "PythonPluginSystemManagerImplementationFactory" will be
  *      sufficient.
+ *
+ *   Note that in some environments, a default manager may not be
+ *   configured. As a host, can error here, or choose to present the
+ *   user with a an additional manual configuration mechanism for your
+ *   specific API session if you so wish.
  *
  * - Create, persist and configure a @ref Context as appropriate for
  *   all calls to the API.
--- a/python/openassetio/hostApi/ManagerFactory.py
+++ b/python/openassetio/hostApi/ManagerFactory.py
@@ -60,3 +60,16 @@ class ManagerFactory(_openassetio.hostApi.ManagerFactory):
         # pylint: disable=protected-access
         return Manager(cppManager._interface(), cppManager._hostSession())
 
+    @staticmethod
+    def defaultManagerForInterface(hostInterface, managerImplementationFactory, logger):
+        """
+        @see @fqref{hostApi.ManagerFactory.defaultManagerForInterface}
+        "ManagerFactory.defaultManagerForInterface".
+        """
+        cppManager = _openassetio.hostApi.ManagerFactory.defaultManagerForInterface(
+            hostInterface, managerImplementationFactory, logger
+        )
+        if cppManager:
+            # pylint: disable=protected-access
+            return Manager(cppManager._interface(), cppManager._hostSession())
+        return None
--- a/python/openassetio/hostApi/ManagerFactory.py
+++ b/python/openassetio/hostApi/ManagerFactory.py
@@ -38,21 +38,15 @@ class ManagerFactory(_openassetio.hostApi.ManagerFactory):
         _openassetio.hostApi.ManagerFactory.__init__(
             self, hostInterface, managerImplementationFactory, logger
         )
-        self.__managerImplementationFactory = managerImplementationFactory
-        self.__hostInterface = hostInterface
-        self.__logger = logger
 
     def createManager(self, identifier):
         """
         @see @fqref{hostApi.ManagerFactory.createManager}
         "ManagerFactory.createManager".
         """
-        return Manager(
-            self.__managerImplementationFactory.instantiate(identifier),
-            _openassetio.managerApi.HostSession(
-                _openassetio.managerApi.Host(self.__hostInterface), self.__logger
-            ),
-        )
+        cppManager = super().createManager(identifier)
+        # pylint: disable=protected-access
+        return Manager(cppManager._interface(), cppManager._hostSession())
 
     @staticmethod
     def createManagerForInterface(identifier, hostInterface, managerImplementationFactory, logger):
@@ -60,9 +54,9 @@ class ManagerFactory(_openassetio.hostApi.ManagerFactory):
         @see @fqref{hostApi.ManagerFactory.createManagerForInterface}
         "ManagerFactory.createManagerForInterface".
         """
-        return Manager(
-            managerImplementationFactory.instantiate(identifier),
-            _openassetio.managerApi.HostSession(
-                _openassetio.managerApi.Host(hostInterface), logger
-            ),
+        cppManager = _openassetio.hostApi.ManagerFactory.createManagerForInterface(
+            identifier, hostInterface, managerImplementationFactory, logger
         )
+        # pylint: disable=protected-access
+        return Manager(cppManager._interface(), cppManager._hostSession())
+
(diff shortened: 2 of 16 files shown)


src/openassetio-core/hostApi/ManagerFactory.cpp
```diff
@@ -66,6 +74,69 @@ ManagerPtr ManagerFactory::createManagerForInterface(
...
 }
+
+ManagerPtr ManagerFactory::defaultManagerForInterface(
+    const HostInterfacePtr& hostInterface,
+    const ManagerImplementationFactoryInterfacePtr& managerImplementationFactory,
+    const log::LoggerInterfacePtr& logger) {
+  const char* configPath = std::getenv(kDefaultManagerConfigEnvVarName.c_str());
+
+  if (!configPath) {
+    Str msg = kDefaultManagerConfigEnvVarName + " not set, unable to instantiate default manager.";
```
feltech, line 85: [consider][minor] Micro-optimisation - declare `msg` at the top-level and reuse, rather than re-declaring it.

src/openassetio-core/hostApi/ManagerFactory.cpp
```diff
@@ -66,6 +74,69 @@ ManagerPtr ManagerFactory::createManagerForInterface(
...
+    throw std::runtime_error(msg);
+  }
+
+  auto config = toml::parse_file(configPath);
+
+  std::string_view identifier = config["manager"]["identifier"].value_or("");
+
+  InfoDictionary settings;
+  if (toml::table* settingsTable = config["manager"]["settings"].as_table()) {
+    for (auto&& [key, val] : *settingsTable) {
```
feltech, line 113: [consider][minor] Make use of [`for_each`](https://marzer.github.io/tomlplusplus/classtoml_1_1table.html#a8a73c50b751f4e20dedfeb14a80f9381) visitor. It looks like you could avoid the `if` block altogether... OK not altogether, since there are types we don't support.

Otherwise, the double `&&` looks a bit odd, though apparently it doesn't cause any problems.  I would probably change this to `const auto&`

feltech (changes requested):
First pass looks good. A good bump in functionality for little code.  A few comments on the implementation as it stands...

.clang-tidy
```diff
@@ -42,8 +42,8 @@ Checks: >
 
 # Turn all the warnings from the checks above into errors.
 WarningsAsErrors: "*"
-# Scan all (non-system) headers.
-HeaderFilterRegex: '.*'
+# Scan all headers in our source tree.
+HeaderFilterRegex: '^src/.*'
```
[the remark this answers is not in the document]
foundrytom: Dropped commit now we're bundling toml++

src/openassetio-core/CMakeLists.txt
```diff
@@ -81,6 +83,11 @@ target_include_directories(openassetio-core
     # Use includes from install tree for installed lib.
     "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>")
 
+# Private header dependency.
+target_include_directories(openassetio-core
+    PRIVATE
+    ${tomlplusplus_INCLUDE_DIR})
```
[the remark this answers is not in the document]
foundrytom: Switched to bundled tom++ [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/608/commits/0d2dc56e507073ee0609f184e70c119609080bd0) to simplify building across different environments.

python/openassetio/hostApi/ManagerFactory.py
```diff
@@ -47,22 +46,46 @@ def createManager(self, identifier):
         @see @fqref{hostApi.ManagerFactory.createManager}
         "ManagerFactory.createManager".
         """
-        return Manager(
-            self.__managerImplementationFactory.instantiate(identifier),
-            _openassetio.managerApi.HostSession(
-                _openassetio.managerApi.Host(self.__hostInterface), self.__logger
-            ),
+        cppManager = super().createManager(identifier)
+        # For reasons TBC, grabbing the HostSession from cpp results in
```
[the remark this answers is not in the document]
foundrytom: Updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/608/commits/9b8d3a62f8e5bf2bc506d4e3271be922071378a2) and [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/608/commits/002bab1a4053734bf0314259c9ddbd0dbd44cc02) now its fixed.

src/openassetio-core/hostApi/ManagerFactory.cpp
foundrytom: Improvements made [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/608/commits/dc5a5eaa1107dd9ea23396a42548de39fb6143ea) based on off-line chat. The short is we can't use the `for_each` visitor as filtering the types to avoid "no matching initializer" ends up being more verbose than this approach.

src/openassetio-core/hostApi/ManagerFactory.cpp
foundrytom: Decided to leave as is as it is infrequently called, and this helps encapsulate each use.

doc/src/ForHosts.dox
```diff
@@ -76,8 +76,9 @@
  *   @fqref{hostApi.HostInterface.identifier} "identifier" and
  *   @fqref{hostApi.HostInterface.displayName} "displayName".
  *
- * - Create a @fqref{hostApi.ManagerFactory} to bootstrap the API,
- *   provide the following objects:
+ * - Use @fqref{hostApi.ManagerFactory.defaultManagerForInterface}
+ *   "ManagerFactory.defaultManagerForInterface" to bootstrap the
+ *   API, provide the following objects:
```
feltech, line 81: [question] Does this need a note that you also need a toml config file?  Or are we assuming some sysadmin has set that up prior, for the purposes of these instructions?

tests/python/openassetio/hostApi/test_managerfactory.py
```diff
@@ -127,6 +128,91 @@ def test_when_has_implementations_then_reports_implementation_details(
...
+            "a_int": 42,
+        }
+
+        mock_manager_interface = create_mock_manager_interface()
+        mock_manager_interface.mock.identifier.return_value = "identifier.from.toml.file"
+
+        # Use a side-effect to check host_session as this is constructed
+        # privately by the factory.
+        def check_initialization(settings, host_session):
+            assert settings == expected_settings
```
feltech, line 200: [consider] If I delete this assertion, then blank out the `default_manager.toml` (the intention was to check if any exceptions were raised), then the tests still pass.  We should `assert_called_once_with` on `instantiate`.

feltech (approved):
Looks good, just a couple of minor points.

doc/src/ForHosts.dox
foundrytom: Added some further notes [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/608/commits/16df21f7d728738136b1f8f4ac34e8ad00d64816), helpful?

tests/python/openassetio/hostApi/test_managerfactory.py
foundrytom: Improved test coverage [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/608/commits/4b2128fd95db54ce809070d6d406b48ad184a555).

feltech: Looks good.  Just a quick question.  A blank .toml file (or equivalently, blank manager identifier) is valid from the perspective of the `ManagerFactory`, but am I correct in thinking that's fine because it's up to the `ManagerImplementationFactory` how to interpret it (and potentially raise an exception)?

feltech (approved):
LGTM can squash and merge when ready.

tests/python/openassetio/hostApi/test_managerfactory.py
foundrytom: A good Q. off-line we pondered that we'd leave as-is for now, and just surface the error from the underlying implementation factory.

PR #619, opened 2022-09-12
Support `ctest --parallel`

### LINKED ISSUE #395: [Test] Improve speed of executing tests from CMake

feltech:
Closes #395

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -151,7 +151,7 @@ jobs:
       - name: Build and test
         run: |
           /usr/bin/ccache -s
-          ctest -VV --test-dir build
+          ctest -VV --test-dir build --parallel 2
           /usr/bin/ccache -s
         env:
           CCACHE_DIR: /tmp/ccache
--- a/src/openassetio-python/module/CMakeLists.txt
+++ b/src/openassetio-python/module/CMakeLists.txt
@@ -103,83 +103,81 @@ endif ()
 #-----------------------------------------------------------------------
 # Create deployment targets
 
-if (Python_Interpreter_FOUND)
-    # Get path to Python executable in the created venv.
-    if (WIN32)
-        set(venv_python_exe "${CMAKE_INSTALL_PREFIX}/Scripts/python.exe")
-    else ()
-        set(venv_python_exe "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/python")
-    endif ()
-
-    # Download core setuptools dependencies so we don't always need an
-    # online connection to local install OpenAssetIO. Need to cache the
-    # build tools explicitly due to build isolation, see:
-    # https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#build-process
-    # This also includes runtime dependencies of packages in the
-    # resources directory, whose install will otherwise fail due to
-    # --no-index.
-    set(python_deps_dir "${CMAKE_BINARY_DIR}/dependencies/python")
-    add_custom_target(
-        openassetio-python-download-deps
-        COMMAND ${CMAKE_COMMAND} -E echo -- Downloading core Python libraries
-        COMMAND
-        ${Python_EXECUTABLE} -m pip download
-        --retries 0 --timeout ${OPENASSETIO_PYTHON_PIP_TIMEOUT} --find-links ${python_deps_dir}
-        --destination-directory ${python_deps_dir}
-        wheel==0.37.1 setuptools==49.0
-        --requirement "${PROJECT_SOURCE_DIR}/resources/codegen/requirements.txt"
-    )
-    # Create a Python environment in the install directory. If the venv
-    # already exists then this is a no-op.
-    add_custom_target(
-        openassetio-python-venv
-        COMMAND ${CMAKE_COMMAND} -E echo -- Creating Python environment in ${CMAKE_INSTALL_PREFIX}
-        COMMAND ${Python_EXECUTABLE} -m venv ${CMAKE_INSTALL_PREFIX}
-        # Install `wheel` so that setuptools can build OpenAssetIO.
-        COMMAND
-        ${venv_python_exe} -m pip install --no-index --find-links ${python_deps_dir}
-        wheel setuptools
-    )
-    add_dependencies(openassetio-python-venv
-        # Ensure dependencies have been downloaded for constructing the
-        # venv.
-        openassetio-python-download-deps)
-
-    # Install OpenAssetIO into the venv.
-    add_custom_target(
-        openassetio-python-py-install
-        COMMAND ${CMAKE_COMMAND} -E echo -- Installing OpenAssetIO
-        COMMAND
-        ${venv_python_exe} -m pip install
-        # For speed, build from the venv rather than copying the whole
-        # project to a temporary environment. Note that this is a pip
-        # feature, setuptools also has its own build cache (see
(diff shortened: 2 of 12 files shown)


tests/CMakeLists.txt
```diff
@@ -1,6 +1,37 @@
 find_package(Catch2 2 REQUIRED)
 find_package(trompeloeil 42 REQUIRED)
 
+# Add a CTest fixture target that runs a CMake build target.
+#
+# Reuses the target name for both the test name and fixture name.
+function(openassetio_add_test_fixture_target target_name)
```
foundrytom, line 7: [question] By introducing these helpers, do we run the risk of making this harder to understand for newcomers - as it is no longer vanilla CMake? They don't contain a lot of logic. How likely is it that we will need to change what they do over time? (I'm just thinking back to how opaque previous projects have been when they become 50% 'custom functions' vs the build systems native calls).

foundrytom (commented):
I'll need to leave proper comments to someone with a better understanding and stronger opinions on CMake. A high-level Q to start :)

tests/CMakeLists.txt
feltech: :thinking: It's a reasonable concern. 

I had hoped that ensuring we use an `openassetio_` namespace would add some clarity at least on what functions are added by us. 

There is an example in this PR of having to change what they do (adding a label), which would have been a larger error-prone refactor without these.  Hard to predict how likely it is we'll need to change these functions going forward, though.

hartmutbehrens-foundry: If it helps to address the concern, this is from the "Functions and Macros" chapter of Professional CMake (written by one of the authors of cmake):

> Looking back on the material covered in this book so far, CMake’s syntax is already starting to look a lot like a programming language in its own right. It supports variables, if-then-else logic, looping and inclusion of other files to be processed. It should be no surprise to learn that CMake also supports the common programming concepts of functions and macros too. Much like their role in other programming languages, functions and macros are the primary mechanism for projects and developers to extend CMake’s functionality and to encapsulate repetitive tasks in a natural way.

foundrytom: A reference from USD: https://github.com/PixarAnimationStudios/USD/blob/release/pxr/base/tf/CMakeLists.txt

feltech:
I've added 619c5087 to reintroduce the `openassetio-python-venv` build target, so that the instructions in `BUILDING.md` are once again valid.  Note that this doc is to be rewritten somewhat in #624 .

.github/workflows/code-quality.yml
```diff
@@ -98,4 +98,4 @@ jobs:
           --preset sanitize
 
       - name: Build and test
-        run: ctest -VV --test-dir build
+        run: ctest -VV --test-dir build --parallel 2
```
cmunro-foundry, line 101: Would it not be better to determine the ideal process count from the cmake, like this in [ProcessorCount cmake docs](https://cmake.org/cmake/help/latest/module/ProcessorCount.html)

```
include(ProcessorCount)
ProcessorCount(N)
if(NOT N EQUAL 0)
  set(CTEST_BUILD_FLAGS -j${N})
  set(ctest_test_args ${ctest_test_args} PARALLEL_LEVEL ${N})
endif()
```

feltech: It would be great if the process count could be automatic.  How could we pass along the `ctest_test_args` variable to ctest, though?  I'll do some doc hunting.

feltech: Ah, the docs you linked use `ctest -S`.  We could add such a script as a convenience, but the ultimate would be to simply run `ctest` and have the `--parallel` determined automatically.

feltech:
Added 509e6534 to upgrade pip. The latest pip versions avoid copying the build directory when creating wheels (i.e. when installing the pure Python component).

This was causing Windows CI breaks due to the now-parallel `ctest`, whereby the C++ component was still being built when `pip` wanted to copy the build directory, and Windows has rather strict file permissions preventing copying whist the build was in progress.

I have (so far) re-run CI 4 times with no error.  Previously the failure happened on two consecutive builds (though not the build prior, when the failure could also have occurred).

src/openassetio-python/module/CMakeLists.txt
```diff
@@ -159,15 +154,21 @@ if (Python_Interpreter_FOUND)
         --no-index --find-links ${python_deps_dir}
         ${PROJECT_SOURCE_DIR}
     )
-    add_dependencies(openassetio-python-py-install
```
foundrytom, line 162: [fix] Removing these dependencies breaks using these targets as part of a build process, as the only dependencies are introduced by `ctest`. Eg: `cmake --build build --target openasetio-python.install` no longer works on a clean build, if you just wanted a python install complete with cmodule, etc:
```
-- Installing OpenAssetIO core Python component
/bin/sh: /Users/tom/dev/OpenAssetIO/build/dist/bin/python: No such file or directory
```
  I think we need to keep this working as a general build system with dependencies between components properly tracked, without having to use tests as the trigger.

foundrytom (changes requested):

src/openassetio-python/module/CMakeLists.txt
foundrytom: Capturing off-line discussions - can we have some kind of prefix/suffix that marks these internal leaf tasks as "do not use" for people inspecting the list of available cmake targets?

feltech:
Added a note about the imminent removal/refactor of the contentious `openassetio-python-deploy` target in 8410b638

foundrytom (approved):
Thanks for those extra changes, feel free to squash and merge at your leisure (once my silly Release notes fixes are in) 🙄

PR #620, opened 2022-09-13
[Core] Fix `RetainPyArgs` for const ref args

feltech:
Discovered during work on #608

`RetainPyArgs` is a utility to decorate C++ functions such that certain `shared_ptr` arguments are swapped out for equivalent `PyRetainingSharedPtr` arguments. These decorated functions are then used for Python bindings in lieu of the original functions. A custom Pybind type caster then detects `PyRetainingSharedPtr` arguments and ensures any associated Python objects are not destroyed prematurely.

The `RetainPyArgs` helper was using strict type matching when comparing function arguments against its `PtrsToPyRetain` (`shared_ptr`) list. This meant, for example, that a `const shared_ptr<SomeType>&` argument would not match because of the const reference qualifiers.

The decorated function returned from `RetainPyArgs::decorator` would then be missing the expected `PyRetainingSharedPtr<SomeType>` argument(s), so the custom Pybind type caster wouldn't trigger on that argument to extend the lifetime of the associated Python instance.

So ensure the argument type is `std::decay`ed, to remove any qualifiers, before checking for a match.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -68,6 +68,14 @@ v1.0.0-alpha.x
   provided is compatible with C++ trait views.
   [#605](https://github.com/OpenAssetIO/OpenAssetIO/pull/605)
 
+- Fixed the `RetainPyArgs` Python binding helper to
+  work with functions that take `shared_ptr`s by const reference, as
+  well as by value. This affected the Python bindings of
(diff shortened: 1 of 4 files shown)


foundrytom (changes requested):
Good detective work! 🕵️  Just needs release notes I think.

Issue #612, opened 2022-09-08
[CI] Fix usage of `actions/cache` for conan packages


feltech:
# What

We occasionally see conan recipes being reinstalled/rebuilt.

Until now we've assumed that it's because a newer recipe revision was posted upstream. However, conan ignores the remote if the package is already installed (unless `-u`/`--update` is given). So a reinstall must mean the cache has disappeared.

We should understand why the cache disappears, and ideally prevent it. 

# Why

This would mitigate issues we have occasionally, whereby conan reinstalls a package whose recipe has updated such that the install fails, and so we're blocked on CI. This problem also spawned #551.

We should still update our conan packages, but on our schedule, not because we're forced to due to mysterious cache eviction logic.

For info: this was discovered whilst investigating #602, which led to the discovery that the cache [is immutable](https://github.com/actions/cache/blob/main/workarounds.md#update-a-cache), so any changes are never persisted. This may be relevant for any fix.

feltech:
It looks like this happens because the first run of any new PR starts with no conan cache for that branch. It [will fall backon the parent branch's cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) , but we don't run the tests on `main` so there is no cache to fall back on.

foundrytom:
It'd also be good to improve/remove the work done in #695 so that it does the minimal amount of redundant work - eg: just populates the build cache.

foundrytom:
And also factor in #689

feltech:
This has been fixed by subsequent work e.g. see linked commits and #680

PR #621, opened 2022-09-13
[CI] Add ccache to speed up linter/sanitizer build

### LINKED ISSUE #613: [CI] Use `ccache` for Github actions
# What

Make use of `ccache` in those Github Actions that must build C++. In particular, the "Code quality / C++ linters" and "Code quality / Sanitizers" actions, which are (currently) the biggest CI bottleneck.

# Why

Following on from https://github.com/OpenAssetIO/OpenAssetIO/issues/602#issuecomment-1241054680, a quick test finds that `ccache` can save us approximately 1m30s for the "C++ linters" job and approximately 2m for the "Sanitizers" job, if there are no code changes.

# Notes

We must use [this strategy](https://github.com/actions/cache/blob/main/workarounds.md#update-a-cache) to use `ccache` with `actions/cache`.

Executing `ccache -s` before and after building gives nice summary statistics, so we can see in the logs how useful our cache is.

feltech:
Closes #613. `ccache` can save us ~30-50% in build step times for the C++ linter and sanitizer CI jobs.

Use the `restore-keys` trick to simulate updating a Github `actions/cache` cache, which is technically immutable. I.e. restore the most recently used cache by giving a prefix match to `restore-keys`, and generate a whole new cache to save any updates by giving the prefix plus run id (i.e. a unique key per run) as the `key`.

This will rapidly fill our cache quota, but the Github cache evicts based on least-recently used cache keys, which works for this use case.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -78,7 +78,8 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/conan
-          key: ubuntu-20.04-conan
+          key: ubuntu-20.04-conan-${{ github.run_id }}
+          restore-keys: ubuntu-20.04-conan-
 
       - name: Install dependencies
         # Configure the system and install library dependencies via
@@ -127,7 +128,8 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/conan
-          key: ubuntu-20.04-conan
+          key: ubuntu-20.04-conan-${{ github.run_id }}
+          restore-keys: ubuntu-20.04-conan-
 
       - name: Install dependencies
         # Configure the system and install library dependencies via
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -54,11 +54,24 @@ jobs:
       - name: Check Python formatting
         run: black --check .
 
+  # Note: in order to keep an `actions/cache` cache up to date, we must
+  # use the approach detailed in
+  # https://github.com/actions/cache/blob/main/workarounds.md#update-a-cache
+  # i.e. load the most recently created cache that matches a prefix,
+  # then create an entirely new cache with every run.
+
   cpp-linters:
     name: C++ linters
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
+
+      - name: Cache ccache cache
+        uses: actions/cache@v2
+        with:
+          path: /tmp/ccache
+          key: ubuntu-20.04-ccache-lint-${{ github.run_id }}
+          restore-keys: ubuntu-20.04-ccache-lint-
 
       - name: Cache conan packages
         id: cache-conan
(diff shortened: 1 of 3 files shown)


.github/workflows/code-quality.yml
```diff
@@ -41,6 +41,13 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Cache ccache cache
+        uses: actions/cache@v2
+        with:
+          path: /tmp/ccache
+          key: ubuntu-20.04-ccache-lint-${{ github.run_id }}
+          restore-keys: ubuntu-20.04-ccache-lint-
```
foundrytom, line 49: [question] Do we not need to do this with Conan as it already namespaces its folder with hashes?

foundrytom:
@feltech - any reason not to do this for all builds?

feltech:
> @feltech - any reason not to do this for all builds?

We could, but

* It wouldn't save us any more CI time overall, based on my experiments, since the C++ linter and sanitizer checks remain the bottleneck even with all the optimisations.
* The remaining builds are the unit test checks (IIRC) for Linux, MacOS and Windows.  I imagined that the non-Linux OSes might be less trivial to configure for ccache (though I didn't try), so was reluctant to expend the effort if it wouldn't give us much benefit.
* I reasoned that it might be a good thing to have a build that didn't use ccache anyway, to help catch any ccache-specific false positives/negatives that might crop up

.github/workflows/code-quality.yml
feltech: Not quite sure what you mean. We don't need to do this with conan because in theory it never changes, unlike build artifacts which change with most PRs.

:thinking: Perhaps we should do this for conan as well, though, otherwise if we add a new package to our conanfile then it will always be reinstalled, until the cache is evicted (which mysteriously happens occasionally, see #612)

I should at least add a comment in the yml to explain what this is doing.

foundrytom: Yeah, that last situation was what I was thinking. But I had noticed there are some hash prefixed dirs inside of the connan folder anyway, thought Id seen that changing when the conanfile changes... hadn't had time to look into what they were used for.

feltech: I added a comment to the yml in 7bd9b696.

I updated the conan cache to use the same strategy as the ccache cache in 6fc03219.

feltech: Frustratingly, the CI checks are not running yet because of a merge conflict (presumably) :roll_eyes:

foundrytom: Yeah, feel free to rebase/squash ready for merge

feltech:
> Is it expected that this is still 6m, with a 73% cache hit rate?
> 

There is a lot of variability, unfortunately.  It might be a particularly slow run - I notice the other "Tests" checks are quite slow too, (they're usually approx 4m with Ninja IIIRC).

There were quite a few cache misses too - approx 50%.  The cache hit/miss rate in the log is all-time, rather than per run, so you have to look at the raw before/after numbers to work it out per run.

I wouldn't worry about it, but to be sure I can trigger a few more builds of this PR, if you like.

foundrytom:
> I wouldn't worry about it, but to be sure I can trigger a few more builds of this PR, if you like.

Cool, thanks. The rebase run is 4m so that's a great improvement! Nice one.

foundrytom (approved):

PR #633, opened 2022-09-22
[Docs] Add DR for project structure refactor

### LINKED ISSUE #628: [Investigation] Modular project file structure
## What

Produce a DR that describes a revised project filesystem structure that better reflects/supports the modularity of the code base.

## Why

The current structure has grown from a pure python project, to a pure python project with some C++ components to a C++ project with optional Python.

The current structure inadvertently has slightly spider-web dependencies that make decoupling the optional components harder, especially with those in `resources`.

Timebox: 1 day

## ACs

- DR outlining chosen structure with rationale.
- Optional (anything in resources, C, Python) components should be self contained.
- Project structure is independent of language/tooling choices used.
- Structure should facilitate the removal of an optional component without needing to update its peers (other than upstream dependencies). 
- Update #655 with tasks for implementing the DR

feltech:
Closes #628

### DIFF
--- a/decisions/DR013-Project-structure.md
+++ b/decisions/DR013-Project-structure.md
@@ -1,0 +1,352 @@
+# DR013 Project structure
+
+- **Status:** Proposed
+- **Impact:** Low
+- **Driver:** @feltech
+- **Approver:** @tomfoundry
+- **Outcome:**
+
+
+## Background
+
+OpenAssetIO started life as fork of the pure Python FnAssetAPI. One of
+the major improvements we have embarked upon is to migrate the
+functionality to a C++ core library, with optional Python bindings.
+
+Performing this migration incrementally has meant that the project
+structure in general, and repository directory structure in particular,
+still largely reflects the primarily Python-based origin.
+
+The project is logically split into weakly coupled components and
+subcomponents, which should ideally be reflected in the file structure,
+to aid navigation and potential extraction into separate repositories.
+
+Instead, the project still reflects a typical Python project, with
+top-level `python`/`tests` directories and associated
+`setup.py`/`pyproject.toml` alongside. This has caused some confusion to
+early adopters who, understandably, assume that `pip install` is the
+primary build mechanism of the project, when in reality the Python
+component is only a small (and optional) slice.
+
+Despite components being logically weakly coupled, there are some
+potentially unnecessary dependencies between them. For
+example, CMake code under `tests/openassetio-python` is currently
+responsible for configuring `ctest` targets for `resources/examples`.
+
+In addition, we are concerned about the clarity of which parts of the
+API surface area across all components should strictly trigger (major)
+version changes, versus being unsupported/example code.
+
+We therefore wish to explore whether an alternative project directory
+structure could help adopters get a clearer at-a-glance picture of the
+project make-up and dependency tree, whilst tidying up some technical
+debt.
+
+
+## Relevant data
+
+### Requirements
+
+Requirements of any proposed structure:
+* Optional (anything in `resources`, C, Python) components should be
+  self-contained.
+* Project structure is independent of language/tooling choices used.
+* Structure should facilitate the removal of an optional component
+  without needing to update its peers (other than upstream
+  dependencies).
+
+### Prior art
+
+Looking at a few of the repositories hosted on the [ASWF
+Github](https://github.com/orgs/AcademySoftwareFoundation/repositories)
+org.
+
+#### [OpenTimelineIO](https://github.com/AcademySoftwareFoundation/OpenTimelineIO)
+
+* C++/Python project. Multiple components.
+* `setup.py` at repo root that drives `CMakeLists.txt` at repo root.
+* Code grouped by component under `src` directory.
+* Python-related code slightly scattered under `src`
+  - Pure Python under `opentimelineview` and
+    `py-opentimelineio/opentimelineio`.
+  - Python bindings under `py-opentimelineio/*-bindings`.
+* Tests under repo-root `tests` directory.
+  - Mostly Python test sources
+  - (Minimal) C++ test sources (with CMakeLists) alongside Python tests.
+* Docs (Sphinx) under repo-root `docs` directory.
+* Third-party code under `src/deps`
+* Separate `contrib` directory for plugins, with independent tests.
+
+#### [MaterialX](https://github.com/AcademySoftwareFoundation/MaterialX)
+
+* C++/Python/Javascript project. Multiple components.
+* Only `CMakeLists.txt` at repo root.
+* C++ code grouped by component under repo-root `source` directory.
+  - Including C++ components for Python and Javascript bindings.
+* Pure Python code under repo-root `python` directory. Similarly for
+  `javascript`.
+  - `python` directory holds a `setup.py.in` template, rendered by CMake
+    into the install tree.
+* Tests in `MaterialXTest` directories, split between `source` (C++),
+  `python` and `javascript` directories.
+* Docs (Doxygen, PDF, Markdown) under repo-root `documents`.
+* Build system helpers under repo-root `cmake/modules`.
+* Examples under repo-root `resources`.
+* Core material library under repo-root `libraries`.
+
+#### [OpenVDB](https://github.com/AcademySoftwareFoundation/openvdb)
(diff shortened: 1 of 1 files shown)


decisions/DR013-Project-structure.md
```diff
@@ -0,0 +1,441 @@
...
+│   │       │   │   └── BasicAssetLibrary
+│   │       │   ├── tests
+│   │       │   │   └── resources
+│   │       │   └── pyproject.toml
+│   │       └── SampleAssetManager
+│   │           ├── python
+│   │           │   └── SampleAssetManager
+│   │           └── tests
+│   └── perftest
+├── src
```
foundrytom, line 334: [fix] Might be good to use an alternate word here - as we use `src` for c++ several layers down.

foundrytom (dismissed):
Like it 👍 just a few superficial suggestions that might make it easier when we discuss later.

decisions/DR013-Project-structure.md
elliotcmorris: @feltech and I went over this, and think leaving the doxygen folder where it is makes the most sense.

There were three viable options we considered : 
1. Move `doxygen` folder into `src` dir, treat it as an individual component
2. Split `doxygen` folder/docs build into individual components, and have a seperate `doxygen` folder under each component.
3. Leave where it is

1 - This makes some sense through an architectural lens. Like the python with pybind, docs are just another target using doxygen as a generator, so could be seen as a seperate component. However, the main argument against this, and in favour of keeping the folder in place, is simply intuitiveness. One would expect docs related items to be in `docs`, not `src`

2 - From the perspective of attempting to keep all modules trivially extractable, a fully developed forumulation may find `doxygen` folders in each component, (ie, generate different docs for core, c-api, and python.) This is challenging to at the moment however, due to the neccesity of generating component level documentation, and jigging up the build system.

3 - This makes the most sense to us, mainly due to intuitiveness, although we agree a super ideal system might use option #2, eventually.

decisions/DR013-Project-structure.md
```diff
@@ -0,0 +1,441 @@
...
+* Move tests under the component that they are testing.
+  - Trickier for pure Python tests that cover multiple components. So
+    simply place alongside the Python package, i.e. in
+    `src/openassetio-python/python/tests`
+* Place `_openassetio_test` Python C++ extension module sources
+  alongside the `_openassetio` module.
+* Conform directory structure for C++ code with `include`, `src` and
+  `tests` subdirectories.
+* Place docs subdirectories under the repo-root `doc` directory (with
+  the addition of a `doxygen` subdirectory).
```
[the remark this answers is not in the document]
elliotcmorris: @feltech and I went over this, and think leaving the doxygen folder where it is makes the most sense.

There were three viable options we considered :

Move doxygen folder into src dir, treat it as an individual component
Split doxygen folder/docs build into individual components, and have a seperate doxygen folder under each component.
Leave where it is
1 - This makes some sense through an architectural lens. Like the python with pybind, docs are just another target using doxygen as a generator, so could be seen as a seperate component. However, the main argument against this, and in favour of keeping the folder in place, is simply intuitiveness. One would expect docs related items to be in docs, not src

2 - From the perspective of attempting to keep all modules trivially extractable, a fully developed forumulation may find doxygen folders in each component, (ie, generate different docs for core, c-api, and python.) This is challenging to at the moment however, due to the neccesity of generating component level documentation, and jigging up the build system.

3 - This makes the most sense to us, mainly due to intuitiveness, although we agree a super ideal system might use option #2, eventually.

decisions/DR013-Project-structure.md
```diff
@@ -0,0 +1,441 @@
...
+currently stands, our `setup.py` is very thin and most of the heavy
+lifting is performed by CMake.
+
+At a high level, we theorise a lot of the confusion comes from the
+project looking like a Python-driven project at first glance, rather
+than a CMake-driven project.
+
+We therefore wish to explore whether an alternative project directory
+structure could help adopters get a clearer at-a-glance picture of
+the project make-up and dependency tree.
```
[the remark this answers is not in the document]
feltech: Added a note in 30a96dfe

decisions/DR013-Project-structure.md
feltech: A top-level `src` directory is common across nearly all the reviewed examples from the ASFW Github org. Intuitively, I feel like that's the location where readers will expect to find our source code. 

Also if we assume all the components, other than core, are extracted, then the core would remain as the only `src` directory.

One option could be to do like OpenVDB and put all components at the top-level, rather than under an `src` subdirectory.  But that pollutes the root of the repo and is inconsistent if we wish to add other groupings (e.g. a `utils`, alongside `src`, with optional convenience utilities).

Another option would be to rename descendent `src` directories.  But then again, devs would expect to find source code under `src`.

Any suggestions?

decisions/DR013-Project-structure.md
```diff
@@ -0,0 +1,457 @@
...
+Estimated cost: Medium
+
+
+## Outcome
+
+```mermaid
+flowchart
+    Start --> What{What are\nyou adding?}
+
+    What -->|Docs| WhatDocType{Who is doc for?}
```
foundrytom, line 432: [fix] Change to simply "Is it docs - put it in `/doc`"

decisions/DR013-Project-structure.md
```diff
@@ -0,0 +1,441 @@
...
+│   └── src
+├── python
+│   └── openassetio
+│       ├── _core
+│       ├── hostApi
+│       ├── managerApi
+│       ├── pluginSystem
+│       ├── test
+│       │   └── manager
+│       └── traits
```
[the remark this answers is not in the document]
feltech: Simplified (after rebased on latest main) in a17aa50c

decisions/DR013-Project-structure.md
feltech: Simplified in a17aa50c

decisions/DR013-Project-structure.md
```diff
@@ -0,0 +1,350 @@
...
+  unmodified ASWF Docker image, which is not a compelling reason.
+- Duplication of `if (OPENASSETIO_ENABLE_<COMPONENT>)` under `tests`
+
+Estimated cost: Small
+
+
+### Option 2
+
+* Place pure Python component under `src/openassetio-python/python`
+  directory, excluding `setup.py` and `pyproject.toml`.
```
feltech, line 219: [question] This used to say "**including** `setup.py`", but I changed this based on recent learnings.  As mentioned in the fixup commit (3a79b6fd)

> `pip install .` won't work out of the box otherwise, since the directory containing `setup.py` is copied for the build process, breaking any relative paths.

elliotcmorris: It still seems desirable to have `setup.py` able to sit inside the python subdirectory, for all the same reasons of project separation that we discussed.

I'm not sure we need to go to any effort to do it right now though, as further project seperation may never occur. We could log an issue for it as potential future work.

decisions/DR013-Project-structure.md
elliotcmorris: Chiming in to agree with David, we discussed this and don't think the reuse of `src` in the directory tree is misleading. Resolving this conversation in the absence of Tom.

decisions/DR013-Project-structure.md
feltech: Added #728, how's that?

decisions/DR013-Project-structure.md
```diff
@@ -0,0 +1,350 @@
...
+
+OpenAssetIO started life as fork of the pure Python FnAssetAPI. One of
+the major improvements we have embarked upon is to migrate the
+functionality to a C++ core library, with optional Python bindings.
+
+Performing this migration incrementally has meant that the project
+structure in general, and repository directory structure in particular,
+still largely reflects the primarily Python-based origin.
+
+This has begun to cause some confusion to early adopters who,
```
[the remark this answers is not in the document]
feltech: Reworded in 93c39c07 @elliotcmorris how does that read?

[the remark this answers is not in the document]
elliotcmorris: Aye, better, no longer reads as the overriding motivation.

elliotcmorris (approved):

Issue #630, opened 2022-09-21
Update `setup.py` to build c module (`pip install .`)


foundrytom:
## What

`pip install .` should produce a fully working python package (including c components) without additional steps.

## Why

This is a pre-requisite for packaging for distribution, and is an expectation from a project with a `setup.py` file.

## Acceptance Criteria

- `pip install .` installs a fully working `openassetio` python package as per standard `pip` behaviour.
- python cmodule is static linked to dependencies
- No additional build steps are required aside from, in the first instance, assuming the starting condition that:
  - build dependencies are satisfied in the current environment (ie: cmake, conan, compiler)
  - cmake find package works to the "desired" dependencies for the user (eg: use toolchain ENV, or ...)
  - Instructions can include simple copy-paste one liners for happy path to achieve the above with conan central or similar.
- additional integration test that pip-installs the result and runs pytest test
  - shouldn't run with existing pytest/ctest tests
  - should be runnable by self-contained means without knowing "magic steps" (cross-platform) (eg. tox, py script).
   -  makes a new venv from build interpreter, installs using `pip install .`, runs pytest against that venv.
- BUILDING.md and other developer guidelines updated
- Follow the `pyproject.toml`/non legacy route described [here](https://pip.pypa.io/en/stable/reference/build-system/).

## Out of scope

- Significant build doc rewrites other than fixing up existing commands that are no longer relevant
- Significant customization of `setup.py` cmake invocation
- Adjusting existing test infrastructure invocation etc..
- Improvements to Window build process (eg: avoiding the need for `vcvars64.bat` or similar).

foundrytom:
https://gist.github.com/JeanChristopheMorinPerso/f5cd307243a8164b1da9ebbf63152c93

and

https://setuptools.pypa.io/en/latest/userguide/extension.html#supporting-sdists-and-editable-installs-in-build-sub-commands

Have much useful info

Issue #624, opened 2022-09-14
Update `BUILDING.md`


foundrytom:
## What

Update `BUILDING.md` to best practices.

## Why

The current version was written before we added C-based tests, and still lists manual steps that won't fully run all tests (now we have C/C++ tests as well as python ones).

foundrytom:
For `BUILDING.md`

- Project Components (indicating optional)
- Quick steps for building, in terms of CMake only (cmake, cmake --build, ctest)n then describe what's in `build/dist`.
- Using Vagrant (just as far as vagrant ssh and configure with conn)
- Detailed description of build steps for debugging (with the details about what's actually happening)

elliotcmorris:
Further discussions, to consider.

In `BUILDING.md`

- "for reference, ... we use conan" - moved above dependency specifics, reworded to be a tad less skippable.
- Link directly to conan docs on `find_package` in section ` packages to be available to the CMake build system.` To make it clearer that this is something you need to consider.
- Consider having collapsible "conan instructions" sections. (--toolchain problem)
    <details>
        <summary>Like this</summary>
        Collapsible summary here
    </details>

- Have vagrant and docker steps at the bottom (as they are)
- Have table of contents, or even just links, so people can jump directly to instructions they're looking for (like docker, etc)

- Don't mention tests/linting etc in initial steps, have that setup in later sections. Initial setup should have absolutely no non-essential fluff.
- "Running tests" section seems pretty solid.


In `README.md` 
- Replace docker section in README quickstart with a link or short snippet. (duplicated from BUILDING.md)

foundrytom:
@elliotcmorris love the collapsible idea!

feltech:
> **Warning**
> This is a warning

feltech:
Github renders bold text "warning" as a special case

> **Warning**
> This is a warning

Similarly for 

> **Note**
> This is a note

foundrytom:
> Github renders bold text "warning" as a special case
> 
> > Warning
> > This is a warning
> 
> Similarly for
> 
> > Note
> > This is a note

Good find!

foundrytom:
> * Consider having collapsible "conan instructions" sections. (--toolchain problem)
>     
>         Like this
>         Collapsible summary here

A Q on the topic of whether adding advanced Markdown breaks readability for raw view on ASWF slack:
 https://academysoftwarefdn.slack.com/archives/CKAH4EU4Q/p1664535553513889

[...]

PR #649, opened 2022-10-07
Decouple CMake from Python somewhat

### LINKED ISSUE #629: Don't use `setup.py` in CMake
## What

De-couple CMake and `setup.py`, so that `setup.py` can be used as a top-level entry point for installation/management of the python component of the API.

## Why

Presently, `cmake` creates a virtual environment, and uses `setup.py` to install `openassetio` into the venv. This complicates python packaging for distribution where it is expected that `setup.py` is the entrypoint for pip et al.

We need a solution that enables simple, local builds that create working python packages, as well as an intuitive install/build process for the python components.

## Acceptance Criteria

- CMake no longer uses the top level setup.py so that setup.py can be invoked to build a python distributable by pip.

## Notes

What needs to happen to make a python package:

 - `libopenasstio-core.so`/`.a` compiled (cmake) 
 - cmodule (`_openassetio.so`) compiled (cmake)
 - `_openasstio.so` copied to python package root

At this point you could extend `PYTHONPATH`/`LD_LIBRARY_PATH` and `import openassetio`, eg: in an embedded application that uses its own interpreter, and links to openassetio from C++.

If this needs to be standalone, ie, with `pip install` and `python`, `libopenasetio-core` needs to be linked by `_openassetio`. Either

- `libopenasetio-core.so` installed such that it is discoverable on the library search path. 
     -  `LD_LIBRARY_PATH`, application's transient `RPATH`, system paths, et al. can be used as a fallback (as above).
     - On Unix, `_openassetio`'s `RPATH` is currently set to a relative path that assumes the shared library is placed under the `lib` ancestor of `site-packages` (matching the CMake install tree). 
     - On Windows we currently have some rather hacky CMake code that copies `libopenasetio-core.dll` into `site-packages` in the CMake install tree. Windows has not been properly considered, beyond just getting CI running the tests.
- `libopenasetio-core.a` statically linked into `_openassetio.so`. Obviously we must then be careful that `libopenasetio-core` has not already been linked into the application through some other route.
## Proposed changes

 - CMake install command manually populates the `dist` structure as before.
   - When `OPENASSETIO_ENABLE_PYTHON_SRC_BUNDLING` is `ON` (default) then also copies `.py` source via CMake's `install(FILES)` (instead of using `pip`/`setup.py`).
   - ~Option to static link `_openassetio.so`~ (done)
- Tests run with modified `PYTHONPATH`.
- `venv.create` build (and CTest fixture) target to create test venv, conditionally built-up via `add_dependencies` as the CMake source tree is traversed. (venv isn't `dist`, can be elsewhere, and `opeanssetio` is still referenced via `PYTHONPATH`)
  - Initially replaces the related fixtures in ctest.
  - `OPENASSETIO_ENABLE_PYTHON_TEST_VENV` option that, when `OFF` (default) causes `OPENASSETIO_PYTHON_VENV_ACTIVATE` to be a no-op and skips adding `venv.create` build target as a fixture.
  - Remove `download-deps` targets and hence remove `--no-index --find-links` from `pip` install step(s).
- ~CI jobs to run Python tests with both statically linked and dynamically linked versions of  `_openassetio.so`.~

feltech:
Closes #629.

* For deployment, use CMake to install pure Python files to the install tree, rather than `pip`. Hence remove the unused and incomplete `setup.py`, for now.
* For tests, only create a Python virtualenv if an `OPENASSETIO_ENABLE_PYTHON_TEST_VENV` option is set. Otherwise, simply expect the user to have appropriately configured an external environment for running tests (installed test dependencies and made the `openassetio` package in the install tree discoverable).
* For tests, roll all Python virtualenv creation targets into a single, easily excludeable, target - useful for subsequent runs where the environment already exists.
* CMake presets by default enable `OPENASSETIO_ENABLE_PYTHON_TEST_VENV`

# WIP notes

* Build instructions updated, but not yet manually tested.
  - Need to remove redundant `pip install [...]/requirements.txt` for test dependencies, if virtualenv fixture is enabled(?).
* Should some of these changes be split into separate PRs?
* ~~Should we have deleted `setup.py`, only for it to return very soon?~~ - yes
* ~~Should we add new CI checks to check test runs without an automatically created venv?~~ - covered by change to always run with augmented PYTHONPATH even if using a test venv
* ~~Should we default-enable `OPENASSETIO_ENABLE_PYTHON_TEST_VENV` in CMake presets, despite it being default-disabled when presets are not used?~~ - yes
* Component tag on some commits should perhaps be `[Tests]`
* Need to review some code comments around changed lines.


foundrytom:
Notes from standup:

- Move test venv from `build/dist` to `build/test-venv` (so `dist` _only_ contains distribution artifacts)
- Always extend python path to `dist`
- Don't `pip install` codegen, ensure CLI tests invoked with `sys.executable -m openassetio_codegen` vs the wrapper (will need `requirements.txt` to install deps to `test-venv`), extend `PYTHONPATH` to resources dir for tests.
- Issue to add test coverage for `setup.py` in codegen.
- Any commit that touches cmake/etc... tagged with `[Build]`

cmake/ThirdParty.cmake
```diff
@@ -114,36 +93,16 @@ if (OPENASSETIO_ENABLE_PYTHON)
     #-------------------------------------------------------------------
     # Target to create a Python virtual environment in the install tree.
 
-    # Download core setuptools dependencies so we don't always need an
-    # online connection to local install OpenAssetIO. Need to cache the
-    # build tools explicitly due to build isolation, see:
-    # https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#build-process
-    # This also includes runtime dependencies of packages in the
-    # resources directory, whose install will otherwise fail due to
-    # --no-index.
```
foundrytom, line 123: [fix] We should add a `RELEASE_NOTES.md` call out for the no-offline aspect I think, as it is quite a significant change.

cmake/ThirdParty.cmake
```diff
@@ -114,36 +93,16 @@ if (OPENASSETIO_ENABLE_PYTHON)
...
-    openassetio_add_pip_download_target(
-        openassetio.internal.python-venv.download-deps
-        "Downloading core Python libraries"
-        wheel==0.37.1 setuptools==49.0
-        # We must upgrade pip so that the build directory is not
-        # copied to a temporary directory, since doing so could cause
-        # Windows to fail with permissions errors when concurrently
-        # installing Python packages and building the C++ component.
```
foundrytom, line 131: [fix] I _think_ this comment block should move to the (new) L105? As it seems unrelated to offline support, and the actual update is done when we make the venv? 

(Also, whilst I remember, does this mean we need to document the above when using your own python env?)

BUILDING.md
```diff
@@ -174,7 +160,6 @@ cmake --build build
 cmake --install build
 cmake --build build --target openassetio-python-venv
 source build/dist/bin/activate
-pip install ./openassetio
 pip install -r openassetio/tests/python/requirements.txt
 pytest openassetio/tests/python
```
foundrytom, line 174: [plan] We should def make sure in #624 we switch this to `ctest` as we now have more than just `pytest` tests (we didn't when we wrote this before).

src/openassetio-python/CMakeLists.txt
```diff
@@ -5,3 +5,9 @@
 add_subdirectory(bridge)
 # Python C extension module making C++ types usable in Python.
 add_subdirectory(module)
+
+# Install OpenAssetIO pure Python into the venv.
+install(
+    DIRECTORY ${PROJECT_SOURCE_DIR}/python/openassetio
+    DESTINATION "${OPENASSETIO_PYTHON_SITEDIR}"
+)
```
foundrytom, line 13: [question] Do we need to be careful to only copy `.py` here? Can we filter it?

src/openassetio-python/CMakeLists.txt
```diff
@@ -5,3 +5,9 @@
 add_subdirectory(bridge)
 # Python C extension module making C++ types usable in Python.
 add_subdirectory(module)
+
+# Install OpenAssetIO pure Python into the venv.
+install(
```
foundrytom, line 10: [question] CMake noob Q - how do we then add this as a dependency later? So when we run the tests, we know this is definitely done?

BUILDING.md
```diff
@@ -77,12 +77,22 @@ binary artifacts are also made available to your Python environment.
...
 
 Testing is disabled by default and must be enabled by setting the
-`OPENASSETIO_ENABLE_TESTS` CMake variable. Assuming that the working
-directory is the repository root, we can configure the CMake build as
-follows
+`OPENASSETIO_ENABLE_TESTS` CMake variable.
+
+In addition, test fixtures can be configured such that a Python
+environment is created during the test run by setting
+`OPENASSETIO_ENABLE_PYTHON_TEST_VENV`. If this is `OFF` (default) then
```
foundrytom, line 84: [question] Im struggling to remember why we said `OFF` by default, rather than `ON` (I know we definitely had good reason at the time!). Once we've de-coupled the venv and `dist` might it make sense to be on by default?

BUILDING.md
```diff
@@ -77,12 +77,22 @@ binary artifacts are also made available to your Python environment.
...
+augmenting your `PYTHONPATH`) before running tests.
+
+Assuming that the working directory is the repository root, we can
+configure the CMake build as follows
 
 ```shell
-cmake -S . -B build -DOPENASSETIO_ENABLE_TESTS=ON
+cmake -S . -B build \
+  -DOPENASSETIO_ENABLE_TESTS=ON \
+  -DOPENASSETIO_ENABLE_PYTHON_TEST_VENV=ON
```
foundrytom, line 95: [question] What do we think about using a preset here instead? For simplez?

cmake/ThirdParty.cmake
```diff
@@ -60,21 +60,31 @@ if (OPENASSETIO_ENABLE_PYTHON)
...
     # Get path to Python executable in the created venv, and a way to activate
     # the venv itself.
-    if (WIN32)
-        set(OPENASSETIO_PYTHON_VENV_ACTIVATE "${CMAKE_INSTALL_PREFIX}/Scripts/activate.bat")
-        set(OPENASSETIO_PYTHON_VENV_EXE "${CMAKE_INSTALL_PREFIX}/Scripts/python.exe")
+
+    if (OPENASSETIO_ENABLE_PYTHON_TEST_VENV)
+        if (WIN32)
+            set(OPENASSETIO_PYTHON_VENV_ACTIVATE "${CMAKE_INSTALL_PREFIX}/Scripts/activate.bat")
+            set(OPENASSETIO_PYTHON_VENV_EXE "${CMAKE_INSTALL_PREFIX}/Scripts/python.exe")
```
foundrytom, line 67: [consider] Should this var still have `VENV` in the name now?

tests/openassetio-python/CMakeLists.txt
```diff
@@ -195,35 +195,41 @@ openassetio_add_pytest_target(
 #-----------------------------------------------------------------------
 # CTest setup fixtures
 
-# Fixture: create Python environment.
-openassetio_add_test_fixture_target(openassetio-python-venv)
+# Optional fixture dependency to create a Python environment in the
+# install tree.
+set(_venv_if_enabled )
```
foundrytom, line 200: [consider][nit-pick] trailing space

tests/openassetio-python/CMakeLists.txt
```diff
@@ -195,35 +195,41 @@ openassetio_add_pytest_target(
...
+    ${_venv_if_enabled}
 )
 
 # Examples tests.
 openassetio_add_test_target(openassetio.internal.resources-examples-pytest)
 openassetio_add_test_fixture_dependencies(
     openassetio.internal.resources-examples-pytest
     openassetio.internal.install
-    openassetio-python-venv
+    ${_venv_if_enabled}
```
foundrytom, line 232: [question] Do we need this for codegen too at this point?

foundrytom (commented):
Great stuff - this is going to be a lot simpler! 👍  These comments are from a quick look through - I'll save full practical testing until the things we chatted about earlier are in.

foundrytom:
Re the whole `codegen`/`pip install` malark, would it work if (ultimately, once #630 is done too), is it just a test option for "tests use pip" that switches from extending PYTHONPATH (for non venv/tdd workflows) to properly installing them into the test venv?

cmake/ThirdParty.cmake
feltech: I don't _think_ we have this problem any more, since we're no longer `pip install`ing our projects. Technically we could remove the `pip` upgrade, but it's still nice to pin versions to avoid nasty surprises.

foundrytom: Cool, we should keep it in mind when we add pip installing for tests back in again :)

src/openassetio-python/CMakeLists.txt
feltech: We can use `FILES_MATCHING PATTERN`, but is that overkill?  Currently, at least, I think we want everything.

src/openassetio-python/CMakeLists.txt
feltech: `install`ed things are added to the built-in `install` build target, which we wrap as `openassetio.internal.install`.  So when CMake installs the binary libraries, it also installs these files.

src/openassetio-python/CMakeLists.txt
foundrytom: Is there some danger of accidentally including `.pyc` or other artifacts from developer tooling? We used to end up with `egg-info` and such as well.  I guess the same argument as the fact we name includes/cpp vs glbobing in CMake?

foundrytom: Also, when we point `setup.py` to that dir, iirc, it will _only_ include `.py` without reconfiguration, so a setup.py assembled package could be different to the cmake assembled one?

BUILDING.md
feltech: Wasn't it that users are most likely to want to run the test suite against their Python environment, to test integration?

BUILDING.md
feltech: I'm less keen on that, since it adds another layer of indirection. Maybe a good idea for quick start though.

cmake/ThirdParty.cmake
feltech: Yeah I guess we can strip the `VENV` now, so the variable name works for both branches.

BUILDING.md
foundrytom: 🤔 Probably was wasn't it. I guess the Q is, if the Quickstart instructions always enable it - should we just enable it?

I could see the argument for it being off if the default instructions for the current early adopters left it so?

cmake/ThirdParty.cmake
foundrytom: Would probably help with readability at the call site too as it becomes clear that the python isn't always in a venv?

tests/openassetio-python/CMakeLists.txt
feltech: I put that to mean "explicitly blank" (seen that pattern elsewhere), but it works without the space, so happy to remove.

tests/openassetio-python/CMakeLists.txt
feltech: It's a transitive dependency of installing the codegen deps.  Should we make it explicit there too? Might help readability and does no harm

feltech:
> Re the whole `codegen`/`pip install` malark, would it work if (ultimately, once #630 is done too), is it just a test option for "tests use pip" that switches from extending PYTHONPATH (for non venv/tdd workflows) to properly installing them into the test venv?

Yeah that could work, but might get messy(er) with more branching logic in the CMake code.

tests/openassetio-python/CMakeLists.txt
foundrytom: Ahh cool, sorry that does make sense, I hadn't groked the implication, but I'm a CMake muppet. Leave as is!

foundrytom:
> > Re the whole `codegen`/`pip install` malark, would it work if (ultimately, once #630 is done too), is it just a test option for "tests use pip" that switches from extending PYTHONPATH (for non venv/tdd workflows) to properly installing them into the test venv?
> 
> Yeah that could work, but might get messy(er) with more branching logic in the CMake code.

Yeah. I guess I was wondering if the "happy place" would be when `ctest` uses a `pip install` by default (when you're not a contributor) such that we have the most e2e coverage (vs needing to run a separate job?)

cmake/ThirdParty.cmake
feltech: Added a note in 59357f4

CMakeLists.txt
```diff
@@ -180,6 +183,14 @@ endif ()
 # Enable unit tests.
 option(OPENASSETIO_ENABLE_TESTS "Create test targets" OFF)
 
+cmake_dependent_option(
+    OPENASSETIO_ENABLE_PYTHON_TEST_VENV
```
foundrytom, line 187: [fix] Can we add this to the status dump output when you run `cmake -S`?

tests/openassetio-python/CMakeLists.txt
```diff
@@ -171,133 +171,86 @@ openassetio_add_pytest_target(
...
 
 # Code generation tests.
 openassetio_add_test_target(openassetio.internal.resources-codegen-pytest)
 openassetio_add_test_fixture_dependencies(
     openassetio.internal.resources-codegen-pytest
     openassetio.internal.install
-    openassetio.internal.python.install
-    openassetio.internal.resources-codegen.install
-    openassetio.internal.resources-codegen-pytest.install-deps
+    openassetio.internal.resources-codegen.install-deps
```
foundrytom, line 255: [fix] I think maybe something is awol, that if I build with `-DOPENASSETIO_ENABLE_PYTHON_TEST_VENV=OFF`, then the codegen deps still get installed by ctest:

```
cmake ... -DOPENASSETIO_ENABLE_PYTHON_TEST_VENV=OFF
cmake --build  build
cmake --install build
cmake --build build --target openassetio-python-venv
. build/test-venv/bin/activate
ctest --test-dir build

    Start 1: openassetio.internal.install
1/7 Test #1: openassetio.internal.install ..........................   Passed    4.39 sec
    Start 2: openassetio.internal.core-cpp-test
2/7 Test #2: openassetio.internal.core-cpp-test ....................   Passed    0.14 sec
    Start 3: openassetio.internal.core-c-test
3/7 Test #3: openassetio.internal.core-c-test ......................   Passed    0.15 sec
    Start 4: openassetio.internal.resources-codegen.install-deps
4/7 Test #4: openassetio.internal.resources-codegen.install-deps ...   Passed    1.83 sec
    Start 5: openassetio.internal.python-pytest
5/7 Test #5: openassetio.internal.python-pytest ....................***Failed    0.24 sec
    Start 6: openassetio.internal.resources-examples-pytest
6/7 Test #6: openassetio.internal.resources-examples-pytest ........***Failed    0.24 sec
    Start 7: openassetio.internal.resources-codegen-pytest
7/7 Test #7: openassetio.internal.resources-codegen-pytest .........***Failed    0.24 sec
```

(ignore the fails)

cmake/ThirdParty.cmake
```diff
@@ -114,49 +96,50 @@ if (OPENASSETIO_ENABLE_PYTHON)
...
     # Target to create a Python environment in the install directory.
     # If the venv already exists then this is a no-op.
-    # Requires: openassetio.internal.python-venv.download-deps
     add_custom_target(
         openassetio.internal.python-venv.create
-        COMMAND ${CMAKE_COMMAND} -E echo -- Creating Python environment in ${CMAKE_INSTALL_PREFIX}
-        COMMAND ${Python_EXECUTABLE} -m venv ${CMAKE_INSTALL_PREFIX}
+        COMMAND ${CMAKE_COMMAND} 
+        -E echo -- Creating Python environment in ${PROJECT_BINARY_DIR}/test-venv
+        COMMAND ${Python_EXECUTABLE} -m venv ${PROJECT_BINARY_DIR}/test-venv
```
foundrytom, line 105: [fix] At present, if `DOPENASSETIO_ENABLE_PYTHON_TEST_VENV` is `OFF` then it uses the environment python to create the venv.  Feels like the venv we make through cmake should always use the build dependency python?  The flag should just change the _test_ code to use that venv or not.

Otherwise, you can't use the approach of setting this to OFF at configure, and manually creating the venv once with `cmake --build build --target openassetio-python-venv` so it targets the build dev python. You have to enable ENABLE_PYTHON_TEST_VENV, then build openassetio-python-venv, then disable it so that tests don't try to create it/install deps, which is a bit of a faf.

If you already have a venv/env with a custom python, you wouldn't want to use the convenience cmake target anyway? And if you did - it'd be assumed that cmake is already pointing to the python you want as a dependency?

foundrytom (changes requested):
Sorry have to scoot, but a few findings from testing from a blank canvas.

src/openassetio-python/CMakeLists.txt
elliotcmorris: Addressed in 0d16ac2784f441217323b2ecf7b1be993fe7851f

cmake/ThirdParty.cmake
elliotcmorris: Removed VENV from name in 9a1951d9dbd56dacecdb245888a7534c81d7f2b1

cmake/ThirdParty.cmake
feltech: This should already work - we don't override `Python_EXECUTABLE` - that comes from `find_package`. Will double-check though.

foundrytom: It uses `python` as `Python_EXECUTABLE` I think... see [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/649/files#diff-aa28751992bba392bf8ddcb87798c0f4b6f24dc31b6030f4237ff1f3f0f9201fR72). I noticed this as on my Mac it means it tries to do it in python 2 🤣

tests/openassetio-python/CMakeLists.txt
```diff
@@ -122,12 +122,13 @@ endif ()
 #
 # Add `--capture=tee-sys` to ensure output shows sanitizer errors (and
 # is useful for debugging regardless).
-function(openassetio_add_pytest_target target_name description target_directory working_directory)
+function(openassetio_add_pytest_target 
+         target_name description target_directory working_directory pythonpath)
     add_custom_target(
         ${target_name}
         COMMAND cmake -E echo -- ${description}
-        COMMAND ${OPENASSETIO_PYTHON_VENV_ACTIVATE} &&
```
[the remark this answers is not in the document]
feltech: Yeah essentially, it was really a YAGNI thing, since it's no longer needed. We can bring it back if you think it's valuable.

BUILDING.md
elliotcmorris: This is enabled by default now. Part of [this commit](https://github.com/feltech/OpenAssetIO/commit/c66b14d9ac9e036690515e4e813c32cd6ce2f23b)

tests/openassetio-python/CMakeLists.txt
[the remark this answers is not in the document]
foundrytom: No that's cool - was just checking my understanding, thanks.

resources/codegen/tests/test___main__.py
```diff
@@ -212,7 +213,7 @@ def test_when_none_set_then_all_languages_are_generated(self, tmp_path, yaml_pat
 
 
 def execute_cli(*args):
-    all_args = ["openassetio-codegen"]
+    all_args = [sys.executable, "-m", "openassetio_codegen"]
```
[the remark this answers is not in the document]
elliotcmorris: [Now a separate commit](https://github.com/feltech/OpenAssetIO/commit/a3deea32fb0e71a8cb8e5ded7e34c6b1cff349f3)

CMakeLists.txt
elliotcmorris: Added. Now part of [[Build] Allow test execution without venv ](https://github.com/feltech/OpenAssetIO/commit/c66b14d9ac9e036690515e4e813c32cd6ce2f23b)

tests/openassetio-python/CMakeLists.txt
elliotcmorris: Aye, that shouldn't have been happening. Dosen't happen anymore, due to the now-correct addition of `${_venv_if_enabled}` to all post-venv targets. (Sorry, with all the force pushing here it's hard to link this to a specific commit)

cmake/ThirdParty.cmake
elliotcmorris: Is this something we need to address? Or is it down to the user to make sure `python` means `python3.9` when they're not using out bundled environment. That'd be my feeling.

foundrytom: We need to make sure that the python venv created by cmake uses the build python, otherwise the whole 'build once' workflow is extraordinarily complicated.  It does use that to make the venv, but then it doesn't use that venv to pip install the dependencies. That was the bit to fix I think. @feltech had a plan for this.

feltech: As discussed offline, the bug was that we were attempting to pin the core venv packages (`setuptools` et al) using the external `python` (`-m pip`) rather than the venv's `python`.  This was fixed and squashed into d95c55ba780be1a01a7747f67e4a14f22b7f19ae

tests/openassetio-python/CMakeLists.txt
feltech: Yeah this was fixed and squashed into d95c55ba

tests/openassetio-python/CMakeLists.txt
feltech: Indeed it was needed for the codegen tests too, since the codegen deps are installed as part of the `openassetio-python-venv` target.  This was added as part of a squashed fixup to d95c55ba.

BUILDING.md
feltech: We removed the whole section in d95c55ba

BUILDING.md
```diff
@@ -128,7 +126,7 @@ Next we must install the test-specific dependencies. Assuming the
 install directory is `dist` under the build directory (the default)
 
 ```shell
-source build/dist/bin/activate
+source build/test-venv/bin/activate
```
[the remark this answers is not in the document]
feltech: We removed the section in d95c55ba

cmake/ThirdParty.cmake
foundrytom: This works a charm now 👍. The make-env-one-when-online-and-reuse trick makes life so much easier. Great work @elliotcmorris and @feltech. No internets for build and test for the win 🎉 .

```
cmake -S .  ... -DOPENASSETIO_ENABLE_PYTHON_TEST_VENV=OFF
cmake --build build  --target openassetio-python-venv
```
Disconnect the internets

```
. ./build/test-venv/bin/activate
ctest --test-dir build --output-on-failure
```
⏲️ 
```
100% tests passed, 0 tests failed out of 6
```

foundrytom (approved):
Top notch! LGTMFFTSAMAYL

The best kind of PR too:
image

elliotcmorris (approved):

PR #656, opened 2022-10-14
[Core] Add logging convenience methods

foundrytom:
The severity + message signature of `LoggerInterface::log` becomes somewhat verbose in regular use. We want to encourage implementations to make good use of `debug` and similar, these conveniences hopefully reduce to overhead of adding "useful logging".

@elliotcmorris @feltech Let me know if there are better/more performant ways of implementing this, went for the "simple" option in my aged brain.


elliotcmorris (approved):
:+1: LGTM. Simple is sweet.

Not merging due to second reviewer assigned.

feltech (approved):

Issue #654, opened 2022-10-13
Conan Center publishing


foundrytom:
## What

As a user I wish to add OpenAssetIO to my dependent C++ project using `conan` to install from Conan Centre

## Why

Conan is a well-used c++ package manager, this will facilitate the development of C++ integrations/PoC.

## ACs

- OpenAssetIO is available via https://conan.io/center as a source-only project
- Static .vs. shared defined via an option
- Python version range 3.7+

## Out of scope

- Prebuilt binary distributions

## Notes

Needs some consideration around how conan center handles options etc... in relation to binary distributions.

See https://github.com/conan-io/conan-center-index/blob/master/docs/how_to_add_packages.md

feltech:
As part of building a Conan package, we may wish to conditionally build and run the tests, based on an environment variable (_not_ `options` since that changes the `package_id` - we have seen prior art that uses this pattern.

foundry-markf:
A general note, Conan Center recipes are being migrating to be Conan 2 supported at the moment.

There is compatibility between 1.x and 2.0 recipes if you're following the bleeding edge of 1.x APIs, and you can build with both major versions, which generates _different_ package_ids for each build of the same recipe + options + settings. The different package_ids are intentional by the Conan devs, as a different algorithm is used under the hood.

What I'm trying to say here, is if you follow the latest and greatest Conan 1.x APIs, you should have a good chance of being Conan 2.0 compliant.

https://docs.conan.io/en/latest/conan_v2.html for migration info

feltech:
We should make pybind11 a `requires` rather than `tool_requires`, otherwise other packages alongside ours in the same dependency graph won't be forced to use the same pybind11 version.  

The disadvantage is that pybind11 will always be installed, even if not needed (since it's only used at build time). This is hopefully addressed in Conan 2, which has a more configurable `requires`.

Note that `tool_requires` is [still available](https://docs.conan.io/en/latest/migrating_to_2.0/recipes.html#requirements) in Conan 2 as a "legacy" hook.

feltech:
PR up in https://github.com/conan-io/conan-center-index/pull/15106

First attempt failed on their CI because we require a newer CMake version.  Adding `tool_requires("cmake` works, but we must pin a version of `openssl`, since there is a `cpython` and `cmake` packages depend on different versions, so we must explicitly choose one.  PR updated with fix, awaiting CI (re)build at time of writing.

feltech:
[...]

Issue #538, opened 2022-07-26
[Investigation] Windows build improvements


feltech:
# What

Update Windows CMake to support Windows as a first-class deployment target.

# Why

Currently the Windows platform is supported in hacky ways.  For example, we have
```
if(WIN32)
    install(TARGETS openassetio-core
        DESTINATION "${OPENASSETIO_PYTHON_SITEDIR}/openassetio")
```
which won't work if `OPENASSETIO_ENABLE_PYTHON` is disabled.

Also targets that depend on the variables defined by `include(GNUInstallDirs)` (e.g. `CMAKE_INSTALL_LIBDIR`) don't make as much sense in Windows.

## Acceptance Criteria
- Investigate "hacky" issues on windows that need to be fixed. (There might be duplicate DLLs)
- Investigate disabled compiler warnings, look into discovering a sensible default set
- Check for genuine problems raised by MSVC compiler warnings.
- Add issue to update the codebase such that we can add a CI check with a MSVC build set to warnings-as-errors.
 
## Out of scope
- Doing any fixes to compiler warnings.

elliotcmorris:
What we want from this pretty much is to do a squiz through our windows environment setup, and make sure there's nothing egregious/fix any windows specific warnings that may have accumulated.

If we manage to squash any and all warnings, consider enabling warning as errors internally so windows keeps being okay.

feltech:
Re. the duplicate `dll` issue.  [Worth a read of the 3.8+ behaviour](https://github.com/python/cpython/blob/main/Doc/whatsnew/3.8.rst#changes-in-the-python-api)

> DLL dependencies for extension modules and DLLs loaded with ctypes on Windows are now resolved more securely. Only the system paths, the directory containing the DLL or PYD file, and directories added with ~os.add_dll_directory are searched for load-time dependencies. Specifically, PATH and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution. If your application relies on these mechanisms, you should check for ~os.add_dll_directory and if it exists, use it to add your DLLs directory while loading your library.

PR #663, opened 2022-10-18
Add setup.py that builds Python extension

### LINKED ISSUE #630: Update `setup.py` to build c module (`pip install .`)

feltech:
Adds a `setup.py` that invokes cmake to install a complete/working `openassetio` python module from source via pip, for python-only integrations.

> **Warning**
We haven't updated the build instructions yet (see #624). At present, `setup.py` requires the C++ build toolchain to work in your environment (ie. cmake installed, and able to find dependencies + compiler).  See [BUILDING.md](BUILDING.md) and [CI bootstrap scripts](resources/build) for tips. 

Once these dependencies are satisfied, then build + installation should only require `python -m pip install .`.

Closes #630


feltech:
> We haven't updated the build instructions yet

I did update the README a little to cover this change, but the updates got lost somewhere (I suspect an errant `git reset --hard`).  I've recovered those changes now (including RELEASE_NOTES) and squashed them in.

feltech:
I removed `get_outputs` and `get_output_mapping` overrides from `setup.py` (squashed out of existence).  I put them in originally because during my investigations of OTIO's `setup.py` I found those overrides were necessary to support `--editable` installs.  However,  it turns out to be unnecessary for us!

pyproject.toml
```diff
@@ -1,3 +1,14 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2021-2022 The Foundry Visionmongers Ltd
+
+[build-system]
+requires = [
+    "setuptools==65.5.0",
```
JeanChristopheMorinPerso, line 6: I would personally just put a minimum version instead of pinning completely that dependency, `setuptools>=65.5.0`.

JeanChristopheMorinPerso (commented):
This look really nice and clean for a first pass! I left some notes in the review.

Additional notes:

* I tested the PR and most things look good! I'm pretty surprised that the editable install works out of the box without any customization in the `setup.py`. Thee is some black magic going on that I'll need to investigate more to understand (for my own curiosity).
* I did not test on Windows and macOs yet.
* You will need a [MANIFEST.in](https://packaging.python.org/en/latest/guides/using-manifest-in/) file. The manifest file is used to
specify what to include in an sdist (source distribution). sdists
are distributed on PyPI alongside the wheels to let users
build a library from source when a matching wheel is not found or
if the user does something like `pip install openassetio --no-binary :all:`.
In this case, pip would download the sdist, extract it and compile+install
the project from source.

   By adding `conan` to your build requirements and running the `conan install`
command from the setup.py, you'll be able to have a fully self contained install.
Note that the comment about not install cpython still applies for this case.
Basically, there is no need to install cpython because when then `setup.py` file is
being executed, Python is already available (obviously).
* The `setup.py` is missing a lot of metadata (author, project url, license, description, etc). See https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#setup-args.

resources/build/conanfile.py
```diff
@@ -0,0 +1,44 @@
...
+    settings = "os"
+    options = {"require_test_deps": [True, False]}
+    # TODO(DF): For minimal friction, default enable install of unit
+    #  test libraries for now - we should switch this off by default
+    #  when we get to distributing a conan package.
+    default_options = {"require_test_deps": True}
+
+    def build_requirements(self):
+        # CY2022
+        self.tool_requires("cpython/3.9.7")
```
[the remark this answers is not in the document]
feltech: Thanks for the detailed explanation! 

We have a parallel dev stream going on looking into building wheels on CI, where we have indeed used `cibuildwheel` and disabled the conan `cpython` requirement in that case, so that the external Python environment is used instead. We'll hopefully have a PR up for that within the next week or so.

Really interesting note on into the glibc version - I hadn't considered that.

feltech:
> * I did not test on Windows and macOs yet.

We're still skipping macOS for now, due to a Conan issue with the PyBind package on macOS: https://github.com/conan-io/conan-center-index/pull/11075 :disappointed: 

> * You will need a [MANIFEST.in](https://packaging.python.org/en/latest/guides/using-manifest-in/) file. The manifest file is used to
>   specify what to include in an sdist (source distribution).

This is something we hadn't considered at all! Thanks for pointing this out.
 
> * The `setup.py` is missing a lot of metadata (author, project url, license, description, etc). See https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#setup-args.

Good spot, thanks for the link.

I'm not sure how much will be added to this PR or if we'll defer to be part of the PyPI work we're doing.

setup.py
```diff
@@ -0,0 +1,86 @@
...
+        @param args: Command-line arguments to pass to `cmake`.
+        """
+        args = ["cmake"] + args
+        self.announce(" ".join(map(shlex.quote, args)), level=2)
+        subprocess.check_call(args, env=os.environ.copy(), stderr=subprocess.STDOUT)
+
+
+setup(
+    name="openassetio",
+    version="1.0.0-alpha.4",
```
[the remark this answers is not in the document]
feltech: Fixed in 0fb1f91a, thanks for the spot.

feltech:
> * The `setup.py` is missing a lot of metadata

Added in 0fb1f91a

elliotcmorris:
Do you think there's anything to be said for automatically trying to remove the build folder? I've come into the problem during dev of having to manually remove it a fair amount, but aware that may be a problem that doesn't become relevant to non contributors.

feltech:
> Do you think there's anything to be said for automatically trying to remove the build folder? I've come into the problem during dev of having to manually remove it a fair amount, but aware that may be a problem that doesn't become relevant to non contributors.

Yeah it's annoying and non-obvious (I'm thinking of the "cannot find ninja" type errors).  But I'd hesitate to assume we have the right to remove it automatically, and I'm not sure where we would add the code to do it.  

Worth noting it's not a problem on CI, since it's a one-shot thing so never tries to re-use the `build` folder.

Maybe it's worth a separate issue to investigate why we need to remove the `build` directory in the first place?

antirotor:
Just general notes I've hit trying to run `python -m build --wheel` - I always hit missing pybind11 error:

```
CMake Error at cmake/ThirdParty.cmake:55 (find_package):
  By not providing "Findpybind11.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "pybind11",
  but CMake did not find one.
```

JeanChristopheMorinPerso:
@antirotor that's because right now the process of running `conan install` and actually building the wheel is a seperate process and require to set `CMAKE_TOOLCHAIN_FILE` to `~/conan/conan_paths.cmake` like it's done at https://github.com/OpenAssetIO/OpenAssetIO/pull/663/files#diff-c26c510aacd5c04c95ead8fd5f471a970232b297bc192cd445fef7e20365835cR62.

Moving the install step into the `setup.py` should fix this.

JeanChristopheMorinPerso:
@feltech I see you didn't set the `author`/`maintainer` and `author_email`/`maintainer_email` (note that it's often redundant to use both author and maintainer at the same time). I would suggest to add them even if it seems redundant. It doesn't hurt and it doesn't force users to use that email. It's just another way of contacting the author and also clarify ownership.

feltech:
> I see you didn't set the `author`/`maintainer` and `author_email`/`maintainer_email`

The `author` field is there in 8580d318 (apologies, linked the wrong commit under the review comment).

I didn't add the `author_email`, reasoning in the commit message

> The `author_email` field has been skipped, for fear of spam, and since
> its unclear who this should point to. It should also hopefully encourage
> people to engage out in the open (i.e. via Github).

Perhaps we should re-evaluate :thinking:

setup.py
```diff
@@ -0,0 +1,134 @@
...
+            This is achieved through the definition of a common set of interactions between a host
+            of the API (eg: a Digital Content Creation tool or pipeline script) and an Asset
+            Management System.
+
+            This common API surface area removes the need for common pipeline business logic to be
+            re-implemented against the native API of each tool, and allows the tools themselves to
+            design new workflows that streamline the creation of complex assets."""
+    ),
+    url="https://github.com/OpenAssetIO/OpenAssetIO",
+    author="The Foundry Visionmongers Ltd",
```
JeanChristopheMorinPerso, line 105: I'm no expert on this, but should this be the OpenAssetIO community or something like that since OpenAssetIO is an ASWF project?

JeanChristopheMorinPerso:
> The author field is there in https://github.com/OpenAssetIO/OpenAssetIO/commit/8580d318519e0fe2475036dc16d8292de722c6d9 (apologies, linked the wrong commit under the review comment).
> 
> I didn't add the author_email, reasoning in the commit message

Ah, I see. I looked too quickly. Sorry about that. I think I would still add an email just as a last resort and for consistency.

BTW, the email won't show up in the PyPI UI. It's just metadata. The UI will show something like this:

Screen Shot 2022-10-20 at 9 29 43 AM

setup.py
feltech: Good point, I suspect you're correct. I'll have to check on that.

setup.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        args = ["cmake"] + args
+        self.announce(" ".join(map(shlex.quote, args)), level=2)
+        subprocess.check_call(args, env=os.environ.copy(), stderr=subprocess.STDOUT)
+
+
+setup(
+    name="openassetio",
+    version="1.0.0a4",
+    description=(
```
feltech, line 82: [quesion] On [further reading](https://packaging.python.org/en/latest/tutorials/packaging-projects/#creating-pyproject-toml) I wonder if we should bundle as much metadata as possible in `pyproject.toml` to do things the "modern" way?

setup.py
```diff
@@ -0,0 +1,134 @@
...
+        @param args: Command-line arguments to pass to `cmake`.
+        """
+        args = ["cmake"] + args
+        self.announce(" ".join(map(shlex.quote, args)), level=2)
+        subprocess.check_call(args, env=os.environ.copy(), stderr=subprocess.STDOUT)
+
+
+setup(
+    name="openassetio",
+    version="1.0.0a4",
```
feltech, line 81: [question] This could be specified automatically if we use [setuptools_scm](https://github.com/pypa/setuptools_scm), which also handles bundling the repo for `sdist` packages. 

We may need to conform our git tag naming convention to [PEP 440](https://peps.python.org/pep-0440/#pre-releases) for that to work properly, though. And if we want to explicitly exclude things from the `sdist` we'd need a `MANIFEST.in` regardless, which would remove some of the appeal.

On balance I think the forcing Python package version to follow the git tag is likely to cause some pain for not much gain. But worth throwing the question out there.

setup.py
JeanChristopheMorinPerso: It would indeed be a good idea.

setup.py
JeanChristopheMorinPerso: I like my sdists to be leaner than the repository itself, meaning only the files required for building the project + tests. But that's going to be up to the OpenAssetIO TSC to decide if you want to use `setuptools_scm` or not.

setup.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def __cmake(self, args: list[str]):
+        """
+        Execute `cmake` as a subprocess with given arguments.
+
+        @param args: Command-line arguments to pass to `cmake`.
+        """
+        args = ["cmake"] + args
+        self.announce(" ".join(map(shlex.quote, args)), level=2)
+        subprocess.check_call(args, env=os.environ.copy(), stderr=subprocess.STDOUT)
```
foundrytom, line 76: [question] What is the reason for re-mapping `stderr` here?

README.md
```diff
@@ -156,15 +156,35 @@ git clone git@github.com:OpenAssetIO/OpenAssetIO
...
+pip install .
+```
+
+This will automatically build and install the Python extension module,
+along with the Python sources.
+
+Note that CMake's `find_package` must be able to locate dependencies.
+Consider using the [`CMAKE_TOOLCHAIN_FILE`](https://cmake.org/cmake/help/v3.24/envvar/CMAKE_TOOLCHAIN_FILE.html)
+environment variable to augment CMake's package search paths.
+
```
foundrytom, line 174: [consider] Given we're going this far, until we sort #624, is it worth just putting the few lines of `conan install` here instead of more general toolchain details? This has been unfamiliar turf for many attempting to get going with python only builds.

src/openassetio-core/CMakeLists.txt
```diff
@@ -41,13 +41,16 @@ set_target_properties(openassetio-core PROPERTIES OUTPUT_NAME openassetio)
 # Add to the set of installable targets.
 install(TARGETS openassetio-core EXPORT ${PROJECT_NAME}_EXPORTED_TARGETS)
 
-if(WIN32)
```
foundrytom, line 44: [consider] Splitting these build changes into a separate PR (or the previous one?) as this one seems CI related otherwise.

resources/build/conanfile.py
```diff
@@ -0,0 +1,44 @@
...
+        # Same as ASWF CY2022 Docker image:
+        # https://github.com/AcademySoftwareFoundation/aswf-docker/blob/master/ci-base/README.md
+        self.tool_requires("pybind11/2.8.1")
+        if self.options.require_test_deps:
```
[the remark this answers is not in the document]
foundrytom: Capturing off-line discussion I (think) we settled on using an env var to avoid changing the `package_id`

setup.py
foundrytom: > We may need to conform our git tag naming convention to [PEP 440](https://peps.python.org/pep-0440/#pre-releases) for that to work properly, though. 

We'd originally followed the examples on [semver.org](https://semver.org/#spec-item-9) for the alpha pre-releases, but it looks like `a` is just as valid as `alpha`, so no harm switching moving forwards if it makes life more consistent.

setup.py
foundrytom, line 105: [fix]
```suggestion
    author="Contributors to the OpenAssetIO project",
```

foundrytom:
> @feltech I see you didn't set the `author`/`maintainer` and `author_email`/`maintainer_email` (note that it's often redundant to use both author and maintainer at the same time). I would suggest to add them even if it seems redundant. It doesn't hurt and it doesn't force users to use that email. It's just another way of contacting the author and also clarify ownership.

@feltech `openassetio-discussion@lists.aswf.io`

setup.py
```diff
@@ -0,0 +1,134 @@
...
+        self.__cmake(
+            [
+                "--install",
+                self.build_temp,
+                "--component",
+                "openassetio-python-module",
+            ]
+        )
+
+    def __cmake(self, args: list[str]):
```
foundrytom, line 68: [fix] Can we stick with 3.7 compatible type hints to save having to revisit this?

foundrytom (changes requested):
This is epic!

```
export CMAKE_TOOLCHAIN_FILE=~/.conan/conan_paths.cmake   
pip install .
```

Worked a charm. Just a few fairly minor tweaks.

foundrytom:
@feltech The CI refactor has upset the required checks linking, we can fix that once this is merged.

resources/build/conanfile.py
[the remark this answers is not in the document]
feltech: Added passing `Python_EXECUTABLE` CMake variable to `setup.py` in 8aaeef81.  Not technically necessary for this PR, but good practice.

pyproject.toml
feltech: Good shout, updated to use a minimum version range in d1496af5

resources/build/conanfile.py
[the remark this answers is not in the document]
feltech: Simply removed the `options` entry and associated conditional in b260bbc2

setup.py
feltech: Moved all the metadata to `pyproject.toml` in e9807bcd

setup.py
feltech: We decided to add a follow-up ticket to deal with `sdist`s, one way or another: https://github.com/OpenAssetIO/OpenAssetIO/issues/666

setup.py
feltech: When debugging build output I noticed some info was still being swallowed, I forget what now, but this fixed it.  I can try to reproduce, if you like, so I can add a comment?

foundrytom: A comment could be good, so we don't forget why once we've slept.

feltech:
> I see you didn't set the [...] `author_email`/`maintainer_email`

Added as part of the migration to `pyproject.toml` in e9807bcd (Tom can deal with any spam :smiling_imp: )

setup.py
feltech: Ah I didn't get there quite fast enough before you started reviewing, already updated in 4161630a

setup.py
feltech: I updated as part of the move to `pyproject.toml` in e9807bcd but I used "The OpenAssetIO Community", should I change that?

foundrytom: This was just to be consistent with OTIO - whichever your prefer.

feltech: Updated author name and email in af4ae75e

feltech:
> > @feltech I see you didn't set the `author`/`maintainer` and `author_email`/`maintainer_email` (note that it's often redundant to use both author and maintainer at the same time). I would suggest to add them even if it seems redundant. It doesn't hurt and it doesn't force users to use that email. It's just another way of contacting the author and also clarify ownership.
> 
> @feltech `openassetio-discussion@lists.aswf.io`

Updated in af4ae75e (see also https://github.com/OpenAssetIO/OpenAssetIO/pull/663#discussion_r1001636498)

resources/build/conanfile.py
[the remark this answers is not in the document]
feltech: > we settled on using an env var

Apologies, that wasn't my take-away, should I update this?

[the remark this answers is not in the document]
foundrytom: We can sort later if its not needed - sorry thought you made the good point that it would speed up the wheel builds.

README.md
feltech: Added some additional instructions in a2cf809f, how does that look?

JeanChristopheMorinPerso: I'd argue that the Conan install should be executed from the setup.py if you want to make a good user experience and will anyway be required for sdists.

foundrytom: We're going to look into this, just in a separate task #666

foundrytom: @JeanChristopheMorinPerso Do you have any experiencing in making this work on Windows too - is it fair to expect that people are already in a developer command prompt?

src/openassetio-core/CMakeLists.txt
foundrytom: Haha sorry yes.

README.md
JeanChristopheMorinPerso: Yep, it is entirely fair to expect that the user has a compiler available and that he is in a developer command prompt properly setup to compile C++ code. You do that same assumption on Linux and macOs, so it would be weird to make a different assumption on Windows.

.github/workflows/build-instructions.yml
```diff
@@ -26,3 +26,43 @@ jobs:
...
+
+  pip_install:
+    name: Pip install on  ${{ matrix.config.os }}
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - os: windows-2019
+            preamble: call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat" x64
```
JeanChristopheMorinPerso, line 38: Cannot CMake usually take care of sourcing the right compiler and environment?

setup.py
feltech: Well, try as I might, analysing my bash history and file history and trying to reproduce the issue, I can't.  So just removed in b74038ca

.github/workflows/build-instructions.yml
feltech: Ah, it might be compounded by us using the Ninja generator - not so much CMake assuming MinGW, but Ninja.

JeanChristopheMorinPerso: Ah, that would make sense.

foundrytom: If there is any way we can remove this malarky that would be _amazing_ (maybe wrap this into #538?)

README.md
```diff
@@ -156,15 +156,45 @@ git clone git@github.com:OpenAssetIO/OpenAssetIO
 Linux is currently the only fully tested platform. Support
 for other platforms is the subject of ongoing work.
 
+#### Quick start: Python only
+
+For use in a Python-only environment, the Python package can be
+installed from the root of the repository using
+
+```shell
+pip install .
```
foundry-markf, line 165: Nit: As a developer, I almost always use `pip install -e .` so I can noodle on Python code. Granted, I do that in a pure Python environment.

Perhaps this needs a little clarification for intended usage?

.github/workflows/build-instructions.yml
feltech: I kinda like having Ninja as a cross-platform consistent generator, which as a bonus trivially parallelizes builds. But if it can completely remove the need for vcvarsall in CI, then removing it (or making it conditional on OS) is probably worthwhile.

JeanChristopheMorinPerso: Ninja is nice indeed as it speeds up builds on Windows. Maybe take a look at OCIO (which I think use Ninja too) and see how they do it?

feltech: > maybe wrap this into #538

That would get my vote, but happy to try it out in this PR if you like.

README.md
feltech: :thinking:  I think people who don't know about develop installs will be wanting to install into a wider system, so the docs are appropriate, whereas people who do know about develop installs will be able to understand that they can add `-e` if they want?

.github/workflows/build-instructions.yml
feltech: > Maybe take a look at OCIO

They skip Ninja on Windows (https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/660cf515faa491fadab188eb0f2ed62b30b4b4f6/setup.py#L93) because

> MSVC would require all variables be exported for Ninja to pick it up, which is a little tricky to do

:laughing:

README.md
foundry-markf: True, although when I first read it (as a developer) I wondered why the develop mode wasn't being used, almost as if it wasn't supported. Perhaps a note on that?

.github/workflows/build-instructions.yml
foundrytom: Lets save this for another day

feltech: Added an investigation ticket to check feasibility of using the MSVC generator #669

README.md
feltech: :+1: added a note in 03d1a46b, is that OK?

foundry-markf: Thanks!

resources/build/conanfile.py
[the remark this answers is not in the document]
feltech: I've added a note on #654 to do this when we get around to creating a Conan package. Does that sound reasonable?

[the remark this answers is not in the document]
foundrytom: LGTM!

src/openassetio-core/CMakeLists.txt
feltech: Moved the CMakeLists changes into the previous (i.e. first) commit e4fdb46e

PR #665, opened 2022-10-20
Package wheels CI workflow

elliotcmorris:
Add workflow that invokes `cibuildwheel` to build a set of python wheels upon push to main branch (or manual invocation)
Prerequisite for having an automated PyPI release pipeline, but useful in its own right. Will archive built wheels into github storage for later retrieval.

Currently, we create 4 wheels, win-x86_64-python3.9, win-x86_64-python3.10, manylinux-x86_64-python3.9, manylinux-x86_64-python3.10
 
---
#### `cibuildwheel' platform differences
It's worth noting that `cibuildwheel` has a very different paradigm when building on linux platforms vs mac and windows. On linux platforms, a docker container is invoked, (using the `manylinux` images,) whereas on windows and mac, a virtual environment is configured directly inside the existing environment.

This is the reason you'll see a new bootstrapping script, as we need to bootstrap this docker container, which is centOS.

---
#### Running the action.
This action uses `workflow_dispatch`, so as well as running upon push/merge to master, is runnable from the `Build Wheels` section on the [Actions](https://github.com/elliotcmorris/OpenAssetIO/actions/workflows/build-wheels.yml) page. Be sure to select the correct branch.

Whilst we wait for that to cook, here's one I made earlier : [Link](https://github.com/elliotcmorris/OpenAssetIO/actions/runs/3291239068)

### DIFF
--- a/.github/bootstrap_platform/action.yml
+++ b/.github/bootstrap_platform/action.yml
@@ -32,7 +32,7 @@ runs:
     run: |
       source resources/build/bootstrap-${{ matrix.config.os }}.sh
     env:
-      WORKSPACE: .
+      WORKSPACE: ${{ github.workspace }}
     # Annoyingly required for Github composite actions.
     shell: bash
 
@@ -43,6 +43,6 @@ runs:
     run: |
       cmd /C resources\build\bootstrap-${{ matrix.config.os }}.bat
     env:
-      WORKSPACE: .
+      WORKSPACE: ${{ github.workspace }}
     # Annoyingly required for Github composite actions.
     shell: cmd
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -1,0 +1,76 @@
+name: Build Wheels
+
+on: 
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+  pull_request:
+    branches-ignore:
+      - docs
+    paths-ignore:
+      - 'docs/**'
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.config.os }} - ${{matrix.python-build}}
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config: 
+          - os: ubuntu-20.04
+            shell: bash
+          # - os: macos-10.15
+          #   shell: bash
+          - os: windows-2019
+            preamble: call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat" x64
+            shell: cmd
+          # These are the platform build strings provided to
+          # cibuildwheel, with wildcarding. See
+          # https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip
+        python-build: ['cp39*64', 'cp310*64']
+    defaults:
+      run:
+        # Annoyingly required here since `matrix` isn't available in the
+        # `shell` property of individual steps.
+        shell: ${{ matrix.config.shell }}
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Bootstrap
+        uses: ./.github/bootstrap_platform
+        env:
+          # As we're building wheels inside a managed environment,
+          # (cibuildwheels) there is no need to install cpython. 
+          OPENASSETIO_CONAN_SKIP_CPYTHON: "True"
+
+      - name: Build wheels
+        run: |
+          ${{ matrix.config.preamble }}
+          pip install cibuildwheel==2.11.1
+          cibuildwheel --output-dir wheelhouse
+        env: 
+          # Windows + Mac runs use this environment variable, as they
+          # execute directly on the runner. Linux runs are somewhat
+          # different, and are containerised fully, therefore they get
+          # their toolchain path from pyproject.toml
+          CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
+          # We matrix the python build here ourselves, rather than
+          # letting cibuildwheel do its regular python matrix, as we
+          # want to allow a full release pipeline to run "vertically" on
+          # CI, on individual agents, rather than building every python
+          # version sequentially on each platform. This is done less for
+          # performance reasons (although it does help) and more so a
+          # single failing python version won't interrupt every other
+          # deploy on that platform.
+          CIBW_BUILD: ${{ matrix.python-build }}
+          CIBW_SKIP: '*musllinux* *arm64*'
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
+          PIP_VERBOSE: 1
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: "openassetio-wheels"
+          path: ./wheelhouse/*.whl
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,9 +20,17 @@ jobs:
         # environments.
         config:
         - os: windows-2019
-          vcvars: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"
+          preamble: call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat" x64
+          shell: cmd
         - os: ubuntu-20.04
+          shell: bash
         - os: macos-10.15
+          shell: bash
+    defaults:
+      run:
+        # Annoyingly required here since `matrix` isn't available in
+        # the `shell` property of individual steps.
+        shell: ${{ matrix.config.shell }}
 
     steps:
     - uses: actions/checkout@v2
@@ -30,36 +38,20 @@ jobs:
     - name: Bootstrap
       uses: ./.github/bootstrap_platform
 
-    - name: Configure CMake build (Unix)
-      if: runner.os != 'Windows'
+    - name: Configure CMake build
       run: >
+        ${{ matrix.config.preamble }}
+
         cmake -S . -B build -G Ninja
-        --install-prefix $GITHUB_WORKSPACE/dist
-        --toolchain $HOME/conan/conan_paths.cmake
+        --install-prefix ${{ github.workspace }}/dist
+        --toolchain ${{ github.workspace }}/.conan/conan_paths.cmake
         --preset test
 
-    - name: Configure CMake build (Windows)
-      if: runner.os == 'Windows'
+    - name: Build, install and test
       run: >
-        call "${{ matrix.config.vcvars }}"
-
-        cmake -G Ninja -S . -B build --install-prefix %GITHUB_WORKSPACE%\dist
-        --toolchain %USERPROFILE%\conan\conan_paths.cmake -DCMAKE_VERBOSE_MAKEFILE=ON
-        --preset test
-
-      shell: cmd
-
-    - name: Build, install and test (Unix)
-      if: runner.os != 'Windows'
-      run: |
+        ${{ matrix.config.preamble }}
+        
         ctest -VV --test-dir build --parallel 2
-
-    - name: Build, install and test (Windows)
-      if: runner.os == 'Windows'
-      run: |
-        call "${{ matrix.config.vcvars }}"
-        ctest -VV --test-dir build --parallel 2
-      shell: cmd
 
   docs:
     runs-on: ubuntu-latest
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -61,6 +61,15 @@ v1.0.0-alpha.x
   `warning`, `error`, `critical` to `LoggerInterface` to log messages of
   the respective severity.
 
+- Added `OPENASSETIO_CONAN_SKIP_CPYTHON` environment variable to prevent
+  conan installing its own python version. This is to support workflows
+  where the user is bringing their own python environment, and does not 
+  want python installations to conflict.
+  [#653](https://github.com/OpenAssetIO/OpenAssetIO/issues/653)]
+
+- Bleeding edge python wheels are now downloadable as artifacts from the
+  `Build wheels` github actions workflow.
+  [[#653](https://github.com/OpenAssetIO/OpenAssetIO/issues/653)]
 
 v1.0.0-alpha.4
 --------------
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -19,6 +19,11 @@ v1.0.0-alpha.x
   `OPENASSETIO_ENABLE_PYTHON_TEST_VENV` is enabled.
   [#629](https://github.com/OpenAssetIO/OpenAssetIO/issues/629)
 
+- Changed `--install-folder` location of conan install in bootstrap
+  scripts from `$CONAN_USER_HOME` to `$WORKSPACE/.conan`. Users who
+  have been using the bootstrap scripts directly may need to update
+  their toolchain CMake arguments.
+
 ### New Features
 
 - Added `simpleResolver.py` example host (under `resources/examples`),
--- a/resources/build/bootstrap-cibuildwheel-manylinux-2014.sh
+++ b/resources/build/bootstrap-cibuildwheel-manylinux-2014.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -xeo pipefail
+
+# This script not runnable from outside a cibuildwheel environment.
+if [ "$CIBUILDWHEEL" != "1" ]
+then
+  echo "Error. Attempting to run cibuildwheel-manylinux bootstrap outside of cibuildwheel"
+  exit 1
+fi
+
+# Install additional build tools.
+pip3 install -r "resources/build/requirements.txt"
+# Use explicit predictable conan root path, where packages are cached.
+export CONAN_USER_HOME="$HOME/conan"
+
+# Create default conan profile so we can configure it before instlibXcomposite-develall.
+# Use --force so that if it already exists we don't error out.
+conan profile new default --detect --force
+# Use old C++11 ABI as per VFX Reference Platform CY2022. Not strictly
+# necessary as this is the default for conan, but we can't be certain
+# it'll remain the default in future.
+conan profile update settings.compiler.libcxx=libstdc++ default
+
+# Install openassetio third-party dependencies from public Conan Center
+# package repo.
+export OPENASSETIO_CONAN_SKIP_CPYTHON="True"
+conan install --install-folder ".conan" --build=missing \
+    "resources/build"
--- a/resources/build/bootstrap-ubuntu-20.04.sh
+++ b/resources/build/bootstrap-ubuntu-20.04.sh
@@ -22,8 +22,7 @@ sudo apt-get install -y --no-install-recommends libfontenc-dev libice-dev libsm-
 
 # Install additional build tools.
 sudo pip3 install -r "$WORKSPACE/resources/build/requirements.txt"
-# Use explicit predictable conan root path, to be used for both packages
-# and conan CMake toolchain config.
+# Use explicit predictable conan root path, where packages are cached.
 export CONAN_USER_HOME="$HOME/conan"
 # Create default conan profile so we can configure it before install.
 # Use --force so that if it already exists we don't error out.
@@ -39,7 +38,7 @@ conan profile update settings.compiler.libcxx=libstdc++ default
 # system packages are already available. In particular, this affects
 # recent versions of the xorg/system recipe (a dependency of cpython).
 # The problem is reported and fixed in https://github.com/conan-io/conan/pull/11712
-conan install --install-folder "$CONAN_USER_HOME" --build=missing \
+conan install --install-folder "$WORKSPACE/.conan" --build=missing \
     -c tools.system.package_manager:mode=install \
     -c tools.system.package_manager:sudo=True \
     "$WORKSPACE/resources/build"
--- a/resources/build/conanfile.py
+++ b/resources/build/conanfile.py
@@ -1,4 +1,4 @@
-from conans import ConanFile, CMake
+from conans import ConanFile, CMake, tools
 
 
 class OpenAssetIOConan(ConanFile):
@@ -7,17 +7,17 @@ class OpenAssetIOConan(ConanFile):
         # which augments package search paths with conan package
         # directories.
         "cmake_paths",
-        # Generate `Find<PackageName>.cmake` finders, required for
-        # various public ConanCenter packages (e.g. pybind11) since they
-        # disallow bundling of `<PackageName>Config.cmake`-like files in
-        # the package.
+        # Generate `Find<PackageName>.cmake` finders, required for various
+        # public ConanCenter packages (e.g. pybind11) since they disallow
+        # bundling of `<PackageName>Config.cmake`-like files in the package.
         "cmake_find_package",
     )
     settings = "os"
 
     def build_requirements(self):
         # CY2022
-        self.tool_requires("cpython/3.9.7")
+        if not tools.get_env("OPENASSETIO_CONAN_SKIP_CPYTHON", False):
+            self.tool_requires("cpython/3.9.7")
         # Same as ASWF CY2022 Docker image:
         # https://github.com/AcademySoftwareFoundation/aswf-docker/blob/master/ci-base/README.md
         self.tool_requires("pybind11/2.8.1")
--- a/.github/workflows/build-instructions.yml
+++ b/.github/workflows/build-instructions.yml
@@ -1,5 +1,4 @@
 # A simple workflow that runs the documented build steps work.
-
 name: Build Instructions
 on:
   pull_request:
@@ -58,7 +57,7 @@ jobs:
         ${{ matrix.config.preamble }}
         pip install .
       env:
-        CMAKE_TOOLCHAIN_FILE: ~/conan/conan_paths.cmake
+        CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
 
     - name: Install test dependencies
       run: pip install -r tests/python/requirements.txt
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -1,6 +1,6 @@
 name: Code quality
 on: pull_request
-
+  
 jobs:
   pylint:
     runs-on: ubuntu-latest
@@ -25,7 +25,7 @@ jobs:
         run: |
           python -m pip install .
         env:
-          CMAKE_TOOLCHAIN_FILE: ~/conan/conan_paths.cmake
+          CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
 
       - name: Lint
         uses: TheFoundryVisionmongers/fn-pylint-action@v1.1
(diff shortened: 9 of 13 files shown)


.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,55 @@
+name: Build Wheels
+
+on: 
+  push:
+    branches:
+      - main
```
elliotcmorris, line 6: We decided to also run this on pull request.

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          ${{ matrix.config.preamble }}
+          pip install cibuildwheel
+          cibuildwheel --output-dir wheelhouse
+        env: 
+          # Windows + Mac runs use this environment variable, as they execute directly
+          # on the runner. Linux runs are somewhat different, and are containerised fully,
+          # therefore they get their toolchain path from pyproject.toml
+          CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
+          CIBW_BUILD: ${{ matrix.python-build }}
```
foundrytom, line 49: [consider] Is it worth documenting the discussion we had around where to paralellize do you think? ie. using a CI matrix vs `cibuildwheel`'s native functionality?

.github/workflows/test.yml
```diff
@@ -36,17 +36,17 @@ jobs:
...
 
     - name: Configure CMake build (Windows)
       if: runner.os == 'Windows'
       run: >
         call "${{ matrix.config.vcvars }}"
 
-        cmake -G Ninja -S . -B build --install-prefix %GITHUB_WORKSPACE%\dist
-        --toolchain %USERPROFILE%\conan\conan_paths.cmake -DCMAKE_VERBOSE_MAKEFILE=ON
+        cmake -G Ninja -S . -B build --install-prefix ${{ github.workspace }}\dist
+        --toolchain ${{ github.workspace }}\.conan\conan_paths.cmake -DCMAKE_VERBOSE_MAKEFILE=ON
```
foundrytom, line 49: [question] Does this work with `/` instead of `\`? As your genius move to switch to yaml subs night mean we could finally get rid of these two independent configure cmake steps?

resources/build/bootstrap-ubuntu-20.04.sh
```diff
@@ -22,8 +22,8 @@ sudo apt-get install -y --no-install-recommends libfontenc-dev libice-dev libsm-
 
 # Install additional build tools.
 sudo pip3 install -r "$WORKSPACE/resources/build/requirements.txt"
-# Use explicit predictable conan root path, to be used for both packages
-# and conan CMake toolchain config.
+# Use explicit predictable conan root path, where packages are 
+# installed.
 export CONAN_USER_HOME="$HOME/conan"
```
foundrytom, line 27: [question] No we're no longer installing into this, does it still make sense to set this? If I understand it right, this is just going to make the conan cache end up in ` ~/conan/.conan` instead of `~/.conan`?  Given we have to hard-code that path in the CI cache step anyway, not sure it makes much difference?

resources/build/bootstrap-ubuntu-20.04.sh
```diff
@@ -22,8 +22,8 @@ sudo apt-get install -y --no-install-recommends libfontenc-dev libice-dev libsm-
 
 # Install additional build tools.
 sudo pip3 install -r "$WORKSPACE/resources/build/requirements.txt"
-# Use explicit predictable conan root path, to be used for both packages
-# and conan CMake toolchain config.
+# Use explicit predictable conan root path, where packages are 
+# installed.
```
foundrytom, line 26: [consider] `where packages are cached` - as we go on to use `--install-folder` in a few lines, and its not to this dir.  Conan documents [this var](https://docs.conan.io/en/latest/reference/env_vars.html#conan-user-home) as being for the cache dir.

resources/build/bootstrap-ubuntu-20.04.sh
```diff
@@ -39,7 +39,7 @@ conan profile update settings.compiler.libcxx=libstdc++ default
 # system packages are already available. In particular, this affects
 # recent versions of the xorg/system recipe (a dependency of cpython).
 # The problem is reported and fixed in https://github.com/conan-io/conan/pull/11712
-conan install --install-folder "$CONAN_USER_HOME" --build=missing \
+conan install --install-folder "$WORKSPACE/.conan" --build=missing \
```
foundrytom, line 42: [question] What are peoples thoughts on this being hidden? Its a critical dir (eg. one you have to set with `--toolchain` etc), could it make more sense to suggest a workflow where it's visible? I don't know what the precedent is in general conan usage in the world at large.

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,55 @@
...
+          CIBW_SKIP: '*musllinux* *arm64*'
+          PIP_VERBOSE: 1
```
foundrytom, line 51: [question] Did we get to the bottom of the `manylinux201` vs `2014` question? (https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/.github/workflows/python-package.yml#L171)

elliotcmorris: No we didn't, i've asked on the OTIO slack : https://academysoftwarefdn.slack.com/archives/CMQ9J4BQC/p1666361161870199

resources/build/bootstrap-ubuntu-20.04.sh
foundrytom: Fine to keep it - but would it be worth setting to `~` or similar instead so we're not adding a random dir? Is this something we should just bring into the workspace too, given we're actively managing its population with a cache stage?

.github/workflows/build-wheels.yml
elliotcmorris: Looks like CentOS 6 support was the main reason, they plan to eventually update. What do you think @foundrytom, how backwards compatible do you wanna be here?

foundrytom: For now, compatibility with Centos 7 is important, as a significant portion of the industry is there.

.github/workflows/test.yml
elliotcmorris: So was the problem before was that we has mismatched separators in the path, so we had to be consistent with what the windows env variable was providing?

Because aye, appears that you're right : https://github.com/elliotcmorris/OpenAssetIO/actions/runs/3298167936/jobs/5439927397

foundrytom: A-mazing, nice one. I _think_ we had it before due to the home dir env var differences ie (`%` and the name). 

If you're so inclined and fancy stripping out the duplicate task - that'd be ace, or make an issue and we can do it later if you'd rather avoid the scope creep.

resources/build/bootstrap-ubuntu-20.04.sh
feltech: I guess I see it as a transient generated directory that's not part of the main source tree.  But happy to remove the `.` if that's the consensus.

.github/workflows/test.yml
foundrytom: There was the VC Vars thing, but I think we have the preamble trick now don't we?

resources/build/conanfile.py
```diff
@@ -0,0 +1,44 @@
...
+    settings = "os"
+    options = {"testing": [True, False], "install_cpython" : [True, False]}
+    # TODO(DF): For minimal friction, default install unit test
+    #  libraries for now - we should switch this off by default when
+    #  we get to distributing a conan package.
+    default_options = {"testing": True, "install_cpython" : True}
+
+    def build_requirements(self):
+        # CY2022
+        if self.options.install_cpython :
```
feltech, line 24: [fix] I think for similar reasons as in https://github.com/OpenAssetIO/OpenAssetIO/pull/663#discussion_r1000539083 (i.e. avoiding affecting the package ID) we should use an environment variable rather than `options`.

We should probably also rename it to something other than install.

So an env var like `OPENASSETIO_CONAN_REQUIRE_CPYTHON` or similar.

.github/workflows/test.yml
feltech: Yeah, removing the branching would be really nice

elliotcmorris: Aye, will do, as it's conceptually part of the action to change over the workspace var.

resources/build/bootstrap-ubuntu-20.04.sh
foundrytom: According to the [docs](https://docs.conan.io/en/latest/reference/env_vars.html#conan-user-home), `CONAN_USER_HOME` only talks about the cache dir - so we can assume its always inside there?

resources/build/bootstrap-ubuntu-20.04.sh
elliotcmorris: Same opinion as David, "feels" nicer as a hidden dir, but very un-strong opinions. I've never found a directory being hidden to be a point of confusion, personally.

resources/build/bootstrap-ubuntu-20.04.sh
foundrytom: I'd certainly be happy to remove it though if you didn't see the specific merit, as it makes it easier to grok what's happening.

elliotcmorris: > then decided there was no point in complicating the PR with that change and so reverted it?

Yeah, this is my recollection, we reasoned the existing setup is equivilant to `~` in all but style, so why bother adding the noise

.github/workflows/test.yml
elliotcmorris: Addressed in : [bb2f723](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/bb2f723942344b33201eba3b7510783195d20a64)

You can see [here](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/3298614794/jobs/5440961991), no more duplicate jobs.

foundrytom: Amazing! nice one

resources/build/bootstrap-ubuntu-20.04.sh
foundrytom: Do you mean why bother removing it? The existing setup explicitly _isnt_ the same as `~`, as it moves the cache from `~/.conan` to `~/conan/.conan`.

.github/workflows/build-wheels.yml
elliotcmorris: Yeah, added a comment in [d9e5dd2](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/d9e5dd2523900e00ca45743c18fd04397fe54439)

resources/build/bootstrap-ubuntu-20.04.sh
foundrytom: An other potential reason to remove it - this bootstrap script is used by Vagrant and as 'suggested' instructions for people. By setting a custom `CONAN_USER_HOME` you then have to remember that this was set before, if you want to interact with conan in future, unrelated commands as the env var may no longer be set.

If we leave it out, then you can use conan 'as normal' in subsequent commands, and not have to remember that it was moved before. So you don't end up with two separate conan folders from differing `install` commands.

foundrytom: Either way though, lets save this for a follow up PR as we'd need to update the CI cache etc...

resources/build/bootstrap-ubuntu-20.04.sh
foundrytom: Cool beans.

.github/workflows/build-wheels.yml
foundrytom: [fix] Could we just pin this though, to save us breaking compatibility at some random point when they update the defaults? (Though hopefully well end up with a CI check that pip installs inside of VFX platform containers).

resources/build/bootstrap-ubuntu-20.04.sh
elliotcmorris: [6fd9e26](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/6fd9e26fd530240f8704c2a1146dd282253e215b) && [2213989](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/2213989fb7dd9a59e779fbc284df9a4243d29e88)

resources/build/bootstrap-ubuntu-20.04.sh
feltech: > you then have to remember that this was set before, if you want to interact with conan in future

That's a pretty convincing argument, shall I make an issue?

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          ${{ matrix.config.preamble }}
+          pip install cibuildwheel
+          cibuildwheel --output-dir wheelhouse
+        env: 
+          # Windows + Mac runs use this environment variable, as they execute directly
+          # on the runner. Linux runs are somewhat different, and are containerised fully,
+          # therefore they get their toolchain path from pyproject.toml
+          CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
+          # We matrix the python build here ourselves, rather than letting cibuildwheel do its
```
feltech, line 54: [consider] We do try to stick to the 72 char limit for comments in non-Python files too.

resources/build/bootstrap-cibuildwheel-manylinux-2014.sh
```diff
@@ -0,0 +1,31 @@
...
+# necessary as this is the default for conan, but we can't be certain
+# it'll remain the default in future.
+conan profile update settings.compiler.libcxx=libstdc++ default
+
+# This script not runnable from outside a cibuildwheel environment.
+if [ "$CIBUILDWHEEL" != "1" ]
+then
+  echo "Error. Attempting to run cibuildwheel-manylinux bootstrap outside of cibuildwheel"
+  exit 1
+fi
```
foundrytom, line 23: [question] What breaks if its not invoked by `cibuildwheel`? Should we error at the top instead if it is broken to avoid partially polluting the calling env?

.github/workflows/build-wheels.yml
elliotcmorris: _grumble_ [9ed6ad9](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/9ed6ad983f3e2418db40c8c4d7de4dad7b468616)

.github/workflows/build-wheels.yml
elliotcmorris: Yeah, i had a similar thought, I think that's sensible. Will do

foundrytom (changes requested):
Thanks for that rejig - a lot easier to keep track of which changes precipitated which now 👍.

resources/build/bootstrap-cibuildwheel-manylinux-2014.sh
elliotcmorris: This was mostly a scope creep avoidance move to stop us having to figure out that very question, as this was originally a `bootstrap-centos` script and it got more specific.

Aye, this should be right at the beginning, will move.

.github/workflows/build-wheels.yml
elliotcmorris: [dd10abd](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/dd10abd68a74b2ecf217001aa5d2261e0718be19)

.github/workflows/build-wheels.yml
feltech: :pray: Beautiful, thanks!

resources/build/bootstrap-cibuildwheel-manylinux-2014.sh
elliotcmorris: [34e1b43](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/34e1b437c487b640aed292abb2e40acc75d1200a)

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,60 @@
...
+          # on the runner. Linux runs are somewhat different, and are containerised fully,
+          # therefore they get their toolchain path from pyproject.toml
+          CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
+          CIBW_BUILD: ${{ matrix.python-build }}
+          CIBW_SKIP: '*musllinux* *arm64*'
+          PIP_VERBOSE: 1
+
+      - uses: actions/upload-artifact@v3
+        with:
+          path: ./wheelhouse/*.whl
```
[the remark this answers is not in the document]
elliotcmorris: [c0c4a19](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/c0c4a197d70705cba9a29d77d5d5471fc1714f8f)

Just called this `openassetio-wheels`

resources/build/bootstrap-ubuntu-20.04.sh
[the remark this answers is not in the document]
elliotcmorris: [54da118](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/54da1189d1c4f7da75c8d6877dfd23ed601a3354)

Aye sorry, you mentioned in the first review about release notes, overlooked em.
I put this in as a breaking change, which I think it technically is, though was unsure.

foundrytom (approved):
Looks good from my perspective - thanks @elliotcmorris

resources/build/conanfile.py
```diff
@@ -13,24 +13,18 @@ class OpenAssetIOConan(ConanFile):
...
-    options = {"testing": [True, False], "install_cpython" : [True, False]}
-    # TODO(DF): For minimal friction, default install unit test
-    #  libraries for now - we should switch this off by default when
-    #  we get to distributing a conan package.
-    default_options = {"testing": True, "install_cpython" : True}
 
     def build_requirements(self):
         # CY2022
-        if self.options.install_cpython :
+        if not tools.getenv("OPENASSETIO_CONAN_SKIP_CPYTHON", False):
```
foundrytom, line 19: I like this name 👍

resources/build/conanfile.py
elliotcmorris: Addressed [48329fc](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/48329fc01c576948332fbb8ded6f4b4c32f7a379)

Worth giving this one a look, made some minor decisions. Called the var `OPENASSETIO_CONAN_SKIP_CPYTHON`, and removed a bit of redundant setting in the bootstrap files.

elliotcmorris: Oh! and conan also had a [get_env](https://docs.conan.io/en/latest/reference/tools.html#tools-get-env) command which looks to be designed to solve that true/false/0/1 problem we were discussing.

foundrytom: I like this name 👍

.github/workflows/test.yml
```diff
@@ -20,48 +20,37 @@ jobs:
...
     - uses: actions/checkout@v2
 
     - name: Bootstrap
       uses: ./.github/bootstrap_platform
 
-    - name: Configure CMake build (Unix)
-      if: runner.os != 'Windows'
+    - name: Configure CMake build
       run: >
+        ${{ matrix.config.preamble }}
```
feltech, line 44: [fix] Despite appearences, this [actually breaks CI](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/3313127660/jobs/5470685372). 

In YAML, a `>` means "join all lines into a single line, unless there is a blank line in-between".  So we should add a blank line between the preamble and `cmake` lines. Or alternatively use `|` instead of `>` and add `^`s to the multiline `cmake` invocation.

There is then of course the more concerning issue that the CI check still passes despite this causing an error :worried:

resources/build/conanfile.py
feltech: Awesome. Actual commit seems to be c88ed675.  Much cleaner.

feltech: I notice the latest [Ubuntu CI job](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/3313127655/jobs/5470684152) is installing `cpython` through conan, though.

feltech: Actually, not quite all good - Windows builds are still installing cpython.

feltech: Tricky though, since we need to set `CIBUILDWHEEL` in the `bootstrap_platform` steps.  Hopefully we can set a global env var for that workflow and it'll work?

foundrytom: As we may have other situations where we want to use system python (Eg: pinning python version for ASAN, once we reinstate 3.7 support), would it make sense to just call that var `OPENASSETIO_CONAN_SKIP_CPYTHON`? and set it from the outside whenever we know we want to inherit python? Vs associating it solely with `cibuildwheel`?

.github/workflows/test.yml
elliotcmorris: :worried:  Oh wow, amazing spot. Possibly the whole workflow still has a bunch of bugs to work out then, :disappointed:

elliotcmorris: Boy i've had a _fun time_ with some seemingly transient errors trying to get this fixed.

Addressed in [41e3197](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/41e319788184270f9ee2e1840a22a7133546d73e), later fixups address problems revealed from the windows tests actually executing.

.github/bootstrap_platform/action.yml
```diff
@@ -32,7 +32,7 @@ runs:
     run: |
       source resources/build/bootstrap-${{ matrix.config.os }}.sh
     env:
-      WORKSPACE: .
+      WORKSPACE: ${{ github.workspace }}
```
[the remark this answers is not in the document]
elliotcmorris: Changed in rebase.

resources/build/conanfile.py
elliotcmorris: @feltech This what you were thinking? Looks to be not treating cpython as a dependency on the runners now : [8baa986](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/8baa9863ea437fd06688ae37268bb79fcab1a978)

(Went back and forth a bit after that fixup on the manylinux bootstrap, which maintains use of the `CIBUILDWHEEL` to setup itself rather than attempting to inject our environment variable through, which in retrospect is likely the entire point of that env variable being in the VM.)

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,76 @@
...
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Bootstrap
+        uses: ./.github/bootstrap_platform
+        env:
+          # As we're building wheels inside a managed environment,
+          # (cibuildwheels) there is no need to install cpython. 
+          OPENASSETIO_CONAN_SKIP_CPYTHON: "True"
```
foundrytom, line 47: I like the approach of doing it here - as its then the same mechanism in both scenarios 👍

feltech: Yeah this is really neat, didn't realise we could do that.

.github/workflows/test.yml
feltech: [plan] Perhaps worth an issue to look into why the CI step passed despite `vcvarsall` failing and if there's anything we can do about it.

Otherwise looks good and can resolve this.

feltech (commented):
Looking great! Feel free to squash ready for merge (may want to take a last quick review after squashing).

.github/workflows/test.yml
elliotcmorris: I'll just have a quick look into this the now rather than putting this off, if it becomes a rabbit hole will bail out and log something.

elliotcmorris:
> (may want to take a last quick review after squashing).

@feltech Be my guest :D

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,76 @@
...
+          # want to allow a full release pipeline to run "vertically" on
+          # CI, on individual agents, rather than building every python
+          # version sequentially on each platform. This is done less for
+          # performance reasons (although it does help) and more so a
+          # single failing python version won't interrupt every other
+          # deploy on that platform.
+          CIBW_BUILD: ${{ matrix.python-build }}
+          CIBW_SKIP: '*musllinux* *arm64*'
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
+          PIP_VERBOSE: 1
```
feltech, line 71: [consider][minor] Presumably this won't be passed along to the manylinux build? We may want to add this to the `[tool.cibuildwheel.linux]` `environment` in `pyproject.toml`.

RELEASE_NOTES.md
```diff
@@ -19,6 +19,11 @@ v1.0.0-alpha.x
   `OPENASSETIO_ENABLE_PYTHON_TEST_VENV` is enabled.
   [#629](https://github.com/OpenAssetIO/OpenAssetIO/issues/629)
 
+- Changed `--install-folder` location of conan install in bootstrap
+  scripts from `$CONAN_USER_HOME` to `$WORKSPACE/.conan`. Users who
+  have been using the bootstrap scripts directly may need to update
+  their toolchain CMake arguments.
+
 ### New Features
```
feltech, line 27: [consider][minor] Should we add a release note that Python wheels are downloadable through the build artifacts of the "Build wheels" Github Actions workflow?  Might be useful to some who want bleeding edge without building themselves.

feltech (approved):
Looking good, just some minor comments.

.github/workflows/test.yml
elliotcmorris: Pretty difficult one to google. I had a look at the `vcvarsall` source itself.

Mostly what that script does is dispatch to other setup files, however, in the case that none of the dispatched scripts error, every return path from `vcvarsall` itself explicitly sets the error code to 0, even in the case where it reports a misuse error via `echo`

This is frustrating, we could scan the output for the word `error` I suppose, but that's seems a tad overly.

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,76 @@
...
+        uses: ./.github/bootstrap_platform
+        env:
+          # As we're building wheels inside a managed environment,
+          # (cibuildwheels) there is no need to install cpython. 
+          OPENASSETIO_CONAN_SKIP_CPYTHON: "True"
+
+      - name: Build wheels
+        run: |
+          ${{ matrix.config.preamble }}
+          pip install cibuildwheel
```
JeanChristopheMorinPerso, line 52: I would recommend to pin the version of cibuildwheel. You can use the GitHub action to make it easier to upgrade (with the help of dependabot).

RELEASE_NOTES.md
elliotcmorris: Yeah, i think that's a good idea.

foundrytom: Improvement works for me 👍

elliotcmorris: [61db753](https://github.com/OpenAssetIO/OpenAssetIO/pull/665/commits/61db7533724539e8d8098c751462fc0d18184078)

.github/bootstrap_platform/action.yml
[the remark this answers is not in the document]
elliotcmorris: Aye, moved that note to the prior commit wholesale, it still applies.

.github/workflows/test.yml
feltech: OK no worries, I guess we just have to keep an eye out, thanks.

.github/workflows/build-wheels.yml
feltech: Unfortunately because of our reliance on `vcvarsall` (due to us using the Ninja generator, it seems), we can't use the Github action. Something we'll to look into in #669.

We're going to look into our version pinning strategy more broadly in #667.

Otherwise I'm happy for `cibuildwheel` version to be pinned, there are arguments either way.

elliotcmorris: Yeah good idea. Have pinned it to 2.11.1 just like's been done in OTIO.

We used the direct install method for the convenience of being able to run `vcvarsall` on windows platforms, which is what that `config.preamble` is doing above, and have all that config bleed into the environment. For the moment we'll leave that as it stands, and forgo dependabots help in this one case.

elliotcmorris: > there are arguments either way

We've been tending to favour pinning over not pinning when the decision has came up, i've noticed, so will continue on with pinning this particular version, with an eye to conforming it to whatever strategy is decided after the fact.

foundrytom: I think for this, as we only find our during a release, I'd def want to opt for pinning - as a failure here is far more of a pain than a PR CI job.

Issue #668, opened 2022-10-21
Update GitHub action versions


foundrytom:
## What

 - Update the pinned versions of git hub actions (Eg: `checkout` etc..)
 - Decide if versions for core GitHub sourced actions should be pinned (vs third party)
 
## Why

- The versions we are using are deprecated, see https://github.com/elliotcmorris/OpenAssetIO/actions/runs/3291239068

JeanChristopheMorinPerso:
I would recommend dependabot to manage updates of GH actions.

foundrytom:
Dependabot configured.

PR #670, opened 2022-10-24
[CI] Enable Dependabot tracking of GitHub Action versions

foundrytom:
Actions deprecation warnings only show up in the detailed logs for an action, so are easily missed.

Part of #668.

Im not entirely sure how best to test this!

Notes: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+# SPDX-License-Identifier: Apache-2.0
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"  # See docs, they say not to use '.github'
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "[CI]"


elliotcmorris (approved):
Wow, so this robot is going to automatically open PRs for us, neat. 

Guess the only way to test this is to set a shorter schedule interval and wait for it to do something on your fork? Depends if you wanna wait a day for this or not.

foundrytom:
> Wow, so this robot is going to automatically open PRs for us, neat.
> 
> Guess the only way to test this is to set a shorter schedule interval and wait for it to do something on your fork? Depends if you wanna wait a day for this or not.



> Wow, so this robot is going to automatically open PRs for us, neat.
> 
> Guess the only way to test this is to set a shorter schedule interval and wait for it to do something on your fork? Depends if you wanna wait a day for this or not.

Well that makes a lot of sense, good idea, thanks. Ill merge it on my fork and we can see!

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: Apache-2.0
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"  # See docs, they say not to use '.github'
+    schedule:
+      interval: "weekly"
+    commit-message:
```
feltech, line 8: [question] I wish the docs were a bit more explicit.  

If you don't include a `commit-message` section, like [OTIO](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/.github/dependabot.yml), then the list of dependencies updated is in the [commit message](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/commits/main).

However, then the docs say:

> `include: "scope"` specifies that any prefix is followed by a list of the dependencies updated in the commit.

This would seem to imply, then, that if you have a `prefix` but don't add `include: "scope"`, then the commit message won't list the updated dependencies?

That seems daft to me, but, like I say, the docs aren't super clear.

I guess we can try without `include: "scope"` and see if the generated messages are reasonable?

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
...
+      prefix: "[CI]"
```
feltech, line 9: [consider] I notice in one [random example](https://birtony.medium.com/breaking-ci-with-dependabot-weekly-telescope-podcast-af9884214187) that there is a space after the prefix, which would make sense.

foundrytom:
> Well that makes a lot of sense, good idea, thanks. Ill merge it on my fork and we can see!

It certainly does its thing:
image

feltech (approved):

.github/dependabot.yml
foundrytom: See what you think in the screenshot above. It also insists on the `:`.... but its prob not tooo bad is it?

.github/dependabot.yml
foundrytom: The ones it made in my account seem ok?

feltech: Yup, looks fine. Why can't docs just make sense.

.github/dependabot.yml
feltech: Yeah that's fine with me, unfortunate about the `:`, maybe it would be clever if you added a space and so not auto-add a `: `?  But anyway, fine as it is.

feltech (approved):
LGTM, for my part can be merged when ready.

elliotcmorris (approved):
Looks good, merge when ready.

PR #676, opened 2022-10-27
PyPI Release Process

### LINKED ISSUE #653: PyPi publishing
## What

PyPi packages published for:

 - `openassetio`

## ACs

- `pip install openassetio` installs a working python package in a clean environments on CY2022 python platforms [~Mac~/win/linux] (such that module are importable).
- Packages are auto-published based on a release being created in GitHub.
- Manual publish steps are known/documented in the case of GitHub downtime.
- Python versions: 3.9, 3.10 (x86_64)
- Linux: Needs to work on (Centos 7, and Ubuntu 20) [tested via pip install + pytest during release]
- Builds steps are idempotent

## Out of scope

- Windows ARM
- ~macOS~
- i386

## Implementation notes

 - It is beneficial to publish per platform/version when succeeded vs blocking all on one failure

- See OTIO for prior art in multi-dist wheel builds and pre-publish testing (testpypi) and check with JC where we plan to diverge.

Do this in same workflow : 
- On merge to `main`/`release` 
  - Using ManyLinux images, build wheels on matrixed platforms.
  - Once all wheels are built, upload all to testPyPI on an Ubuntu runner.
  - matrix test the uploaded wheels using pip install from testpypi and `python -m pytest` on all built platforms
- Only On `release`
  - upload to PyPi, after all prior steps have finished.

elliotcmorris:
Closes #653 

Release workflow for uploading to pypi, including verification tests.
Runs upon a release being created, or upon being manually triggered.

### [Release in action](https://github.com/elliotcmorris/OpenAssetIO/actions/runs/3339481952)

Above release run was done prior to this PR. (I had the changeset on main at the time to most fully replicate the expected environment.) Notice that the real pyPI upload fails, the reason for this being that this fork intentionally dosen't have the real pyPI secret on it.

Putting this up as a draft as there's a few things I want to run by some folks and perhaps change, and seeing the source is helpful. Although it's not really a lot of code anyway...

---
### Questions and Concerns

**1.** Python Venvs. We discussed doing the testing in a virtual environment, but i'm really not sure if it's necessary in this formulation. As the testing is done in a dependent job, one should be be able to assume a clean environment, and not really need to insulate. If it were up to me i'd remove the usage of venvs here. Thoughts?

**2.** This workflow uses 2 thirdparty actions, which i don't relish depending on, although i'd relish rolling our own even less.

-- **2.1.** The first, SebRollen/toml-action@v1.0.1, is used to parse our toml file, and extract the version string. This lets us pin the version for the test pip install, so we can be certain we're querying for the correct version. (This is good as it verifies we've actually uploaded all the architectures we're planning to test.)

-- **2.2.** The second, nick-fields/retry@v2, is used due to `testPyPI` not immediately reporting downloadable packages after upload, it can take a little while to propagate. During testing, occasionally some machines didn't report the new version for the platform as yet uploaded, but on retry, it functioned as expected. Normally a retry is not required, but occasionally I have been seeing this step needing be be reran once. A little bit dubious about this, because having time dependencies is never good. Whilst during testing I was observing gaps of 5-10s being the difference maker, it's no guarantee that we'll be able to see the libs even within the minute timeframe I've given the install step.

[here's an example of a run that failed due to this ](https://github.com/elliotcmorris/OpenAssetIO/actions/runs/3337524613/jobs/5524066932)

**3.** I could still be convinced that doing any testing at all on built wheels isn't worth it, even to this more limited, release time only degree. It would remove the dependency on testPyPI entirely, and allow us to remove practically all of the complexity from this workflow.

... Also added some emojis because I like the way they look in the github UI.

### DIFF
--- a/.github/workflows/deploy-pypi.yml
+++ b/.github/workflows/deploy-pypi.yml
@@ -1,0 +1,56 @@
+name: Deploy_PyPI
+
+on: 
+  release:
+    types: [published]
+  workflow_dispatch:
+
+jobs:
+  # testpypi and pypi are done as two seperate jobs due to a potential
+  # fail in the pypi upload, and wanting to be able to rerun just the
+  # mainline pypi upload. This can happen when pypi goes into read 
+  # only mode, which happens sometimes.
+  publish_testpypi:
+    name: Publish distribution 📦 to TestPyPI
+    runs-on: ubuntu-20.04
+    steps:
+      # This action will download the artifact for workflow runs that
+      # happen for this actual commit (the commit that the tag points to).
+      # It also restores the files timestamps.
+      # This does however mean that when making a release, you must be
+      # sure that the build-wheels job on main has completed
+      - name: Download wheels from commit ${{ github.sha }}
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: build-wheels.yml
+          workflow_conclusion: success
+          commit: ${{ github.sha }}
+          name: openassetio-wheels
+          path: dist
+
+      - name: Upload to TestPyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.TEST_PYPI_ACCESS_TOKEN }}
+          repository_url: https://test.pypi.org/legacy/
+
+  publish_pypi:
+    name: Publish distribution 📦 to PyPI
+    needs: publish_testpypi
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Download wheels from commit ${{ github.sha }}
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: build-wheels.yml
+          workflow_conclusion: success
+          commit: ${{ github.sha }}
+          name: openassetio-wheels
+          path: dist
+
+      - name: Upload to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_ACCESS_TOKEN }}


feltech (commented):
Looking good, just minor comments

.github/workflows/deploy-pypi.yml
```diff
@@ -0,0 +1,124 @@
...
+jobs:
+  call_build_wheels:
+    name: Call 🗣️🔨 Build Wheels
+    uses: ./.github/workflows/build-wheels.yml
+
+  publish_testpypi:
+    name: Publish distribution 📦 to TestPyPI
+    needs: call_build_wheels
+    runs-on: ubuntu-20.04
+    if: ${{ github.event_name == 'release' || github.event_name == 'workflow_dispatch'}}
```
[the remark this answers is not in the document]
elliotcmorris: Ah, no, this is leftover from when these steps were all inside `build-wheels`. Will remove.

.github/workflows/deploy-pypi.yml
```diff
@@ -0,0 +1,124 @@
...
+          path: dist
+
+      - name: Upload to TestPyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.TEST_PYPI_ACCESS_TOKEN }}
+          repository_url: https://test.pypi.org/legacy/
+
+  verify_test_pypi:
```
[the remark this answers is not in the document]
elliotcmorris: I agree that the benefits of uploading to TestPyPI at this point are somewhat ephemeral. 

The formal reason I'd think to do it is checking for metadata corruption on wheels before touching the real endpoint, as I believe uploads are done sequentially, so if uploading 6 files, and it rejects the 5th, you'll still have the initial 4 up. However, even this i'm not certain of...

I know we discussed this as not really a problem, but given the additional context we have from the pypa docs, I'd argue it's not a good idea to allow half-releases to go out, at least without making the decision to do it consciously.

However, this aside, a dry run for a release does _seem_ sensible, though I could be convinced to abandon it as yeah, I agree the justifications for doing it at all are falling away.

.github/workflows/deploy-pypi.yml
```diff
@@ -0,0 +1,124 @@
...
+            ${{ matrix.config.venv_activate }}
+            pip install -i https://test.pypi.org/simple/ openassetio==${{ steps.deployed_version.outputs.value }}
+
+      - name: Run Pytest
+        run : |
+          ${{ matrix.config.venv_activate }}
+          pip install -r ${{ github.workspace }}/tests/python/requirements.txt
+          pytest ${{ github.workspace }}/tests/python/openassetio
+
+  publish_pypi:
```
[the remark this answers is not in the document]
foundrytom: The problem with only being on `if release` is that we may need a way to re-try it in the event of some failure not related to the code in the SHA the release is anchored to (eg one of our occasional transient CI failures). As long as there is no other _automated_ way to publish to PyPI I _think_ that leaves us in a good place.

.github/workflows/deploy-pypi.yml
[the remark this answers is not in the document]
foundrytom: Lets go with
- Keep upload to TestPyPI as a canary for real PyPI upload failure.
- Don't run the retry'd tests
- Upload to PyPI

.github/workflows/deploy-pypi.yml
```diff
@@ -0,0 +1,124 @@
...
+        id: deployed_version
+        with:
+          file: '${{ github.workspace }}/pyproject.toml'
+          field: 'project.version'
+
+      - name: Setup Venv
+        run : |
+          python -m venv .venv
+
+      # Sometimes, PiPY won't publish platform end points quick enough,
```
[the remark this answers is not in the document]
elliotcmorris: Aye, that section now removed.

.github/workflows/deploy-pypi.yml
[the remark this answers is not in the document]
elliotcmorris: [06c64b6](https://github.com/OpenAssetIO/OpenAssetIO/pull/676/commits/06c64b64158d6052dd157735227086b130f8cbce)

.github/workflows/deploy-pypi.yml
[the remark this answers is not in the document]
elliotcmorris: Have removed all steps save for
- Call build wheels
- publish_testpypi
- publish_pypi

[6e9f93e](https://github.com/OpenAssetIO/OpenAssetIO/pull/676/commits/6e9f93ef8a46c242676b7142b675190a5f8febf6)

.github/workflows/deploy-pypi.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Upload to TestPyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.TEST_PYPI_ACCESS_TOKEN }}
+          repository_url: https://test.pypi.org/legacy/
+
+  publish_pypi:
+    name: Publish distribution 📦 to PyPI
+    needs: verify_test_pypi
```
foundrytom, line 38: [fix] Still referencing old verify - is it worth just combining these two jobs, so you get a fail on the step before publish to real PyPI? Then we only download the wheels once, etc..

.github/workflows/deploy-pypi.yml
```diff
@@ -0,0 +1,51 @@
+# - Build Python Wheels
+# - Deploy to TestPyPI
+# - Download from TestPyPI
+# - Run tests on downloaded artifacts
+# - Deploy to real PyPI
```
foundrytom, line 5: [fix] Remove?

elliotcmorris: Yep, no comment at all is really neccesary now as workflow is much less complex.

elliotcmorris: [33fc41d](https://github.com/OpenAssetIO/OpenAssetIO/pull/676/commits/33fc41d0b33b170428cca2d70ea1216ec697fdef)

setup.py
```diff
@@ -55,6 +55,7 @@ def build_extension(self, _ext: Extension):
                 # Ensure expected Python environment is discovered by
                 # CMake's `find_package` during the build.
                 f"-DPython_EXECUTABLE={sys.executable}",
+                "-DCMAKE_OSX_DEPLOYMENT_TARGET=10.15",
```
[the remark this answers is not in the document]
elliotcmorris: [0c7c64f](https://github.com/OpenAssetIO/OpenAssetIO/pull/676/commits/0c7c64fd370895af8cac3dee9e2b4c4cf7c160ce)

.github/workflows/deploy-pypi.yml
elliotcmorris: Hmm, i see your reasoning. There's more flexibility if we leave them separate, as you can re-run individual failed steps in the github UI if we needed to, (was helpful when testing.)

foundrytom: Cool - but the uses still needs updating right?

elliotcmorris: fixup changing the dependency name only : [c074e84](https://github.com/OpenAssetIO/OpenAssetIO/pull/676/commits/c074e84681e95e5bea09b5a5aff69f8e8275e3ea)

foundrytom (approved):
LGTM

elliotcmorris:
Dry run : https://github.com/elliotcmorris/OpenAssetIO/actions/runs/3345534955/jobs/5541256080

.github/workflows/deploy-pypi.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on: 
+  release:
+    types: [published]
+  workflow_dispatch:
+
+jobs:
+  call_build_wheels:
+    name: Call 🗣️🔨 Build Wheels
+    uses: ./.github/workflows/build-wheels.yml
```
JeanChristopheMorinPerso, line 11: Note that this is somewhat risky. There could be changes between the time your last R is merged and the time you trigger a release, which means you are never really sure if it's really gonna work or not. This is not a problem for pure python wheels, but it's a different story for CPython extensions.

JeanChristopheMorinPerso: It will run against the SHA of the release, but unless you pin everything (which is impossible because you can't pin to a specific version of the VM image), then you are at the mercy of GH. It's not much a problem for Linux since you build in a docker image (that if I'm not mistaken is aggressively pinned by cibuildwheel, but I might be mistaken and this would have to be checked), Windows and macOS builds are not containerized. Additionally, the actions versions are not pinned to the SHA commits. Action versions are moving versions (if you say you want an action version 1.1, it's not pinned at all. Bugfixes, etc can be pushed into 1.1).

foundrytom: This seems to be the prescribed way to share data between jobs within a single workflow? (See [here](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow)).

foundrytom: This log shows the wheel build is agains the release tag: https://github.com/elliotcmorris/OpenAssetIO/actions/runs/3345534955/jobs/5541185466#step:2:419

foundrytom: Right, got you - sorry thought you meant that it would be picking up an artifact from another run.

foundrytom: @JeanChristopheMorinPerso - thanks - Its a good point you raise.  To unblock the people waiting for wheels, we're going to use this as it is right now, as we always do a PR to bump the version almost immediately before release, and it seemed more likely that the wheel build action for the merge commit might not have finished yet which could cause the download step to fail. We'll def look into this as a follow up though as you make a really good point about the entirely non-deterministic nature of the build environment.

elliotcmorris: @JeanChristopheMorinPerso After some intense discussions, (and macos 10.15 being brownouted across github) we all came to the conclusion that you were right and are changing it to be just like what OTIO does.

foundrytom: Switched to the right way to do things [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/676/commits/e1f722c7db1543f8754fa24277eedfaa272b9899).

PR #680, opened 2022-10-31
Bump CI runners to macOS-11

foundrytom:
The issue was that 11 comes with Clang 13, for which there are no re-built packages on Conan. This adds `automake` tools so that Conan can build them locally. They should then be cached so hopefully this won't impact normal CI run times.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,8 @@ cmake_minimum_required(VERSION 3.21)
 # Additional include directories for CMake utils.
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
 
-# TODO(DF): Set CMAKE_OSX_DEPLOYMENT_TARGET when it's time to tackle OSX
+# Minimum required for the C++17 features used.
+set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15)
 
 # Forbid in-source builds.
 if (${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
--- a/resources/build/bootstrap-macos-11.sh
+++ b/resources/build/bootstrap-macos-11.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+# This bootstrap script is used by both the Github CI action and the
+# Vagrant VM configuration.
+
+set -xeo pipefail
+
+# macOS 11 runners ship with clang 13. Conan Center only has pre-built
+# binaries for clang 12. This allows packages to be built locally.
+brew install automake autoconf
+
+# Install additional build tools.
+pip3 install -r "$WORKSPACE/resources/build/requirements.txt"
+# Use explicit predictable conan root path, where packages are cached.
+export CONAN_USER_HOME="$HOME/conan"
+# Create default conan profile so we can configure it before install.
+# Use --force so that if it already exists we don't error out.
+conan profile new default --detect --force
+
+# Install openassetio third-party dependencies from public Conan Center
+# package repo.
+conan install --install-folder "$WORKSPACE/.conan" --build=missing \
+    "$WORKSPACE/resources/build"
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -12,7 +12,8 @@ cmake_minimum_required(VERSION 3.21)
 # Additional include directories for CMake utils.
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
 
-# TODO(DF): Set CMAKE_OSX_DEPLOYMENT_TARGET when it's time to tackle OSX
+# Minimum required for the C++17 features used.
+set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15)
```
elliotcmorris, line 16: [question] Isn't this the same variable as `MACOSX_DEPLOYMENT_TARGET` set in `build-wheels`? Does it need to be set twice?

elliotcmorris (approved):
:+1:

CMakeLists.txt
foundrytom: It needs to be set for any build in CMake to ensure that the Clang is set to the right macOS SDK version (otherwise it errors as we're using features in `std::filesystem` that are only available in macOS 10.15 or later). It need to be set for `cibuildwheel` so that it sets the correct metadata in the wheel so it won't install on systems earlier than 10.15 - as that process is the parent environment for CMake so it can't see the vars CMake sets.

elliotcmorris: :+1:

feltech:
> They should then be cached so hopefully this won't impact normal CI run times.

But first run of any PR, before the caching happens, will show

> Tests / macos-11 (pull_request) Successful in **11m **

We already have #612 to look into better Github caching.

resources/build/bootstrap-macos-11.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+# This bootstrap script is used by both the Github CI action and the
+# Vagrant VM configuration.
+
+set -xeo pipefail
+
+# macOS 11 runners ship with clang 13. Conan Center only has pre-built
+# binaries for clang 12. This allows packages to be built locally.
```
feltech, line 8: [question] Is this the case in general, or specifically the cpython package?

feltech (approved):
Looks good, just a minor question

resources/build/bootstrap-macos-11.sh
foundrytom: cpython, and many of its dependencies.

foundrytom:
> But first run of any PR, before the caching happens, will show

Ahh, I thought you had that clever cache fallback where it should re-use the cache from a previous PR?

feltech:
> > But first run of any PR, before the caching happens, will show
> 
> Ahh, I thought you had that clever cache fallback where it should re-use the cache from a previous PR?

No don't think we added anything like that.  IIRC the cache will fall back to `main`'s cache, but we don't do any building on `main` so there is no cache to fall back on.

foundrytom:
> > Ahh, I thought you had that clever cache fallback where it should re-use the cache from a previous PR?
> 
> No don't think we added anything like that. IIRC the cache will fall back to `main`'s cache, but we don't do any building on `main` so there is no cache to fall back on.

I thought that's what [this](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/.github/bootstrap_platform/action.yml#L26) was for - as it uses that mega unintuitive GitHub feature of sub-string matching.

feltech:
> > > Ahh, I thought you had that clever cache fallback where it should re-use the cache from a previous PR?
> > 
> > 
> > No don't think we added anything like that. IIRC the cache will fall back to `main`'s cache, but we don't do any building on `main` so there is no cache to fall back on.
> 
> I thought that's what [this](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/.github/bootstrap_platform/action.yml#L26) was for - as it uses that mega unintuitive GitHub feature of sub-string matching.

That's a trick to simulate "overwriting" the cache within the same PR.  I think it's only really useful at the moment if you change Conan dependencies during a PR.  But will be useful if/when `main` has a cache that can be used as a fallback.

PR #681, opened 2022-10-31
Bump pybind to 2.10.0, build for Python 3.11

### LINKED ISSUE #683: Support Python 3.11
## What

Ensure `openassetio` wheels can be built for Python 3.11

## Why

Ensures we don't get too far out of sync with Python as it evolves

## Tasks

- Update pybind to 2.10.0
- Update tests to cover re-worded exceptions [should we be validating the actual message when it originates from outside of our code?].

foundrytom:
Hopefully should allow us to support Python 3.11 wheels

Closes #683

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -33,7 +33,7 @@ jobs:
           # These are the platform build strings provided to
           # cibuildwheel, with wildcarding. See
           # https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip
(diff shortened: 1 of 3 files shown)


foundrytom:
Low priority - on hold until 3.7 support is in place

foundrytom:
Capturing note from #726 - The text of the exception message changes between Python versions - we should perhaps not check the exact wording if it is not under our control, and just check the exception is of the correct type.

foundrytom:
Closing pending 3.11 being available in Conan

PR #694, opened 2022-10-31
Reinstate Python 3.7 support

### LINKED ISSUE #660: Re-instate support for Python 3.7
## What

~Decide which Python (minor) versions to support~

Early adopters have stated the need for Python 3.7. We need to ensure there are no newer python features in the code base, and update CI and build infrastructure to use 3.7 by default, but also build for latest. Packages should be published for 3.7 -> latest.

## Why

This affects which Python features we can use (e.g. `list[str]` type annotation is 3.9+) , our CI config, and other ongoing maintenance burdens.

## Tasks

- [ ] Wrap implicit casts of Enums to int
- [ ] #662

foundrytom:
Fortunately this wasn't anything too onerous.

Closes #660


resources/codegen/tests/test___main__.py
```diff
@@ -213,8 +213,10 @@ def test_when_none_set_then_all_languages_are_generated(self, tmp_path, yaml_pat
 
 
 def execute_cli(*args):
-    all_args = [sys.executable, "-m", "openassetio_codegen"]
-    all_args.extend(args)
+    # Str wrapping prevents issues with Path objects in Windows/3.7
```
elliotcmorris, line 216: [Consider][Minor] Windows/3.7 could be a tad confusing. Maybe Windows/Python 3.7?

tests/python/openassetio/test_log.py
```diff
@@ -114,17 +114,27 @@ def test_when_shouldColorOutput_set_false_then_messages_are_not_colored(self, ca
...
+            == "progress"
+        )
+        assert lg.LoggerInterface.kSeverityNames[int(lg.LoggerInterface.Severity.kInfo)] == "info"
+        assert (
+            lg.LoggerInterface.kSeverityNames[int(lg.LoggerInterface.Severity.kDebug)] == "debug"
+        )
+        assert (
+            lg.LoggerInterface.kSeverityNames[int(lg.LoggerInterface.Severity.kDebugApi)]
+            == "debugApi"
         )
```
elliotcmorris, line 138: [Consider][Minor] I suspect the inconsistent indentation here is a consequence of consistent line wrapping. Due to how some of these statements split into 4 lines now while some still only take one, maybe consider slightly breaking the rule here for a consistent, (and i think more readable) block.

elliotcmorris (approved):
Looks great. Two unimportant style comments.

tests/python/openassetio/test_log.py
foundrytom: Yeah, it does my nut, but the issue with manually massaging it is that we can't the use auto formatters, and the CI black check will fail as it'll want to reformat it for us...

resources/codegen/tests/test___main__.py
foundrytom: Updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/694/commits/819f332c17bfafb11dd74261254121a439f03cf1).

pyproject.toml
```diff
@@ -12,7 +12,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "openassetio"
 version = "1.0.0a5"
-requires-python = ">=3.9"
+requires-python = ">=3.7"
```
feltech, line 15: [consider][minor] Do we want to update the `classifiers` below too?

BUILDING.md
```diff
@@ -16,7 +16,7 @@ following packages are installed:
 
 Binary builds additionally require the following packages to be
 available to the CMake build system.
-- Python 3.9 (development install)
+- Python 3.7 (development install)
```
feltech, line 19: [plan] In order to test this PR I wanted to run 
```
conan install -if .conan --require-override cpython/3.7.12 resources/build
````
i.e., making use of `--require-override` to force use of 3.7. 

However, our conanfile has a mistake (my bad), similar to that alluded to in https://github.com/OpenAssetIO/OpenAssetIO/issues/654#issuecomment-1290756213, but for other reasons.  

Since `cpython` is a `tool_requires` rather than normal `requires`, the `--require-override` has no effect. [The docs](https://docs.conan.io/en/latest/reference/commands/consumer/install.html#require-override) say this explicitly

> This feature affects only to regular requires, not to tool_requires or python_requires, as those don’t have such an overriding mechanism, and they are private to their consumer, not propagating downstream nor upstream.

We need to split our `build_requirements` method in `conanfile.py` into separate `requirements` and `build_requirements`, where the `requirements` method contains `requires()` calls for `cpython` and `pybind11`, and the `build_requirements` method contains the `tool_requires()` calls for the unit testing libraries.

File: resources/codegen/tests/test___main__.py [hunk 5]
```diff
@@ -213,8 +213,11 @@ def test_when_none_set_then_all_languages_are_generated(self, tmp_path, yaml_pat
 
 
 def execute_cli(*args):
-    all_args = [sys.executable, "-m", "openassetio_codegen"]
-    all_args.extend(args)
+    # Str wrapping prevents issues with Path objects in Windows
+    # Python 3.7
+    all_args = [str(sys.executable), "-m", "openassetio_codegen"]
```
feltech, line 218: [consider][minor] Pretty sure stringifying `sys.executable` is unnecessary

`sys.executable` [should be a string](https://docs.python.org/3.7/library/sys.html#sys.executable)

> A string giving the absolute path of the executable binary for the Python interpreter

feltech (approved):
A particular shame to lose the implicit conversion from enum to int, but hopefully that's not something we'll use often (esp. since we now have conveniences like `log.debug` to use instead).

pyproject.toml
foundrytom: Good spot, thanks, yes. Added 3.7 [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/694/commits/488c8ad2c31e511c3e946731d0eb5e7e7b658546).

resources/codegen/tests/test___main__.py
foundrytom: Good spot - it seems like it must have been a red-herring before. Removed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/694/commits/2173016989f7d996d36a9ec79d74f646174fb382.

BUILDING.md
foundrytom: https://github.com/OpenAssetIO/OpenAssetIO/issues/699

feltech (approved):
LGTM, ready to squash and merge for my part

PR #697, opened 2022-10-31
[CI] Set concurrency groups for CI jobs

### LINKED ISSUE #696: CI: Add concurrency group to PR jobs
## What

Add a concurrency group, with preemption(?) that will cancel existing runs for a PR when it is force-pushed to.

## Why

Now we have many jobs, it wastes resource, and waiting for runners is a more of an issue.

foundrytom:
This should ensure that we don't waste runners repeatedly running jobs for out-of-date commits.

Anchoring to `github.ref`[1] seems like it should "do the right thing" so they are unique to each PR/etc.

[1] https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

Closes #696 

It can be seen in action [here](https://github.com/foundrytom/OpenAssetIO/actions).

### DIFF
--- a/.github/workflows/build-instructions.yml
+++ b/.github/workflows/build-instructions.yml
@@ -7,6 +7,10 @@ on:
     paths-ignore:
       - 'docs/**'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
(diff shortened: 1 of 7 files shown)


feltech (approved):
LGTM

PR #698, opened 2022-10-31
Remove BAL

### LINKED ISSUE #672: Extract BAL to independent repo
## What

Extract `resources/examples/manager/BasicAssetLibrary` into `OpenAssetIO-Manager-BAL`.

## Why

- Dog-foods the process of developing a manager plugin with OpenAssetIO as an external dependency.
- BAL is used by integrations for testing and so it more easily installed as a standalone project vs somehow embedding it in the openassetio release.
- As part of consolidating the core library, we want to keep it's versioning as stable as possible. This allows iteration of BAL independently of API release semver.
- BAL makes use of the `ManagedTrait` as part of `managementPolicy`, we soon plan to split those out of the core API into OpenAssetIO-MediaCreation, this breaks what otherwise becomes a cyclic dependency between that and OpenAssetIO.

## ACs

- BAL is extracted to its own sem-ver managed repository - `OpenAssetIO/OpenAssetIO-Manager-BAL`
- Integration workflow added to OpenAssetIO CI that pulls bal and runs it's API compliance suite against each PR
- `otio-openassetio` instructions/tests updated to clone BAL as pre-step
- `simpleResolver` example instructions/tests updated to clone BAL as pre-step

## Notes

This does also pave the way to `python -m pip` install as a package and automatically load in the `PythonPluginSystemManagerImplementationFactory` as per OTIO plugins, which would greatly simplify setup for a variety off user and end-user cases.

foundrytom:
Closes #672.

Takes a stab at an interim testing solution that, combined with https://github.com/OpenAssetIO/OpenAssetIO-Manager-BAL/pull/3, should get us good coverage in day-to-day operations. This idea of an umbrella repo that uses submodules or similar to put together a sensible 'TDD environment' might be a nice extension/evolution of this.]

~NB: I think there is something suspicious going on with the windows build in integrations, will check in the morning.~ @elliotcmorris sorted it 👍 🙏


foundrytom:
🤔 I wonder if there is a way we can make the integrations workflow depend on the wheel build, and then we can just use those, which would save redundant work, though probably not that much time.

elliotcmorris:
> NB: I think there is something suspicious going on with the windows build in integrations, will check in the morning.

This'll be because it's trying to use gcc as a compiler, because you're not in a dev shell (havn't run VCVARSALL), you'll need to do the prefix thing we do elsewhere before doing `pip install .`

.github/workflows/integrations.yml
```diff
@@ -0,0 +1,71 @@
...
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Bootstrap
+        uses: ./.github/bootstrap_platform
+
+      - name: Build OpenAssetIO (setup.py)
+        run: |
+          python -m pip install .
```
elliotcmorris, line 45:
```suggestion
        run: |
          ${{ matrix.config.preamble }}
          python -m pip install .
```

foundrytom:
> > NB: I think there is something suspicious going on with the windows build in integrations, will check in the morning.
> 
> This'll be because it's trying to use gcc rather than msbuild as a compiler, because you're not in a dev shell (havn't run VCVARSALL), you'll need to do the preamble thing we do elsewhere before doing `pip install .`

Legend, thanks! Sorted 👍

resources/examples/host/simpleResolver/README.md
```diff
@@ -24,14 +24,13 @@ optional arguments:
 ## Example
 
 > Note: This script requires the `openassetio` module to be available to
-> Python. Presently, this requires manual installation, see
-> [BUILDING.md](../../../../BUILDING.md) in the main OpenAssetIO
-> repository for more details.
+> Python. This can be installed via
+> `python -m pip install -r requirements.txt`.
```
feltech, line 28: [question] It seems odd that the instructions for `simpleResolver` would have you install `openassetio` from PyPI, at least whilst it's part of this repo?

tests/openassetio-python/CMakeLists.txt
```diff
@@ -173,12 +173,18 @@ openassetio_add_pytest_target(
 # - openassetio-python-venv
 openassetio_add_pytest_target(
     openassetio.internal.resources-examples-pytest
-    "Running pytest for examples"
+    "Running pytest for Manager examples"
```
feltech, line 176: [question] Should we bother with any tests under `examples` now?  There's only SAM, and that's so bare bones it doesn't really test much of anything.  Maybe SAM should just disappear altogether, for now?

feltech (approved):
Looks good, I like the CI behaving like a real integration. Just minor questions.

resources/examples/host/simpleResolver/README.md
foundrytom: A lot easier for people than having to build it from source?

tests/openassetio-python/CMakeLists.txt
foundrytom: I'd be happy with that, maybe migrate it to a "template manager plugin"?

foundrytom: Happy to do this in a follow up PR?

.github/workflows/integrations.yml
```diff
@@ -0,0 +1,72 @@
...
+        uses: ./.github/bootstrap_platform
+
+      - name: Build OpenAssetIO (setup.py)
+        run: |
+          ${{ matrix.config.preamble }}
+          python -m pip install .
+        env:
+          CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
+
+      - name: Test dependencies
```
elliotcmorris, line 50: [Consider][Minor]
```suggestion
      - name: Install test dependencies
```

elliotcmorris (approved):
Looks great.

resources/examples/host/simpleResolver/README.md
foundrytom: .. as in, if they're just looking for an example to guide their python dev

feltech: OK that's fine, it just looked a bit tangential to this PR, and I'd imagined `simpleResolver` being used to test a build of the project, rather than treated as a standalone project, at least whilst it's in the same repo.

PR #711, opened 2022-11-02
Update Building.md with new build instructions.

### LINKED ISSUE #624: Update `BUILDING.md`

elliotcmorris:
Closes #624 

Updates Building.md to contain the new build instructions following the pip work, and attempts to refresh the entire document to be clearer and more cohesive.

Rendered version of the new doc [here](https://github.com/OpenAssetIO/OpenAssetIO/blob/a6c3c6d6bc11b4abe477f765b3086eeefca3649c/BUILDING.md).

### DIFF
--- a/.github/workflows/build-instructions.yml
+++ b/.github/workflows/build-instructions.yml
@@ -59,7 +59,46 @@ jobs:
     - name: Install using pip
       run: |
         ${{ matrix.config.preamble }}
-        pip install .
+        python -m pip install .
+      env:
+        CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
+
+    - name: Install test dependencies
+      run: pip install -r tests/python/requirements.txt
+
+    - name: Run Python tests
+      run: pytest tests/python/openassetio
+
+  editable_pip_install:
+    name: Editable pip install on  ${{ matrix.config.os }}
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - os: windows-2019
+            preamble: call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat" x64
+            shell: cmd
+          - os: ubuntu-20.04
+            shell: bash
+          - os: macos-11
+            shell: bash
+    defaults:
+      run:
+        # Annoyingly required here since `matrix` isn't available in
+        # the `shell` property of individual steps.
+        shell: ${{ matrix.config.shell }}
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Bootstrap
+      uses: ./.github/bootstrap_platform
+
+    - name: Install using pip
+      run: |
+        ${{ matrix.config.preamble }}
+        python -m pip install --editable .
       env:
         CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -121,9 +121,6 @@ the Python module.
 In installing the Python module, OpenAssetIO creates a valid Python
 package, placed at
 `build/dist/lib/{python-version}/site-packages/openassetio`.
-This generated python package can then be used directly (e.g. by
-extending `PYTHONPATH` and/or `LD_LIBRARY_PATH`), or as the source for a
-further packaging step.
 
 > **Note**
 >
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,50 +1,28 @@
 # Building OpenAssetIO
 
-## Quick start: Python only
+## Contents
 
-For use in a Python-only environment, the Python package can be
-installed from the root of the repository using
-
-```shell
-pip install .
-```
-
-This will automatically build and install the Python extension module,
-along with the Python sources.
-
-> **Note**
->
-> CMake's `find_package` must be able to locate dependencies.
-> Consider using the [`CMAKE_TOOLCHAIN_FILE`](https://cmake.org/cmake/help/v3.24/envvar/CMAKE_TOOLCHAIN_FILE.html)
-> environment variable to augment CMake's package search paths.
->
-> For example: to use `conan` to resolve dependencies, from the root of
-> the repository run (assuming a Linux host, translate as necessary):
->
-> ```shell
-> conan install -if .conan resources/build --build=missing
-> export CMAKE_TOOLCHAIN_FILE=`pwd`/.conan/conan_paths.cmake
-> pip install .
-> ```
-
-This will install a single binary module with the OpenAssetIO core
-library statically linked into it, and so is not suitable for use in a
-C++ application.
-
-For users that wish to modify the Python sources and test their changes
-without reinstalling, [editable installs](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs)
-are also supported
-
-```shell
-pip install --editable .
-```
+- [System requirements](#system-requirements)
+- [Library dependencies](#library-dependencies)
+  - [Build dependencies](#build-dependencies)
+  - [Test dependencies](#test-dependencies)
+- [Building](#building)
+  - [Building via pip](#building-via-pip)
+- [CMake options](#cmake-options)
+  - [Presets](#presets)
+- [Running tests](#running-tests)
+  - [Using `ctest`](#using-ctest)
+- [Sandboxed builds](#sandboxed-builds)
+  - [Docker](#docker)
+  - [Vagrant](#vagrant)
 
 ## System requirements
 
 We assume the following system packages are installed
 
 - CMake 3.21+
-- GCC 9.3
+- C++17 or later compliant compiler
+(GCC 9.3+ / MSVC 16.9+ / macOS 10.15+)
 
 For linting and other build verification options, we assume the
 following packages are installed:
@@ -53,30 +31,37 @@ following packages are installed:
 
 ## Library dependencies
 
-Binary builds additionally require the following packages to be
-available to the CMake build system.
+### Build dependencies
 
-- Python 3.7 (development install)
-- [pybind11](https://pybind11.readthedocs.io/en/stable/) 2.6.1+
+Binary builds additionally require the following packages to be
+available.
+
+- [Python 3.7+](https://www.python.org/) (development install)
+- [pybind11](https://pybind11.readthedocs.io/en/stable/) 2.8.1+
+
+### Test dependencies
 
 Building and running tests requires the following additional packages:
 
-- catch2 2.13
-- trompeloeil 42
+- [catch2](https://catch2.docsforge.com/) 2.13
+- [trompeloeil](https://github.com/rollbear/trompeloeil) 42
 
 We use the CMake build system for compiling the C++ core library and
-its Python bindings. As such, the library dependencies listed above must
-be discoverable by CMake's `find_package`.
+its Python bindings. As such, a library being available means that it
+must be discoverable by CMake's [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html).
 
-For reference, during development of OpenAssetIO we use the [conan](https://conan.io/)
-package manager, and source the dependencies from the default public
-[ConanCenter](https://conan.io/center/) repository. We therefore assume
-the presence of CMake build targets as exported by these packages.
+During development, we tend use [ConanCenter](https://conan.io/center/)
+to fulfill the [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html) requirement.
+See [Building](#building) for more information.
 
-> Warning: if attempting to build on CentOS 7 using the ConanCenter
+> **Warning**
+>
+> If attempting to build on CentOS 7 using the ConanCenter
 > Python 3.7 package you will likely hit an [issue installing pkgconf](https://github.com/conan-io/conan-center-index/issues/8541).
 
-> Warning: On macOS, by default, CMake prefers Framework installations
+> **Warning**
+>
+> On macOS, by default, CMake prefers Framework installations
 > (see: [$CMAKE_FIND_FRAMEWORK](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_FRAMEWORK.html)).
 > This means dependencies such as Python may default to the system
 > installation despite alternatives being present. In addition, if you
@@ -84,22 +69,15 @@ the presence of CMake build targets as exported by these packages.
 > alternative, we have encountered missing header issues that may
 > require additional work/configuration to resolve.
 
-## Building with CMake
+## Building
 
-We use the CMake build system for compiling the C++ core library and
-its Python bindings.
+OpenAssetIO consists of a core C++ project, as well as a set of C and
+Python bindings to that core library. These modules are organized into a
+CMake project, and are all built together by default.
 
-By default, the project will build shared libraries and set the
-installation directory to `dist` under the build directory. This can be
-overridden by setting the built-in CMake variables `BUILD_SHARED_LIBS`
-and `CMAKE_INSTALL_PREFIX`, respectively.
-
-If shared library builds are disabled, then the core library will be
-built and installed as a static library, and statically linked into the
-Python module.
-
-In order to build and install the binary artifacts and Python sources,
-assuming the current working directory is at the repository root, run
+Assuming your build environment has the required dependencies,
+[(see above)](#build-dependencies), OpenAssetIO can be built and installed by
+running:
 
 ```shell
 cmake -S . -B build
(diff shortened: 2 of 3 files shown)


BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+> locations via referencing the generated `conan_paths.cmake`.
+
+The modules are installed to `/build/dist`. This default location can
+be overridden by setting `CMAKE_INSTALL_PREFIX` prior to build.
+
+By default, OpenAssetIO builds as a shared library. This can be
+overriden by setting `BUILD_SHARED_LIBS` prior to build.
+If shared library builds are disabled, then the core library will
+be built and installed as a static library, and statically linked into
+the Python/C modules.
```
elliotcmorris, line 117: I've tried to reference `Python/C` instead of just Python as it was before, I'd appreciate some verification that this is actually correct in the places I've written it, because I'm not sure.

BUILDING.md
```diff
@@ -1,43 +1,19 @@
...
-are also supported
-
-```shell
-pip install --editable .
-```
+## Contents
+
+- [System requirements](#system-requirements)
+- [Library dependencies](#library-dependencies)
+  - [Build Dependencies](#build-dependencies)
```
foundrytom, line 7:
```suggestion
  - [Build dependencies](#build-dependencies)
```

BUILDING.md
```diff
@@ -1,43 +1,19 @@
...
+  - [Test Dependencies](#test-dependencies)
```
foundrytom, line 8:
```suggestion
  - [Test dependencies](#test-dependencies)
```

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
 
 ## Library dependencies
 
+### Build Dependencies
```
foundrytom, line 32:
```suggestion
### Build dependencies
```

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+
 Binary builds additionally require the following packages to be
-available to the CMake build system.
+available.
 
 - Python 3.7 (development install)
 - [pybind11](https://pybind11.readthedocs.io/en/stable/) 2.6.1+
 
+### Test Dependencies
```
foundrytom, line 40:
```suggestion
### Test dependencies
```

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
 
 ## Library dependencies
 
+### Build Dependencies
+
 Binary builds additionally require the following packages to be
-available to the CMake build system.
+available.
 
 - Python 3.7 (development install)
```
foundrytom, line 37: [consider] Linking to Python here, given we're linking to pybind?

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
-built and installed as a static library, and statically linked into the
-Python module.
+OpenAssetIO consists of a core C++ project, as well as a set of C and
+Python bindings to that core libary. These modules are organized into a
+CMake project, and are all built together by default.
 
-In order to build and install the binary artifacts and Python sources,
-assuming the current working directory is at the repository root, run
+Assuming your build environment has the required dependencies,
+[(See Above)](#build-dependencies), OpenAssetIO can be built and installed by
```
foundrytom, line 77:
```suggestion
Assuming your build environment has the required dependencies
[(see above)](#build-dependencies), OpenAssetIO can be built and installed by
```

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+>
+
+### Building via pip
+
+OpenAssetIO can alternately be built and installed via pip, which may be
+preferable to the exclusively python focused developer.
+Check out the repository, and from the root, run :
+
+```shell
+pip install .
```
foundrytom, line 139:
```suggestion
python -m pip install .
```

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+> exclusively python focused developer. Anyone wishing to write
+> non-python hosts or managers should build via the CMake method
+> described above.
+
+For users that wish to modify the Python sources and test their changes
+without reinstalling, [editable installs](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs)
+are also supported
+
+```shell
+pip install --editable .
```
foundrytom, line 167:
```suggestion
python -m pip install --editable .
```

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+```
+
+## CMake Options
```
foundrytom, line 170: ❤️

BUILDING.md
```diff
@@ -144,18 +224,40 @@ ctest
 This will build and install binary artifacts and Python sources, create
```
foundrytom, line 224: [fix] I can't add a comment to the right line, but how about changing `ctest` to `ctest --test-dir build` then you don't need to change directory (given we run all the other commands from the root of the repo).

BUILDING.md
```diff
@@ -171,10 +273,3 @@ cmake -S openassetio -B build --install-prefix ~/dist \
 
 Then we can run the usual steps (see above) to build and install
 OpenAssetIO, and run the tests (if enabled).
-
-## CMake presets
```
foundrytom, line 175: [question] Any reason not to mention the presets any more? Maybe they could do with the options as its very closely related - ie; so people aren't bamboozled by the options list?

README.md
```diff
@@ -158,35 +158,18 @@ For pure Python projects, OpenAssetIO is available on PyPI, simply run:
...
 For detailed instructions see [BUILDING](BUILDING.md).
 
+> **Note**
+>
+> Currently, C++ projects require a from-source build.
+> A convenient way to achieve this out of the box is to use a sandboxed
+> build workflow, with the [ASWF Docker image](https://github.com/AcademySoftwareFoundation/aswf-docker).
+>
+> You can find instructions for that build workflow [here](BUILDING.md#sandboxed-builds).
+
```
foundrytom, line 172: [consider] It feels like the order of information might be a little inverted in this section. Can we add the bit about C++ builds needing a source build before the link to `BUILDING.md`?

eg, open with:

> C++ projects require a from-source build, the most convenient way to achieve this is to use the [ASWF Docker image](BUILDING.md#sandboxed-builds) workflow.
> 
> For detailed build instructions and information on dependency requirements, see [BUILDING](BUILDING.md).

foundrytom (changes requested):
👌 

This is great @elliotcmorris love the new structure. Mostly very minor formatting consistency fixes.

README.md
elliotcmorris: I guess I was thinking that keeping the "C++ require you to build manually" thing as a suffix was appropriate due to the hopefully temporary nature of that statement, but I guess there's no reason to alter the layout of the docs based on what's about to happen.

BUILDING.md
feltech: Not quite. Assuming `BUILD_SHARED_LIBS=OFF` (i.e. static library build), then `openassetio-core`, `openassetio-core-c` and `openassetio-python` will all be static libraries; whereas `_openassetio` (and `_openasset_test`) will remain dynamic libraries (because they're Python extension modules).

So it is not correct to say that that "the core library will be [...] statically linked into the C module".  The C library will depend on the core library (in the CMake target dependencies sense), but not merged into it in any way.

README.md
foundrytom: Right got you - that makes sense - I suppose which ever way - it might be good to add a little more context for the "other building methods" (eg why you might need them?)

BUILDING.md
feltech: In general I'd shy away from mentioning the C API much at all, since it doesn't do anything useful at the moment.

BUILDING.md
feltech: I'm curious about this preference for `python -m pip`?  Is it just that it works for a venv without having to activate it?

BUILDING.md
```diff
@@ -1,43 +1,19 @@
...
-For users that wish to modify the Python sources and test their changes
-without reinstalling, [editable installs](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs)
-are also supported
-
-```shell
-pip install --editable .
-```
+## Contents
+
+- [System requirements](#system-requirements)
```
feltech, line 5: [question] Was this TOC generated, or did you create it manually? Would be good to know the auto-generator used so we can use it again when this doc is updated.  Not sure where to put that info though.

BUILDING.md
elliotcmorris: casing fixup : [53349c6](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/53349c64da59bf0be52ae1a6fdc6682cb97a1443)

BUILDING.md
elliotcmorris: casing fixup : [53349c6](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/53349c64da59bf0be52ae1a6fdc6682cb97a1443)

BUILDING.md
elliotcmorris: casing fixup : [53349c6](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/53349c64da59bf0be52ae1a6fdc6682cb97a1443)

BUILDING.md
elliotcmorris: casing fixup : [53349c6](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/53349c64da59bf0be52ae1a6fdc6682cb97a1443)

BUILDING.md
elliotcmorris: casing fixup : [53349c6](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/53349c64da59bf0be52ae1a6fdc6682cb97a1443)

BUILDING.md
```diff
@@ -1,43 +1,19 @@
...
+- [Building](#building)
+  - [Building via pip](#building-via-pip)
+- [CMake Options](#cmake-options)
+- [Running tests](#running-tests)
+  - [Using `ctest`](#using-ctest)
+- [Sandboxed builds](#sandboxed-builds)
+  - [Docker](#docker)
+  - [Vagrant](#vagrant)
 
 ## System requirements
```
feltech, line 19: [consider] I wonder if we should  be a bit more relaxed here, or at least note that these are the system requirements we assume for the following sections, but "MacOS and Windows works too, translate as necessary".

BUILDING.md
feltech: [consider]
```suggestion
- Python 3.7+ (development install)
```

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
 ## Library dependencies
 
+### Build Dependencies
+
 Binary builds additionally require the following packages to be
-available to the CMake build system.
+available.
 
 - Python 3.7 (development install)
 - [pybind11](https://pybind11.readthedocs.io/en/stable/) 2.6.1+
```
feltech, line 39: [consider] Not sure that pybind11 will still work, we're currently using 2.8.1 (but then that'll change too once we have MacOS 11 support from #681)

BUILDING.md
elliotcmorris: I'd actually linked all the dependencies, before undoing them with a "I don't wanna deal with discussions around that" thought.

Added them back in (Not system dependencies, that seemed inappropriate)
[a270dbb](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/a270dbb71d57199b1db2a80c2ca7abbba501ee71)

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
-Python module.
+OpenAssetIO consists of a core C++ project, as well as a set of C and
+Python bindings to that core libary. These modules are organized into a
+CMake project, and are all built together by default.
 
-In order to build and install the binary artifacts and Python sources,
-assuming the current working directory is at the repository root, run
+Assuming your build environment has the required dependencies,
+[(See Above)](#build-dependencies), OpenAssetIO can be built and installed by
+running :
```
feltech, line 79: [consider]
```suggestion
running:
```

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+> **Note**
+>
+> A common error to encounter is to find that CMake cannot resolve
+> its dependencies via [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html)
+>
+> A convenient way to solve this is to use the [`CMAKE_TOOLCHAIN_FILE`](https://cmake.org/cmake/help/v3.24/envvar/CMAKE_TOOLCHAIN_FILE.html)
+> environment variable to augment CMake's package search paths.
+>
+> For example, to use [`conan`](https://docs.conan.io/en/1.46/installation.html)
+> to resolve dependencies, (as we do during internal development) run
```
feltech, line 96: [consider]
```suggestion
> to resolve dependencies (as we do during internal development), run
```

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+> the following commands from the root of the repository prior to
+> building :
```
feltech, line 98: [consider]
```suggestion
> building:
```

BUILDING.md
foundrytom: https://snarky.ca/why-you-should-use-python-m-pip/

foundrytom: eg, on my Mac, some of my venvs don't have `pip` in (not looked into why yet), so activating them , uses the wrong python when you run `pip` but not `python -m pip`, so its universally safer.

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+>
+> ```shell
+> conan install -if .conan resources/build --build=missing
+> export CMAKE_TOOLCHAIN_FILE=`pwd`/.conan/conan_paths.cmake
```
feltech, line 102: [consider][minor] An alternative, if you prefer, is to use the `--toolchain` argument to `cmake` at configure time, i.e. ```cmake -S . -B build --toolchain `pwd`/.conan/conan_paths.cmake```.

Not saying that's any better, just making sure we've considered the options.

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+> ```
+>
+> (These commands assume a Linux host, translate as necessary).
+>
+> This installs OpenAssetIO's dependencies as described in
+> `/resources/build/conanfile.py`, and allows CMake to deduce the
+> locations via referencing the generated `conan_paths.cmake`.
+
+The modules are installed to `/build/dist`. This default location can
+be overridden by setting `CMAKE_INSTALL_PREFIX` prior to build.
```
feltech, line 112: [consider][minor] Or by passing `--install-prefix` to `cmake` at configure time.  We probably don't want to confuse things by adding several alternatives, though.

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+
+By default, OpenAssetIO builds as a shared library. This can be
+overriden by setting `BUILD_SHARED_LIBS` prior to build.
+If shared library builds are disabled, then the core library will
+be built and installed as a static library, and statically linked into
+the Python/C modules.
+
+In installing the python module, OpenAssetIO creates a valid python
```
feltech, line 120: [consider]
```suggestion
In installing the Python module, OpenAssetIO creates a valid Python
```

BUILDING.md
elliotcmorris: [d571a5a](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/d571a5aee9a1749f91b0c15284e6f4b5371f50d3)

BUILDING.md
elliotcmorris: [2881a25](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/2881a25efb589ea3df43a4260a287dfa49bd9314)

BUILDING.md
elliotcmorris: Yeah great idea, love the consistency. [5a4e6b2](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/5a4e6b2a4a782217b76cf5ec2890827a026ef7d1)

BUILDING.md
elliotcmorris: Hmm, not sure why we originally decided to omit them, i guess they seemed a little unmoored. Nonetheless I love the idea of framing them as extentions to the cmake options section, [cbec5fa](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/cbec5fa797f4bdb8417b3cf9d4f4ea59454061bd)

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+install tree, placed at `build/dist/{python-version}/site-packages`.
```
feltech, line 121: [consider] Not quite convinced on the wording here.  I think a "valid Python install tree" isn't really a thing - you can extend `PYTHONPATH` into any directory you want.  Oh I think you mean a Python package?  Yeah perhaps best to use the word "package"
```suggestion
package, placed at `build/dist/lib/{python-version}/site-packages/openassetio`.
```
It's also obfuscating the C++ component. The C/C++ libraries is what the reference to `LD_LIBRARY_PATH` was originally about, and the "further packaging step" was also originally about packaging along with the non-Python artifacts (e.g. Conan, Rez, tar.gz).

feltech, line 121: [fix]
```suggestion
install tree, placed at `build/dist/lib/{python-version}/site-packages`.
```

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+>
+> When built this way, the `openassetio` Python package will not
+> function unless the binary artifacts are also made available to your
+> Python environment.
+>
+
+### Building via pip
+
+OpenAssetIO can alternately be built and installed via pip, which may be
+preferable to the exclusively python focused developer.
```
feltech, line 136: [consider]
```suggestion
OpenAssetIO can alternately be built and installed via `pip`, which may be
preferable to the exclusively Python focused developer.
```
A while ago we did discuss a policy for capitalisation of words, I think we said that if referring to the command you're executing, then it should be in backticks, e.g. `python`, otherwise capitalised as the project does themselves, e.g. "Python" vs. "pybind11" - @foundrytom does that sound right?

README.md
elliotcmorris: Yeah. I've taken a bit of a less-is-more approach in the end. Added a piece of context, but removed as much build specific information from the README as I can, favouring rather to defer out almost entirely to BUILDING, as this section is about installing, not building.

[5f3796a](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/5f3796a7d3cdeaa11c907f6a4f27823484e2200a)

What do you think? We lose the hint to try a containerised build, but i'm not sure how valuable being explicit about that was anymore anyhow, given how much easier the regular build steps are to follow now.

BUILDING.md
```diff
@@ -53,66 +29,169 @@ following packages are installed:
...
+OpenAssetIO provides several custom CMake options to control the
+behaviour of the build.
+
+| Option      | Description | Default |
+| ----------- | ----------- | ------- |
+| `BUILD_SHARED_LIBS` | `ON` to build shared libraries. `OFF` to build static libraries. | `ON` |
+| `OPENASSETIO_ENABLE_PYTHON` | Additionally build python bindings | `ON` |
+| `OPENASSETIO_ENABLE_C` | Additionally build C bindings | `ON` |
+| `OPENASSETIO_ENABLE_TESTS` | Additionally build tests | `OFF` |
+| `OPENASSETIO_ENABLE_PYTHON_TEST_VENV` | Automatically create environment when running tests | `ON` if `OPENASSETIO_ENABLE_TESTS` is enabled, `OFF` otherwise. |
```
feltech, line 182: [consider] It's technically true, but reading it here makes me think we shouldn't have bothered with a `cmake_dependent_option`, since it's already implicitly dependent.
```suggestion
| `OPENASSETIO_ENABLE_PYTHON_TEST_VENV` | Automatically create environment when running tests | `ON` |
```

README.md
elliotcmorris: (slight wording changes in subsequent fixups)

BUILDING.md
elliotcmorris: It was autogenerated, but I don't think it's too important to record the generator, it was just from the top of google, they're a dime a dozen. (Good thing about being open source is not having to worry about using those sorts of things.)

I've massaged it myself by hand anyway after the fact when moving sections about.

BUILDING.md
elliotcmorris: Is this just GCC we're talking about here? I wonder if it's possible to list what specific minimum versions of MSVC/XCode you need installed? Do we know?

BUILDING.md
elliotcmorris: Will update version to 2.8.1 for now.

BUILDING.md
elliotcmorris: I think generally we've favoured the environment variable approach, at least from my perspective.

I don't think one can be said to be better than the other, so will leave as is.

BUILDING.md
elliotcmorris: Aye, I don't think one can be said to be better than the other, and better not to introduce branching options.

BUILDING.md
elliotcmorris: Yeah, still unclear on this terminology. I thought the term "install tree" meant like, we generated an output structure that is identical to the structure expected in python-land when libraries are installed.

Nonetheless, will change to your phrasing as it's obviously correct.

elliotcmorris: It's still correct to follow on with ... "the generated install tree can then be used directly..."? that comes from the original docs.

foundrytom:
Reminder: Check that `build-instructions.yml` is up to date

BUILDING.md
elliotcmorris: It's still correct to say Python module though right? As the extension module itself does statically link the core libs when it's built?

foundrytom (commented):
Looking really good!

BUILDING.md
foundrytom: I think it's "install tree" that is confusing in a python context.  It would be "the generated python package can then be..."

BUILDING.md
elliotcmorris: [0e3ee0b](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/0e3ee0bba5ca3470c7a46e8427f23ffc951e49dc)

BUILDING.md
elliotcmorris: [b57de1a](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/b57de1a613d2275edcaab104eedbfea3827cb2ea)

BUILDING.md
foundrytom: Should we just change it to `C++17 compliant(?) compiler`

BUILDING.md
foundrytom: > [consider] Not sure that pybind11 will still work, we're currently using 2.8.1 (but then that'll change too once we have MacOS 11 support from #681)

*Python 3.11

README.md
```diff
@@ -152,41 +152,20 @@ Windows, macOS, and Linux are all supported platforms.
 
 ### Package Installation
 
+OpenAssetIO can be installed either as a pure Python package, or as a
+full install, which builds a core C++ component, as well as optional
```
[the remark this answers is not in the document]
elliotcmorris: I think it's an install, the instructions ask you to run `cmake --install` afterall. which places the build artifacts into a structure thats know and usable by your application.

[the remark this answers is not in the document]
foundrytom: ... only if you specify a custom install path, otherwise its in `build/dist` which is of no use to anyone directly.

BUILDING.md
elliotcmorris: [bd3a6db](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/bd3a6db46c6ca142d77b71699f7c470756f4f218)

README.md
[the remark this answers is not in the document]
foundrytom: Im just not sure, that, someone reading this fresh, who doesn't even know about cmake yet, would think of it as a "full install". In that, a pip install produces a resulting package that is setup in your environment such that it is available for python to use.  If you just build with our instructions, the result isn't available for anything to use.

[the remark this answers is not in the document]
elliotcmorris: Or even "Getting OpenAssetIO" as the title, if getting started implies too much tutorialisation.

BUILDING.md
elliotcmorris: I'd quite like to list exact versions, I think it helps, even as an afterthought.
What do you think about
`C++17 compliant compiler (GCC 9.3 / MSVC 16.9 / macOS 10.15)`
?
These are CY2022 specs.

feltech: ^ I like that.  Another option is to link to VFX reference platform and say CY22.

foundrytom: "or later" as there are no issues building with newer compilers (I'm using clang 14 all the time)?

BUILDING.md
feltech: That's cool. As per our process, feel free to resolve `[consider]` or `[consider][minor]` comments if you don't want to make any changes (after posting a comment saying why).

elliotcmorris:
Added a commit that changes `pip install .` to `python -m pip install .` as is now the format, as well as adding an `--editable` job.

BUILDING.md
feltech: The rewrite is OK in and of itself, but I think we've lost some of the original intention of that paragraph. It's now very Python-centric, when the original intention was that the install tree as a whole could be used e.g.
* Directly in a C++ application, by extending `LD_LIBRARY_PATH` (or `PATH` on Windows).
* Directly in a Python application, by extending `PYTHONPATH` (plus probably `PATH` on Windows).
* For consumption in a CMake project using `find_package`, by extending `CMAKE_PREFIX_PATH` (or equivalent).
* For archiving as a zip file for redistribution (since all RPATHS are relative, so it's theoretically relocatable).
* For consumption by an packaging tool such as Conan, Rez, vcpkg, for further redistribution.

Admittedly what was there before was rather terse to cover all these options.

.github/workflows/build-instructions.yml
```diff
@@ -59,7 +59,46 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Bootstrap
+      uses: ./.github/bootstrap_platform
+
+    - name: Install using pip
+      run: |
+        ${{ matrix.config.preamble }}
+        python -m pip install --editable .
```
feltech, line 101: [consider] Not 100% sure on the value of this. The problem with editable installs isn't generally that they fail to install, its that once installed trying to use the package doesn't work (because search paths to the binaries aren't configured correctly).  To make this valuable we'd have to run the (Python) tests too.

README.md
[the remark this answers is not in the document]
foundrytom: happy with hybrid package in the test, and Getting Started as the title 👍

BUILDING.md
elliotcmorris: [fbca361](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/fbca36113879859bfdf31fbf38ad773915bad3f5)

BUILDING.md
elliotcmorris: Pretty terse indeed. I'm rather not sure what to do here. Now that we've listed the output formulations of both the C++ and python libs, i'm somewhat tempted to just delete the

> This generated python package can then be used directly (e.g. by
extending `PYTHONPATH` and/or `LD_LIBRARY_PATH`), or as the source for a
further packaging step.

fragment that is causing all this consternation, is letting people know they can use the output dirs directly helpful at all?

elliotcmorris: Also includes a couple of spelling fixes [2f0ec37](https://github.com/OpenAssetIO/OpenAssetIO/pull/711/commits/2f0ec37ce2bd7579a18f82aaa581c32c1ff877b7)

feltech (approved):

PR #726, opened 2022-11-08
[Build] Update Pybind to support Python 3.11

### LINKED ISSUE #683: Support Python 3.11

elliotcmorris:
Closes #683 

In order to support building python 3.11 wheels, update pybind. This causes the error messages to change, which we test against, so also update those messages.


elliotcmorris:
Closing because this didn't work out of the box, moving to fork to iterate.

foundrytom:
We also have https://github.com/OpenAssetIO/OpenAssetIO/pull/681

PR #732, opened 2022-11-08
[Build] Use `requires` rather than `tool_requires` in conanfile.py

### LINKED ISSUE #699: Support conan `--require-override`
## What
In order to test [this PR](https://github.com/OpenAssetIO/OpenAssetIO/pull/694), @feltech wanted to run 
```
conan install -if .conan --require-override cpython/3.7.12 resources/build
````
i.e., making use of `--require-override` to force use of 3.7. 

However, our conanfile has a mistake (my bad), similar to that alluded to in https://github.com/OpenAssetIO/OpenAssetIO/issues/654#issuecomment-1290756213, but for other reasons.  

Since `cpython` is a `tool_requires` rather than normal `requires`, the `--require-override` has no effect. [The docs](https://docs.conan.io/en/latest/reference/commands/consumer/install.html#require-override) say this explicitly

> This feature affects only to regular requires, not to tool_requires or python_requires, as those don’t have such an overriding mechanism, and they are private to their consumer, not propagating downstream nor upstream.

We need to split our `build_requirements` method in `conanfile.py` into separate `requirements` and `build_requirements`, where the `requirements` method contains `requires()` calls for `cpython` and `pybind11`, and the `build_requirements` method contains the `tool_requires()` calls for the unit testing libraries.

_Originally posted by @feltech in https://github.com/OpenAssetIO/OpenAssetIO/pull/694#discussion_r1010332175_

### PR DESCRIPTION
**Pull request by feltech**
Closes #699. We were using `tool_requires` for all dependencies, but that function establishes a build requirement, rather than a runtime requirement. This has a few implications

* Dependency versions cannot be overridden by `--require-override`.
* Other packages in the same dependency tree will not be forced to have compatible versions of shared dependencies.
* The package hash is unaffected by changes in the dependencies.

The first issue is more immediate, since we now support a range of Python versions, and it is useful to be able to override the Python version for testing.

The latter two are important more in the future, when we submit a package to ConanCenter.

So split out the "runtime" dependencies from `build_requirements` into `requirements`, allowing `--require-override` to work.

Leave the unit testing libraries as build-only requirements - a future-proofing choice, since when we eventually create a Conan package of OpenAssetIO we will only need the unit testing libraries at build time. The [Conan docs](https://docs.conan.io/en/latest/devtools/build_requires.html) recommend this for unit testing libraries.

### DIFF
--- a/resources/build/conanfile.py
+++ b/resources/build/conanfile.py
@@ -7,9 +7,10 @@ class OpenAssetIOConan(ConanFile):
         # which augments package search paths with conan package
         # directories.
         "cmake_paths",
-        # Generate `Find<PackageName>.cmake` finders, required for various
-        # public ConanCenter packages (e.g. pybind11) since they disallow
-        # bundling of `<PackageName>Config.cmake`-like files in the package.
+        # Generate `Find<PackageName>.cmake` finders, required for
+        # various public ConanCenter packages (e.g. pybind11) since they
+        # disallow bundling of `<PackageName>Config.cmake`-like files in
+        # the package.
         "cmake_find_package",
     )
     settings = "os"
--- a/resources/build/conanfile.py
+++ b/resources/build/conanfile.py
@@ -15,13 +15,15 @@ class OpenAssetIOConan(ConanFile):
     )
     settings = "os"
 
-    def build_requirements(self):
+    def requirements(self):
         # CY2022
         if not tools.get_env("OPENASSETIO_CONAN_SKIP_CPYTHON", False):
-            self.tool_requires("cpython/3.9.7")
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by feltech** [comment]
For testing, e.g. override cpython to 3.7.12
```shell
conan install -if .conan --build=missing --require-override cpython/3.7.12 resources/build
cmake -S . -B build --toolchain .conan/conan_paths.cmake -DOPENASSETIO_ENABLE_TESTS=ON && ctest --parallel 8 --test-dir build
```

elliotcmorris (approved):
Ran the install steps as described, worked out of the box.

Rebase & Merge when ready.

PR #734, opened 2022-11-09
[CI] Don't build wheels on PR

feltech:
CI is very slow (approx 12 mins). By not building wheels on PR (only `main`) we can halve the time taken (first push to this PR took 7 mins)

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -5,11 +5,6 @@ on:
     branches:
       - main
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


foundrytom:
I'm not 100% sure on this one, its only only check against Python 3.7. If we're going to do this, we should move main CI to Python 3.7 (which has asan issues).

foundrytom:
Related to #756

foundrytom:
Closing as 3.7 checks are valuable and we have #756

Issue #743, opened 2022-11-15
Incorrect arg labels in `ManagerInterface` bindings


foundrytom:
## What

Python binding arg names for `ManagerInterface` do not match c++.

Eg `managementPolicy` `traitSet` should be `traitSets`.  There may be others

Python args should match the c++ headers.

## Why

This policy was decided jn DR001 to unify naming so that documentation is usable across c++ and Python where possible.

feltech:
I decided to have some fun and see if we can detect such mismatches automatically. 

Attached is a script that uses `libclang` (`pip install libclang`) to parse the C++ headers corresponding to the classes in the `_openassetio` Python extension module, then compares method arguments against the docstring signature created by pybind11: 

[check_args.zip](https://github.com/OpenAssetIO/OpenAssetIO/files/10022689/check_args.zip)

The output is a list of arguments found in the C++ signature but not in the pybind11 signature, plus any methods found to be defined in Python but missing from C++ (to prompt a manual check).

We have quite a few bindings where we haven't given the argument a name, so pybind11 uses `arg0`.  

There are several instances of using `id` instead of `traitId`.  

Methods reported as missing from C++ are expected and are just due to the script not being clever enough (e.g. can't see that a method is declared in a base class).  Most of that output is stripped from the dump below.

The raw output is as follows:
```
TraitsData: 'traitId' missing from addTrait(self: openassetio._openassetio.TraitsData, arg0: str) -> None
TraitsData: 'traitSet' missing from addTraits(self: openassetio._openassetio.TraitsData, arg0: Set[str]) -> None
TraitsData: 'out' missing from getTraitProperty(self: openassetio._openassetio.TraitsData, id: str, key: str) -> Optional[Union[bool, int, float, str]]
TraitsData: 'traitId' missing from getTraitProperty(self: openassetio._openassetio.TraitsData, id: str, key: str) -> Optional[Union[bool, int, float, str]]
TraitsData: 'propertyKey' missing from getTraitProperty(self: openassetio._openassetio.TraitsData, id: str, key: str) -> Optional[Union[bool, int, float, str]]
TraitsData: 'traitId' missing from hasTrait(self: openassetio._openassetio.TraitsData, id: str) -> bool
TraitsData: 'traitId' missing from setTraitProperty(self: openassetio._openassetio.TraitsData, id: str, propertyKey: str, propertyValue: Union[bool, int, float, str]) -> None
[...]

PR #752, opened 2022-11-17
[Docs] Add release notes for Python arg renames

feltech:
Part of #743. Release notes were missing from the original PR.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,6 +13,14 @@ v1.0.0-alpha.x
 - Removed `toml++` as vendored library. Is now an external dependency
   similar to other external dependencies.
 
+- Updated various Python method argument names to match their C++
+  equivalent. Specifically this affects
+  `Manager`/`ManagerInterface.managementPolicy`,
+  `SeverityFilter.setSeverity`, `TraitsData.hasTrait` / `.addTrait` /
+  `.addTraits` / `.setTraitProperty` / `.getTraitProperty` /
+  `.traitPropertyKeys` / copy-constructor.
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -13,6 +13,10 @@ v1.0.0-alpha.x
 - Removed `toml++` as vendored library. Is now an external dependency
   similar to other external dependencies.
 
+- Updated various Python method argument names to match their C++
```
elliotcmorris, line 16: [Consider] It might be best to explicitly list the changed method signatures, for searchability down the line.

feltech: :thinking: There are quite a few, but I'll have a go and see how it looks.

elliotcmorris: Hmm, personally i'd have it as you've done it there, despite the length, but that's becoming a philosophical point. What _are_ release notes.
I don't think we've got standards on this, so I'll yield to your judgement. What I _wouldn't_ do is compromise and only list a subset just to satisfy my point, go with your gut and commit :P

elliotcmorris (approved):

RELEASE_NOTES.md
elliotcmorris: Whatever your choice, merge when ready.

elliotcmorris: Ah, so you've omitted the arguments. but all the args change are for the listed methods. I think that's smart yeah, if you're confused why something isn't working between versions, you'd probably search for the method rather than the arg anyhow.

elliotcmorris (approved):

Issue #655, opened 2022-10-13
Update project file structure


foundrytom:
## What

Update the filesystem layout as per #633.

## AC

- The layout matches the DR

## Notes

It's fine for the final change to be in a single commit if it simplifies matters, otherwise component-wise is fine.

(Incomplete) checklist:
- Link paths in Markdown
- CI/Actions
- Build/install documentation et. al.
- Any consuming projects affected by layout change.

PR #754, opened 2022-11-17
[Build] Move setup.py under openassetio-python

### LINKED ISSUE #728: [Build] Move `setup.py` into `openassetio-python`
# What

~Decide on and then possibly~ Move the top-level `setup.py` under the Python component in the directory structure.

# Why

This follows on from https://github.com/OpenAssetIO/OpenAssetIO/pull/633#discussion_r1014326262. 

When originally designing #628 we had a `setup.py` that was - unexpectedly to outside adopters - insufficient to build a working Python package.  So it made a lot of sense to hide it away under the Python component.  However, this argument no longer holds - `pip install .` can create and install a fully functioning Python package (assuming C++ library dependencies can be located).

The other main argument for moving `setup.py` is that the project looked too Python-centric, when the Python component is an optional add-on.  We also wanted to "physically" separate components into different directories, and having a rogue `setup.py` outside of the Python component is inconsistent with that vision.

feltech:
Closes #728 and follows on from the project restructure in #655. 

During the project restructure we were not sure if wheel builds could continue to work if we moved setup.py; and had an open question of whether we wanted to move setup.py anyway, since it helps discoverability if we locate setup.py at the root of the project.

After a little investigation, the main headache in the build process is resolved by simply upgrading pip. Later pip versions build in-tree (rather than copying the package to a temporary directory), so that relative paths (e.g. to `CMakeLists.txt`) can be resolved.

In addition, we decided that on balance it's better to colocate Python artifacts together under the `openassetio-python` component directory. This has the added benefit of further disambiguating the project at-a-glance to make it clear that this is a CMake C++ project primarily, with an optional Python component.


src/openassetio-python/package/pyproject.toml
```diff
@@ -119,7 +119,7 @@ target-version = ["py39"]
```
feltech, line 119: [question] I was in two minds whether to move these files under `openassetio-python/package` or just under `openassetio-python`.  

In particular, any configuration related to running tests or linters applies to `openassetio-python/tests/package` too.

But having this under `package` seems neater, since the primary responsibility of `pyproject.toml`/`setup.py` is related to  packaging the `package`.

:question:

elliotcmorris: Hmm, i'm not really in a strong mind about it. My intuition would be to have them directly under `openassetio-python`, rather than under package, but I don't have strong arguments. What I've got is: 

- There's a `CMakeLists.txt` at that level already, meaning it dosen't "look" like an expectable python project when you go into the python directory, without these files
- Not totally sure how well the idea that because `setup.py`/`pyproject.toml` are involved in packaging, that means they must go in `package`. `package` is the noun here, whilst `setup.py` is the verb, one could almost make the strained argument that the packager should be outside of the package.
- I can imagine a possibility`setup.py` wanting to reach into adjacent folders at some point in the future, for whatever reason, and having it top level makes that a tad more natural

As usual when there's no strong arguments either way, I yield to the author, happy either way.

src/openassetio-python/package/setup.py
```diff
@@ -39,10 +40,22 @@ def build_extension(self, _ext: Extension):
...
+        cmake_project_path = pathlib.Path("../../..")
+
+        if not os.path.isfile(cmake_project_path / "CMakeLists.txt"):
+            # Unfortunately there doesn't seem to be a way to detect the
+            # version of pip that launched this process. So assume if
+            # CMakeLists.txt is missing then its probably because the
+            # wrong version of pip was used.
+            raise FileNotFoundError(
+                "CMakeLists.txt not found. OpenAssetIO wheel building requires an in-tree build"
+                " with setup.py at the cwd. If using pip, ensure pip>=21.3."
```
elliotcmorris, line 52: [Consider] I'm guessing `cwd` is current working directory? Took me a little while to puzzle it out, worth expanding the jargon away I'd say.

elliotcmorris (approved):
Looks fab.

src/openassetio-python/package/setup.py
feltech: :grin: Fair enough, expanded in cad3ac91

src/openassetio-python/package/pyproject.toml
feltech: Good arguments, moved one level up in 22b434d9.  The `package` directory looks a bit lonely and sparse now, but I do kinda like that `setup.py`/`pyproject.toml` is above the `module`, `package` and `tests`, which are all related.

I guess @foundrytom will be back before this is merged anyway, so we can get a tie-break opinion.  Easy enough to drop the fixup if the previous location is preferable.

foundrytom: We chatted together about this, and decided on keeping it under `openassetio-python`, as it helps convey that the `package` subdir is not self-contained.

src/openassetio-python/setup.py
```diff
@@ -39,10 +40,22 @@ def build_extension(self, _ext: Extension):
 
         @param _ext: Extension to build (we only have one, so ignored).
         """
+        cmake_project_path = pathlib.Path("../..")
```
foundrytom, line 43: [question] Could we use `os.path.abspath(__file__)` or similar to avoid `cwd` sensitive paths?

foundrytom (changes requested):

src/openassetio-python/setup.py
```diff
@@ -39,10 +40,22 @@ def build_extension(self, _ext: Extension):
...
+
+        if not os.path.isfile(cmake_project_path / "CMakeLists.txt"):
+            # Unfortunately there doesn't seem to be a way to detect the
+            # version of pip that launched this process. So assume if
```
foundrytom, line 47: Would `import pip, pip.__version__` be any use?

RELEASE_NOTES.md
```diff
@@ -13,6 +13,10 @@ v1.0.0-alpha.x
 - Removed `toml++` as vendored library. Is now an external dependency
   similar to other external dependencies.
 
+- Moved `setup.py` and `pyproject.toml` under the Python component's
+  directory, i.e. `src/openassetio-python`.
```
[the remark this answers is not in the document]
foundrytom: Could/should it go [here](https://github.com/OpenAssetIO/OpenAssetIO/blob/22b434d937a1d7f66b7d9bfbe886599a7bcce196/src/openassetio-python/pyproject.toml#L6) too?

src/openassetio-python/setup.py
feltech: IIRC `__file__` won't work because `build_ext` seems to be ultimately called from an `exec` statement.  It's all very weird.  I'll double-check though.

RELEASE_NOTES.md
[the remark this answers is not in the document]
feltech: > Could/should it go [here](https://github.com/OpenAssetIO/OpenAssetIO/blob/22b434d937a1d7f66b7d9bfbe886599a7bcce196/src/openassetio-python/pyproject.toml#L6) too?

As mentioned in https://github.com/OpenAssetIO/OpenAssetIO/pull/754#discussion_r1028282735 - if you put `pip` in there then that'll be installed as a package, but won't be the pip that is actually used (if you think about it, it's a bit chicken-and-egg)

src/openassetio-python/setup.py
foundrytom: ahh right - yeah, sorry forgot that it wasn't available there.

src/openassetio-python/setup.py
foundrytom: I tried a quick print and it did give me `setup.py`, but I hadn't tried with `abspath`

RELEASE_NOTES.md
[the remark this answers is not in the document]
foundrytom: Maybe worth a note in the release notes, and perhaps a slight rewording of the text - as I got the exception that it couldn't find CMAKE lists that only talks about wheels when I tried with an old version?

src/openassetio-python/setup.py
feltech: Confirmed, you get `setup.py`, but `__file__` is normally the absolute path to the file, not just the file name. `abspath` will then prefix the cwd to that.  So it's kinda equivalent to just using `.`.

I'll see if there's a better option available, though.

I was a little wrong about the `exec` statement thing - it is ultimately called from `exec`, but that's not the reason we can't get the path of `setup.py` from inside `build_ext`.  It is ultimately running `build_ext` as if under `exec` (i.e. like executing a string), but I'd have to spend a bit more time debugging to really get to the bottom of it.

foundrytom: Don't worry for now if its not easily solvable. I was just curious if it was an option. I was more confused that it _is_ working when you do `python -m pip install src/openassetio-python` if that is making a wheel!

RELEASE_NOTES.md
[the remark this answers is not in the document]
feltech: Ah OK, yeah I see that could be confusing.  I'll just remove the "working directory" bit from the error message, since it's really advanced/unsupported usage to install using anything other than pip.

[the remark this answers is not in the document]
foundrytom: Ahh does the "invocation magic" cd you in there with you use `python -m pip`? If that is the case then the `__file__` thing I was waffling about below seems a little redundant.

src/openassetio-python/setup.py
feltech: Yeah, it's a really long winding process, e.g. with a breakpoint in `built_ext` I get
```
sys.argv
['/home/dave/workspace/cloud/assetapi/OpenAssetIO/.venv/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py', 'bdist_wheel', '--dist-dir', '/tmp/pip-wheel-eed3kpwi/tmpxuecirt8']
```
and then early on in the stack trace there is an `exec` call adding yet more confusion...

Anyway, I tried all sorts in the debugger to see if there was some variable available somewhere that could give the directory of `setup.py`, but can't find anything.

Is it OK as it is, or does it need some clarification? - could be explicit and use `os.cwd`, or could add a code comment.

foundrytom: If I've understood what you were saying above re: "not needing to be in the dir" is only for advanced non-standard wheel builds, then I think it'd be fine as is - especially if `python -m pip install` and others do the right thing for you? Sorry for the noise here. Good job exhausting options!

RELEASE_NOTES.md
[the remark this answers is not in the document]
feltech: Updated release notes and exception message in 99cce633

doc/BUILDING.md
```diff
@@ -130,12 +130,12 @@ package, placed at
...
-OpenAssetIO can alternately be built and installed via `pip`, which may be
-preferable to the exclusively Python focused developer.
-Check out the repository, and from the root, run :
+OpenAssetIO can alternately be built and installed via `pip`, which may
+be preferable to the exclusively Python focused developer. Check out the
+repository, and from the root, run:
 
 ```shell
-python -m pip install .
+python -m pip install src/openassetio-python
```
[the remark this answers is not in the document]
feltech: Good spot, fixed in bcf80b4a

PR #755, opened 2022-11-18
[Build] Split CMake package variables to own file

feltech:
Part of #654. ConanCenter does not allow `OpenAssetIOConfig.cmake` et al. files and instead insists that such files are autogenerated by Conan itself.

But then Conan's autogenerated CMake config files will be missing the additional variables we added to our config files.

So move the variables to a different .cmake file, which can be bundled in the Conan package using the `cmake_build_modules` property.

Note that Conan generators, which produce importable CMake config files, will create an `OpenAssetIO_INCLUDE_DIR` variable in the generated scripts. So don't move that particular variable out of `OpenAssetIOConfig.cmake` (which will not be included in the Conan package), to avoid duplication.

Also, remove the `OpenAssetIO_DIR` variable, since that shouldn't be set within the config file - it is set by CMake itself when the package is found.


elliotcmorris:
Adding @foundrytom re: slack discussion about not using Maintainers.

foundrytom (approved):

Issue #528, opened 2022-07-21
Fix Python linking


feltech:
# What

#### Ensuring Python version

`find_package(Python`, when given a specific version, only guarantees a version greater than or equal to. This means an unexpected Python can be found (e.g. system 3.8 when requested conan's 3.7).  We should add `EXACT` to the `find_package` call.

We also still default to `3.7` when it should be `3.9` for CY2022.

#### Using shared `libpython`

This is only a problem if we have a need to link to `libpython`, which for most purposes we don't (though see next point re. embedding)

The current conancenter `cpython` package recipe defaults to `shared=False`, meaning `libpython` is a static library and the `python` interpreter (presumably) has the library compiled in.

Setting `shared=True` gives a shared library, but the RPATH of the interpreter is not set, so it does not function (e.g. for creating the venv and running tests) without patching. 

`find_package(Python` prefers shared libraries but will silently use static libraries if that is all that is found. If we want to force shared libraries then we must set `Python_USE_STATIC_LIBS` to `false`.  
  
#### Using embedded Python interpreter (for testing / app integration) - FIXED (see comments)

When creating a target with an embedded Python interpreter (e.g. for tests, as used in a [prototype for testing C++ loading the Python plugin system](https://github.com/feltech/OpenAssetIO/tree/prototype/loadingPythonPluginSystemInCpp)) we must add `Development.Embed` to the list of `COMPONENTS` passed to `find_package(Python`. 

This has the effect of making any Python-related target link to `libpython` (even if `pybind::module` is the only linked library, whose whole purpose is to avoid linking `libpython`). This is bad - Python extension modules should _not_ link directly to `libpython`.  This appears to be a bug in either pybind or Python's CMake helpers. 

If `Development.Embed` is not specified (i.e. only  `Development.Module`) then everything works fine, but we cannot create a (test) executable with an embedded Python interpreter.

Beyond use for testing, this may be a problem when adopters of OpenAssetIO come to use it in their application with an embedded Python interpreter.

feltech:
Re. `libpython` being erroneously linked in with `pybind11::module` (i.e. Python extension modules) when the `Development.Embed` component of Python is available:

I discovered the problem is the autogenerated conan `Findpybind11.cmake`, which makes some odd choices that should probably be reported upstream.  

[...]

PR #764, opened 2022-11-23
[Build] Split Python "internal" tests into "module" and "bridge" tests

### LINKED ISSUE #739: [Tests] Split `_openassetio_test`-related tests into two
## What

Split the tests that depend on the `_openassetio_test` Python extension module into two independent test components

## Why

During the project restructure in #655 symmetry in the directory structure between tests and the (sub)components under test was impossible to achieve, since we bundle tests of two quite separate components under the one umbrella, namely `bridge` (`createPythonPluginSystemManagerImplementationFactory`) tests and `module` (`PyRetainingSharedPtr`) tests.

Ideally we would split those tests into their own CTest targets, under their own subdirectory, with their own `_openassetio_test`-like helper module (if required - arguably the `bridge` test should be a C++ application with embedded Python interpreter)

## ACs

- `tests` dir mirrors structure of `bridge`/`module` etc
- Tests for `bridge` are done using a simple host with it's own embedded python interpreter

feltech:
Closes #739.

A little refactoring sneaked in, especially the first few commits.

In particular, there's an update to the Conan generator we use, so that the frustrating pybind11 Conan package bug, which prevents use of pybind11 when embedding a Python interpreter (see commit message and https://github.com/OpenAssetIO/OpenAssetIO/issues/528), is fixed.


feltech:
Converted to draft to address CI failures.

foundrytom:
Quick Q - whilst we're rejigging things here - what do people think about `module` -> `cmodule` to middle disambiguate that we have `module` and `package` dirs (as `package` technically also contains other "modules")?

feltech:
> Quick Q - whilst we're rejigging things here - what do people think about `module` -> `cmodule` to disambiguate that we have `module` and `package` dirs (as `package` technically also contains other "modules")?

Added a commit at the end (512ed5a) to do this, but can move it to the beginning if you think the git wrangling is worthwhile.

foundrytom (approved):
Looks good from my perspective in terms of the resulting structure. I'll have to leave any specific assessment of the CMake changes to someone better qualified :)

elliotcmorris (approved):
Similar thoughts to Tom, bit of a reverse-bikeshed review.

foundrytom:
> > Quick Q - whilst we're rejigging things here - what do people think about `module` -> `cmodule` to disambiguate that we have `module` and `package` dirs (as `package` technically also contains other "modules")?
> 
> Added a commit at the end ([512ed5a](https://github.com/OpenAssetIO/OpenAssetIO/commit/512ed5afffdae4dc595e4c4e6b2e4f870637f8a3)) to do this, but can move it to the beginning if you think the git wrangling is worthwhile.

Im happy for this to go later. I was wondering what the final commit granularity was here too, happy to squash some of the "similar" cmake tweaks together too.

feltech:
> happy to squash some of the "similar" cmake tweaks together too.

I've squashed a few of the later commits together (renaming `module`/`internal` to `cmodule` and moving the test config to its own `CMakeLists.txt`).  How does that look?

PR #768, opened 2022-11-24
[PluginSystem] Support entry point based plugin discovery

### LINKED ISSUE #762: [PluginSystem] Support "pkg_resources" style plugin discovery
## What

Allow the python plugin system to load manager plugins from "pkg_resources" style plugins, as well as `OPENASSETIO_PLUGIN_PATH`

## Why

Presently, the distribution and installation of manager plugins is somewhat manual. It is potentially arduous for some end users. OTIO has a clever trick where it uses `pkg_resources` to discover `pip` installed packages that provide the OTIO plugin entry point. This greatly simplifies installation and version management of these plugins.

#712 demonstrates this is relatively trivial to add to OpenAssetIO

## Notes

- ~`pkg_resources` is deprecated in favor of `importlib_metadata`~ The new API is not available until 3.8. and not stable till 3.10, so we should stick with `pkg_resources` for now, as we'll need to support 3.7 for a while, and 3.9 for a long while.

## ACs

- The Python plugin system will discover pip-installed manager plugins if they provide the `openassetio.manager_plugin` entry point - which provides a module exporting a top-level `plugin` var (i.e. a standard manager plugin).
- `$OPENASSETIO_PLUGIN_PATH` plugins with the same identifier should take precedence over pip installed ones.
- If `$OPENASSETIO_DISABLE_PKG_RESOURCES_PLUGINS` is set, then only `$OPENASSETIO_PLUGIN_PATH` plugins should be loaded (var name formulated to match OTIO).
- When Python 3.7 is used, then a suitable debug statement notes the entry point plugins are not available.

foundrytom:
Well, this has been a fun adventure. But I hope were in a better place now.

This started out as the planned pkg_resources based implementation, with naming conventions matching OTIO. Since we did our PoC investigations OTIO have now [switched to `importlib`](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1469). So the implementation has been updated to use the same control env var name as the new OTIO implementation.

We opted to use the upstream/backport `importlib_metadata` package for _all_ python versions (as it supports 3.7+). We chose this over a split code path approach as it felt like there were just too many differences between 3.7, 3.8 - 3.10 that would result in harder to maintain code.

`importlib_metadata` _is_ the code that ends up in future python releases anyway (I think), so we're already implementing against the future API of the `importlib.metadata` module once we can drop older python versions.

Closes #762.


src/openassetio-python/pyproject.toml
```diff
@@ -13,6 +13,9 @@ build-backend = "setuptools.build_meta"
 name = "openassetio"
 version = "1.0.0a6"
 requires-python = ">=3.7"
+dependencies = [
+    "importlib_metadata == 5.0"
```
foundrytom, line 17: [question] My brain is a bit fried - what is the preferred way to specify this, should we pin to a specific version or greater than?

RELEASE_NOTES.md
```diff
@@ -28,6 +42,18 @@ v1.0.0-alpha.x
   [#743](https://github.com/OpenAssetIO/OpenAssetIO/issues/743)
 
 
+### New features
+
+- Added entry point based discovery of Python manager plugins. This
+  allows pure Python manager plugins to be deployed and managed using
+  `pip`, without the need to wrangle `OPENASSETIO_PLUGIN_PATH`. The
```
JeanChristopheMorinPerso, line 49: It is mainly pip yes (since pip is the main installer), but not only. The entry point mechanism isn't specific to pip. So maybe "deployed and managed as Python packages"?

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystemManagerImplementationFactory.py
```diff
@@ -37,48 +37,97 @@ class PythonPluginSystemManagerImplementationFactory(ManagerImplementationFactor
     derived plugins. Not usually used directly by a @ref host, which
     instead uses the @fqref{hostApi.ManagerFactory} "ManagerFactory".
 
+    The factory loads plugins from exposed `openassetio.manager_plugin`
+    entry points, as well as additional plugins found under paths
```
JeanChristopheMorinPerso, line 41: What do you think of:
```suggestion
    The factory loads plugins from entry points registered within the `openassetio.manager_plugin`
    entry point group, as well as additional plugins found under paths
```
?

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystemManagerImplementationFactory.py
```diff
@@ -37,48 +37,97 @@ class PythonPluginSystemManagerImplementationFactory(ManagerImplementationFactor
...
     """
 
     ## The Environment Variable to read the plug-in search path from
     kPluginEnvVar = "OPENASSETIO_PLUGIN_PATH"
+    ## The Environment Variable to control the discovery of entry point based plugins
+    kDisableEntryPointsEnvVar = "OPENASSETIO_DISABLE_ENTRYPOINTS_PLUGINS"
+
+    ## The name of the ManagerPlugin entry point for entry point
+    ## discovered plugins.
+    kPackageEntryPointName = "openassetio.manager_plugin"
```
JeanChristopheMorinPerso, line 69: It's more the group name than an entry point name. See https://packaging.python.org/en/latest/specifications/entry-points/#data-model.

JeanChristopheMorinPerso (commented):
I think it looks pretty god. I left only some minor comments which are mostly pedantic.

src/openassetio-python/pyproject.toml
foundrytom: Good point, thanks. [Downgraded to `>=3.6.0` here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/c1d5cbbbcf476aa05d411950d4b3bebe65c28a38).

RELEASE_NOTES.md
foundrytom: Good idea, that's much better - updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/0d2f50c200a87e35ab38b09d121aa8c8ee48d1ff).

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystemManagerImplementationFactory.py
foundrytom: Ahh yes, thanks, renamed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/029267c8474d54ff04c4dfa4c9f1af3c5935a4bb).

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
```diff
@@ -110,6 +111,64 @@ def scan(self, paths):
...
+
+                if not hasattr(module, "plugin"):
+                    self.__logger.error(
+                        f"PythonPluginSystem: No top-level 'plugin' variable in {debugIdentifier}"
+                    )
+                    continue
+
+            except Exception as ex:  # pylint: disable=broad-except
+                self.__logger.warning(
+                    f"PythonPluginSystem: Caught exception loading {debugIdentifier}:\n{ex}"
```
elliotcmorris, line 155: [Consider] Perhaps better practice would be to limit the `try` scope to just around `entryPoint.load()`, which I assume is the expectation. Exceptions in the surrounding code are unexpected, (and unlikely!), but would be misdiagnosed.

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
```diff
@@ -110,6 +111,64 @@ def scan(self, paths):
...
+                )
+                continue
+
+            path = getattr(module, "__file__", name)
```
elliotcmorris, line 159: [Question] Is using `name` as a default path good enough here? Is the absence of the `__file__` attribute an expected case, or should we exit instead?

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystemManagerImplementationFactory.py
```diff
@@ -1,5 +1,5 @@
 #
-#   Copyright 2013-2021 The Foundry Visionmongers Ltd
+#   Copyright 2013-2020 The Foundry Visionmongers Ltd
```
elliotcmorris, line 2: [Fix] Assuming this is a typo, should be 2022?

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystemManagerImplementationFactory.py
```diff
@@ -37,48 +37,97 @@ class PythonPluginSystemManagerImplementationFactory(ManagerImplementationFactor
...
     @envvar **OPENASSETIO_PLUGIN_PATH** *str* A PATH-style list of
-    directories to search for @ref openassetio.pluginSystem.PythonPluginSystemManagerPlugin
-    based plugins. It uses the platform-native delimiter.  Searched left
-    to right.
+    directories to search for @ref
+    openassetio.pluginSystem.PythonPluginSystemManagerPlugin based
+    plugins. It uses the platform-native delimiter.  Searched left to
+    right. Plugins found here will take precedence over those discovered
+    through package entry points. Note: this search path is entirely
+    independent of **PYTHONPATH** and doesn't not need to be a subset of
```
elliotcmorris, line 51: [Consider] Oddly phrased double negative, `dosen't not need`, consider simplifying.

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystemManagerImplementationFactory.py
```diff
@@ -37,48 +37,97 @@ class PythonPluginSystemManagerImplementationFactory(ManagerImplementationFactor
...
         Scans for PythonPluginSystemManagerPlugins, and registers them
         with the factory instance.
-
-        @param paths `str` A searchpath string to search for plug-ins.
-        If None, then the contents of the Environment Variable
-        @ref kPluginEnvVar is used instead.
         """
 
-        if not self.__paths:
+        if self.__disableEntryPointsPlugins is None:
```
elliotcmorris, line 102: [Consider][Minor] Took me a second to parse, maybe a quick "if disable entry points member not set, attempt to use environment variable, otherwise false" comment?

elliotcmorris: [Consider] Could you do this in `init`, making `self.__disableEntryPointsPlugins` an effective invariant?

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystemManagerImplementationFactory.py
```diff
@@ -37,48 +37,97 @@ class PythonPluginSystemManagerImplementationFactory(ManagerImplementationFactor
...
+        # Construct this here, so we have this even if we early out
         self.__pluginManager = PythonPluginSystem(self._logger)
 
-        # We do this after instantiating, so that the lifetime of the manager is
-        # consistent with cases where some paths were set.
+        if not self.__paths and self.__disableEntryPointsPlugins:
+            self._logger.log(
+                self._logger.Severity.kWarning,
+                "No search paths specified and entry point plugins are disabled, no plugins "
+                f"will load - check ${self.kPluginEnvVar} is set.",
```
elliotcmorris, line 117: :+1:  Love this. Avoid people bashing their heads against silent terminals wondering why nothing is working.

elliotcmorris (changes requested):
Looks fab. Using the backport universally was a good choice.

Technically a request changes, but the [fix] is the most minor thing in the world. Approval otherwise.

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
foundrytom: Ahh yes - excellent point, thanks. Updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/5613c18b54059936a4e2a89f8a4c15e45c06dc34).

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
foundrytom: This raised an interesting Q, which tipped things to the point of "we really need some tests for the failure cases code". So have added coverage, and improved message generation and the way we determine path etc... [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/e8a6a467da66d7b57fe091e7494e88c92f3372e0).

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystemManagerImplementationFactory.py
foundrytom: Lol yes, that was `<Ctrl-A>` instead of `<Ctrl-X>`... great spot. Fixed [here]()https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/72c05c583b2cb37d4a06d3c8aab9fa1dd96aa9c1).

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystemManagerImplementationFactory.py
foundrytom: Fixed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/21dd2256bfbb9bc4e7e54831516cf9ab448422b7)

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystemManagerImplementationFactory.py
foundrytom: Yeah, I think we did it that way originally to allow the env to be changed after construction, but before scanning. Given that the paths/etc.. are constructor params, I think you're right, and its better there. Changed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/4a5fc9e5ce493567c8b771f6104e9701f19ea3c3).

src/openassetio-python/pyproject.toml
JeanChristopheMorinPerso: Great. You might have forget to modify it in the pyproject.toml file though.

foundrytom: Good spot - thanks! Fixed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/bced6d2cd8f6d83753190425614e66cd1857fce8).

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
elliotcmorris: Wow, significant extra test case.

elliotcmorris (approved):

src/openassetio-python/tests/package/pluginSystem/resources/broken/site-packages/missing_plugin-0.0.0.dist-info/INSTALLER
```diff
@@ -0,0 +1 @@
+pip
```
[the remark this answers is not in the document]
feltech: OK just saw the `pyproject.toml`s :thinking:

src/openassetio-python/tests/package/pluginSystem/resources/entryPoint/README.md
```diff
@@ -0,0 +1,50 @@
+# Entry Point discovery test plugin
```
feltech, line 1: [question] I'm in two minds about whether we want to use "real" packages, which is admittedly more representative, vs. the absolute minimum to test the feature, which is apparently just the plugin module and a `dist-info` directory containing `entry_points.txt`.

It sounds like for privacy purposes we want to delete half the contents anyway.

feltech (changes requested):
Looking good.

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
```diff
@@ -92,29 +86,79 @@ def scan(self, paths):
...
+            f"PythonPluginSystem: Searching packages for '{entryPointName}' entry points."
+        )
+
+        for entryPoint in importlib_metadata.entry_points(group=entryPointName):
+
+            self.__logger.debug(f"PythonPluginSystem: Found entry point in {entryPoint.name}")
+            try:
+                module = entryPoint.load()
+            except Exception as ex:  # pylint: disable=broad-except
+                self.__logger.error(
```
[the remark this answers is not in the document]
foundrytom: Yeah, it would be better wouldn't it.  (Don't forget that `self.__logger` isn't a python logger, its a `LoggerInterface` logger, so you'd need to grab the lines from `format_traceback` or similar, but not a big deal. 

I think we did it this way before with some mild notion that log entries were single-line, but that doesn't really need to be the case I dont think?

src/openassetio-python/tests/package/pluginSystem/resources/entryPoint/README.md
foundrytom: Yeah - I was thinking that too, but wondered if keeping it a "real package" would be more representative when it comes to the fact that parsing is handled by an external lib.

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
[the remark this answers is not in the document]
feltech: :man_facepalming: of course it's not a normal Python logger, I wasn't thinking.  Yeah I think a full stack trace is really useful when debugging.

src/openassetio-python/tests/package/pluginSystem/resources/broken/src/missing_plugin/build/lib/missing_plugin.py
```diff
@@ -0,0 +1,5 @@
+"""
+Provides a broken plugin (missing top-level var)
+"""
+# pylint: disable=invalid-name
+something = None
```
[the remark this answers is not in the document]
foundrytom: Removed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/791587ef808b5e50b47801a13dd1e95bb21bef2e).

src/openassetio-python/tests/package/pluginSystem/resources/broken/src/raises_exception/build/lib/raises_exception.py
```diff
@@ -0,0 +1,4 @@
+"""
```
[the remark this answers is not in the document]
foundrytom: Removed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/791587ef808b5e50b47801a13dd1e95bb21bef2e).

src/openassetio-python/tests/package/pluginSystem/resources/entryPoint/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Making changes
+
+Update the package code in `src`, then, from this directory:
+
+```bash
+python -m pip install ./src -t ./site-packages
+```
+
+The `dist-info` dir will contain an additional `direct_url.json` file
```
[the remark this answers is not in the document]
foundrytom: Combined [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/16f399b984b7133f7dc641dcd46ee0bdbc9d03bb)

src/openassetio-python/tests/package/pluginSystem/resources/entryPoint/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Use in tests
+
+Extend `sys.path`:
+
+```python
+sys.path.append('/path/to/entryPoint/site-packages')
+```
+
+## Making changes
```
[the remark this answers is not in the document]
foundrytom: This this is fixed as part of [this](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/16f399b984b7133f7dc641dcd46ee0bdbc9d03bb).

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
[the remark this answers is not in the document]
foundrytom: Full tb printed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/020d67c7411a179dfc57d35c65b30bc5fdf13213).

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
```diff
@@ -215,3 +256,12 @@ def __load(self, path):
...
 
         self.register(module.plugin, path)
+
+    def __log_exc(self, ex):
+        """
+        Logs information about the supplied exception into the logger
+        with appropriate formatting.
+        """
+        for frame in traceback.format_tb(ex.__traceback__):
+            for line in frame.splitlines():
```
feltech, line 266: [fix] I don't think we should prefix each line of the stack trace. 

The example that springs to mind (from my experience) is centralised logging, e.g. the [ELK stack](https://www.elastic.co/what-is/elk-stack) where individual log entries are database records in their own right.  It can be a pain when you filter for a specific multi-line message if the lines aren't grouped under the same record.

src/openassetio-python/tests/package/pluginSystem/resources/broken/site-packages/missing_plugin-0.0.0.dist-info/INSTALLER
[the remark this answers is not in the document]
feltech: On balance yeah, it's rather opaque that way, but being reliant on knowledge of the internals of a third-party tool for our own unit tests is probably an anti-pattern.

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
```diff
@@ -64,23 +66,15 @@ def scan(self, paths):
         @note Precedence order is undefined for plugins sharing the
         same identifier within the same directory.
 
-        Any existing plugins registered with the PythonPluginSystem will be
-        cleared before scanning.
-
         @param paths `str` A list of paths to search, delimited by
         `os.pathsep`.
         """
-        self.reset()
```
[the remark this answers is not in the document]
foundrytom: We do use it in tests at the mo

src/openassetio-python/tests/package/pluginSystem/resources/entryPoint/README.md
feltech: Resolving, as discussed in https://github.com/OpenAssetIO/OpenAssetIO/pull/768#discussion_r1036996344

feltech (changes requested):
Neat, just one small gripe

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
foundrytom: Hows [this](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/95d33bb5a32498dabc7446caab65d02e370fb880)?

src/openassetio-python/tests/package/pluginSystem/resources/broken/site-packages/missing_plugin-0.0.0.dist-info/direct_url.json
```diff
@@ -0,0 +1 @@
+{"dir_info": {}, "url": "file:///Users/tomcowland/dev/OpenAssetIO/src/openassetio-python/tests/package/pluginSystem/resources/broken/src/missing_plugin"}
```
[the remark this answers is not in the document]
foundrytom: Removed this and other unused files [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/5da66d5a267a22ad642c01a7368364bf35137b77).

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
```diff
@@ -257,11 +258,10 @@ def __load(self, path):
...
         """
-        Logs information about the supplied exception into the logger
-        with appropriate formatting.
+        Provides a multi-line string with information about the supplied
+        exception with appropriate formatting.
         """
-        for frame in traceback.format_tb(ex.__traceback__):
-            for line in frame.splitlines():
-                self.__logger.error(line)
+        return "/n".join(traceback.format_tb(ex.__traceback__))
```
feltech, line 267: [consider] I'm a little wary of directly accessing magic attributes. I think `traceback.format_exc()` works for us, since we're always printing the exception during the stack unroll?

feltech (approved):
Minor niggle.  Apologies for the churn, missed this before.

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
foundrytom: Yeah I was wondering about this, but it is explicitly documented [here](https://docs.python.org/3/reference/datamodel.html?highlight=__traceback__):

> [Traceback objects] ... It is accessible as the third item of the tuple returned by sys.exc_info(), and as the __traceback__ attribute of the caught exception.

I didn't like a helper function assuming it was in the stack unroll, and I didn't like duplicating how we handled the exceptions, so this felt reasonable.

feltech: Since `traceback.format_exc()` gives us a string, there would be no need for a helper function.  But yeah it is odd and annoying that there is no non-stateful way to do this.  Feel free to resolve this and squash+merge if you want.

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
foundrytom: (An elk)

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
foundrytom: Ahh sorry, id been looking at `traceback.format_exception()` which needs `sys.last...` 🙃 . Much [simpler](https://github.com/OpenAssetIO/OpenAssetIO/pull/768/commits/bf873888c5796e6222d268880147f716173ea790). Thanks.

PR #800, opened 2023-01-09
[Build] Don't build C bindings by default

### LINKED ISSUE #796: [Build] Disable C API by default in CMake
## What

Disable building of the C API artifacts by default in CMake, i.e. set `OPENASSETIO_ENABLE_C` to `OFF ` by default. 

## Why

The C API is woefully incomplete and cannot be used in any meaningful way at the moment. We should disable it by default, but ensure we still run its tests in CI.

## Acceptance Criteria

* `OPENASSETIO_ENABLE_C` defaults to `OFF `.
* The `test` CMake preset overrides `OPENASSETIO_ENABLE_C=ON `.  Hence CI/TDD builds continue to run the C API unit tests.
* Release note mentioning that C API is now disabled.

feltech:
## Description

Closes #796

- ~~[ ] I have updated all relevant user documentation.~~

## Test Instructions

CI passing should be a good indication that all is well.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,15 @@
 Release Notes
 =============
 
+v1.0.0-alpha.X
+--------------
+
+### Breaking changes
+
+- Disabled the (nascent) C bindings by default. To enable, the
+  `OPENASSETIO_ENABLE_C` CMake option must be explicitly set to `ON`.
+
+
 v1.0.0-alpha.8
 --------------
 
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -122,10 +122,14 @@ jobs:
         uses: ./.github/bootstrap_platform
 
       - name: Build/install OpenAssetIO
-        run: |
+        run: >
           ${{ matrix.config.preamble }}
+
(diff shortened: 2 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -1,6 +1,15 @@
 Release Notes
 =============
 
+v1.0.0-alpha.X
+--------------
+
+### Breaking changes
+
+- Disabled the (nascent) C bindings by default. To enable, the
+  `OPENASSETIO_ENABLE_C` CMake option must be explicitly set to `TRUE`.
```
elliotcmorris, line 10: [Consider][Minor] Rather than `TRUE`, should we say `ON`, as that's the vernacular used in the scripts?

elliotcmorris (approved):
LGTM :+1:

RELEASE_NOTES.md
feltech: Yeah I think you're right. My thinking at the time was that `TRUE` is a bit less CMake-specific to may be a bit more readable, but I think consistency with the rest of the code is more valuable.

feltech: Updated and force-pushed, rather than fixupped, to b56c7209

PR #803, opened 2023-01-26
[Core] Acquire GIL before C++ imports Python plugin system

### LINKED ISSUE #797: Acquire the GIL within `createPythonPluginSystemManagerImplementationFactory`
### What

Ensure the Python GIL is acquired for the duration of `createPythonPluginSystemManagerImplementationFactory`, and restored to it's previous state when that function exits.

Update OpenAssetIO-Test-CMake to release the GIL before calling this function.

### Why

We don't acquire the GIL within `createPythonPluginSystemManagerImplementationFactory`, so currently the host must acquire it before calling this. 

The symptom in a Release build is an unhelpful segfault. In a Debug build you get a failed assertion telling you that the GIL must be acquired before importing modules.

_Originally posted by @feltech in https://github.com/OpenAssetIO/OpenAssetIO/issues/554#issuecomment-1375559910_

## Notes

We should remove the warning added by https://github.com/OpenAssetIO/OpenAssetIO/pull/799 once this is done.

feltech:
Fixes #797. 

If the host thread does not hold the GIL when `createPythonPluginSystemManagerImplementationFactory` is called, then segfaults abound when trying to `import` the Python plugin system via pybind11.

So ensure the GIL is acquired first. pybind11's `gil_scoped_acquire` is clever enough to restore the state to whatever it was before, upon function exit.

## Reviewer Notes

Note that, as described below, this can be tested using OpenAssetIO's CTest suite, so no need for changes to the OpenAssetIO-Test-CMake integration test project to check this fix.

## Test Instructions

The `openassetio.internal.python-bridge-test` CTest in OpenAssetIO has been updated to trigger the bug.  If you leave that change, but remove the fix, and run CTest, then the segfault (or exception, if in Debug mode) will be apparent.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -9,6 +9,14 @@ v1.0.0-alpha.X
 - Disabled the (nascent) C bindings by default. To enable, the
   `OPENASSETIO_ENABLE_C` CMake option must be explicitly set to `ON`.
 
+### Bug fixes
+
+- Ensured that the Python GIL is acquired within
+  `createPythonPluginSystemManagerImplementationFactory`, so that it is
+  no longer necessary to acquire externally by the calling (host)
+  thread.
+  [#797](https://github.com/OpenAssetIO/OpenAssetIO/issues/797)
+
 
 v1.0.0-alpha.8
 --------------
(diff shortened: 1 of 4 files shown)


foundrytom (approved):
I'm having some build issues at the mo, so haven't been able to test locally, but this lookls exactly as I'd expect.

PR #805, opened 2023-01-30
[Core] Fix PyObject cleanup if interpreter gone

feltech:
The `PyRetainingSharedPtr` mechanism in OpenAssetIO allows a C++ `shared_ptr` chain to keep a `PyObject` alive, even if that object is not currently visible from the Python interpreter. This allows e.g. a C++ host to hold a reference to (and call) a Python manager plugin.

A side effect of this mechanism is that it is possible for a C++ object to keep a `PyObject` alive beyond the lifetime of the Python interpreter. Then, when the C++ object is finally destroyed (e.g. on application exit), attempting to clean up the `PyObject` is a use-after-free error, causing segfaults or hangs.

So ensure cleanup of the `PyObject` is trivial, in the case that the Python interpreter has been shut down, by not trying to acquire the GIL and by `release`ing the pybind11 handle.

- ~~[ ] I have updated all relevant user documentation.~~

## Test Instructions

Undoing the fix in `pointers.hpp` and running CTest (specifically, `openassetio.internal.pytest.cmodule`) will demonstrate the problem. The issue is use-after-free, so the symptoms can vary. On my system I see a hang (hence the test timeout added in the CMakeLists), you may see a segfault, or (if running a debug build of Python) an `assert` failure.

### DIFF
--- a/src/openassetio-python/private/include/openassetio/private/python/pointers.hpp
+++ b/src/openassetio-python/private/include/openassetio/private/python/pointers.hpp
@@ -4,6 +4,7 @@
 #include <memory>
 #include <utility>
 
+#include <Python.h>
 #include <pybind11/pybind11.h>
 
 #include <openassetio/export.h>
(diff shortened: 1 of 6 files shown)


src/openassetio-python/private/include/openassetio/private/python/pointers.hpp
```diff
@@ -48,10 +49,19 @@ Ptr createPyRetainingPtr(const py::object& pyInstance,
...
+                                      // If the Python interpreter is
+                                      // gone, clear the internal
+                                      // PyObject* so pybind11 won't
+                                      // attempt to clean it up.
+                                      pyObjectPtr->release();
+                                      delete pyObjectPtr;
+                                    } else {
+                                      // Acquire the GIL, in case deleter
+                                      // runs in a non-Python thread.
+                                      py::gil_scoped_acquire gil;
```
foundrytom, line 62: [question] Asking to make sure we've considered it: What are the failure modes of this call, and do we have to worry about them happening in a destructor?

src/openassetio-python/private/include/openassetio/private/python/pointers.hpp
```diff
@@ -48,10 +49,19 @@ Ptr createPyRetainingPtr(const py::object& pyInstance,
   // of the original shared_ptr holder stored on the PyObject.
   auto pyInstancePtr =
       std::shared_ptr<py::object>{new py::object{pyInstance}, [](py::object* pyObjectPtr) {
-                                    // Acquire the GIL, in case deleter
-                                    // runs in a non-Python thread.
-                                    py::gil_scoped_acquire gil;
-                                    delete pyObjectPtr;
+                                    if (_Py_IsFinalizing()) {
```
foundrytom, line 52: [question] Im struggling to find any docs on this method. Is it misleadingly named? In that it will also return true if it is already finalized?

foundrytom (approved):
Sorry I wont have a chance to properly verify this from an eng perspective today, so didn't want to hold this up. But looks sensible from a high-level p.o.v.

src/openassetio-python/private/include/openassetio/private/python/pointers.hpp
feltech: It's a bit of a wierd one. The leading underscore and lack of explicit docs makes it sound like something internal that shouldn't be called. However, [other docs](https://docs.python.org/3/c-api/init.html) explicitly say we can use it.

And yeah it's misleadingly named as you say.

src/openassetio-python/private/include/openassetio/private/python/pointers.hpp
feltech: Looking at the code of `gil_scoped_acquire`, it's mostly calls to CPython. The only exceptions that can be thrown are catastrophic errors with CPython that shouldn't ever occur, as far as I can tell.

foundrytom: Ace, thanks for the extra due dill.

feltech: Not a bad idea to wrap in a `try`/`catch`, though :thinking:

src/openassetio-python/private/include/openassetio/private/python/pointers.hpp
foundrytom: 🤣 solid foundations to the core.

src/openassetio-python/private/include/openassetio/private/python/pointers.hpp
foundrytom: That sounds more important than the try/catch? Given your notes on "the general badness that would be happening at the same time anyway"?

PR #809, opened 2023-02-02
[Core] Add ability to create a new manager for each harness test case

foundrytom:
The `openassetio.test.manager` harness used to create a single manager for the entire test run, reusing it for each case. This was mainly for performance reasons.

This does however, preclude the ability to test manager initialization behaviour.

This adds a class variable that instructs the harness to create a new, uninitialized manager for each case instead.

Part of OpenAssetIO/OpenAssetIO-Manager-BAL#26.

Note: I've had a world of a time getting builds working again on my laptop, so hopefully I haven't missed something obvious!

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -9,6 +9,16 @@ v1.0.0-alpha.X
 - Disabled the (nascent) C bindings by default. To enable, the
   `OPENASSETIO_ENABLE_C` CMake option must be explicitly set to `ON`.
 
+### New features
+
+- The `FixtureAugmentedTestCase` class of the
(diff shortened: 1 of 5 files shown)


elliotcmorris (approved):
Looks good to me.

A wee bit of a shame the need to add this extra complexity, but needs must.

feltech (approved):
Not 100% on `shareManager` as the variable name, but then nothing I come up with is any better.

An alternative approach would be to provide the `managerCreateFn` as a method on the `FixtureAugmentedTestCase`, rather than using class variable magic. But again, not sure that would be any better.

PR #811, opened 2023-02-07
Add CMake Lint

elliotcmorris:
Add a cmake-lint configuration file, as well as CI that runs it.
Update current cmake documents to standard style.

## Description

- [x] I have updated the release notes. I don't think something like this gets a release note?
- [x] I have updated all relevant user documentation. : The cmake config variables remain unchanged, but there was a slight update to a process document

## Reviewer Notes

Included in this PR is a big configuration document. This is the default configuration, taken from `cmake-lint` itself via dump command. There are a couple of modifications to it only
 - Line length set to 99
 - Tab width set to 2
 - Set to ignore the "No COMMENT provided" warning, as I did some research on this, and `COMMENT` vars are unreliable between generators, and the current echo approach is superior.

## Test Instructions

To test, clone repo, and run `find . \( -name "CMakeLists.txt" -o -wholename "*/cmake/*.cmake" \) | xargs cmake-lint`


cmake/CompilerWarnings.cmake
```diff
@@ -40,12 +41,12 @@ function(set_default_compiler_warnings target_name)
         -Wimplicit-fallthrough)
 
     if (OPENASSETIO_WARNINGS_AS_ERRORS)
-        set(CLANG_WARNINGS ${CLANG_WARNINGS} -Werror)
-        set(MSVC_WARNINGS ${MSVC_WARNINGS} /WX)
+        set(clang_warnings ${clang_warnings} -Werror)
+        set(vs_msvc_warnings ${vs_msvc_warnings} /WX)
```
elliotcmorris, line 45: I know this is an awkward name, but `MSVC_WARNINGS` and `MSVCPP_WARNINGS` are both built in symbols, so we were getting complaints about overlap.

CMakeLists.txt
```diff
@@ -231,13 +231,13 @@ endif ()
 
 include(StaticAnalyzers)
 if (OPENASSETIO_ENABLE_CLANG_FORMAT)
-    enable_clang_format()
+    SET_CLANG_FORMAT()
```
elliotcmorris, line 234: Renamed these macros to conform with the style, but also not cause confusion with `ENABLE_CLANG_FORMAT` variables

.github/workflows/code-quality.yml
```diff
@@ -100,6 +100,12 @@ jobs:
           clang-format --version
           cpplint --version
 
+      - name: CMake lint
+        shell: bash
+        run: |
+          python -m pip install cmakelang pyyaml
+          find . \( -name "CMakeLists.txt" -o -wholename "*/cmake/*.cmake" \) | xargs cmake-lint
```
feltech, line 107: [consider][minor] I had to look up `-wholename`, and `man` said `-path` is more portable

>        -wholename pattern
>              See -path.  This alternative is less portable than -path.

It works on CI so all good, but if someone e.g. on MacOS tried to lift this command to run locally, it might not work.

feltech, line 107: [question] Is there a reason to limit the `*.cmake` search to only files under the `cmake` directory, whereas `CMakeLists.txt` are searched throughout the repo?

Perhaps `-name "*.cmake"` would be better (we can add exclusions if there's anywhere we don't want it to search).

feltech, line 107: [consider][minor] Might be good to exclude hidden (i.e. `.`-prefixed) files, e.g. from [askubuntu](https://askubuntu.com/a/1127030)
```
find <start-point> -path '*/.*' -prune -o <expression> -print
```

CMakeLists.txt
```diff
@@ -92,9 +92,9 @@ endif ()
 # Note: this must come after `project(...)` otherwise
 # `CMAKE_CXX_COMPILER_ID` is undefined.
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    set(IS_GCC_OR_CLANG TRUE)
+    set(_is_gcc_or_clang TRUE)
```
feltech, line 95: [fix] `IS_GCC_OR_CLANG` is used in multiple cmake files, so I think should be uppercased.

CMakeLists.txt
feltech: If we're renaming anyway, we should namespace them, i.e. prefix with `OPENASSETIO_`.

feltech, line 234: [question] Not really loving the uppercase here, looks too much like a variable, and pretty much all functions I've seen in other projects are lowercased. How easy would it be to allow `lower_snake_case` for functions/macros? (is there a distinction between functions and macros?).

cmake-format.yaml
```diff
@@ -0,0 +1,246 @@
+_help_parse: Options affecting listfile parsing
+parse:
+  _help_additional_commands:
+    - Specify structure for custom cmake functions
+  additional_commands:
+    foo:
```
feltech, line 6: [fix] Presumably some leftover copy-paste example?

cmake/CompilerWarnings.cmake
feltech: Interesting, already a good catch by the linter!

cmake/CompilerWarnings.cmake
```diff
@@ -1,9 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2013-2022 The Foundry Visionmongers Ltd
 
+# Set platform compiler to have relevent warnings and checks enabled.
 function(set_default_compiler_warnings target_name)
 
-    set(CLANG_WARNINGS
+    set(clang_warnings
```
feltech, line 7: [question] How come no leading `_`?  Is it because it has function scope?

src/openassetio-core/tests/CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ target_link_libraries(
 # Create CTest target
 
 # Requires: openassetio.internal.install
+# (Disable line too long as it makes this command awkward.)
+# cmake-lint: disable=C0301
```
feltech, line 47: [question] Can we disable linter checks using a more human-readable tag?  With other linters you can disable by either the code or a string tag.

feltech (commented):
Looks really good, mostly just some questions.

CMakeLists.txt
elliotcmorris: Fair enough, I was just sticking to the default style, but will change macros to have similar requirements to functions.

That being the case, there's no need in my mind to distinguish these symbols from the variables, so will change the names back to ENABLE_

cmake-format.yaml
elliotcmorris: No, this cmake-format.yaml is a dump `cmake-lint --dump-config`. I think we should leave it as it stands rather than cutting it down, considering the default structure is autogenerated.

cmake/CompilerWarnings.cmake
elliotcmorris: Yeah.
`variable` is for local scope vars, whilst `_variable` is for what cmake-lint calls "private directory variables"

I guess this is the distinction between private and "really private" that you were upholding before.

src/openassetio-core/tests/CMakeLists.txt
elliotcmorris: I don't think so, it's why I've put the additional comment string above. I did try to put the english language description as the disable arg, but to no effect.

The only disable docs i can find are on using the code : https://cmake-format.readthedocs.io/en/latest/lint-inline.html

.github/workflows/code-quality.yml
elliotcmorris: There was a lot of stuff in the conan directory it was picking up initially, I'll try an exclusion first approach instead.

elliotcmorris: Good spot, replaced with path https://github.com/OpenAssetIO/OpenAssetIO/pull/811/commits/c37e253dcc71eb7d0c1f9c5c67b6a2002b21aa5e

elliotcmorris: The exclusion of the hidden folders solved this, https://github.com/OpenAssetIO/OpenAssetIO/pull/811/commits/c37e253dcc71eb7d0c1f9c5c67b6a2002b21aa5e

elliotcmorris: Yeah good idea.

Couldn't quite get that formulation working, but went with 
```
find . -not -path '*/.*' -a -name "CMakeLists.txt" -o -not -path '*/.*' -name "*.cmake"
 ```
which I think is okay.

https://github.com/OpenAssetIO/OpenAssetIO/pull/811/commits/c37e253dcc71eb7d0c1f9c5c67b6a2002b21aa5e

CMakeLists.txt
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/811/commits/99f68bddec6ad40fa6d7ec416f40c6fcc44b5e8e

elliotcmorris: Ended up setting these back to what they were before, so leaving them as is.

cmake-format.yaml
foundrytom: I think I side with David - having this here blurs what is intentional vs illustrative boilerplate. If it was commented out, that might be different. I totally get having parameters not left to implicit defaults, but active foo-barry seems to offer no real benefit. Would it make more sense to use this section to actually define our custom commands? (Assuming I understand its purpose correctly)

elliotcmorris: Removed the additional command example.

https://github.com/OpenAssetIO/OpenAssetIO/pull/811/commits/b6ca58463e247f22e3b5ef0e59f19f6d2cf39f0e

Reading the [docs](https://cmake-format.readthedocs.io/en/latest/format-features.html#custom-commands), it seems that the intent here is to handle custom commands defined by `add_custom_command` ,which we don't have.

This section may or may not be intended to also manage `add_custom_target`, which we have a lot of, the docs don't make that clear. I didn't manage to produce any formatting variations adding them there. Given this, I'd rather not get really drawn in trying to figure out this niche concern, and leave this section blank.

CMakeLists.txt
elliotcmorris: Despite the discussion at standup, I'd rather go with your intuition that most macros seem to follow function styling (lowercase) from the cmake you've seen, rather than the cmake-lint default, if you're agreeable.

.github/workflows/code-quality.yml
feltech: Just for fun, `find . -not -path '*/.*' \( -name "CMakeLists.txt" -o -name "*.cmake" \)` seems to produce the same result.

cmake-format.yaml
feltech: Yeah the docs aren't super clear.  Based on the example, I'm pretty certain this is a way to lint `function`s that use [`cmake_parse_arguments`](https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html) (we don't currently use that, but we really should for some of our functions that take multiple arguments).

src/openassetio-core/tests/CMakeLists.txt
feltech: One thing to be aware of, based on the docs, that `disable` applies for the whole (rest of the?) file, i.e. from the docs you've linked:

> must be specified at body scope [...] and it applies for the duration of the scope

So we should use sparingly, at least until it's made more configurable:

> behavior will likely change in the future

elliotcmorris: Aye, point well taken.

(Awkwardly, thanks cmake,) Wrapped these lines, rather than just disabling them.

https://github.com/OpenAssetIO/OpenAssetIO/pull/811/commits/d5be5c5aa62255eb7a8c54f4e686868ffbf3bb3e

feltech (approved):

PR #820, opened 2023-02-23
[CI] Fix CI for latest version of Conan/recipes

feltech:
The latest version of Conan does not install the CMake config files for `tool_requires` dependencies when using `conan install -g CMakeDeps`

We must use the latest version of Conan due to a recipe change in the `ncurses` package (a dependency of `cpython`).

`tool_requires` are intended for build requirements, when building a package via Conan.  We abuse this slightly during dev/CI by stopping short of building via Conan, instead using it just to provide dependencies. The latest Conan has changed its behaviour to no longer install `tool_requires` build system artifacts (i.e. CMake config files) as part of a `conan install`, which makes some sense since they shouldn't be required by consumers, only by builders.


feltech:
Closed to continue debugging on my fork.

PR #821, opened 2023-02-24
[CI] Fix CI for latest version of Conan/recipes

feltech:
The latest Conan recipe for the `ncurses` dependency of `cpython` is incompatible with Conan V1, so we must pin it.

One of the initial attempted fixes was to upgrade to the latest Conan. This may not actually have been required, but is worth keeping.  This necessitated a slight change in our conanfile.py to no longer use `tool_requires`, since their CMake config files are no longer installed (into the `.conan` directory).

Also fixed a couple of minor issues discovered whilst debugging
* The linter CI job was using the wrong GitHub runner (Ubuntu 22.04 rather than 20.04), meaning, at best, duplicated packages in the Conan package cache, or at worst, cache misses due to the caching race condition between jobs.
* A long-standing TODO comment was resolved, since the latest Conan versions no longer have the bug, where we must allow Conan to install system packages, even if none actually need to be installed.

See commit messages for more info.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -7,7 +7,7 @@ concurrency:
 
 jobs:
   pylint:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.config.os }}
     name: Pylint
     strategy:
       matrix:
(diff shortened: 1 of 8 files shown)


foundrytom (approved):
Nice one!

PR #827, opened 2023-03-03
[Docs] Add DR for `abortOnBatchElementError`

foundrytom:
A first pass... may need some tweaks!

Before this DR is merged, we need to add illustrative implementations for both the Host and Manager sides.

### DIFF
--- a/doc/decisions/DR016-Hinting-batch-failure-strategy.md
+++ b/doc/decisions/DR016-Hinting-batch-failure-strategy.md
@@ -1,0 +1,162 @@
+# DR016 Hinting batch failure strategy
+
+- **Status:** Decided
+- **Impact:** Medium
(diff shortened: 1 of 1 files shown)


foundrytom:
Note, we have since moved away from using the Context as a dumping ground for options that affect specific API calls. Instead, reserving it for "things that categorize the workflow the context represents" (#1054). As such this conclusion may no longer be valid.

foundrytom:
> Note, we have since moved away from using the Context as a dumping ground for options that affect specific API calls. Instead, reserving it for "things that categorize the workflow the context represents" (#1054). As such this conclusion may no longer be valid.

Closing, DR captured [here](https://github.com/OpenAssetIO/OpenAssetIO/issues/859) for posterity.

PR #828, opened 2023-03-09
[Core] Remove [[nodiscard]] from getTraitProperty

### LINKED ISSUE #825: [Core] `TraitsData::getTraitProperty` should not be `[[nodiscard]]`
## What

Remove `[[nodiscard]]` attribute from `TraitsData::getTraitProperty`, ~or otherwise redesign `getTraitProperty`.~

## Why

The design of `getTraitProperty` uses an out-parameter for the value the user is most interested in. The return value is a `bool` indicating whether the property is set.  A possible valid use case is ignoring the return value, e.g. in a "value or default" flow:

```c++
openassetio::Str value = "fallback";
traitsData->getTraitProperty(&value, traitId, key);
```
where `value` will be the stored value if set, or will remain "fallback" if not.  In this case, the returned `bool` is not interesting and can be safely ignored.

elliotcmorris:
## Description

The value of getTraitProperty is returned via an out param. It's actual return value is merely a status flag.

It's possible the user may wish to employ a "value or default" style, which would not be possible with [[nodiscard]]

Closes #825

## Reviewer Notes
Also applied the same logic to `getTraitProperty` in `TraitBase`, it applies the same I think.

## Test Instructions

To test, building and running `ctest` is sufficient.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,6 +16,13 @@ v1.0.0-alpha.X
   mechanism for testing whether a `TraitsData` is imbued with a trait.
   [#815](https://github.com/OpenAssetIO/OpenAssetIO/issues/815)
 
+### Bug fixes
+
+- Removed `nodiscard` from `TraitsData::getTraitProperty`, and
+  `TraitBase::getTraitProperty`, to allow "value or default" style use
+  cases.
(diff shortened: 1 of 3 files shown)


feltech (approved):
LGTM

PR #840, opened 2023-03-20
[CI] Fix TraitGen integration test dependencies

feltech:
TraitGen now includes C++ tests. This causes two issues. Firstly, the C++ AST is tested with a third-party library, "tree-sitter". So we must install the appropriate `requirements.txt` to get that dependency.

Secondly, the C++ CMake CTest tests require an OpenAssetIO C++ install tree. This will be fixed as a follow-up issue (#841). In the meantime they will be automatically skipped, since we don't specify the required `OPENASSETIO_TRAITGENTEST_CMAKE_PRESET` environment variable.

### DIFF
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -55,7 +55,10 @@ jobs:
         run: |
           # We don't use the local requirements.txt for each
           # integration as they will try to install openassetio
-          # and we want to use the working tree
+          # and we want to use the working tree.
+          # Note: if we do need to install specific test dependencies
+          # for an integration, ensure that it does not override our
+          # `openassetio` package.
           python -m pip install pytest
 
       - name: Checkout BAL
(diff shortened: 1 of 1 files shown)


.github/workflows/integrations.yml
```diff
@@ -85,6 +88,11 @@ jobs:
         run : |
           python -m pip install external/TraitGen
 
+      - name: Install TraitGen test dependencies
+        run : >
+          python -m pip install -r external/TraitGen/tests/requirements.txt
+          | grep "Requirement already satisfied: openassetio"
```
foundrytom, line 94: [question] What is the purpose of the grep here?

foundrytom: If you can please, otherwise it just look like were filtering output...  So the intention is that this step will error if thats the case and fail the rest of the job - so it is a canary rather than something that actually affects what is installed? Also, have we double checked this actually happens (i.e. is `pipefail` definitely set? We've been burnt by that before now.

elliotcmorris (approved):
Good for me, apart from echoing what tom said.

foundrytom (approved):

PR #842, opened 2023-03-20
[Docs] Add more examples to displayName and identifier docs

### LINKED ISSUE #829: Improve documentation for Host Interface returned strings
## What

Give more context and better examples of `identifier` and `displayName`.

foundrytom:
Our previous examples lead people to think they need to include 'Host' or similar in these strings. This attempts to make it clearer that it's supposed to just be the common name for host/manager in question.

Closes #829

### DIFF
--- a/src/openassetio-core/include/openassetio/hostApi/HostInterface.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/HostInterface.hpp
@@ -62,16 +62,22 @@ class OPENASSETIO_CORE_EXPORT HostInterface {
    * example:
    *
    *    "org.openassetio.test.host"
+   *    "io.aswf.openrv"
+   *    "com.foundry.nuke"
    *
    * @return host identifier.
+   *
+   * @see https://en.wikipedia.org/wiki/Reverse_domain_name_notation
    */
   [[nodiscard]] virtual Identifier identifier() const = 0;
 
   /**
    * Returns a human readable name to be used to reference this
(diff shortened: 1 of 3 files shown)


src/openassetio-core/include/openassetio/hostApi/HostInterface.hpp
```diff
@@ -72,6 +74,8 @@ class OPENASSETIO_CORE_EXPORT HostInterface {
    * specific host in user-facing presentations.
    *
```
elliotcmorris, line 75: [consider][minor] This might read better like above, with a "for example:" before the list of examples. Unlikely, but possible someone might think this list is the set of supported hosts or something.

elliotcmorris (approved):

src/openassetio-core/include/openassetio/hostApi/HostInterface.hpp
foundrytom: Good spot - thanks. Definitely. Added [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/842/commits/d5843ecb70ca8a929e2431799fd88a0035b907fa)

elliotcmorris (approved):

feltech (approved):
LGTM

foundrytom:
@elliotcmorris added you RDNS note here too https://github.com/OpenAssetIO/OpenAssetIO/pull/842/commits/0d5c037fed0f04e6193baa1564e106fbb499097c

PR #843, opened 2023-03-22
Update Integrations to support new MediaCreation

elliotcmorris:
Updates integrations to support the new mediacreation which requires traitgen.

~~- [ ] I have updated the release notes.~~
~~- [ ] I have updated all relevant user documentation.~~

### DIFF
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -147,17 +147,32 @@ jobs:
         env:
           CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
 
+      - name: Checkout TraitGen
+        uses: actions/checkout@v3
+        with:
+          repository: OpenAssetIO/OpenAssetIO-TraitGen
+          path: external/TraitGen
+
+      - name: Install TraitGen
+        run : |
+          python -m pip install external/TraitGen
+
       - name: Checkout OpenAssetIO-Test-CMake
         uses: actions/checkout@v3
         with:
           repository: OpenAssetIO/OpenAssetIO-Test-CMake
           path: external/OpenAssetIO-Test-CMake
 
+      - name: "Get Submodules"
+        run: |
+          cd external/OpenAssetIO-Test-CMake
+          git submodule update --init --recursive --remote OpenAssetIO-MediaCreation
+
       - name: Test CMake package
         run: |
           ${{ matrix.config.preamble }}
           cd external/OpenAssetIO-Test-CMake
-          cmake -S . -B build -G Ninja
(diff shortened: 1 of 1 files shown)


.github/workflows/integrations.yml
```diff
@@ -147,17 +147,33 @@ jobs:
...
         with:
-          repository: OpenAssetIO/OpenAssetIO-Test-CMake
+          repository: elliotcmorris/OpenAssetIO-Test-CMake
           path: external/OpenAssetIO-Test-CMake
+          ref: work/mediacreation
+
+      - name: "Get Submodules"
+        run: |
+          cd external/OpenAssetIO-Test-CMake
+          git submodule update --init --recursive OpenAssetIO-MediaCreation
```
feltech, line 170: [question] Submodules screw with my head.  I think this will use the last commit referenced in the `.gitmodules` file of the parent repo (Test-CMake)?  

Presumably you want to get the latest MediaCreation. I think you need to add `--remote --checkout`. See [the docs](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---remote) and/or [stack overflow](https://stackoverflow.com/a/42175412/535103).

.github/workflows/integrations.yml
```diff
@@ -147,17 +147,33 @@ jobs:
...
+
+      - name: Install TraitGen
+        run : |
+          python -m pip install external/TraitGen
+
       - name: Checkout OpenAssetIO-Test-CMake
         uses: actions/checkout@v3
         with:
-          repository: OpenAssetIO/OpenAssetIO-Test-CMake
+          repository: elliotcmorris/OpenAssetIO-Test-CMake
```
feltech, line 163: [fix] Just to ensure this isn't forgotten.

feltech (approved):
I think this is a good balance for now.

Ideally we'll find a way to (parametrise and) trigger the CI of Test-CMake and report back the results, rather than having this reduced matrix.

.github/workflows/integrations.yml
elliotcmorris: Aye i'm just gonna drop that commit.

.github/workflows/integrations.yml
feltech: Oh but in `.gitmodules` in Test-CMake we have `branch = main`, so does that mean this works as is?  Like I said, submodules hurt my brain.

elliotcmorris: It was my understanding that since the CI runners are always "fresh", it will always fetch the head of main.
The --remote thing it only needed (to my understanding, submodules are also new to me,) when you need to update to remote when your local cache from a previous `init` says otherwise.

PR #860, opened 2023-03-27
[Docs] Document plugin loading and config mechanisms

### LINKED ISSUE #770: Document entry point based plugin loading
## What

Document how the new entry point based plugin loading can be used.

## Why

Or no-one will know!

## ACs

- Doxygen documentation updated with:
   -  An "end user" focused section that covers "how to install plugins", that explains both the search path and entry point based mechanism.
   -  Glossary entry for "entry point plugins"
   - Any other references to plugin loading updated to include relevant links.

foundrytom:
An attempt to capture the snippets of setup notes scattered around the docs in once place.

Pre-built version here: https://github.com/OpenAssetIO/OpenAssetIO/suites/11835979369/artifacts/618358855

`index.html` -> Configuring OpenAsset IO

Closes #770

### DIFF
--- a/doc/doxygen/src/Configuration.dox
+++ b/doc/doxygen/src/Configuration.dox
@@ -1,0 +1,172 @@
+/*!
+ * @page runtime_configuration Configuring OpenAssetIO
+ *
+ * This page covers how to configure OpenAssetIO for a specific
+ * end-user working environment.
+ *
+ * @section runtime_configuration_overview Overview
+ *
+ * @startuml
+ * node Host {
+ *   package OpenAssetIO {
+ *     collections ManagerPlugin
+ *   }
+ * }
+ *
+ * cloud "Asset Management System" as ams
+ *
+ * ManagerPlugin -right-> ams
+ * ams -left-> ManagerPlugin
+ * @enduml
+ *
+ * An OpenAssetIO enabled tool or application is known as a @ref host
+ * "Host" (as it "hosts" the API). Through the use of manager plugins,
+ * OpenAssetIO allows the host to communicate with one or more
+ * @ref asset_management_system "Asset Management Systems", called
+ * @ref manager "Managers".
+ *
+ * This page covers hows to configure OpenAssetIO so that the hosts you
+ * use (e.g. @ref DCC "DCC tools", applications and other scripts) use a
+ * specific manager of your choice to @ref resolve and potentially @ref
+ * publish data.
+ *
+ * OpenAssetIO itself is usually built into any host that supports it,
+ * this means you don't normally need do to anything to ensure the API
+ * itself is present. The most common tasks in configuration are:
+ *
+ * - Installing manager plugins and making them available to a host.
+ * - Configuring which manager a host should use.
+ *
+ *
+ * @section installing_manager_plugins Installing Manager plugins
+ *
+ * @note To aid problem solving, the plugin system logs debug messages
+ * when it searches for, and loads plugins.
+ *
+ * Presently, OpenAssetIO only supports Python based plugins.
+ * These can be installed in one of two ways.
+ *
+ * @subsection installing_manager_plugins_pip Using pip
+ *
+ * Most Python-based plugins should support what is known as
+ * <a href="https://packaging.python.org/en/latest/specifications/entry-points/" target="_blank">"entry point" based loading</a>.
+ * This means they can be installed using the package installer for
+ * Python, and will then be automatically detected.
+ *
+ * These plugins are usually published to
+ * <a href="https://pypi.org" target="_blank">PyPI</a>, and so are
+ * easy to install. Simply use the `pip` module of the Python
+ * installation you wish to make the plugin available to:
+ *
+ * @code{.sh}
+ * python -m pip install <manager-plugin-package>
+ * @endcode
+ *
+ * If the plugin is not published, you may be able to install it from a
+ * local checkout of its source code:
+ *
+ * @code{.sh}
+ * python -m pip install <path/to/src>
+ * @endcode
+ *
+ * @warning Many hosts use their own embedded Python interpreter. You
+ * may need to consult the relevant documentation to determine its
+ * location, version, `sys.path` etc. and adjust the invocation
+ * of `python` accordingly, so that the plugin is installed to the
+ * correct Python instance.
+ *
+ *
+ * @subsection installing_manager_plugins_manual Manual installation
+ *
+ * The OpenAssetIO plugin system can also be explicitly configured to
+ * look in specific locations. The @ref plugin_path_var environment
+ * variable can be set to point to one or more directories containing
+ * manager plugins.
+ *
+ * It is a standard 'left-most wins' search-path, that uses platform
+ * specific delimiters (see @ref plugin_path_var "here" for more
+ * information).
+ *
+ * This variable is completely independent from `$PYTHONPATH` or any
+ * other language/host specific configuration.
+ *
+ * This can simplify deployment by allowing a common location to be
+ * specified to service multiple hosts and Python versions (assuming the
+ * manager plugins have been suitably coded).
+ *
+ * To install a plugin using this approach:
+ *
+ * - Download or otherwise obtain the required manager plugin.
+ * - Place this in your chosen location. This can be anywhere on a
+ *   filesystem that is visible to the hosts that you wish to use.
+ * - Set, or extend @ref plugin_path_var to include this location.
+ *
+ *
+ * @section host_settings Host configuration
+ *
+ * In order to make use of OpenAssetIO to manage data, the host needs to
+ * know which manager plugin it should use, and any associated settings
+ * that plugin many need (e.g. server configuration).
+ *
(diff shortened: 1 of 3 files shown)


doc/doxygen/src/Configuration.dox
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * cloud "Asset Management System" as ams
+ *
+ * ManagerPlugin -right-> ams
+ * ams -left-> ManagerPlugin
+ * @enduml
+ *
+ * An OpenAssetIO enabled tool or application is known as a @ref host "Host"
+ * (as it "hosts" the API, tenuous - but naming things is hard). Through
+ * the use of a plugin system, OpenAssetIO allows the host to
```
elliotcmorris, line 24: [Consider]
```suggestion
 * (as it "hosts" the API). Through
 * the use of a plugin system, OpenAssetIO allows the host to
```
I don't think this aside is that helpful for comprehension.

elliotcmorris, line 24: [Consider]
```suggestion
 * (as it "hosts" the API, tenuous - but naming things is hard). Through
 * the use of a manager plugin, OpenAssetIO allows the host to
```
Unsure on this one. My thinking is that we use the term `manager plugin` below without really defining it, and that the "plugin system", isn't really what lets us communicate. The manager plugins let us communicate, and the plugin system lets us load the manager plug ins.

doc/doxygen/src/Configuration.dox
```diff
@@ -0,0 +1,174 @@
...
+ * local checkout of its source code:
+ *
+ * @code{.sh}
+ * python -m pip install <path/to/src>
+ * @endcode
+ *
+ * @warning Many hosts use their own embedded Python interpreter. You
+ * may need to consult the relevant documentation to determine its
+ * location, version, `sys.path` etc. and adjust the invocation
+ * of `python` accordingly.
```
elliotcmorris, line 75: [Consider]
```suggestion
 * location, version, `sys.path` etc. and adjust the invocation
 * of `python` accordingly, so that the plugin is installed to the
 * correct python instance.
```

I think that's the reason this warning exists right, to make sure folks install the libs to the correct python version? If so, I think the extra explanation reduces ambiguity.

doc/doxygen/src/Configuration.dox
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * This mechanism is easiest, but as `pip` installs packages into a
+ * specific python version, it can be harder to manage with multiple
+ * hosts if each has its own interpreter/version.
```
elliotcmorris, line 79: [Question][Minor] Is this actually true you think? I have no experience with the matter, but it seems to me managing this via the manual workflow would be similarly difficult, albeit different problems. (I wonder about plugins built to support only python 3.11> for example, pip would help guard against this, but the manual workflow wouldn't.)

elliotcmorris (commented):
:+1: Big fan of this page.

doc/doxygen/src/Configuration.dox
```diff
@@ -0,0 +1,174 @@
...
+ * use this configuration to setup the API at startup.
+ *
+ * The format for manager-specific sections of this file is documented
+ * @fqref{hostApi.ManagerFactory.defaultManagerForInterface} "here".
+ *
+ * To configure OpenAssetIO using this approach:
+ *
+ * - Create a TOML file with the identifier and any settings in a
+ *   location of your choosing. This can be anywhere on a filesystem
+ *   that is visible to the hosts that you with to use.
```
feltech, line 136: [fix]
```suggestion
 *   that is visible to the hosts that you wish to use.
```

doc/doxygen/src/Configuration.dox
feltech, line 1: [consider] Consistent capitalisation of "Python" vs. "python"

doc/doxygen/src/Configuration.dox
```diff
@@ -0,0 +1,174 @@
...
+ * @subsection pip_plugins_missing Pip installed Python plugins not being found
+ *
+ * - Check any messages logged by the plugin system during startup.
+ * - Check with the plugin's documentation that it does support entry
+ *   point based loading.
+ * - Check that the Python interpreter used by your host is the same
+ *   version as the one used to install via `pip`.
+ * - Check that the appropriate directory is on `sys.path` in the
+ *   host's interpreter.
+ * - Check that importlib_metadata` is available to the host interpreter
```
feltech, line 159: [fix]
```suggestion
 * - Check that `importlib_metadata` is available to the host interpreter
```

feltech (approved):
Looks good, minor typographical things.

doc/doxygen/src/Configuration.dox
feltech: Yeah different versions of the interpreter would need a different plugin installation.  But different applications, each with their own interpreter, but all using the same Python version, would be easier to configure with a system-wide env var, than by independently installing the plugin into all of them.

foundrytom: If you have mimumum python version support for a python-only plugin (a common approach, as we do), then this is _much_ easier with the plugin path approach.

Eg, BAL is 3.7-3.10 compatible by design. This means it could be installed once and used with a singular point of consistent configuration, with any host from CY2020-CY2023, without needing to worry about how each particular host manages its `sys.path`.

foundrytom: Version management can also be easier with a non `pip` approach, as with, e.g. Rez, you can have a single installation point for all available plugin versions and the appropriate version for a project selected at runtime (though this is true with `PYTHONPATH` or `OPENASSETIO_PLUGIN_PATH`, just not using `pip`). 

Maybe worth qualifying that it obviously requires supporting plugins?

We will also need the manual install approach for non-python plugins, and that section will be common, so its partially setting the stage for later in terms of doc structure.

foundrytom: Ill try and rephrase this para to be less opinionated and more factual, then people can make their own mind up.

doc/doxygen/src/Configuration.dox
foundrytom: But the whimsy, the whimsy! 🤣 Removed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/860/commits/da4b28697cbc15b230dc72e29c13af56e80bdb50)

foundrytom: Good idea, updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/860/commits/da4b28697cbc15b230dc72e29c13af56e80bdb50).

doc/doxygen/src/Configuration.dox
foundrytom: https://github.com/OpenAssetIO/OpenAssetIO/pull/860/commits/3ebc71bdda17d53e4ed89bec330ae8a93f9c1846

doc/doxygen/src/Configuration.dox
foundrytom: Simplified [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/860/commits/3f224a8f12b49f018a5381170791b3904fd76a09)

doc/doxygen/src/Glossary.dox
```diff
@@ -277,6 +284,14 @@
  * of the standard plug-in.
  *
  *
+ * @section logging_severity_var $OPENASSETIO_LOGGING_SEVERITY
+ *
+ * This is the environment variable used to control the default logging
+ * severity of the API. Note that many @ref host "hosts" re-direct
+ * logging to their own system. Usually in these cases, logging severity
+ * is then controlled by the hosts own configuration mechanism(s).
```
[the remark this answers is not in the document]
foundrytom: Fixed [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/860/commits/c3e8ea37d87024cc2bf0a697add119312cccc152#diff-98ec7fb4d9eedc0416b178d2e5c4d4f64f8364259f67823a627880b690872895), please ignore stray file.

doc/doxygen/src/Configuration.dox
foundrytom: https://github.com/OpenAssetIO/OpenAssetIO/pull/860/commits/398a6c1cc19a7321693a7383f46db96888d5aa04

doc/doxygen/src/Configuration.dox
foundrytom: Ill have a `P` please Bob https://github.com/OpenAssetIO/OpenAssetIO/pull/860/commits/3de584fe302f9e5d1f0f08daf12fc6402f7c3105

doc/doxygen/src/Configuration.dox
foundrytom: https://github.com/OpenAssetIO/OpenAssetIO/pull/860/commits/a85a902c2e2bdb1538ae473cbd6eff6989df8758

elliotcmorris (approved):

elliotcmorris:
LGTMSAMWR

feltech (approved):
LGTM

PR #861, opened 2023-03-27
[CI] Fix docs build artifact path

foundrytom:
When we refactored the repo, we inadvertently neglected to update the path in the step that publishes the built docs as an artifact of the run.

Test from CI run: https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/4532689159


elliotcmorris (approved):
ty for noticing this.

PR #868, opened 2023-03-31
Remove `entityName` and `entityDisplayName`

### LINKED ISSUE #837: Remove `entityDisplayName`
## What

Remove `ManagerInterface.entityDisplayName`, and replace with a `DisplayName` trait in MediaCreation.

## Why

Now we have a composable traits system, that allows a host to express the entity data they need, in a concise (and reflectable form, via `managementPolicy`), making the entity name a first-class concept has several flaws:

- It requires an additional API call to service, which can add non-trivial work.
- In many situations where it would be called, a `resolve` call would also be made to provide other information.
- It prohibits the addition of additional fields, eg, providing a shot name and a long name (maybe for UI + tooltip).

By making a trait, we can hopefully reduce the total number of API calls, and add more flexibility over time.

foundrytom:
This is better served by a resolvable trait to reduce API calls and allow industry specific flexiblity.

Closes #837

See https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/pull/30


feltech (approved):
LGTM, I can't find any trace of them after this.

elliotcmorris (approved):
`EntityName` is mentioned tangentially in the usdResolver wiki, but that was a planning document and I don't think qualifies as staying up to date.

PR #869, opened 2023-03-31
Switch to single line spacing in `RELEASE_NOTES`

foundrytom:
This means markdownlint and others can be used to keep an eye on formatting.


RELEASE_NOTES.md
```diff
@@ -32,7 +32,6 @@ v1.0.0-alpha.X
   cases.
   [#825](https://github.com/OpenAssetIO/OpenAssetIO/issues/825)
 
-
```
feltech, line 35: [question] I really don't feel strongly about it, but a double-line break between version headings would look better aesthetically.  

Do you know if this is even possible with markdownlint?  

A quick look at the [default config example](https://github.com/DavidAnson/markdownlint/blob/main/schema/.markdownlint.yaml#L105) seems to imply that all heading levels are treated the same, so we couldn't enforce different spacing for different heading levels.

feltech (approved):
LGTM other than minor question

RELEASE_NOTES.md
foundrytom: Yeah, I agree on readability, I couldn't find a sensible way - and it was just becoming a source of churn...

elliotcmorris (approved):

PR #872, opened 2023-04-03
[Examples] Clarify use of ConsoleLogger in simpleResolver

foundrytom:
Just to help make sure that people realize they should bridge to their own system in real-world scenarios.


feltech (approved):
LGTM

PR #873, opened 2023-04-03
[Build] Always set `-std=c++17`

foundrytom:
By defalut, CMake ommits this if the compiler default matches the project. We run clang-tidy as part of build, version mismatches can resut in failed checks if the discoverd linter has an older default std. Explicitly setting this ensures checks run with the right std version.


feltech (approved):
LGTM

PR #874, opened 2023-04-03
Fix clang-tidy v15 warnings

feltech:
Although we have yet to move to the newer `clang-tidy` on CI, we aim to support devs using a newer `clang-tidy` than v12.

## Test Instructions

We seem to be getting cross-platform differences coming from a `std::rethrow_exception` call in `src/openassetio-python/cmodule/src/BatchElementErrorBinding.cpp`.  That is, MacOS gives a warning that the `std::exception_ptr` cannot be `std::move`d, for some reason.  See also https://github.com/OpenAssetIO/OpenAssetIO/pull/867#issuecomment-1492211562

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -44,6 +44,10 @@ v1.0.0-alpha.X
 - Made `BatchElementError` a copyable type in the C++ API.
   [#849](https://github.com/OpenAssetIO/OpenAssetIO/issues/849)
 
+- Made the C++ codebase compliant with Clang-Tidy v15. Note that this is
+  not yet enforced on CI.
+  https://github.com/OpenAssetIO/OpenAssetIO/pull/874
+
 ### Bug fixes
 
 - Removed `nodiscard` from `TraitsData::getTraitProperty`, and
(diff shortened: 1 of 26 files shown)


foundrytom:
Just noting:
```
 rc/openassetio-core/include/openassetio/BatchElementError.hpp|48 col 7| error: an exception may be thrown in function 'BatchElementError' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
src/openassetio-core/include/openassetio/EntityReference.hpp|35 col 7| error: an exception may be thrown in function 'EntityReference' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp|43 col 10| error: an exception may be thrown in function 'ManagerDetail' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
``` 
On macOS which @feltech discovered as being https://github.com/llvm/llvm-project/issues/54668.

foundrytom (approved):
Good fixes, thanks. We're no worse off on mac/etc than we were, so can tackle those in followups if you like. Do we need an Improvements entry in the release notes?

feltech:
> Good fixes, thanks. We're no worse off on mac/etc than we were, so can tackle those in followups if you like. Do we need an Improvements entry in the release notes?

Updated release notes in 3194a59f

elliotcmorris (approved):

PR #878, opened 2023-04-04
Default to `BatchElementError` in `variant` returns

### LINKED ISSUE #876: Feature Title
## What

Spun off from #859. Make `BatchElementError` the default type in returned `variant`s.

## Why

Currently `resolve` can return a `variant<TraitsDataPtr, BatchElementError>`. This seems the wrong way around, in particular when default-constructed, where the current formulation would create a `variant` holding a `nullptr` `TraitsData`, which is dangerous and less informative than a `BatchElementError` with a suitable default error code.

In the Notes for #859. the following idea was mooted:
* Make `BatchElementError` default construction have a `kSkipped` (or similar) status.
* Make `BatchElementError` the default type in variant returns.
* Then a default-constructed `std::vector<std::variant<BatchElementError, ...>>` will be a vector of "skipped" states, to be replaced if possible with "real" values.

It may be a while before we get around to that. But changing the `variant` ordering is an ABI break, so would be good to do at least this minimal change sooner rather than later.

_cough_ typedef _cough_

feltech:
Closes #876. Currently `variant`s returned from `resolve` seem the wrong way around, in particular when default-constructed, where the current formulation would create a variant holding a `nullptr` `TraitsDataPtr`, which is dangerous and less informative than a `BatchElementError` with a default error code.

## Description

Closes # (issue)

## Reviewer Notes

Does the volume of changes, and the fact that it's easy to get the ordering wrong, lend some weight to the idea of having a `typedef` for this?

### DIFF
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -664,7 +664,7 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * @return Object containing either the populated data or an error
    * object.
    */
-  std::variant<TraitsDataPtr, BatchElementError> resolve(
+  std::variant<BatchElementError, TraitsDataPtr> resolve(
       const EntityReference& entityReference, const trait::TraitSet& traitSet,
       const ContextConstPtr& context, const BatchElementErrorPolicyTag::Variant& errorPolicyTag);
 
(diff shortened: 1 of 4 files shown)


elliotcmorris (approved):
I could equally make the argument that if the type wasn't exposed, we wouldn't have noticed the ordering thing until later on.

foundrytom (dismissed):
Needs `RELEASE_NOTES`

foundrytom:
> Needs `RELEASE_NOTES`

Sorry I forgot we haven't released this yet.

PR #880, opened 2023-04-05
Python bindings for `preflight` conveniences

feltech:
## Description

Closes #852

Add Python bindings to `preflight` overloads that allow single-element and/or exception-less workflows.

- [x] I have updated the release notes.
- [x] I have updated all relevant user documentation.

PR #881, opened 2023-04-05
[Core] Add eq comparison to BatchElementError

### LINKED ISSUE #862: [Core] Add equality comparison operator to BatchElementError
## What

Add equality operator to `BatchElementError`.

## Why

This would be useful for tests, if nothing else.

## ACs

- Given two different error objects (with different addresses) `error1 == error2`
- No temp comparison implementations remain in tests/etc
- Tests that compare BEEs (🐝) updated.
- `!=` behaviour also tested if required but not if inherent in the language (C++/Python, create issue for TraitsData if were missing anything there)

## Notes

Implement as member function to follow prior art

lpapp-foundry:
## Description

Closes #862

- [X] I have updated the release notes.
- [X] I have updated all relevant user documentation.

### DIFF
--- a/src/openassetio-core/include/openassetio/BatchElementError.hpp
+++ b/src/openassetio-core/include/openassetio/BatchElementError.hpp
@@ -110,6 +110,23 @@ class BatchElementError final {
     kEntityResolutionError = OPENASSETIO_BatchErrorCode_kEntityResolutionError
   };
 
+  /**
+   * Compares two instances to see if they match.
+   *
+   * @return `true` if the batch elements errors match, otherwise `false`.
+   */
+  constexpr bool operator==(const BatchElementError& other) const {
+    return code == other.code && message == other.message;
+  }
+
+  /**
+   * Compares two instances to see if they do not match.
+   *
+   * @return `true` if the batch elements errors do not match, otherwise
+   * `false`.
+   */
+  constexpr bool operator!=(const BatchElementError& other) const { return !(*this == other); }
+
   /// Error code indicating the class of error.
   ErrorCode code;
   /// Human-readable error message.
--- a/src/openassetio-core/tests/BatchElementErrorTest.cpp
+++ b/src/openassetio-core/tests/BatchElementErrorTest.cpp
@@ -26,5 +26,27 @@ SCENARIO("BatchElementError usage") {
         CHECK(error.message == "some message");
       }
     }
+
+    WHEN("Two BatchElementErrors are constructed wrapping the same code and message") {
+      const BatchElementError error{code, message};
+      const BatchElementError error2{code, message};
+
+      THEN("The errors instances match") { CHECK(error == error2); }
+    }
+
+    WHEN("Two BatchElementErrors are constructed wrapping different code and the same message") {
+      const BatchElementError error{code, message};
+      const BatchElementError error2{BatchElementError::ErrorCode::kEntityResolutionError,
+                                     message};
+
+      THEN("The errors instances do not match") { CHECK(error != error2); }
+    }
+
+    WHEN("Two BatchElementErrors are constructed wrapping the same code and different message") {
+      const BatchElementError error{code, message};
+      const BatchElementError error2{code, "another message"};
+
+      THEN("The errors instances do not match") { CHECK(error != error2); }
+    }
   }
 }
--- a/src/openassetio-python/cmodule/src/BatchElementErrorBinding.cpp
+++ b/src/openassetio-python/cmodule/src/BatchElementErrorBinding.cpp
@@ -3,6 +3,7 @@
 #include <string_view>
 
 #include <pybind11/eval.h>
+#include <pybind11/operators.h>
 #include <pybind11/pybind11.h>
 
 #include <openassetio/BatchElementError.hpp>
@@ -24,6 +25,7 @@ void registerBatchElementError(py::module &mod) {
   batchElementError
       .def(py::init<BatchElementError::ErrorCode, openassetio::Str>(), py::arg("code"),
            py::arg("message"))
+      .def(py::self == py::self)  // NOLINT(misc-redundant-expression)
       .def_readonly("code", &BatchElementError::code)
       .def_readonly("message", &BatchElementError::message);
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -51,6 +51,9 @@ v1.0.0-alpha.X
 - Made `BatchElementError` a copyable type in the C++ API.
   [#849](https://github.com/OpenAssetIO/OpenAssetIO/issues/849)
 
+- Added equality/inequality comparison operators to `BatchElementError`.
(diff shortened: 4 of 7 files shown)


src/openassetio-python/cmodule/src/BatchElementErrorBinding.cpp
```diff
@@ -24,6 +25,8 @@ void registerBatchElementError(py::module &mod) {
   batchElementError
       .def(py::init<BatchElementError::ErrorCode, openassetio::Str>(), py::arg("code"),
            py::arg("message"))
+      .def(py::self == py::self)
+      .def(py::self != py::self)
```
feltech, line 29: [consider] I believe (though not 100% sure) that `py::self != py::self` is unnecessary if `py::self == py::self` is defined.

src/openassetio-python/cmodule/src/BatchElementErrorBinding.cpp
```diff
@@ -24,6 +25,8 @@ void registerBatchElementError(py::module &mod) {
   batchElementError
       .def(py::init<BatchElementError::ErrorCode, openassetio::Str>(), py::arg("code"),
            py::arg("message"))
+      .def(py::self == py::self)
```
feltech, line 28: [fix] Would be good to have a little `pytest` unit test for `==` and `!=` .

src/openassetio-core/include/openassetio/BatchElementError.hpp
```diff
@@ -116,6 +116,24 @@ class BatchElementError final {
   Str message;
 };
 
+/**
+ * Compares two batch element error instances to see if they match.
+ *
+ * @return `true` if the batch elements errors match, otherwise `false`.
+ */
+constexpr bool operator==(const BatchElementError& lhs, const BatchElementError& rhs) {
```
feltech, line 124: [question] Curious why you chose a free function rather than a member function?

feltech (commented):
Really good stuff. Just some minor comments/questions.

src/openassetio-core/include/openassetio/BatchElementError.hpp
lpapp-foundry: Scott Meyers one of the known and trusted C++ members seems to think that non-member functions increase encapsulation: https://www.drdobbs.com/cpp/how-non-member-functions-improve-encapsu/184401197

Whilst investigating, I have also found these stack overflow references:

https://stackoverflow.com/a/24938533/2682142
https://stackoverflow.com/a/4622467/2682142

foundrytom: Fwiw, We had made a note when estimating the issue to do it as a member function for consistency with the other exiting implementations. Also, in other classes it has to consider private members.

src/openassetio-python/cmodule/src/BatchElementErrorBinding.cpp
foundrytom: As we have an explicit implementation of `!=` which we know isn't covered by `pytest` would be good to have a cpp test case too?

src/openassetio-core/include/openassetio/BatchElementError.hpp
feltech: The Meyers article has a compelling argument, which boils down to: when deciding the impact of changing the private API you have to consider all member functions. So removing member functions, where possible, reduces work when redesigning/refactoring the private API. I.e. member functions that only access the public API just reduces encapsulation.

foundrytom: Totally makes sense, just wondering for this specific case, does the complexity of friending undo that - and do we want to have diverging approaches across the code base?

feltech: The downside as I see it is discoverability/readability.  I think for this super-simple class with no real moving parts, a member function is probably slightly better for that reason.  But I think the encapsulation argument is definitely worth keeping in mind in future designs.

foundrytom: @lpapp-foundry Sorry to be a total pain, would you mind making this a member function so it is consistent with the other classes we have so far? Thanks!

lpapp: Done, thanks.

foundrytom:
Just a note that #880 added some more tests 🙃

src/openassetio-python/cmodule/src/BatchElementErrorBinding.cpp
lpapp-foundry: C++ done, thanks: https://github.com/OpenAssetIO/OpenAssetIO/pull/881/commits/d3692d300f310c58624c523552c86c8b3d6a09f3

feltech: We still don't have explicit `pytest` tests for the equality operator. However I'm not so worried, since it's tested implicitly by the fact we use it in other tests.

feltech:
Note that for [git autosquash](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash) to work, the fixup commits should be prefixed with `fixup!`, not `!fixup`.

src/openassetio-python/cmodule/src/BatchElementErrorBinding.cpp
foundrytom: As we have to explicitly implement it on the python side, can we make sure we have python coverage too? Its quite a core object and we have coverage for other classes.

src/openassetio-core/tests/BatchElementErrorTest.cpp
```diff
@@ -26,5 +26,19 @@ SCENARIO("BatchElementError usage") {
...
     }
+
+    WHEN("Two BatchElementErrors are constructed wrapping the same code and message") {
+      const BatchElementError error{code, message};
+      const BatchElementError error2{code, message};
+
+      THEN("The errors instances match") { CHECK(error == error2); }
+    }
+
+    WHEN("Two BatchElementErrors are constructed wrapping different code and message") {
```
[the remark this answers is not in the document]
lpapp-foundry: Added in https://github.com/OpenAssetIO/OpenAssetIO/pull/881/commits/d44c697601bb1cfb6fa5304435046bdae43b2f54 - thanks!

src/openassetio-python/cmodule/src/BatchElementErrorBinding.cpp
lpapp-foundry: Added in https://github.com/OpenAssetIO/OpenAssetIO/pull/881/commits/718a813119f6d70d3fb06e6ae87206f71d95de52 - thanks!

foundrytom: Awesome, thanks!

foundrytom (approved):

PR #883, opened 2023-04-06
Update `pybind` version

lpapp-foundry:
## Description

Closes # 683

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -44,6 +44,9 @@ v1.0.0-alpha.X
   provide more context when using it as a starting point for an
   OpenAssetIO integration.
 
+- Update the pybind dependency version to 2.10.1.
+  [#863](https://github.com/OpenAssetIO/OpenAssetIO/issues/863)
+
 - Made `BatchElementError` a copyable type in the C++ API.
   [#849](https://github.com/OpenAssetIO/OpenAssetIO/issues/849)
 
--- a/resources/build/conanfile.py
+++ b/resources/build/conanfile.py
@@ -39,9 +39,7 @@ class OpenAssetIOConan(ConanFile):
             self.requires("cpython/3.9.7")
             # Pin recipe as latest not compatible with Conan 1.59
             self.requires("ncurses/6.2@#54f100a8e4a94700d4123ed31420506c")
-        # Same as ASWF CY2022 Docker image:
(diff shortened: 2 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -41,6 +41,9 @@ v1.0.0-alpha.X
   provide more context when using it as a starting point for an
   OpenAssetIO integration.
 
+- Support Python 3.11 for building the Python API.
+  [#863](https://github.com/OpenAssetIO/OpenAssetIO/issues/863)
+
```
foundrytom, line 46: [fix] Technically, this change doesn't fully support 3.11. It will allow the project to build, but until we update the tests as per #683, and have 3.11 in the CI matrix, then we can't fully claiming support.  Can we reword this, and the commit title to be just about the `pybind` version update?

foundrytom (changes requested):

resources/build/conanfile.py
```diff
@@ -41,7 +41,7 @@ def requirements(self):
             self.requires("ncurses/6.2@#54f100a8e4a94700d4123ed31420506c")
         # Same as ASWF CY2022 Docker image:
         # https://github.com/AcademySoftwareFoundation/aswf-docker/blob/master/ci-base/README.md
```
foundrytom, line 43: [fix] This comment is also incorrect with the version update so needs removing.

lpapp: Done, thanks.

RELEASE_NOTES.md
lpapp: Done, thanks.

foundrytom: Thanks, can we update `RELEASE_NOTES` too please to mention the Pybind version change in the  supplied `conanfile` instead of that we support Python 11 - as we can't really say that till we fix the tests. Our build instructions cover how to test your build.

foundrytom (changes requested):

RELEASE_NOTES.md
foundrytom: Awesome, thanks! Don't worry, its a somewhat fiddly process until you get used to it :)

foundrytom (approved):
Thanks for all the tweaks here @lpapp-foundry. Please feel free to rebase, squash and force-push these ready for merge.

Issue #848, opened 2023-03-23
Host Convenience implementation.


elliotcmorris:
## Convenience Signatures

We want to add additional call-methodologies for methods that currently require use of the callback
based batch api. The motivation for doing this is convenience and ease-of-rampup.

```[tasklist]
### Tasks
- [ ] #849
- [ ] #850
- [ ] #851
- [ ] #852
- [ ] #853
- [ ] #854
```

We have chosen to exploit language specific features to have a more natural interface, at the
expense of perfect cross-language consistency.

To that end, using `resolve` as the canonical example, this is the pattern we shall choose :

### Cpp
```cpp
// Singular Callback
virtual void resolve(const EntityReference& entityReferences, const trait::TraitSet& traitSet,
                     const ContextConstPtr& context,
                     const ResolveSuccessCallback& successCallback,
                     const BatchElementErrorCallback& errorCallback);

// Singular, throws BatchElementErrors
virtual TraitsDataPtr resolve(const EntityReference& entityReference,
                              const trait::TraitSet& traitSet,
                              const ContextConstPtr& context,
                              const errors::BatchElementErrorPolicyTag::Except = {});

// Singular, returns BatchElementError object
virtual std::variant<TraitsDataPtr, BatchElementError> resolve(
  const EntityReference& entityReference,
  const trait::TraitSet& traitSet,
  const ContextConstPtr& context,
  const errors::BatchElementErrorPolicyTag::Return);

// Multi, Callback
virtual void resolve(const std::vector<EntityReference>& entityReferences, const trait::TraitSet& traitSet,
                     const ContextConstPtr& context,
                     const ResolveSuccessCallback& successCallback,
                     const BatchElementErrorCallback& errorCallback);

// Multi, throws BatchElementErrors
virtual std::vector<TraitsDataPtr> resolve(const std::vector<EntityReference>& entityReferences,
                                           const trait::TraitSet& traitSet,
                                           const ContextConstPtr& context,
                                           const errors::BatchElementErrorPolicyTag::Except = {});

// Multi, returns BatchElementError object
virtual std::vector<std::variant<TraitsDataPtr, BatchElementError>> resolve(
  const std::vector<EntityReference>& entityReference,
  const trait::TraitSet& traitSet,
  const ContextConstPtr& context,
  const errors::BatchElementErrorPolicyTag::Return);

```

Note here how we've chosen to use `std::vector<EntityReference>` rather than `EntityReferences`
[...]

PR #884, opened 2023-04-06
[Tests] Reusable C++ fixture for Manager tests

feltech:
Part of #848. In preparation for adding many more C++-specific tests of `Manager` member functions, move common `Manager` instance creation boilerplate into a fixture struct.


foundrytom (approved):
Thanks for splitting that out :)

PR #900, opened 2023-04-18
Deep-copy locale in `createChildContext`

### LINKED ISSUE #896: `Manager.createChildContext` should deep-copy the locale
## What

`createChildContext` should copy the locale vs sharing.

## Why

This method is used to create 'related but detached' contexts, sharing a locale couples them

## Acceptance Criteria

- The new child context does not share any ptr-referenced properties with the parent context (aside from the managerState).
- The new context has a locale equal to the parent.

## Out of scope

- New host documentation

foundrytom:
Closes #896

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,16 @@ Release Notes
 v1.0.0-alpha.x
 --------------
 
+### Breaking changes
+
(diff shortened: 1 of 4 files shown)


feltech (approved):
LGTM

Issue #16, opened 2021-09-22
RFC 08: `setRelatedReferences` should be `add` and `remove`


foundrytom:
# What

Replace [`setRelatedReferences`](https://github.com/TheFoundryVisionmongers/OpenAssetIO/blob/14dc134703fd9b3431c933bdb992e85bd67c8dce/python/openassetio/managerAPI/ManagerInterface.py#L974) with `addRelatedReferences` and `removeRelatedReferences`.

**NOTE: This is an old ticket, We should investigate whether add/remove is needed at all.**

# Why

It was highlighted that hard `set` updates in mult-tennanted APIs, especially with lists, are prime candidates for race condition issues.

elliotcmorris:
Closing as out of date.

Add/Remove is implicit as part of the access mode used for publishing.
Closing this ticket dosen't mean we're outright rejecting the concept, just that we're not planning to reconsider any time soon.

PR #901, opened 2023-04-18
Remove `setRelatedReferences`

foundrytom:
Part of #16.

`set` was not well considered, as it has the potential for race conditions in real-world scenarios. A re-design is in the works, so remove for now to avoid dead code in the release.

### DIFF
--- a/doc/doxygen/src/ForManagers.dox
+++ b/doc/doxygen/src/ForManagers.dox
@@ -193,14 +193,6 @@
  * places any constraints on the implementation of any given asset
  * management system.
  *
- * There are however times when a host will explicitly add a
- * relationship to an entity. Generally speaking, this is when an
- * existing entity is known to be relevant to a newly created
- * one. For example, if an image has been rendered from an assetized
- * script. The host may call @ref openassetio.managerApi.ManagerInterface.ManagerInterface.setRelatedReferences
- * "setRelatedReferences", to add the script's reference to the newly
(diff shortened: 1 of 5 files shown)


doc/doxygen/src/ForManagers.dox
```diff
@@ -211,18 +203,6 @@
...
- *   @ref openassetio.managerApi.ManagerInterface.ManagerInterface.setRelatedReferences
- *   "setRelatedReferences" to update any internal relationships that
- *   may be affected by the registered change.
- *
- * @note It is currently in flux as to whether relationship creation
- * should be a hard requirement of the API, or considered hints as set
- * by the host. There is even question as to the validity of the API
- * allowing sets at all, please contribute to the discussion here if you
- * have any opinion on the matter:
- * https://github.com/OpenAssetIO/OpenAssetIO/discussions/18
```
feltech, line 224: [consider][minor] Perhaps worth retaining this link for now?

elliotcmorris (approved):
LGTM. Annoying release notes conflict though since we did a release :(

feltech (approved):
LGTM

doc/doxygen/src/ForManagers.dox
foundrytom: I was pondering this - and thought it might be worth leaving out for now as we definitely have use cases that require a specific method (eg. connecting a render to the existing, un-changed, assetized script that produced it).

PR #905, opened 2023-04-20
[CI] Add libxcb-cursor-dev to ubuntu bootstrap.

elliotcmorris:
This got added to the upstream conan recipe for xorg, so we need it installed.


feltech (approved):
LGTM

PR #907, opened 2023-04-21
[Python] Always require Development

lpapp-foundry:
Ideally, we would only require Development.Embed when it is really needed. This is when running the OpenAssetIO tests.

However, FindPython.cmake may have a bug when using this module with this configuration. It would find Python 2.7 for some reason, which does not have some symbols that we use.

    error: use of undeclared identifier '_Py_IsFinalizing'

Please refer to:

https://gitlab.kitware.com/cmake/cmake/-/issues/24834

So, the workaround is to require the whole of Development for now until we have a working way to customise and modularise it further down.

The downside of the fix is only that we require a bit more than necessary, but this probably does not have such a big impact as the current code, which does not work in certain scenarios.

I have also tried to make sure that it picks up Python 3 by being explicit about it, but in the end, it seems to have suffered from the same issue.

## Description

Closes # (issue)

### DIFF
--- a/cmake/ThirdParty.cmake
+++ b/cmake/ThirdParty.cmake
@@ -10,14 +10,8 @@ if (OPENASSETIO_ENABLE_PYTHON)
     #-------------------------------------------------------------------
     # Locate packages
 
-    list(APPEND _components Interpreter)
-    list(APPEND _components Development.Module)
-    if (OPENASSETIO_ENABLE_TESTS)
-        list(APPEND _components Development.Embed)
-    endif ()
-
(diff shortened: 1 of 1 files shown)


feltech:
As can be seen from CI, there are Python environments that don't support `Development.Embed` (in CI's case it's the Python environment in the `pypa/manylinux2014_x86_64` Docker image), and indeed `Development.Embed` isn't required for building OpenAssetIO (other than for some of our tests).

foundrytom:
Closing as this change would break our current wheel builds (see check logs)

PR #908, opened 2023-04-24
[DR] Result paging strategy for related entities query

### LINKED ISSUE #188: [DR] Result paging for `getRelatedReferences` and `getEntityVersions`
## What/Why

These methods may return significant quantities of results, we need a packing mechanism that is simple and supports the breadth of languages and runtimes OpenAssetIO targets.

Determine:

- Do we need to include some result-window mechanism in the v1 methods, or is it best added later as an _additional_ implementation method?
- Do we opt for simple controls or extended result types? (may be related)
- What guarantees around stability do we care about? 

## Background

`getRelatedReferences`, as an example, takes the following (in our usual batch formulation):

- An `EntityReference`
- A relationship defined in a `TraitsData`

And produces a list of `EntityReferences`.

The success callback looks like this:

```python
def success(batchElementIndex: Int, relatedReferences: List<EntitiyReference>):
```

The resulting `relatedRefefernces` may be millions of elements long. It has been noted that we need for the Host to define a window on interest into these results, specifically to avoid:

- Excess work being done by the manager.
- Onerous data transfer/processing requirements for the host/manager.

It can be assumed that the ordering of items in the result list is considered stable.

Note also that we already have the `managerInterfaceState` mechanism in the supplied context that can be used to facilitate stability between queries.

## Out of scope

- Ordering predicate mechanism
- The manager influencing the result limit aside from lack of data.
- The Context as window control

## Requirements

- The solution should be amenable to implementations that use remote proxy strategies such as gRPC, OpenAPI etc... and not introduce significant additional back-end calls (and the associated latency) to access a single result window.
- The solution should be readily compatible with all of our target languages.
- The solution should not add undue implementation complexity in any integration. SIMPLES.
- The solution should avoid requiring the host to conditionally invoke different code paths based on supported capabilities of the manager.
- The solution should do as much as possible to make API calls atomic/stateless operations.`
- The solution ideally provides stability across window moves but is not required if complexity if prohibitive.
- Prior art is used wherever possible.

feltech:
## Description

Closes #188. Add decision record considering options for result paging, in particular for the related entities query.

[Rendered version](https://github.com/OpenAssetIO/OpenAssetIO/blob/2c14e29fde34e464ca190ace49b3917af9901834/doc/decisions/DR016-One-to-many-result-paging.md) for 👀

### DIFF
--- a/doc/decisions/DR016-One-to-many-result-paging.md
+++ b/doc/decisions/DR016-One-to-many-result-paging.md
@@ -1,0 +1,691 @@
+# DR016 One-to-many result paging
+
+- **Status:** Decided
+- **Impact:** Medium
+- **Driver:** @feltech
+- **Approver:** @elliotcmorris @foundrytom
+- **Contributors:** @feltech @elliotcmorris @foundrytom @mattdaw
+- **Informed:** @mattdaw
+- **Outcome:** Split batching from paging and page using a cursor object
+
+## Background
+
+The OpenAssetIO API contains functions that can return an arbitrarily
+long list of results.  In particular, the current core API has functions
+for retrieving a list of versions of a given entity, and retrieving a
+list of entities that are "related to" another entity in some way.
+
+Taking the related entities query as an example: this could be sibling
+or parent-child relationships (e.g. shot and sequence), or more loose
+relationships (e.g. the project file that was executed to generate a
+render).
+
+The current API design provides the host with a list of entity
+references that correspond to an input entity-relationship pair.
+
+There may be multiple entity-relationship pairs provided as input to
+a single API call, allowing the manager plugin to leverage batching
+optimizations. Note that, typically, batching primarily addresses
+network latency concerns, in addition to database query optimizations.
+
+In the most general case, the host cannot know the size of the list of
+related entity references that will be returned to it. This clearly
+poses a scalability issue.
+
+Database and web APIs have had to deal with this problem for decades,
+and invariably provide some mechanism of returning slices of result sets
+or "pages" that can be iterated through, lazily fetching chunks of data
+in a controlled way.
+
+In an ideal situation, requesting subsequent pages has performance
+benefits over repeating the same query twice (or more) with different
+range arguments, and yet work is performed lazily as pages are requested
+so that no unnecessary processing of unused pages is performed.  Also,
+pages should be stable, in that results are not skipped nor appear more
+than once on successive pages.
+
+Ultimately, the implementation must be provided by the manager plugin
+authors. The mechanism chosen in this DR can only present an abstract
+interface that the plugin must satisfy. As a result, characteristics
+such as page stability and performance cannot be guaranteed, and nor
+do we desire to place such restrictions upon the implementation. However
+we can at least ensure that the chosen mechanism does not preclude such
+optimisations.
+
+The mechanism of paging tends to take three broad forms
+
+- (Integer) offsetting
+- (Opaque) page tokens
+- Cursor objects / iterators
+
+The addition of a paging mechanism to the batched query API raises a
+number of questions around host and plugin implementation complexity.
+
+## Relevant data
+
+A (certainly non-exhaustive) selection of API design guides and
+implementations for the paging strategies outlined above can be found in
+the following links.
+
+* A broad discussion of paging strategies:
+  - [Patterns for API Design](https://microservice-api-patterns.org/patterns/quality/dataTransferParsimony/Pagination)
+* Offset based designs:
+  - [Atlassian](https://developer.atlassian.com/server/framework/atlassian-sdk/atlassian-rest-api-design-guidelines-version-1/#standard-query-parameters-in-uris)
+  - [Australian Government](https://apiguide.readthedocs.io/en/latest/build_and_publish/batching_results.html)
+  - [Microsoft](https://github.com/Microsoft/api-guidelines/blob/master/Guidelines.md#982-client-driven-paging)
+* Page token based designs:
+  - [Google Cloud guidelines](https://cloud.google.com/apis/design/design_patterns#list_pagination)
+  - [Google Cloud Datastore](https://cloud.google.com/datastore/docs/concepts/queries#datastore-datastore-cursor-paging-python)
+  - [Twitter](https://developer.twitter.com/en/docs/twitter-api/pagination)
+  - [Slack](https://slack.engineering/evolving-api-pagination-at-slack/)
+* Cursor object designs:
+  - [Cisco](https://github.com/CiscoDevNet/api-design-guide#363-get)
+  - [Cloudfoundry](https://github.com/cloudfoundry/cc-api-v3-style-guide#pagination)
+  - [JDBC](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html)
+  - [MongoDB](https://www.mongodb.com/docs/manual/reference/method/js-cursor/)
+  - [Python PEP 239 DB-API](https://peps.python.org/pep-0249/#fetchmany)
+    used by many database libraries.
+  - [Ruby on Rails](https://api.rubyonrails.org/v7.0.4.2/classes/ActiveRecord/Batches.html#method-i-find_in_batches)
+  - [W3C IndexedDB Cursor API spec](https://w3c.github.io/IndexedDB/#idbcursor)
+
+
+## Options considered
+
+### Paging option 1 - Offseting
+
+Host provides an integer offset argument to locate the start of the next
+page, and a limit argument providing the page size.
+
+#### Pros
+
+- Conceptually simple for the host.
+- Jumping arbitrarily to nth page is trivial.
+- Transactions, indexing and caching technologies exist that can
+  alleviate stability and performance problems, to some extent.
+
+#### Cons
+
+- Anomalous page results if insertions/deletions can happen before next
+  page is retrieved, or if ordering is not stable.
+- Server cannot negotiate page size - there is no mechanism to signal
+  that more results are available other than by exhausting the page
+  limit.
+- Poor performance on the back-end for large offsets into non-trivial
+  (i.e. non-indexed/cached) queries, where many rows must be iterated
+  through and skipped over.
+- No mechanism to inform the manager that no further pages are required.
+
+### Paging option 2 - Page token
+
+Page tokens (also often called "cursors") allow the server to retain
+control over where to find the next page (typically they are a pointer
+to an element in the overall result list), alleviating stability
+problems.
+
+Each query returns a page token as well as a page of results. The
+original query is repeated to retrieve subsequent pages, but with an
+updated page token provided by the previous query.
+
+#### Pros
+
+- Conceptually simple.
+- Matches the most common pattern of pageable HTTP APIs allowing e.g.
+  the page token to be a thin wrapper around a "real" page token(s).
+- Manager plugin can remain conceptually stateless, since required state
+  is bundled up in the page token.
+- Page token is (or can be) immutable, aiding thread-safety.
+- Using the page token to determine if more pages are available (e.g.
+  a `None` token means exhausted) allows the server to override the
+  page size, opening up an avenue for optimisation.
+- Destruction hooks on a page token (e.g. destructor of a class-based
+  token) could signal to the manager that the page is no longer
+  required, allowing e.g. cache cleanup.
+
+#### Cons
+
+- Adding more functionality (e.g. previous page) requires additional
+  sibling functions or arguments.
+- Adding metadata (e.g. total results) requires additional sibling
+  functions or arguments.
+- Function invocations to fetch subsequent pages must pass all the same
+  arguments, such that they are valid for the page token. This is
+  error-prone and/or redundant.
+
+### Paging option 3 - Cursor objects / iterators
+
+Cursor objects are an abstraction over some unknown internal paging
+mechanism, which may be a page-token, or some other mechanism.
+This cursor could be an object-oriented wrapper with methods, e.g.
+`.next()`, or a data structure with opaque links, e.g. containing a URL
+to the "next" page.
+
+For example, a simple paging cursor object may look like:
+```python
+class AbstractCursor:
+
+    def getPage(self, pageSize: int) -> List[EntityReference]:
+        # Get next page of refs.
+        ...
+
+    def skip(self, pageSize: int):
+        # Skip over records without fetching.
+        ...
+
+    def remaining(self) -> Union[int, None]:
+        # None means "unsupported".
+        ...
+```
+
+#### Pros
+
+- Retrieving the data can be decoupled from traversal of pages, allowing
+  for lazy network retrieval of desired pages.
+- Additional metadata can be included in the cursor (e.g. remaining
+  pages) in an extensible way.
+- A mutable, stateful cursor object maximizes opportunities for the
+  manager plugin to optimize query caches.
+- Destruction of a cursor object could be used as a signal to the
+  manager that no further pages are required, allowing e.g. cache
+  cleanup.
+
+#### Cons
+
+- Manager plugin must implement yet another interface.
+- Complexity of binding to multiple target languages.
+- A mutable, stateful cursor violates the stateless design principle of
+  interactions with manager plugins. This causes complications, e.g.
+  thread-safety.
+- An immutable cursor object (such that `.next()` returns a new cursor
+  object), could increase optimisation complexity for the manager
+  implementation, e.g. managing query cache sharing, lifetime and
+  updates across multiple "live" cursors.
+- The temptation to directly wrap a "real" DB cursor object in our
+  (potentially long-lived) paging cursor object could lead to connection
+  pool starvation.
+
+### Paging option outcome
+
+The pros and cons considered are insufficient to decide on a final
+design.  However, we can at least rule out any offset-based solution,
+given the lack of flexibility afforded to the manager's implementation,
+resulting in poor stability and likely manager overhead of record
+counting.
+
+More factors will be considered in the following sections, taking into
+account only page token and cursor object based solutions.
+
+## Considering batching
+
+Additional considerations arise when considering how batching can be
+handled algongside paging. There are three broad strategies
+
+- Split batching from paging, using a different API method for each.
+- Page through the results from all the batched input elements in
+  lockstep, i.e. whole-batch paging.
+- Page through the results of each batched input element independently,
+  using individual sub-queries within the whole-batch query, i.e.
+  per-element paging.
+
+Each of these batched paging designs combines more or less well with the
+various paging options presented above.
+
+The above strategies are thus considered in the following sections with
+respect to page token and cursor object designs.
+
+### Batched paging option 1 - Split batching from paging
+
+E.g. Non-paged batch signature
+```python
+# Callback taking index of batch element and all results for that element.
+SuccessCallback = Callable[[int, List[EntityReference]], None]
+
+def relatedEntities(
+    batch: List[EntityReference], relationship,
+    successCb: SuccessCallback, errorCb): ...
+```
+
+E.g. Page token paged signature
+
+```python
+def relatedEntities(
+    input: EntityReference, relationship, limit=-1, pageToken=None
+) -> Tuple[AbstractPageToken, List[EntityReference]]: ...
+```
+
+E.g. Cursor object paged signature
+```python
+class AbstractCursor:
+    def getPage(self, pageSize: int) -> List[EntityReference]:
+        ...
+    def skip(self, pageSize: int):
+        ...
+
+def relatedEntities(
+    input: EntityReference, relationship) -> AbstractCursor: ...
+```
+
+#### Pros
+
+- Simple direct interface for the host application.
+- Simpler for manager plugin implementors to reason about (as evidenced
+  by ad-hoc discussions with collaborators).
+- Hosts should usually have a good idea whether batching or paging is
+  more appropriate for the query they are making.
+- Manager and host do not need to consider both paging and batching at
+  the same time, which is a complex convergence of concepts.
+- Manager implementation can typically be a lightweight wrapper around
+  common HTTP/DB API pagination patterns.
+
+#### Cons
+
+- Precludes batching optimisations by the manager when paging is
+  required.
+- Requires implementation of two methods by the manager: one for
+  batched, one for paged.
+- Potential difficulty for some hosts/queries to determine a priori
+  which API (paging or batch) is the most optimal choice.
+- Having the choice adds a foot-gun for host developers to shoot
+  themselves with by accident.
+- Precludes hosts atomically expressing more complex queries that may
+  be required by spreadsheet-type UIs.
+
+### Batched paging option 2 - Whole-batch paging
+
+E.g. Page token
+
+```python
+# Callback taking index of batch element and selected page of results.
+SuccessCallback = Callable[[int, List[EntityReference]], None]
+
+def relatedEntities(
+    batch: List[EntityReference], relationship,
+    successCb: SuccessCallback, errorCb,
+    limit=-1, pageToken=None
+) -> AbstractPageToken: ...
+```
+
+E.g. Cursor object
+
+```python
+class AbstractCursor:
+    def getPage(self, pageSize: int) -> List[List[EntityReference]]:
+        ...
+    def skip(self, pageSize: int):
+        ...
+
+def relatedEntities(
+    batch: List[EntityReference], relationship) -> AbstractCursor: ...
+```
+
+#### Pros
+
+- Whole batch is paged at once, retaining batch-awareness in the
+  manager for subsequent pages.
+- Conceptually simpler API for both host and manager if only a single
+  page is active across the whole batch.
+
+#### Pros (page token specific)
+
+- Per-element callback means it's trivial to skip processing elements of
+  the input batch whose pages are exhausted.
+- Retains existing design in line with OpenAssetIO design principles,
+  i.e. non-paged signature is simply the same function but ignoring
+  optional arguments.
+
+#### Cons
+
+- Batch elements cannot enter an error state in isolation, since paging
+  is whole-batch. E.g. cursor objects may raise an exception, but that
+  would preclude getting pages for non-failing elements; and page tokens
+  may fail independently, but then _some_ token must still be passed
+  into the next call, and it is not clear how that should work.
+- Collaborators working on manager plugins raised this as being
+  particularly tricky to implement.
+- Likely that the manager plugin must combine multiple back-end page
+  tokens into one, adding complexity to ensure they are maintained in
+  sync.
+
+#### Cons (cursor object specific)
+
+- Additional metadata in the object is limited to whole-batch data.
+- Wasted processing of empty lists by the host when pages are exhausted
+  i.e. when there are still pages remaining for some inputs but not
+  others.
+
+### Batched paging option 3 - Per-element paging
+
+E.g. Page token
+
+```python
+# Callback taking index of batch element and selected page of results.
+SuccessCallback = Callable[[int, List[EntityReference]], None]
+
+def relatedEntities(
+    batch: List[EntityReference],
+    relationship,
+    successCb: SuccessCallback,
+    errorCb,
+    limit: -1,
+    pageTokens: List[AbstractPageToken]
+): -> List[AbstractPageToken]: ...
+```
+
+E.g. Cursor object
+
+```python
+class AbstractCursor:
+    def getPage(self, pageSize: int) -> List[EntityReference]:
+        ...
+    def skip(self, pageSize: int):
+        ...
+
+# Callback taking index of batch element and cursor object.
+SuccessCallback = Callable[[int, AbstractCursor], None]
+
+def relatedEntities(
+    batch: List[EntityReference],
+    relationship,
+    successCb: SuccessCallback,
+    errorCb): ...
+```
+
+#### Pros
+
+- Manager is under no obligation to consider errors in other batch
+  elements when implementing query behaviour, as pagination method need
+  not be synchronized.
+
+#### Pros (cursor object specific)
+
+- Individual cursors can be incremented independently, potentially
+  reducing wasted queries.
+
+#### Cons
+
+- Batching is very difficult to optimize for the manager, especially
+  after the first page.
+- Batch-optimized queries have unpredictable performance profile for
+  the host, given the wide range of potential implementations.
+
+#### Cons (cursor object specific)
+
+- Cannot advance to the next page as a batch operation.
+
+### Batched paging option outcome
+
+Given the reported difficulty of crafting queries that are both batch
+and paging-aware, we opt to go with Option 1 and split the paging API(s)
+from the batched API(s).
+
+This then leaves only a batch-only signature, plus a decision between
+two remaining paging solutions: (non-batched) page token or cursor
+objects.
+
+## Outcome
+
+We opt for a non-batched cursor object design.
+
+Avoiding conflating batching and pagination simplifies manager plugin
+implementations.
+
+Whilst we may miss out on some theoretical performance, we have been
+unable to find any real-world examples showing this sort of batched-page
+optimization.
+
+A cursor object maximises future extensibility, where new functionality
+can be added without further polluting the `Manager`/`ManagerInterface`
+class. They also naturally capture arbitrary state that the manager
+may wish to store as part of the ongoing query.
+
+A cursor object also encapsulates the arguments required for requesting
+subsequent pages, whereas a page token requires (potentially
+redundantly) re-specifying the same arguments for each subsequent page
+along with the page token. This could be mitigated for the page token
+design by having a separate API for page token operations, but this
+starts to closely resemble a cursor object solution (with the `this`
+pointer fed from outside).
+
+Cursor objects also reflect the most common programmatic approach to
+paging through the results of a database query, presenting a familiar
+interface for plugin authors, and mapping well to common back-end SDKs.
+
+The additional complexity of the interface compared to a page token
+solution, for plugin authors, host authors and future language bindings,
+is small compared to the advantages in extensibility. For example, we do
+not foresee any issues in following the existing pattern of the
+(prototype) C API.
+
+Enforcing excessively stateful or complex functionality on the cursor
+object, e.g. by mandating `skip` or `remaining` methods, as used in the
+sketched examples, is a cause for concern. We do not propose to mandate
+such methods. Methods may be added, but most if not all will be optional
+such that the host will have to test the capabilities of the cursor
+object before using them.
+
+## Appendix: Considering manager implementations
+
+The following subsections sketch some possible implementations of
+manager-side queries for the remaining options under consideration. They
+are not to be considered canonical.
+
+### Using Python's DB-API standard
+
+Python's [PEP 239 DB-API](https://peps.python.org/pep-0249/#fetchmany)
+standard provides a canonical example of an API for interacting with a
+back-end SQL-like database.
+
+#### Page token signature
+
+The page token design, as envisioned in earlier sketches, is further
+elaborated below, as a possible manager implementation using DB-API.
+
+```python
+class PageToken(AbstractPageToken):
+    def __init__(self, cursor):
+        self.cursor = cursor
+
+
+def relatedEntities(
+    input: EntityReference, relationship, limit=-1, pageToken=None
+) -> Tuple[AbstractPageToken, List[EntityReference]]:
+
+    if pageToken is None:
+        my_relationship_type: str = get_my_relationship_type(relationship)
+
+        cursor = get_my_connection().cursor()
+        cursor.execute(
+            f"""
+            SELECT ToRef FROM Relationship
+            WHERE RelationshipType = ?
+            AND FromRef = ?
+            """, my_relationship_type, str(input))
+
+        pageToken = PageToken(cursor)
+
+    if limit <= 0:
+        result_set = pageToken.cursor.fetchall()
+    else:
+        result_set = pageToken.cursor.fetchmany(limit)
+
+    related_refs = [EntityReference(result[0]) for result in result_set]
+    return pageToken, related_refs
+```
+
+#### Page token alternative - command pattern
+
+One of the major Cons of page tokens as sketched so far is that the
+original query arguments must be redundantly re-specified when fetching
+subsequent pages. To overcome this we can split the construction of page
+tokens from the operations performed on them.
+
+This could be done as a series of methods, however that then looks
+extremely similar to a cursor object design.
+
+Instead, the sketch below takes a command pattern approach. We add a
+single new method, with the operation to perform given as an additional
+argument.
+
+```python
+def relatedEntities(
+        input: EntityReference, relationship) -> AbstractPageToken:
+
+    my_relationship_type: str = get_my_relationship_type(relationship)
+
+    cursor = get_my_connection().cursor()
+    cursor.execute(
+        f"""
+        SELECT ToRef FROM Relationship
+        WHERE RelationshipType = ?
+        AND FromRef = ?
+        """, my_relationship_type, str(input))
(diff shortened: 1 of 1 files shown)


doc/decisions/DR016-Related-entities-result-paging.md
```diff
@@ -0,0 +1,333 @@
...
+  manager for subsequent pages.
+- Per-element callback means it's trivial to skip processing elements of
+  the input batch whose pages are exhausted.
+- Matches the most common pattern of pageable HTTP APIs allowing e.g.
+  the page token to be a thin wrapper around a "real" page token(s).
+- Retains existing design in line with OpenAssetIO design principles,
+  i.e. non-paged signature is simply the same function but ignoring
+  optional arguments.
+- Manager plugin can remain conceptually stateless, since required state
+  is bundled up in the page token.
```
elliotcmorris, line 222: [Question] Is this not true of the cursor based solutions as well? (More on this below at the cursor part)

doc/decisions/DR016-Related-entities-result-paging.md
```diff
@@ -0,0 +1,333 @@
...
+        relationship: TraitsData,
+        successCb: SuccessCallback,
+        errorCb,
+        limit: int = -1): ...
+```
+
+#### Pros
+
+- Individual cursors can be incremented independently, potentially
+  reducing wasted queries.
```
elliotcmorris, line 290: [consider]
```suggestion
#### Pros

- Individual cursors can be incremented independently, potentially
  reducing wasted queries.
- Manager is under no obligation to consider errors in other streams 
  when implementing query behaviour, as pagination method need not be synchronized
```

doc/decisions/DR016-Related-entities-result-paging.md
```diff
@@ -0,0 +1,333 @@
...
+- Retrieving the data can be decoupled from traversal of pages, allowing
+  for lazy network retrieval of desired pages.
+- Additional metadata can be included in the cursor (e.g. remaining
+  pages) in an extensible way.
+
+#### Cons
+
+- Manager plugin must implement yet another interface.
+- Cursor objects are stateful by design, breaking the stateless
+  design principle of interactions with manager plugins. This causes
```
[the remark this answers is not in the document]
feltech: Ah good point, when I wrote this I assumed a mutable cursor object, which does seem to mirror DB cursor objects - probably need to do some more research there.  

A mutable cursor object vs. an immutable page token is what I had in mind here, where sharing a mutable cursor across threads introduces thread-safety issues - the very issues `ManagerInterface` claims to avoid by being stateless.

But an immutable cursor object does seem to solve this problem :thinking:

doc/decisions/DR016-Related-entities-result-paging.md
```diff
@@ -0,0 +1,333 @@
...
+
+In order to avoid considering a long list of options that are either
+clearly inferior, or broadly similar to other options, some valid
+solutions were rejected early.
+
+#### Whole-batch offset
+
+The classic offset-based solution was rejected due to being clearly
+inferior to page token (cursor) based pagination, which has a similar
+signature but fewer critical drawbacks.
```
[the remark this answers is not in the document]
feltech: Changed the wording in b85d9a44, how's that?

doc/decisions/DR016-Related-entities-result-paging.md
```diff
@@ -0,0 +1,333 @@
...
+        relationship: TraitsData,
+        successCb: SuccessCallback,
+        errorCb,
+        limit: int = -1,
+        pageToken: AbstractPageToken = None) -> AbstractPageToken: ...
+```
+
+#### Pros
+
+- Conceptually simple.
```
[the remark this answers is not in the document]
feltech: Added "for the host" in 9a0217db

doc/decisions/DR016-Related-entities-result-paging.md
```diff
@@ -0,0 +1,333 @@
...
+- Matches the most common pattern of pageable HTTP APIs allowing e.g.
+  the page token to be a thin wrapper around a "real" page token(s).
+- Retains existing design in line with OpenAssetIO design principles,
+  i.e. non-paged signature is simply the same function but ignoring
+  optional arguments.
+- Manager plugin can remain conceptually stateless, since required state
+  is bundled up in the page token.
+- Page token is (or can be) immutable, aiding thread-safety.
+
+#### Cons
```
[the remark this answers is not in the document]
feltech: Added a couple more cons in 4856a84a, does that cover it?

doc/decisions/DR016-Related-entities-result-paging.md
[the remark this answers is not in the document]
elliotcmorris: :+1:

doc/decisions/DR016-Related-entities-result-paging.md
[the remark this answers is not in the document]
elliotcmorris: :+1:

doc/decisions/DR016-Related-entities-result-paging.md
elliotcmorris: (Resolving in favour of the other thread)

doc/decisions/DR016-Related-entities-result-paging.md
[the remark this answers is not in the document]
feltech: I've split the cons to hopefully cover both immutable and mutable cursor types: 73795d74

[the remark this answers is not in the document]
feltech: Further updated the relevant pros and cons in cf837603

elliotcmorris (approved):

mattdaw:
Hi @feltech... if I'm understanding correctly, a manager can internally figure out how it wants to break up network requests such that its back-end will not be overwhelmed. Then it calls the success callback when it's sure it has acquired all the data for a particular EntityReference. Is that right? If so, then is pagination more for a host/UI to communicate the limits of what it wants to handle?

feltech:
> Hi @feltech... if I'm understanding correctly, a manager can internally figure out how it wants to break up network requests such that its back-end will not be overwhelmed. Then it calls the success callback when it's sure it has acquired all the data for a particular EntityReference. Is that right? If so, then is pagination more for a host/UI to communicate the limits of what it wants to handle?

Thanks for the question, I'll try to break it down since I'm not 100% sure I understand:

> a manager can internally figure out how it wants to break up network requests such that its back-end will not be overwhelmed

Yeah, a manager plugin can do what it wants, as long as the host application can retrieve slices/pages of data lazily, through whatever mechanism we decide on, in chunks of the requested size. 

Obviously it's in the manager's interest to optimise around the fact that data is being lazily requested, page by page.  But a naive implementation may well fetch all the data up-front, cache it, then regurgitate page by page as requested (in fact we have discussed this as a possible default behaviour, to be overridden by plugins that support more intelligent paging).

> it calls the success callback when it's sure it has acquired all the data for a particular EntityReference

The manager should call the success callback when it has acquired the _requested page_ of data for a particular `EntityReference` (assuming we're talking about Option 1).

> is pagination more for a host/UI to communicate the limits of what it wants to handle?

Yes indeed, as mentioned it's the host that selfishly wants paged data, regardless of how that's implemented by the manager plugin.

Having said that, host applications will have some concern for back-end performance too - that's the whole point of batching requests, after all.

mattdaw:
Ok perfect thanks, I think we're aligned:

- Code on the host side declares the data that it wants, which includes paginated subsets of the data
- Code on the manager side determines the (hopefully optimal) execution plan for acquiring that data. Or as you suggested, it may well over-fetch data that the host hasn't asked yet.

I'll give some more thought to the options with that mental model in mind...

doc/decisions/DR016-Related-entities-result-paging.md
```diff
@@ -0,0 +1,352 @@
+# DR016 Related entities result paging
```
foundrytom, line 1: [fix] It would be good to call out that this is a more general mechanism for any one-to-many type methods, eg we would also want to apply it to `entityVersions`.

doc/decisions/DR016-Related-entities-result-paging.md
```diff
@@ -0,0 +1,352 @@
...
+
+- Manager plugin must implement yet another interface.
+- A mutable, stateful cursor violates the stateless design principle of
+  interactions with manager plugins. This causes complications, e.g.
+  thread-safety.
+- An immutable cursor object (such that `.next()` returns a new cursor
+  object), could increase optimisation complexity for the manager
+  implementation, e.g. managing query cache sharing, lifetime and
+  updates across multiple "live" cursors.
+
```
foundrytom, line 283: [fix] - Cursor objects increase complexity when binding to multiple target languages.

doc/decisions/DR016-Related-entities-result-paging.md
```diff
@@ -0,0 +1,352 @@
...
+  page is retrieved, or if ordering is not stable.
+- Poor performance on the back-end for large offsets into non-trivial
+  (i.e. non-indexed/cached) queries, where many rows must be iterated
+  through and skipped over.
+
+#### Whole-batch cursor object
+
+Paging through an entire batch via a cursor object was rejected mainly
+due to being broadly similar to a whole-batch page token. A whole-batch
+cursor object offers few additional benefits, whilst presenting a
```
foundrytom, line 141: Im not 100% convinced that this option should be so easily discarded if we are also considering a whole-batch token. The ability of offer proper navigation of pages without new methods is quite significant I think.

foundrytom (changes requested):
Great stuff @feltech @elliotcmorris, thanks.

Reading through, it does seem like there are perhaps two main questions that are almost discreet concerns. 

 - Token vs Object.
 - batch or singular input.

There are certainly some elements in the single/batch vs token/object matrix that probably aren't practical, but I wonder if we're mixing up the pros/cons a little when ruling out some of the options when considering specific combinations from the matrix in isolation.

I think it would be extending the form you took for Option 2 - i.e. looking at the pros/cons of Object vs token in isolation, then separately discussing the pros/cons of keeping the batch semantic?

doc/decisions/DR016-Related-entities-result-paging.md
feltech: We were desperately trying to keep the number of options to a minimum. This was made difficult because we didn't think the two concerns of batch and paging strategy were orthogonal enough to consider in isolation.  

So we deliberately considered only our favoured options that sample across both the batch and paging strategy axes, accepting the possibility that the chosen solution will be some hybrid.

Back on topic... do you think this particular rejected solution is worth promoting to an "Option 4"?

foundrytom: I think perhaps rephrasing as comparing options on the two axis that make up the matrix might be easier to grok?

doc/decisions/DR016-Related-entities-result-paging.md
feltech: Added con in 1003d8be

doc/decisions/DR016-Related-entities-result-paging.md
feltech: Split pros/cons of page tokens in 76f85a78

Also tentatively promoted this option to "Option 2.3" in c98a580e

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,501 @@
...
+        batch: List[EntityReference],
+        relationship,
+        successCb: SuccessCallback,
+        errorCb,
+        limit: int = -1,
+        pageToken: AbstractPageToken = None) -> AbstractPageToken: ...
+```
+
+```python
+# Usage:
```
feltech, line 196: Added usage examples (here and for other options) after offline conversation in e9e343ab

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,501 @@
...
+- Jumping arbitrarily to nth page is trivial.
+- Retains batch-awareness for each page.
+- Transactions, indexing and caching technologies exist that can
+  alleviate stability and performance problems, to some extent.
+
+##### Cons
+
+- Anomalous page results if insertions/deletions can happen before next
+  page is retrieved, or if ordering is not stable.
+- Server cannot negotiate page size - there is no mechanism to signal
```
feltech, line 129: Added page size negotiation Con after offline conversation (also corresponding Pro for page token) in 042c17a5

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,491 @@
...
+  lockstep, i.e. whole-batch paging.
+- Page through the results of each batched input element independently,
+  using individual sub-queries after the initial whole-batch query,
+  i.e. per-element paging.
+
+Each of these batched paging designs combines more or less well with the
+various paging options presented above.
+
+Henceforth the offset-based solution is rejected, given the lack of
+flexibility afforded to the manager's implementation.
```
foundrytom, line 198: [fix] I'd be good to call out that we also reject it for other reasons too (lack of stablity, potential overhead of record counting)

feltech: Updated the text in ac493901

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,615 @@
...
+- Manager plugin must implement yet another interface.
+- Complexity of binding to multiple target languages.
+- A mutable, stateful cursor violates the stateless design principle of
+  interactions with manager plugins. This causes complications, e.g.
+  thread-safety.
+- An immutable cursor object (such that `.next()` returns a new cursor
+  object), could increase optimisation complexity for the manager
+  implementation, e.g. managing query cache sharing, lifetime and
+  updates across multiple "live" cursors.
+- Destruction of a cursor object could be used as a signal to the
```
foundrytom, line 185: [question] Not a pro?

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,615 @@
...
+  page is active across the whole batch.
+
+#### Cons
+
+- Wasted processing of empty lists by the host when pages are exhausted
+  i.e. when there are still pages remaining for some inputs but not
+  others.
+- Batch elements cannot enter an error state in isolation, since paging
+  is whole-batch.
+
```
foundrytom, line 305: [fix] Something about how no one manager side could figure out in their heads out how you'd actually implement this

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,615 @@
...
+- Manager is under no obligation to consider errors in other streams
+  when implementing query behaviour, as pagination method need not be synchronized
+
+#### Cons
+
+- Batching is very difficult to optimize for the manager, especially
+  after the first page.
+- Batch-optimized queries have unpredictable performance profile for
+  the host, given the wide range of potential implementations.
+
```
foundrytom, line 354: [fix] The host cant express cursor modifications as a batch operation.

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,615 @@
...
+                related += page
+
+    return relateds
+```
+
+#### Cons
+
+- Additional metadata in the object is limited to whole-batch data.
+
+## Outcome
```
foundrytom, line 613: [fix] Great work on the restructure. Can we move this before the appendix with out conclusions to date

- no batch paging
- cursor or token

doc/decisions/DR016-One-to-many-result-paging.md
foundrytom: Don't worry about spending too much time on the batch, paged illustrations,  as we know were going to exclude them for now, def need to document manager implementation sketches for token vs cursor in singular form.

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,615 @@
...
+#### Cons
+
+- Precludes batching optimisations by the manager when paging is
+  required.
+- Requires implementation of two methods by the manager: one for
+  batched, one for paged.
+- Potential difficulty for some hosts/queries to determine a priori
+  which API (paging or batch) is the most optimal choice.
+- Having the choice adds a foot-gun for host developers to shoot
+  themselves with by accident.
```
foundrytom, line 261: - Precludes hosts atomically expressing more complex queries that may be required by spreadsheet-type UIs.

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Indeed, whoops.

doc/decisions/DR016-One-to-many-result-paging.md
feltech: I wonder if pros/cons that are specific to a particular paging strategy should go in a new section further down, and ideally only after we've rejected one or more of the options here, so we can reduce the matrix?

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Have we definitively concluded no batch paging?  I had spent some time trying (mostly failing) to sketch how that would look on the manager side, but if we're confident we've rejected batch paging then I can skip that.

foundrytom: We were strongly advised to keep the two things separate if we can. The only noted 'con' would be that it precludes complex UIs requesting their data in one go. It was noted though that we have no concrete use cases for that yet though, and can always add it later once we have some real world scenarios.

doc/decisions/DR016-One-to-many-result-paging.md
foundrytom: Was just trying to keep it a bit more grokable, we know people get lost in these when there are a lot of sections.

feltech: OK that's fine.  Though I'm now struggling to understand what your suggested Con means, please can you elaborate?

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Fixed in 7d540968

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,615 @@
...
+- Page through the results of each batched input element independently,
+  using individual sub-queries within the whole-batch query, i.e.
+  per-element paging.
+
+Each of these batched paging designs combines more or less well with the
+various paging options presented above.
+
+Henceforth the offset-based solution is rejected, given the lack of
+flexibility afforded to the manager's implementation, resulting in poor
+stability and likely manager overhead of record counting.
```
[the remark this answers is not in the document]
feltech: Moved to an intermediate Outcome section in 2c1020e3

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Added in 86010fc2

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Added an(other) intermediate Outcome section in 3854ec13

doc/decisions/DR016-One-to-many-result-paging.md
foundrytom: Sorry was trying to describe how the cursor-per-element approach precludes the host from getting the next page for multiple cursors simultaneously (as they're distinct methods on each individual cursor) which breaks the batch. If you use tokens, then you can supply multiple tokens at once to the second invocation which keeps the batch-ness intact.

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,522 @@
...
+```
+
+#### Pros
+
+- Whole batch is paged at once, retaining batch-awareness in the
+  manager for subsequent pages.
+- Conceptually simpler API for both host and manager if only a single
+  page is active across the whole batch.
+
+### Pros (page token specific)
```
foundrytom, line 310: [fix]
```suggestion
#### Pros (page token specific)
```

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,522 @@
...
+
+- Per-element callback means it's trivial to skip processing elements of
+  the input batch whose pages are exhausted.
+- Retains existing design in line with OpenAssetIO design principles,
+  i.e. non-paged signature is simply the same function but ignoring
+  optional arguments.
+
+#### Cons
+
+- Wasted processing of empty lists by the host when pages are exhausted
```
foundrytom, line 320: [fix] I think this needs to go to cursor specific?

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,522 @@
...
+  i.e. when there are still pages remaining for some inputs but not
+  others.
+- Batch elements cannot enter an error state in isolation, since paging
```
foundrytom, line 323: [question] Is this only true if the cursor returns a list of lists, vs calling the callback for each element as before? Would probably be good to keep the callback mechanism.

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,522 @@
...
+  is whole-batch.
+- Collaborators working on manager plugins raised this as being
+  particularly tricky to implement.
+
+#### Cons (page token specific)
+
+- Likely that the manager plugin must combine multiple back-end page
```
foundrytom, line 330: [question] This is true of the cursor approach too right? Just they're aggregated in the cursor object not the token object? Or is this really touching on the topic of is a token a value or an object?

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,522 @@
...
+def relatedEntities(
+    batch: List[EntityReference],
+    relationship,
+    successCb: SuccessCallback,
+    errorCb): ...
+```
+
+#### Pros
+
+- Manager is under no obligation to consider errors in other streams
```
foundrytom, line 377: [question]
```suggestion
- Manager is under no obligation to consider errors in other batch elements
```

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,522 @@
...
+from the batched API(s).
+
+This then leaves only two remaining solutions: (non-batched) page token
+or cursor objects.
+
+## Considering manager implementations
+
+### Using Python's DB-API standard
+
+#### Batch signature
```
foundrytom, line 406: [question] The preceding sentence rules out batching, be good to remove or move to the end?

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,522 @@
...
+#### Cursor object signature
+
+The design of DB-API cursors have a somewhat different control flow to
+the `AbstractPage` example used throughout, and in practice we would
+probably like to mirror a standard such as DB-API in any final design.
+However, shimming the DB-API cursor to match our contrived and
+ill-conceived `AbstractPage` is a useful exercise.
+
+```python
+class Page(AbstractPage):
```
foundrytom, line 489: [fix] Can we align the limit behaviour between the two implementations? As iirir, this version doesn't have a way to define the size of the first page.

ie. pass limit in `relatedEntities` and set the initial page size there?

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,522 @@
...
+        self.__is_pending = False
+        self.__page_size = 0
+
+    def get(self) -> List[EntityReference]:
+        self.__is_pending = False
+        result_set = self.__cursor.fetchmany(self.__page_size)
+        return [
+            EntityReference(result) for result[0] in result_set]
+
+    def next(self, pageSize):
```
foundrytom, line 501: [question] Is this illustrating the additional cursor functionality (ie arbitrarity advancing without fetching) as assuming (its not entirely clear in the docs) `fetchmany` advances the cursor anyway? If so - maybe calling it `skip` or similar may disambiguate the a host needs to call next to get the next page?

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,522 @@
...
+#### Page token signature
+
+```python
+class PageToken(AbstractPageToken):
+    def __init__(self, cursor):
+        self.cursor = cursor
+
+
+def relatedEntities(
+    input: EntityReference, relationship, limit=-1, pageToken=None
```
foundrytom, line 455: [consider]
```suggestion
    input: EntityReference, relationship, limit=-1, prevPageToken=None
```

doc/decisions/DR016-One-to-many-result-paging.md
feltech: It's tricky.  If one of the batch elements enters an error state in the page token design, then what's the correct way to get the next page of results for the remaining batch elements?  

You need to provide _some_ page token for the erroring element.  Perhaps the last one that was successful (if any were)?  Perhaps a `None` page token can mean "ignore"?  Perhaps you just have to deal with the error callback being called again and again for that failing batch element.

In any case, it's not really possible to consider them erroring in isolation.

foundrytom: Got you - might be good to elaborate in the con a little to explain in more detail

doc/decisions/DR016-One-to-many-result-paging.md
feltech: :thinking: It came from considering REST APIs that use page tokens and assuming that ultimately (whether in the plugin or some back-end) the batch request will actually be several requests with different page tokens, so a single uber page token must be a concatenation of some kind.

But like you say, if that was the back-end behaviour then wrapping in a cursor object doesn't solve the problem.

I'll "promote" that Con to a general problem with whole-batch paging.

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Batching isn't ruled out, instead it is split out from paging, so there's a separate sketch that considers batching only without paging.

doc/decisions/DR016-One-to-many-result-paging.md
feltech: The `next` function determines the page size applied to the next `get` call.

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Yeah, the whole `AbstractPage` design is a bit wonky and ill-conceived.  

I didn't want to go back and change all the previous sketches, and thought actually the fact that it doesn't conform to DP-API cursor objects provided a useful exercise in adapting a "real" cursor to an awkward design.  

Given that we can't predict the API of back-end cursor objects, I thought that was a valid exercise.

doc/decisions/DR016-One-to-many-result-paging.md
feltech: :thinking: Would it be `prev` or is it more like `current` or even `next`?  Since the token represents the page you haven't got yet.  Depends how you think about it.  Perhaps best to leave it generic?

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Added a Con in 3619805a

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Fixed in e8ca9f26

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Good spot, fixed in a4c2b516

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Moved in cef989e7

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Re-worded in 7df1b89b

doc/decisions/DR016-One-to-many-result-paging.md
foundrytom: > ... we opt to go with Option 1 and split the paging API(s)
from the batched API(s).
>
> This then leaves only two remaining solutions: (non-batched) page token
or cursor objects.

Reads as excluded!

doc/decisions/DR016-One-to-many-result-paging.md
foundrytom: Yeah- maybe thats the Q - what does the token represent!

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,584 @@
...
+    if limit <= 0:
+        result_set = pageToken.cursor.fetchall()
+    else:
+        result_set = pageToken.cursor.fetchmany(limit)
+
+    related_refs = [EntityReference(result[0]) for result in result_set]
+    return pageToken, related_refs
+```
+
+#### Page token alternative - command pattern
```
feltech, line 496: After offline conversaion, added this alternative sketch

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Oh to actually answer your question, yeah this is illustrating advancing without fetching, and indeed `fetchmany` also advances the cursor.

However, `next` _must_ be called by the user to set the page size for the `get` call, so "skip" doesn't really work, since it's only skipping if `get` isn't called.

doc/decisions/DR016-One-to-many-result-paging.md
foundrytom: But if I just call `get` on the cursor im given by the method, it gets everything defeating paging.

doc/decisions/DR016-One-to-many-result-paging.md
foundrytom: Yeah - but I think we have to balance confusing a reader of the DR who will consider this as a proposed cursor design.  The one presented would certainly put me off choosing that option.

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Ah OK, I'll update that text.

doc/decisions/DR016-One-to-many-result-paging.md
feltech: In the current wonky design the initial page size is 0, so it'll be `fetchmany(0)` and you'll get nothing.

Having said that, I agree with your assessment in https://github.com/OpenAssetIO/OpenAssetIO/pull/908#discussion_r1198864935 and will redesign the wonky `AbstractPage` to make a bit more sense.

doc/decisions/DR016-One-to-many-result-paging.md
foundrytom: I thought that was good though - that were trying to limit the options in the DR?

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Agreed, will redesign throughout.

doc/decisions/DR016-One-to-many-result-paging.md
foundrytom: 🙏

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,550 @@
...
+network latency concerns, in addition to database query optimizations.
+
+In the most general case, the host cannot know the size of the list of
+related entity references that will be returned to it. This clearly
+poses a scalability issue.
+
+Database and web APIs have had to deal with this problem for decades,
+and invariably provide some mechanism of returning slices of result sets
+or "pages" that can be iterated through, lazily fetching chunks of data
+in a controlled way.
```
elliotcmorris, line 38: [Consider] I realise we no longer (or never did?) mention stability as a problem in the Background. I wonder if we should prime the concept before we start mentioning it in pros and cons, especially as it's a large part of the reason for rejecting offset.

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,550 @@
...
+
+We opt for a non-batched cursor object design.
+
+Avoiding conflating batching and pagination simplifies manager plugin
+implementations.
+
+A cursor object maximises future extensibility, where new functionality
+can be added without further polluting the `Manager`/`ManagerInterface`
+class.
+
```
elliotcmorris, line 546: [Consider]
```suggestion
A cursor object maximises future extensibility, where new functionality
can be added without further polluting the `Manager`/`ManagerInterface`
class.

Whilst we may miss out on some theoretical optimization, we have been 
unable to find any real-world examples showing this sort of batched-page
optimization.
```

elliotcmorris (approved):
Looks fabulous.

Just two optional suggestions, feel free to resolve yourself if you wanna reject em.

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Originally it was mentioned in the Background, since we enumerated the three different paging styles and made brief mention of the stability issues of offset-based solutions.

I've added a little paragraph on the features of an ideal paging solution in a6785840

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,669 @@
...
+in a controlled way.
+
+In an ideal situation, requesting subsequent pages has performance
+benefits over repeating the same query twice (or more) with different
+range arguments, and yet work is performed lazily as pages are requested
+so that no unnecessary processing of unused pages is performed.  Also,
+pages should be stable, in that results are not skipped nor appear more
+than once on successive pages.
+
+Ultimately, the implementation must be provided by the manager plugin
```
feltech, line 47: After offline conversation, added this clarification in 724e9dc0

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,669 @@
...
+- Manager plugin must implement yet another interface.
+- Complexity of binding to multiple target languages.
+- A mutable, stateful cursor violates the stateless design principle of
+  interactions with manager plugins. This causes complications, e.g.
+  thread-safety.
+- An immutable cursor object (such that `.next()` returns a new cursor
+  object), could increase optimisation complexity for the manager
+  implementation, e.g. managing query cache sharing, lifetime and
+  updates across multiple "live" cursors.
+- The temptation to directly wrap a "real" DB cursor object in our
```
feltech, line 201: After offline conversation/feedback, added this Con and updated the Outcome slightly in 5d85986d

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,669 @@
...
+A cursor object maximises future extensibility, where new functionality
+can be added without further polluting the `Manager`/`ManagerInterface`
+class. They also naturally capture arbitrary state that the manager
+may wish to store as part of the ongoing query.
+
+Cursor objects also reflect the most common programmatic approach to
+paging through the results of a database query, presenting a familiar
+interface for plugin authors, and mapping well to common back-end SDKs.
+
+## Appendix: Considering manager implementations
```
feltech, line 443: After offline conversation, to reduce confusion re. dictating a particular implementation, moved all examples to this Appendix section in 0b9b211c

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,669 @@
...
+        f"""
+        SELECT ToRef FROM Relationship
+        WHERE RelationshipType = ?
+        AND FromRef = ?
+        """, my_relationship_type, str(input))
+
+    return Cursor(cursor)
+```
+
+### Using a REST API
```
feltech, line 575: After offline conversation, added a sketch using a theoretical REST API in 22c9f811

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,669 @@
...
+objects.
+
+## Outcome
+
+We opt for a non-batched cursor object design.
+
+Avoiding conflating batching and pagination simplifies manager plugin
+implementations.
+
+Whilst we may miss out on some theoretical performance, we have been
```
feltech, line 430: Reorganised slightly in 44831de0

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,669 @@
...
+so that no unnecessary processing of unused pages is performed.  Also,
+pages should be stable, in that results are not skipped nor appear more
+than once on successive pages.
+
+Ultimately, the implementation must be provided by the manager plugin
+authors. The mechanism chosen in this DR can only present an abstract
+interface that the plugin must satisfy. As a result, characteristics
+such as page stability and performance cannot be guaranteed, but we can
+at least ensure that the chosen mechanism does not preclude such
+optimisations.
```
elliotcmorris, line 52: [Minor][Consider]
Super subjective thing, but this reads to me almost like we're saying "Yes, managers _should_ be stable but we can't force them to be", when I think in actuality we mean more like "Manager may want to be stable, and we don't want to get in the way of that, whilst we also don't want force it."

Maybe?

```suggestion
such as page stability and performance cannot be guaranteed, and nor
do we desire to place such restrictions upon the implementation. However
we can at least ensure that the chosen mechanism does not preclude such
optimisations.
```

feltech: Yeah I like it, will commit your suggestion as-is.

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,691 @@
+# DR016 One-to-many result paging
+
+- **Status:** Proposed
```
foundrytom, line 3: [fix] Can we move this to approved

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,691 @@
...
+- **Impact:** Medium
+- **Driver:** @feltech
+- **Approver:** @elliotcmorris @foundrytom
+- **Contributors:** @feltech @elliotcmorris
```
foundrytom, line 7: [fix] Can we add contributors from the threads?

foundrytom (approved):
Looking good. Thanks. Based on community feedback can we move this to approved and squash down for merge?

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Do you mean what GitHub refers to as "participants" (shown at the top right) on this PR?

doc/decisions/DR016-One-to-many-result-paging.md
```diff
@@ -0,0 +1,691 @@
...
+
+Given the reported difficulty of crafting queries that are both batch
+and paging-aware, we opt to go with Option 1 and split the paging API(s)
+from the batched API(s).
+
+This then leaves only a batch-only signature, plus a decision between
+two remaining paging solutions: (non-batched) page token or cursor
+objects.
+
+## Outcome
```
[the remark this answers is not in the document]
feltech: Disambiguated in da871bc3

doc/decisions/DR016-One-to-many-result-paging.md
feltech: Added GitHub participants in 426f7c93

foundrytom (commented):
Thank you kindly! FFTSAMAYDL

PR #911, opened 2023-05-03
Create new Contexts with pre-initialized locale

### LINKED ISSUE #903: Create Contexts with empty `TraitsData` in the `locale`
## What

☝️ 

## Why

Makes it easier for customisations as there will then always be a valid locale, vs a null ptr.

foundrytom:
It is a lot easier to work with the locale if it is never null. Otherwise, appending to a locale ends up needing branches which adds unneccesary complexity.

Closes #903


elliotcmorris (approved):
:+1: LGTM

Issue #573, opened 2022-08-15
[Core] `EntityReference` container interoperability


feltech:
## What

Design and implement convenience functionality for handling and querying `EntityReference` objects.

Especially in Python, Eg: as keys in a dict etc.

```
TypeError: unhashable type: 'openassetio._openassetio.EntityReference'
```

## Why

Entity references used to be simple strings.  Now they are a strong type and lose the conveniences that come with strings. See https://github.com/OpenAssetIO/OpenAssetIO/pull/567#discussion_r945649223

### User

- As a developer integrating OpenAssetIO, I need to be able to use entity references where I would normally use strings, such as map keys, or passing to a string formatting library. Without this, it adds significant unnecessary complexity to my integration.

- As a developer integrating OpenAssetIO I should still be protected by the strong-typing protections for OpenAssetIO API methods, to avoid programming errors.

## Acceptance Criteria
- Implement all possible dunder methods for python EntityReference
- EntityReference is usable in STL containers, implementing std::hash etc

PR #920, opened 2023-05-09
[Python] Add `__str__` implementation for `EntityReference`

foundrytom:
Had a moment to chip away at the backlog a little, so was looking for a small task.

Allows them to be used with fstrings, print, etc. Retains the `toString` method for compatibility and/or more explicit programming.

Part of #573.


feltech (approved):
LGTM

PR #922, opened 2023-05-10
[Core] Add Ptr alias members to classes

### LINKED ISSUE #918: Provide generically accesible pointer type on classes
## What
Several of our types have their own pointer types, defined by the macro `OPENASSETIO_DECLARE_PTR`, which creates aliased shared_ptr types for that specific class, which are used heavily in their interfaces.

```cpp
#define OPENASSETIO_DECLARE_PTR(Class)             \
  /* NOLINTNEXTLINE(bugprone-macro-parentheses) */ \
  class Class;                                     \
  using Class##Ptr = std::shared_ptr<Class>;       \
  using Class##ConstPtr = std::shared_ptr<const Class>;
```

For example, in [Manager.hpp](https://github.com/OpenAssetIO/OpenAssetIO/blob/6c6690d854c4261fa3a72cc48848d8d94cdbf19f/src/openassetio-core/include/openassetio/hostApi/Manager.hpp#L32) you can see how we create a `ManagerPtr` and `ManagerConstPtr` via this macro.

We should create a alias on the type that allows us to reference the pointer type in a compile time context when we only have the main type. ie, `Manager::PtrType` (name not final) should resolve to `ManagerPtr`

## Why
This is useful in generic programming, as generic functions can be written that accept the primary type, and work freely with interfaces that accept and receive the `Ptr` types. This came up recently with regard to writing converters to/from python types, which want to apply broadly generic code across the entire api surface.

## Sketch
A possible implementation may be : 
```
OPENASSETIO_DECLARE_PTR(Manager)     

class Manager {
public:
...
using PtrType = ManagerPtr;
...
```

and a usage, taken from WIP python converters. This allows us the option to define the frontend interface as requiring either `Manager` or `ManagerPtr`, depending on our preference.
```cpp
template <typename T>
T::Ptr_Type castFromPyObject(PyObject* theObject) {
  // get Python object.
  const auto pyInstance = py::reinterpret_borrow<py::object>(theObject);

  // Extract the underlying C++ base class pointer.
  auto* cppInstancePtr = py::cast<T*>(pyInstance);

  // Use aliasing constructor, linking Python instance and C++ instance
  // lifetimes via PyObject refcount.
  return pointers::createPyRetainingPtr<T::Ptr_Type>(pyInstance, cppInstancePtr);
}
```

## Acceptance Criteria
- Generic ptr type mechanism added to all types that have pointer type declarations

feltech:
## Description

Closes #918

## Test Instructions

Tests are all compile-time `static_assert` checks.

### DIFF
--- a/doc/contributing/CODING_STANDARDS.md
+++ b/doc/contributing/CODING_STANDARDS.md
@@ -93,8 +93,8 @@ and consistency trade-offs.
 #### Smart pointers
 
 C++ classes that represent system components with reference semantics
-(as opposed to 'value' types) should inherit
-`std::enable_shared_from_this` and define a peer `Ptr` alias using
+(as opposed to 'value' types) should define both a qualified peer
+`ClassNamePtr` alias and an unqualified member `Ptr` alias, using
 `std::shared_ptr`. Their constructors should be private, and a static
 `make` method provided.
 
@@ -105,6 +105,9 @@ There is a convenience macro `OPENASSETIO_DECLARE_PTR` available in
 `typedefs.hpp` that should be used to declare the `shared_ptr`. A macro
 gives us a single point of change should we wish to alter or add more
 declarations.
+
+The `OPENASSETIO_ALIAS_PTR` macro should then be used to declare
+`shared_ptr` alias members.
 
 Usage example:
 
@@ -117,6 +120,8 @@ OPENASSETIO_DECLARE_PTR(Host)
 
 class OPENASSETIO_CORE_EXPORT Host final {
 public:
+    OPENASSETIO_ALIAS_PTR(Host)
+
     static HostPtr make(...);
 
 private:
--- a/src/openassetio-core/include/openassetio/typedefs.hpp
+++ b/src/openassetio-core/include/openassetio/typedefs.hpp
@@ -88,6 +88,21 @@ using Identifiers = std::vector<Identifier>;
   using Class##ConstPtr = std::shared_ptr<const Class>;
 
 /**
+ * Create unqualified shared_ptr aliases of qualified shared_ptr
+ * aliases.
+ *
+ * Used to define `Ptr`/`ConstPtr` aliases within class definitions,
+ * thus allowing their use in generic programming, e.g. `typename
+ * T::Ptr`.
+ *
+ * @note This requires OPENASSETIO_DECLARE_PTR to have been invoked for
+ * the target class.
+ */
+#define OPENASSETIO_ALIAS_PTR(Class) \
+  using Ptr = Class##Ptr;            \
+  using ConstPtr = Class##ConstPtr;
+
+/**
  * Forward declare a class and its smart pointers.
  *
  * Wraps the declaration in the top-level openassetio and ABI
--- a/src/openassetio-core/tests/CMakeLists.txt
+++ b/src/openassetio-core/tests/CMakeLists.txt
@@ -17,6 +17,7 @@ install(
 target_sources(openassetio-core-cpp-test-exe
     PRIVATE
     main.cpp
+    typedefsTest.cpp
     BatchElementErrorTest.cpp
     ContextTest.cpp
     TraitsDataTest.cpp
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -48,6 +48,10 @@ v1.0.0-alpha.xx
   allowing more intuitive use with `format`, `print`, and others.
   [#573](https://github.com/OpenAssetIO/OpenAssetIO/issues/573)
 
+- Added `Ptr`/`ConstPtr` alias members to all appropriate C++ classes,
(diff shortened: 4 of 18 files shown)


doc/contributing/CODING_STANDARDS.md
```diff
@@ -117,7 +120,9 @@ OPENASSETIO_DECLARE_PTR(Host)
 
 class OPENASSETIO_CORE_EXPORT Host final {
 public:
-    static HostPtr make(...);
+    OPENASSETIO_ALIAS_PTR(Host)
+
+    static Ptr make(...);
```
elliotcmorris, line 125: [Consider][Minor] Do you think maybe leaving the more descriptive type when working concretely (ie HostPtr instead of Ptr) might be clearer, means the reader dosen't have to burrow into a macro. (or rather, only has to burrow into one).

src/openassetio-core/include/openassetio/typedefs.hpp
```diff
@@ -87,6 +87,18 @@ using Identifiers = std::vector<Identifier>;
   using Class##Ptr = std::shared_ptr<Class>;       \
   using Class##ConstPtr = std::shared_ptr<const Class>;
 
+/**
+ * Create unqualified shared_ptr aliases of qualified shared_ptr
+ * aliases.
+ *
+ * Used to define `Ptr`/`ConstPtr` aliases within class definitions,
+ * thus allowing their use in generic programming, e.g. `typename
+ * T::Ptr`.
```
elliotcmorris, line 96: [Consider] Maybe add something like 

> Requires `OPENASSETIO_DECLARE_PTR` to have been invoked previously in the document.

src/openassetio-core/tests/CMakeLists.txt
```diff
@@ -20,9 +20,16 @@ target_sources(openassetio-core-cpp-test-exe
...
+    hostApi/HostInterfaceTest.cpp
+    hostApi/ManagerFactoryTest.cpp
+    hostApi/ManagerImplementationFactoryInterfaceTest.cpp
     hostApi/ManagerTest.cpp
+    log/ConsoleLoggerTest.cpp
+    log/LoggerInterfaceTest.cpp
+    log/SeverityFilterTest.cpp
     managerApi/HostTest.cpp
     managerApi/HostSessionTest.cpp
+    managerApi/ManagerInterfaceTest.cpp
```
elliotcmorris, line 32: [Question] This dosen't do any harm, but as I mentioned before, I question if there's much value in these static checks, I'm not sure what we're testing, other than that we haven't forgotten to declare the macro (which has the problem that you're equally likely to forget the test.)

elliotcmorris (approved):

doc/contributing/CODING_STANDARDS.md
feltech: Good point, I had the same thought with the rest of the codebase, but obviously didn't change the .md to match.

src/openassetio-core/tests/CMakeLists.txt
feltech: In my TDD loop I added the test to a couple of the classes (red), then did the naive solution of manually adding the aliases without using a macro (green), then wrote and used the macro (refactor).  

So by adding the tests I guarded against a bad refactor.  Though that particular benefit only applied to the first couple of classes. Perhaps an explicit test of the macro would be better.

Another benefit is guarding against accidental breakage (e.g. deletion) of the aliases at some point in the future. Though if the aliases were tested implicitly by being used in other bits of code, then that would be sufficient (IMO), but as of this PR they're not used anywhere.

We could perhaps refactor the individual tests into a single templated test?  That might also tick the box of "tests as documentation", since we would maintain a list of classes that have a `Ptr` alias member.

elliotcmorris: I think a single template test would be the best form of this, we could test not just the aliases, but also the existence of the pointer types in general.

feltech: I think the Python converter stuff will have to be separate, since it may not be compiled if Python support is disabled.

doc/contributing/CODING_STANDARDS.md
feltech: Updated docs in c0566f8b

src/openassetio-core/include/openassetio/typedefs.hpp
feltech: Updated docstring in 9e6ad1dd

src/openassetio-core/tests/CMakeLists.txt
feltech: Refactored tests in d7632df7

elliotcmorris: Thanks for that, seems much nicer to me.

elliotcmorris:
LGTMSAMWR

PR #923, opened 2023-05-11
[Core] Add python conversion functions

### LINKED ISSUE #798: Provide C++ functions to expose objects to Python
## What

As a hybrid C++/Python host application author
I want functions to expose C++ `Manager`, `Context`, `LoggerInterface` ... objects to Python that were created in C++ (and vice versa?)
Because assets need to be queried and published in both languages sharing the same underlying OpenAssetIO state.

## Why

If a host application uses a compatible version of pybind11, then exposing a (bound) OpenAssetIO object to Python is fairly trivial (`pybind11::cast(...)` et al.).  However, we cannot rely on a host application using pybind11 (or a compatible version of it). In OpenAssetIO we deliberately hide the dependency on pybind11 as an internal tool, so it is not accessible directly through the OpenAssetIO library.

So we should expose functions whose only external dependency is CPython (`PyObject*` et al) and that will expose a C++-created `Manager` object (for example) to the host application's Python environment.

## Acceptance Criteria

* `exposeToPython`-like functions for all required cross-language objects (to be designed).
* Documentation examples updated.
* Integration tests added (OpenAssetIO-Test-CMake or otherwise).
* If `py::cast` fails, raise the string held in `PyErr` (no custom exception needed)
* The refcount strategy for the method is conceptually aligned with the CPython API usage.
* GIL is not acquired by the method (unless required by error handling etc).
* GIL and ref counting behaviour is documented as part of the method docs.
* Test lifetimes (including python derived class unowned in Python called from C++)
* Mis-matched casts raise appropriate non-pybind exception
* Converters do not assume ownership of the castee

## Notes

Workflow sketches required to figure out the API surface area, including which objects must be explicitly exposed by the host application.

elliotcmorris:
## Description

There is a need as a hybrid C++/Python host application author to convert types between an embedded python interpreter and a c++ context, as some hosts will want to do asset resolution in both languages, sharing the same configuration.

This adds two methods to python bridge, `castToPyObject` and `castFromPyObject` that facilitates this.

Prior, the way to do this was to do a pybind cast, which is what we do here behind the scenes, but by providing methods to raw CPython types, we eliminate the need to link against pybind for the consumer.

Closes #798 

~~[ ] I have updated all relevant user documentation.~~

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -31,6 +31,11 @@ v1.0.0-alpha.xx
 
 ### Improvements
 
+- Added utility methods `castToPyObject` and `castFromPyObject` to
+  facilitate converting between cpp and python objects for hosts
+  seeking to support mixed language workflows.
+  [#798](https://github.com/OpenAssetIO/OpenAssetIO/issues/798)
+
 - Added support for running `ctest` when a python venv is used to
   determine which Python distribution to build against.
 
--- a/src/openassetio-python/bridge/CMakeLists.txt
+++ b/src/openassetio-python/bridge/CMakeLists.txt
@@ -55,7 +55,8 @@ endif ()
 # Source file dependencies.
 target_sources(openassetio-python-bridge
     PRIVATE
-    src/python/hostApi.cpp)
+    src/python/hostApi.cpp
+    src/python/converter/converters.cpp)
 
 # Public header dependency.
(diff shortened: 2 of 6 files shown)


elliotcmorris:
Havn't review this myself quite yet, just putting this up as a draft to get a jump on the test. (I forgot there's an automatic reviewer assignment now.)

PR #929, opened 2023-05-17
Provide C++ functions to expose objects to Python

### LINKED ISSUE #798: Provide C++ functions to expose objects to Python

feltech:
## Description

Closes #798. Adds new `openassetio::python::converter` namespace with functions `castToPyObject` and `castFromPyObject`.

- [ ] I have updated all relevant user documentation (TBD see below)

## Reviewer Notes

AC "Documentation examples updated" - other than docstrings (which contain snippets), we don't have C++ example docs to update(?).

AC "Integration tests added (OpenAssetIO-Test-CMake or otherwise)" - do we still think this is necessary in addition to included Catch2 tests?

## Test Instructions

Unit tests provided. Tricky to test externally without hacking together a hybrid C++/Python OpenAssetIO host.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -29,6 +29,13 @@ v1.0.0-alpha.xx
   `ResolveSuccessCallback` from reference types to value types.
   [#858](https://github.com/OpenAssetIO/OpenAssetIO/issues/858)
 
+### New Features
+
+- Added utility methods `castToPyObject` and `castFromPyObject` to
+  `openassetio-python-bridge` to facilitate converting between C++ and
+  Python objects for hosts seeking to support mixed language workflows.
+  [#798](https://github.com/OpenAssetIO/OpenAssetIO/issues/798)
+
 ### Improvements
 
 - Added support for running `ctest` when a python venv is used to
--- a/src/openassetio-python/bridge/include/openassetio/python/converter.hpp
+++ b/src/openassetio-python/bridge/include/openassetio/python/converter.hpp
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2023 The Foundry Visionmongers Ltd
+#pragma once
+
+#include <Python.h>
+
+#include <openassetio/export.h>
+#include <openassetio/python/export.h>
+#include <openassetio/typedefs.hpp>
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+/**
+ * Converter functionality for going between C++ and CPython objects.
+ */
+namespace python::converter {
+/**
+ * Casts a C++ API object to the equivalent Python object.
+ *
+ * This template is explicitly instantiated to only the OpenAssetIO
+ * types, and is not intended to be a generic converter.
+ *
+ * The purpose of this function is to provide a Python/C++ conversion
+ * without requiring/exposing a particular CPython binding library
+ * implementation.
+ *
+ * The returned `PyObject` owns a hidden reference to the input
+ * `shared_ptr`, ensuring it is kept alive. The reference will be
+ * released when the `PyObject` is destroyed.
+ *
+ * @note This function does not acquire the GIL.
+ *
+ * @warning A Python environment, with `openassetio` imported, must be
+ *          available in order to use this function.
+ *
+ * @throws std::invalid_argument if the input is null.
+ * @throws std::runtime_error if the cast fails.
+ *
+ * @param objectPtr A non-const OpenAssetIO pointer type, (eg, @needsref
+ * ManagerPtr). The returned `PyObject` takes shared ownership of this
+ * input \p objectPtr, and will keep the C++ instance alive until the
+ * `PyObject` is destroyed.
+ * This parameter must be the non-const OpenAssetIO pointer type as
+ * casting to python erases constness.
+ *
+ * @return A `PyObject` pointer to an object of the Python API type
+ * associated with the C++ API object provided. The reference count of
+ * the `PyObject` will be incremented, and must be decremented by the
+ * caller when done.
+ */
+template <typename T>
+OPENASSETIO_PYTHON_BRIDGE_EXPORT PyObject* castToPyObject(const T& objectPtr);
+
+/**
+ * Casts a Python object to the equivalent C++ API object.
+ *
+ * This template is explicitly instantiated to only the OpenAssetIO
+ * types, and is not intended to be a generic converter.
+ *
+ * The purpose of this function is to provide a Python/C++ conversion
+ * without requiring/exposing a particular CPython binding library
+ * implementation.
+ *
+ * The returned `shared_ptr` owns a hidden reference to the input
+ * `PyObject`, ensuring it is kept alive. The reference will be released
+ * when the `shared_ptr` is destroyed.
+ *
+ * Using this function requires specifying the template argument of
+ * the C++ API type equivalent to the type of the object referred to by
+ * the \p pyObject pointer.
+ *
+ * @code{.cpp}
+ * ManagerPtr manager = castFromPyObject<Manager>(pyManager);
+ * @endcode
+ *
+ *
+ * If the types of the template argument and the \p pyObject are not
+ * equivalent, an exception will be thrown due to inability to perform
+ * the cast.
+ *
+ * @note This function does not acquire the GIL.
+ *
+ * @warning A Python environment, with `openassetio` imported, must be
+ *          available in order to use this function.
+ *
+ * @throws std::invalid_argument if the function fails due to inability
+ * to cast between types, or if the input is null.
+ *
+ * @param pyObject A `PyObject` pointer to a Python object that must be
+ * of equivalent type to the template argument.
+ *
+ * @return An OpenAssetIO pointer to a C++ API object cast from the
+ * provided \p pyObject. The lifetime of the PyObject will be extended
+ * to at least the lifetime of the returned `shared_ptr`.
+ */
+template <typename T>
+OPENASSETIO_PYTHON_BRIDGE_EXPORT typename T::Ptr castFromPyObject(PyObject* pyObject);
+
+}  // namespace python::converter
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
--- a/src/openassetio-python/bridge/src/python/converter.cpp
+++ b/src/openassetio-python/bridge/src/python/converter.cpp
@@ -1,0 +1,109 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 The Foundry Visionmongers Ltd
+#include <openassetio/python/converter.hpp>
+
+#include <pybind11/embed.h>
+
+#include <openassetio/Context.hpp>
+#include <openassetio/TraitsData.hpp>
+#include <openassetio/hostApi/HostInterface.hpp>
+#include <openassetio/hostApi/Manager.hpp>
+#include <openassetio/hostApi/ManagerFactory.hpp>
+#include <openassetio/hostApi/ManagerImplementationFactoryInterface.hpp>
+#include <openassetio/log/ConsoleLogger.hpp>
+#include <openassetio/log/LoggerInterface.hpp>
+#include <openassetio/log/SeverityFilter.hpp>
+#include <openassetio/managerApi/Host.hpp>
+#include <openassetio/managerApi/HostSession.hpp>
+#include <openassetio/managerApi/ManagerInterface.hpp>
+#include <openassetio/managerApi/ManagerStateBase.hpp>
+
+// Private headers
+#include <openassetio/private/python/pointers.hpp>
+
+namespace py = pybind11;
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace python::converter {
+
+template <typename T>
+PyObject* castToPyObject(const T& objectPtr) {
+  // Stash the errors for restoring after cast.
+  // An error_scope calls PyErr_Fetch (which clears the active error),
+  // and stashes them, restoring the error state when it falls out of
+  // scope
+  const py::error_scope previousErrorState;
+
+  if (objectPtr == nullptr) {
+    throw std::invalid_argument(
+        "Attempting to cast a nullptr objectPtr in "
+        "openassetio::python::converter::castToPyObject");
+  }
+
+  py::object pybindObj = py::cast(objectPtr);
+
+  // Check to see if the py::cast has spawned a python error
+  // We use another error_scope here for convenience of access, relying
+  // on the fact that the original error scope will be destructed last,
+  // resetting to the initial error.
+  const py::error_scope castErrorState;
+
+  if (castErrorState.value != nullptr) {
+    throw std::runtime_error(py::str(castErrorState.value));
+  }
+
+  // `release()` to avoid decrementing the PyObject refcount on leaving
+  // this function. We explicitly want +1 refcount.
+  return pybindObj.release().ptr();
+}
+
+template <typename T>
+typename T::Ptr castFromPyObject(PyObject* pyObject) {
+  if (pyObject == nullptr) {
+    throw std::invalid_argument(
+        "Attempting to cast a nullptr PyObject in "
+        "openassetio::python::converter::castFromPyObject");
+  }
+
+  // get Python object.
+  const auto pyInstance = py::reinterpret_borrow<py::object>(pyObject);
+
+  try {
+    // Extract the underlying C++ base class pointer.
+    auto* cppInstancePtr = py::cast<T*>(pyInstance);
+
+    // Use aliasing constructor, linking Python instance and C++ instance
+    // lifetimes via PyObject refcount.
+    return pointers::createPyRetainingPtr<typename T::Ptr>(pyInstance, cppInstancePtr);
+  } catch (const py::cast_error& castError) {
+    // Rethrow here to avoid bleeding pybind exception dependencies.
+    throw std::invalid_argument(castError.what());
+  }
+}
+
+// To/from explicit specialization convenience.
+// As these specializations are being used across library boundaries,
+// we must export them to prevent undefined-symbol link errors.
+#define OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(Class)                          \
+  template OPENASSETIO_PYTHON_BRIDGE_EXPORT PyObject* castToPyObject<Class::Ptr>( \
+      const Class::Ptr&);                                                         \
+  template OPENASSETIO_PYTHON_BRIDGE_EXPORT Class::Ptr castFromPyObject<Class>(PyObject*);
+
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(openassetio::Context)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(openassetio::TraitsData)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(hostApi::HostInterface)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(hostApi::Manager)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(hostApi::ManagerFactory)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(hostApi::ManagerImplementationFactoryInterface)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(log::ConsoleLogger)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(log::LoggerInterface)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(log::SeverityFilter)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(managerApi::Host)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(managerApi::HostSession)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(managerApi::ManagerInterface)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(managerApi::ManagerStateBase)
+
+}  // namespace python::converter
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
--- a/src/openassetio-python/tests/bridge/CMakeLists.txt
+++ b/src/openassetio-python/tests/bridge/CMakeLists.txt
@@ -31,6 +31,7 @@ target_sources(
     openassetio-python-bridge-test-exe
     PRIVATE
     main.cpp
+    python/test_converter.cpp
     python/test_hostApi.cpp
 )
 
@@ -97,7 +98,7 @@ separate_arguments(_envvars NATIVE_COMMAND ${_envvars})
 
 
 #-----------------------------------------------------------------------
-# Create CTest target.
+# Create CTest target for tests that may import _openassetio
 add_custom_target(
     openassetio.internal.python-bridge-test
     COMMAND
@@ -116,5 +117,33 @@ $<TARGET_FILE_NAME:openassetio-python-bridge-test-exe>"
 openassetio_add_test_target(openassetio.internal.python-bridge-test)
 openassetio_add_test_fixture_dependencies(
     openassetio.internal.python-bridge-test
+    openassetio.internal.install
+)
+
+#-----------------------------------------------------------------------
+# Create CTest target for tests that intentionally exclude _openassetio
+# This came about from wanting to test the error state of functions
+# that use the python interpreter, specifically the errors that happen
+# when `_openassetio` hadn't been imported into the python environment.
+# We can't just rely on ordering for this, as test execution order
+# is non-consitent across platforms.
+add_custom_target(
+    openassetio.internal.python-bridge-no-openassetio-module-test
+    COMMAND
+    # Note: environment variables set in-line here rather than using
+    # `set_tests_properties`. If that function is used then the calling
+    # environment is affected. In particular, PYTHONHOME can interfere
+    # with CMake itself (if installed as a pip package and so
+    # /usr/local/cmake is actually a Python console_script).
+    # Note: as above, we must not have trailing whitespace after setting
+    # an env var on Windows.
+    ${_envvars}&&
+"${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/\
+$<TARGET_FILE_NAME:openassetio-python-bridge-test-exe>" "[no_openassetio_module]"
+)
+
+openassetio_add_test_target(openassetio.internal.python-bridge-no-openassetio-module-test)
+openassetio_add_test_fixture_dependencies(
+    openassetio.internal.python-bridge-no-openassetio-module-test
     openassetio.internal.install
 )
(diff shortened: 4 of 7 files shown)


src/openassetio-python/bridge/src/python/converter.cpp
```diff
@@ -0,0 +1,105 @@
...
+  return pybindObj.release().ptr();
+}
+
+template <typename T>
+typename T::Ptr castFromPyObject(PyObject* pyObject) {
+  if (pyObject == nullptr) {
+    throw std::invalid_argument("Attempting to cast a nullptr PyObject");
+  }
+
+  // get Python object.
```
foundrytom, line 65: [fix] (unless its obvious/the only way) can we explain why `reinterpret_borrow` specifically?

src/openassetio-python/bridge/src/python/converter.cpp
```diff
@@ -0,0 +1,105 @@
...
+    // Rethrow here to avoid bleeding pybind exception dependencies.
+    throw std::invalid_argument(castError.what());
+  }
+}
+
+// To/from explicit specialization convenience.
+// As these specializations are being used across library boundaries,
+// we must export them to prevent undefined-symbol link errors.
+#define OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(Class)                          \
+  template OPENASSETIO_PYTHON_BRIDGE_EXPORT PyObject* castToPyObject<Class::Ptr>( \
```
foundrytom, line 85: [question] Do we need to consider `ConstPtr` here at all? I know pybind erases constness, I guess C++ is happy to pop the non-const one into a const one on assignment?

src/openassetio-python/bridge/src/python/converter.cpp
```diff
@@ -0,0 +1,105 @@
...
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(hostApi::Manager)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(hostApi::ManagerFactory)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(hostApi::ManagerImplementationFactoryInterface)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(log::ConsoleLogger)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(log::LoggerInterface)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(log::SeverityFilter)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(managerApi::Host)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(managerApi::HostSession)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(managerApi::ManagerInterface)
+OPENASSETIO_SPECIALIZE_PYTHON_CONVERSIONS(managerApi::ManagerStateBase)
```
foundrytom, line 101: ❤️

src/openassetio-python/tests/bridge/CMakeLists.txt
```diff
@@ -118,3 +119,26 @@ openassetio_add_test_fixture_dependencies(
     openassetio.internal.python-bridge-test
     openassetio.internal.install
 )
+
+#-----------------------------------------------------------------------
+# Create CTest target for tests that intentionally exclude _openassetio
```
foundrytom, line 124: [fix] Can we explain why?

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
```diff
@@ -20,49 +20,51 @@ namespace python::converter {
...
  *
- * @return A `PyObject` pointer to an object of the Python api type
- * associated with the c++ api object provided.
- * The pointer is not an RAII type, and will be returned with an
- * incremented reference count for the caller to manage,  and must
- * must be released via `Py_DECREF` or similar mechanisms.
+ * @return A `PyObject` pointer to an object of the Python API type
+ * associated with the C++ API object provided. The reference count of
+ * the `PyObject` will be incremented, and ownership of the reference
+ * transferred to the caller.
```
elliotcmorris, line 47: [Consider] I'm unsure whether saying that the ownership is transferred is enough help for these docs. This is the only place in the API where you have to take manual ownership of a type and call a cleanup function, so I don't think assumed CPython knowledge should be taken for granted. I'd rather be more explicit and inform the caller that they need to call PY_DECREF.

src/openassetio-python/tests/bridge/python/test_converter.cpp
```diff
@@ -0,0 +1,268 @@
...
+        CHECK(Py_REFCNT(pyTraitsData) == 1);
+        PyObject* resultBool = PyObject_CallMethod(pyTraitsData, "hasTrait", "s", kTestTraitId);
+
+        // Use CHECK rather than REQUIRE in order to ensure the Decrefs
+        // below actually execute.
+        CHECK(PyBool_Check(resultBool));
+        CHECK(PyObject_IsTrue(resultBool));
+        Py_DECREF(resultBool);
+      }
+      Py_DECREF(pyTraitsData);
```
foundrytom, line 115: [question] Should this be up one level in the scope, as the incref happens at 99?

src/openassetio-python/tests/bridge/python/test_converter.cpp
```diff
@@ -0,0 +1,268 @@
...
+        CHECK(PyObject_IsTrue(resultBool));
+        Py_DECREF(resultBool);
+      }
+    }
+
+    WHEN("data is set via the Python object") {
+      PyObject_CallMethod(pyTraitsData, "addTrait", "s", kTestTraitId);
+      THEN("C++ object reflects the data set") { CHECK(traitsData->hasTrait(kTestTraitId)); }
+    }
+    Py_DECREF(pyTraitsData);
```
foundrytom, line 91: [question] Should this be up one level in the scope, as the incref happens at 69?

src/openassetio-python/tests/bridge/python/test_converter.cpp
```diff
@@ -0,0 +1,268 @@
...
+    }
+  }
+}
+
+SCENARIO("Casting to a C++ object binds object lifetime") {
+  GIVEN("A python object") {
+    py::module_::import("openassetio");
+
+    // Create a python object and release it, to simulate an unmanaged
+    // PyObject being provided to us.
```
foundrytom, line 125: [fix] We don't make the same qualifying comment about L69 - is this relevant still? Is it that we're making a _pybind_ object that motivates the callout to explain the `release()`  as we'd just `PyObject*` natively if it wasn't cumersome? if so, can we say pybind and it's convenient, and add the comment before too (+ update 150)?  I understood that we wanted to use raw CPython where possible in these tests, so good to have the exception explained (and maybe that philosophy documented at the top too?)

src/openassetio-python/tests/bridge/python/test_converter.cpp
```diff
@@ -80,25 +83,26 @@ SCENARIO("Mutations in one language are reflected in the other") {
...
-      }
-      THEN("The python object remains alive and can be operated on via the python interpreter") {
+  GIVEN("A Python object casted from a C++ object") {
+    TraitsDataPtr traitsData = TraitsData::make();
+    traitsData->addTrait(kTestTraitId);
+    PyObject* pyTraitsData = openassetio::python::converter::castToPyObject(traitsData);
+
+    WHEN("C++ reference is destroyed") {
+      CHECK(Py_REFCNT(pyTraitsData) == 1);  // Initial condition.
+      traitsData.reset();
```
elliotcmorris, line 103: :+1: Much better, wasn't possible when we were using Manager due to there not being a default constructor

src/openassetio-python/tests/bridge/python/test_converter.cpp
```diff
@@ -124,26 +128,28 @@ SCENARIO("Casting to a C++ object binds object lifetime") {
...
+
+      AND_WHEN("C++ reference is destroyed") {
+        traitsData.reset();
+
+        THEN("Python reference is released") { CHECK(Py_REFCNT(pyTraitsData) == 1); }
       }
     }
-    AND_WHEN("The c++ object falls out of scope") {
-      THEN("The python reference count is reduced") { CHECK(Py_REFCNT(pyTraitsData) == 1); }
-    }
```
elliotcmorris: [Question] Ah was this the scope error? Does `AND_X` not cause the sections to be run sequentially? I guess not. Does that mean `AND` is just flavour, with no mechanical effect?

elliotcmorris (approved):
Great improvements. Thats Feltell quality for ya.

src/openassetio-python/tests/bridge/python/test_converter.cpp
```diff
@@ -0,0 +1,268 @@
...
+  GIVEN("A Python environment without openassetio loaded") {
+    REQUIRE_FALSE(py::module_::import("sys").attr("modules").contains("openassetio"));
+
+    AND_GIVEN("an OpenAssetIO C++ API object") {
+      const TraitsDataPtr traitsData = TraitsData::make();
+
+      WHEN("the C++ object is casted to a Python object") {
+        THEN("cast throws expected exception") {
+          REQUIRE_THROWS_WITH(openassetio::python::converter::castToPyObject(traitsData),
+                              std::string("Unregistered type : openassetio::v1::TraitsData"));
```
foundrytom, line 192: [consider] And below, any way we can avoid baking `v1` in there to reduce future maintenance?

foundrytom (changes requested):
Awesome stuffs, lovely implementation. Thanks for wrangling all the ins and outs on this one too - know it's been a bit of a mission.

Sorry, mostly just wordsmithing docs, and trying to make sure we capture a little more "working knowledge" for 6m down the line when our brains have moved onto other things. Please feel free to push back if you want.

src/openassetio-python/tests/bridge/python/test_converter.cpp
feltech: Yeah all those `THEN`, `AND_GIVEN` etc etc are just synonyms for `SECTION`, only difference being how they render when there's an error.

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
feltech: OK that's fair enough.  I updated this based on nomenclature in the [CPython docs](https://docs.python.org/3/extending/extending.html#ownership-rules).

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
```diff
@@ -0,0 +1,105 @@
...
+ * @throws std::runtime_error if the cast fails.
+ *
+ * @param objectPtr An OpenAssetIO pointer type, (eg, @needsref
+ * ManagerPtr). The returned pointer takes shared ownership of this
+ * input \p objectPtr, and will keep the C++ instance alive until the
+ * `PyObject` is destroyed.
+ *
+ * @return A `PyObject` pointer to an object of the Python API type
+ * associated with the C++ API object provided. The reference count of
+ * the `PyObject` will be incremented, and ownership of the reference
```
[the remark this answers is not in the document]
feltech: Another case where I was just following the nomenclature in the [CPython docs](https://docs.python.org/3/extending/extending.html#ownership-rules), but happy to be more explicit than that.

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
elliotcmorris: My thinking is that the context is different for us, in the CPython docs everything inherently must be DECREF'd so you don't mention it all the time, but here it's an exception, so we should mention it.

src/openassetio-python/tests/bridge/python/test_converter.cpp
elliotcmorris: I don't believe so, this is on the same level of scope as the incref, meaning it gets run after every test-leaf, as far as my understanding goes.

src/openassetio-python/bridge/src/python/converter.cpp
feltech: I think pybind11 casting away constness is a bit eyebrow-raising.  I think allowing a cast from a `ConstPtr` to a non-const `PyObject*`, effectively providing a backdoor way to bypass constness, possibly without realising it, would just be confusing.

foundrytom: Cool. I take it its a compile error if you you try to cast a ConstPtr to py? Be good to note this situation in the docs - seems useful.

src/openassetio-python/tests/bridge/python/test_converter.cpp
foundrytom: Sorry misread the indents there, stands for cases below though I think.

src/openassetio-python/tests/bridge/python/test_converter.cpp
foundrytom: FWIW, is it worth just using the pybind object to save this kind of thing even coming up?

elliotcmorris: Yeah, although not technically a different call-order here because there's only one case, this'll break if someone adds another leaf without thinking.

> is it worth just using the pybind object to save this kind of thing even coming up?
I thought, and still think, that it's worth dogfooding the way people will need to interact with our methods when we write our tests, otherwise we don't have any context about the burdens of using these interfaces as we provide them. Since the whole point is to not have pybind, I think it's good that we put ourselves through it.

On the other hand, one could argue test stability is more important than my abstract gesture at building design-sense...

foundrytom: > Since the whole point is to not have pybind,

Don't know if its relevant, but it's perhaps more to hide _our_ use of pybind. I'd imagine most calling contexts will use some management layer (pybind or other), so in a way having the tests use pybind and having to interact with our raw CPython object is actually more representative. I think most people avoid raw CPython where they can due to 🦶 🔫 🔫 🔫

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
```diff
@@ -0,0 +1,105 @@
...
+ * without requiring/exposing a particular CPython binding library
+ * implementation.
+ *
+ * The returned `PyObject` owns a hidden reference to the input
+ * `shared_ptr`, ensuring it is kept alive. The reference will be
+ * released when the `PyObject` is destroyed.
+ *
+ * @note This function does not acquire the GIL.
+ *
+ * @warning A Python environment, with openassetio imported, must be
```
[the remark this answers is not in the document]
elliotcmorris: d6aebae94bbc53be47a0457b5534147516b9b671

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
```diff
@@ -0,0 +1,105 @@
...
+ *          available in order to use this function.
+ *
+ * @throws std::invalid_argument if the input is null.
+ * @throws std::runtime_error if the cast fails.
+ *
+ * @param objectPtr An OpenAssetIO pointer type, (eg, @needsref
+ * ManagerPtr). The returned pointer takes shared ownership of this
```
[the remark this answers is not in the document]
elliotcmorris: Yeah, whilst it is a `pyobject` ptr, python object is more descriptive I think.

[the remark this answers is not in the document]
elliotcmorris: d99b4d87dd76294b7d89b8412823bc4ab024a0c8

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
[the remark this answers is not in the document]
elliotcmorris: Gone ahead and taken the suggestion to be more explicit. Other discussion (https://github.com/OpenAssetIO/OpenAssetIO/pull/929#discussion_r1197644984)

8d7ee1e526876611cf4f042afa49e21b1d8b4785

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * The returned `shared_ptr` owns a hidden reference to the input
+ * `PyObject`, ensuring it is kept alive. The reference will be released
+ * when the `shared_ptr` is destroyed.
+ *
+ * Specifically, this function will increment the reference count of the
+ * provided Python object for the lifetime of the returned C++ object.
+ * When the returned C++ object falls out of scope or is otherwise
+ * cleaned up, the Python object will have its reference count reduced
+ * by one, potentially invoking cleanup.
```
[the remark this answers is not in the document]
elliotcmorris: I think it matters if you're concered with looking at the reference count value specicially, is "owning a hidden reference" enough to tell you that the reference count has been incremented? ... probably yes if you think about it

Will come back to this one in a hot minute, but I think you're right, and will likely remove this paragraph.

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
```diff
@@ -0,0 +1,105 @@
...
+ * @endcode
+ *
+ *
+ * If the types of the template argument and the \p pyObject are not
+ * equivalent, an exception will be thrown due to inability to perform
+ * the cast.
+ *
+ * @note This function does not acquire the GIL.
+ *
+ * @warning A Python environment, with openassetio imported, must be
```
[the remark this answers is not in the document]
elliotcmorris: 171563ca05b8a2f3f651d1e0921a471af1235f1d

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @throws std::invalid_argument if the function fails due to inability
+ * to cast between types, or if the input is null.
+ *
+ * @param pyObject A `PyObject` pointer to a Python object that must be
+ * of equivalent type to the template argument.
+ *
+ * @return An OpenAssetIO pointer to a C++ API object cast from the
+ * provided \p pyObject. The lifetime of the PyObject will be extended
+ * to at least the lifetime of the returned C++ object.
```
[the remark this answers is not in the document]
elliotcmorris: This seems contray to a prior suggestion.

In that, when referring to a `PyObject*`, we call it `Python Object`
So, when referring to a `std::shared_Ptr<T>`, should we not call it a `C++ Object`?

src/openassetio-python/tests/bridge/python/test_converter.cpp
elliotcmorris: Did a thing where the Stringified namespace macro is stripped from the expected string, I think it's pretty neat.

https://github.com/OpenAssetIO/OpenAssetIO/pull/929/commits/f2604f7d8ce1ca79fd2bfd5406e0f2425ae266b8

src/openassetio-python/tests/bridge/CMakeLists.txt
elliotcmorris: Sure can.

b872947d3c3c394c8eb97005fa7093af3ac269b5

src/openassetio-python/bridge/src/python/converter.cpp
```diff
@@ -0,0 +1,105 @@
...
+template <typename T>
+PyObject* castToPyObject(const T& objectPtr) {
+  // Stash the errors for restoring after cast.
+  // An error_scope calls PyErr_Fetch (which clears the active error),
+  // and stashes them, restoring the error state when it falls out of
+  // scope
+  const py::error_scope previousErrorState;
+
+  if (objectPtr == nullptr) {
+    throw std::invalid_argument("objectPtr cannot be null");
```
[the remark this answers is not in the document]
elliotcmorris: You mean the attempting to cast phrasing? Sure. Also added the method signiature in the message to your point about lack of stack trace.

I don't think we can get a string of the template type easily though, if that's what you were gesturing at.

1932c26216446e7434b8ddb3ca6530499bd5a161

src/openassetio-python/tests/bridge/python/test_converter.cpp
foundrytom: Oh clever! I'll have to defer to @feltech as to the cunning involved :)

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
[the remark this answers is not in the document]
foundrytom: For me it was just that "C++ object" suggests that it might be tied to the specific instance, whereas saying `shared_ptr` hopefully conveys that the lifetime is based on shared pointer semantics (ie, including all subsequent copies).

src/openassetio-python/bridge/src/python/converter.cpp
elliotcmorris: 72bf6bbc0faa10c76baf4052c50dd44f4acff929

RELEASE_NOTES.md
```diff
@@ -29,6 +29,13 @@ v1.0.0-alpha.xx
   `ResolveSuccessCallback` from reference types to value types.
   [#858](https://github.com/OpenAssetIO/OpenAssetIO/issues/858)
 
+### New Features
+
+- Added utility methods `castToPyObject` and `castFromPyObject` to
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/929/commits/f80379e0050cccafdfcedf2e7382ef7b8873caa8

src/openassetio-python/bridge/src/python/converter.cpp
elliotcmorris: I'm not actually sure, I took this from the other method that uses `PyRetainingSharedPtr`.

From my breif pybind foray, I think we could replace this with `std::cast<py::object>(pyObject);` and it would do the same thing, all the tests pass if you do.

@feltech What's your thoughts?

src/openassetio-python/bridge/include/openassetio/python/converter.hpp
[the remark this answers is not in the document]
elliotcmorris: Deleted the paragraph, but also changed `owns` to `borrows` in the prior one, which I think is more technically correct, and I think is the thing that was niggling me to write the expounding text in the first place, as `owns` didn't seem to tell the whole story.

cd3222ae480128e7d7561b1e34ad33978ac73431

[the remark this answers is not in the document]
foundrytom: Great thanks - are we sure it borrows one though? The shared pointer concretely own an additional ref on the python object for its lifetime (ie, it doesnt affect anyone elses ref, and it lives as long as the shared ptr (or copies of it) does).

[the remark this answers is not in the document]
elliotcmorris: 67a4e862f66b94a0a718bb9b1cb2c766656a4ab6

Undid to `owns` to `borrows` change after our discussion

src/openassetio-python/tests/bridge/python/test_converter.cpp
feltech: Another alternative would be to build the expected value, rather than strip the actual value.  I.e.

`"Unregistered type : openassetio::" STRINGIFY(OPENASSETIO_CORE_ABI_VERSION) "::TraitsData"`

src/openassetio-python/tests/bridge/python/test_converter.cpp
elliotcmorris: 3e2f4b05e9126bbb58a95b61593fdd2a7fa0412b

As per our discussion, simply fixed the existing formulation via moving the DECREF up a scope.

src/openassetio-python/tests/bridge/python/test_converter.cpp
```diff
@@ -0,0 +1,305 @@
...
+    }
+    Py_DECREF(pyTraitsData);
+  }
+}
+
+SCENARIO("Casting to a C++ object binds object lifetime") {
+  GIVEN("A python object") {
+    py::module_::import("openassetio");
+
+    // Use pybind to conveniently create a released CPython object.
```
foundrytom, line 150: [consider]
```suggestion
    // Use pybind to conveniently create a CPython object with a ref count of 1.
```
Sorry to be a total pedant, but is 'released' confusing here? To me that wound be one with no ref count. Whereas its deliberately making one with an inc ref. As I understood it, "release" is the (imho confusingly named) method of `py::object` rather than a CPython thing - that abandons the underlying pyobject it previously increfd, which forms a 'trick' to give us a `PyObject` with a non-zero ref.

src/openassetio-python/bridge/src/python/converter.cpp
feltech: One option that could sidestep this issue is to replace with
```
const auto pyInstance = py::handle(pyObject);
```

src/openassetio-python/tests/bridge/python/test_converter.cpp
elliotcmorris: Oh duh that's obviously better. b96ae62ea3cb732b190d559812cc8c88b67e45f6

(I did think about regex but std::regex has a deserved poor reputation so I try to avoid it.)

foundrytom: LEBTM 👍  (@feltech feel free to close thread when happy)

src/openassetio-python/bridge/src/python/converter.cpp
elliotcmorris: All that said, is there anything to say here? We're not really making a choice to borrow, we're just making a choice not to steal, which is the default choice. If this was written as `py::cast<py::object>` would we be having this discussion?

foundrytom: fine to leave it if its the obvious choice

src/openassetio-python/tests/bridge/python/test_converter.cpp
elliotcmorris: acae406ef82b7f0342494d6bb37d19d12177ef07

Yeah, was just trying to convey how "pybind is done, CPython is king now". Have accepted your suggestion in the spirit of getting the thing across the line, it's more technically descriptive anyhow.

foundrytom: 🙏 thanks for the patience here!

src/openassetio-python/bridge/src/python/converter.cpp
foundrytom: Happy with whatever you folks decide, closing this from a docs perspective.

foundrytom (approved):

feltech:
I can't add a review as an author, but I've gone through all the changes and they look good to me.

Issue #124, opened 2021-12-03
Prefix glossary refs in Doxygen comments


feltech:
# What

Prefix the cross-reference labels of glossary page items with some token and adjust the ref text case to match the glossary title, e.g. `attributes` -> `gloss:Attributes`.

# Why

* When reading the documentation in the source it would be useful to know at a glance where a ref points to, in particular whether it points to the glossary or some class/method/variable.
* When writing the documentation it would be useful to have the case of the ref match the case of the rendered output, to help visualise how the paragraph will look.

feltech:
Some of this was done as part of #1037 by adding a `glossary_` prefix to disambiguate.

Looking at the `@section`s in `Glossary.dox`, there are still some ambiguous reference names. I.e. `Context`, `HostInterface`, `ManagerInterface`.

In general, I think we would be well served by prefixing _all_ `@section` references with `glossary_`, so its easy to tell when the glossary is being mentioned.

Gmaj9:
I would like to have a go at this for ASWF Dev Days on 9/25.

marivenkatakrishnan:
is it still open? i would like to contribute.

Gmaj9:
> is it still open? i would like to contribute.

Yes, I did not finish this. I can provide what I did complete.

marivenkatakrishnan:
> Yes, I did not finish this. I can provide what I did complete.

sure, i will enhance from where you left

Gmaj9:
@marivenkatakrishnan I just opened a PR with what I completed, alongside the regex I used to locate references.

Gmaj9:
@marivenkatakrishnan my PR partially resolving this issue has been merged. You are free to pick up where I left off.

PR #931, opened 2023-05-19
[Docs] Docs detailing python cast exceptions

elliotcmorris:
We forgot to add these, wanna do it before the release.

Some methods are implemented in python, temporarily, and this means that casted PyObjects wont have access to those methods. Add docs to describe this niggle, even though this problem will disappear soon as these methods get ported to c++.

## Description

Closes # (issue)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -34,6 +34,10 @@ v1.0.0-alpha.xx
 - Added utility methods `castToPyObject` and `castFromPyObject` to
   `openassetio-python-bridge` to facilitate converting between C++ and
   Python objects for hosts seeking to support mixed language workflows.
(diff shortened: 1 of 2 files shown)


foundrytom (changes requested):
Thanks @elliotcmorris looks good aside from the broken ref.

elliotcmorris:
Annoying, linked locally, but all the `Manager` types are lowercased when @ref'd throughout the codebase so I guess that's the thing, (ManagerInterface is capital everywhere so beats me why :shrug: )

foundrytom:
> Annoying, linked locally, but all the `Manager` types are lowercased when @ ref'd throughout the codebase so I guess that's the thing, (ManagerInterface is capital everywhere so beats me why 🤷 )

lowercase will prob take you to the glossary entry (see #124) vs the class

PR #932, opened 2023-05-19
Move `getRelated` refactor release notes into feature-specific section

foundrytom:
This hopefully helps avoid accidentally merging feature branch changes into older releases, if a release happens on main whilst the branch is alive.

Experimenting with a `feat!` commit prefix that indicates that we need to massage the commit history before merging to `main` what do people thing?

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,23 @@
 Release Notes
 =============
 
+feature/getRelatedReferencesRefactor
+------------------------------------
+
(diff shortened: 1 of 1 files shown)


feltech (approved):
LGTM

elliotcmorris (approved):
Yeah I like it, especially the commit prefix, makes it obvious.

Means it'd behoove us to make sure we initially do the release notes in a separate commit. in all feature branches, just something to remember

PR #935, opened 2023-05-22
Migrate `getWithRelationship[s]` to callbacks

### LINKED ISSUE #919: Migrate relationship methods to callbacks
## What

Following on from #847, migrate the resulting python implementation to callbacks.

## Why

Consistency with other methods

foundrytom:
Still Python only, but should now be the final signature.

Fixes a few typos, and has to disable the `pylint` similarity checker due to  false positives.

Closes #919

### DIFF
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -592,12 +592,12 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * @param successCallback Callback that will be called for each
    * successful resolution of an entity reference. It will be
    * given the corresponding index of the entity reference in
-   * `entityRefs` along with its `TraitsData`. The callback will be
-   * called on the same thread that initiated the call to `resolve`.
+   * `entityReferences` along with its `TraitsData`. The callback will
+   * be called on the same thread that initiated the call to `resolve`.
    *
    * @param errorCallback Callback that will be called for each
    * failed resolution of an entity reference. It will be given the
-   * corresponding index of the entity reference in `entityRefs`
+   * corresponding index of the entity reference in `entityReferences`
    * along with a populated @fqref{BatchElementError}
    * "BatchElementError" (see @fqref{BatchElementError.ErrorCode}
    * "ErrorCodes"). The callback will be called on the same thread
@@ -901,7 +901,7 @@ class OPENASSETIO_CORE_EXPORT Manager {
    *
    * @param errorCallback Callback that will be called for each
    * failed preflight of an entity reference. It will be given the
-   * corresponding index of the entity reference in `entityRefs`
+   * corresponding index of the entity reference in `entityReferences`
    * along with a populated @fqref{BatchElementError}
    * "BatchElementError" (see @fqref{BatchElementError.ErrorCode}
    * "ErrorCodes"). The callback will be called on the same thread
@@ -1147,7 +1147,7 @@ class OPENASSETIO_CORE_EXPORT Manager {
    *
    * @param errorCallback Callback that will be called for each
    * failed registration of an entity reference. It will be given the
-   * corresponding index of the entity reference in `entityRefs`
+   * corresponding index of the entity reference in `entityReferences`
    * along with a populated @fqref{BatchElementError}
    * "BatchElementError" (see @fqref{BatchElementError.ErrorCode}
    * "ErrorCodes"). The callback will be called on the same thread
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -249,7 +249,7 @@ class OPENASSETIO_CORE_EXPORT Manager {
    *
    * @param context The calling context.
    *
-   * @return a `TraitsData` for each element in `traitSets`.
+   * @return a `TraitsData` for each element in @p traitSets.
    */
   [[nodiscard]] trait::TraitsDatas managementPolicy(const trait::TraitSets& traitSets,
                                                     const ContextConstPtr& context) const;
@@ -592,12 +592,12 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * @param successCallback Callback that will be called for each
    * successful resolution of an entity reference. It will be
    * given the corresponding index of the entity reference in
-   * `entityReferences` along with its `TraitsData`. The callback will
+   * @p entityReferences along with its `TraitsData`. The callback will
    * be called on the same thread that initiated the call to `resolve`.
    *
    * @param errorCallback Callback that will be called for each
    * failed resolution of an entity reference. It will be given the
-   * corresponding index of the entity reference in `entityReferences`
+   * corresponding index of the entity reference in @p entityReferences
    * along with a populated @fqref{BatchElementError}
    * "BatchElementError" (see @fqref{BatchElementError.ErrorCode}
    * "ErrorCodes"). The callback will be called on the same thread
@@ -894,14 +894,14 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * @param successCallback Callback that will be called for each
    * successful preflight of an entity reference. It will be given
    * the corresponding index of the entity reference in
-   * `targetEntityRefs` along with an updated reference to use for
+   * @p entityReferences along with an updated reference to use for
    * future interactions as part of the publishing operation. The
    * callback will be called on the same thread that initiated the
    * call to `preflight`.
    *
    * @param errorCallback Callback that will be called for each
    * failed preflight of an entity reference. It will be given the
-   * corresponding index of the entity reference in `entityReferences`
+   * corresponding index of the entity reference in @p entityReferences
    * along with a populated @fqref{BatchElementError}
    * "BatchElementError" (see @fqref{BatchElementError.ErrorCode}
    * "ErrorCodes"). The callback will be called on the same thread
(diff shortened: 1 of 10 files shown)


src/openassetio-python/package/openassetio/hostApi/Manager.py
```diff
@@ -430,6 +436,23 @@ def getWithRelationship(
 
         @param context Context The calling context.
 
+        @param successCallback Callback that will be called for each
+        successful relationship query. It will be given the
+        corresponding index of the entity reference in
+        `entityReferences` along with a list of entity references for
+        entities that have the specified relationship. If there are no
+        relations, an empty list will be supplied. The callback will be
```
elliotcmorris, line 444: [Consider][Minor]

Maybe just a quirk I have, but I continually find the mapping between the concept of a `Relationship` and the argument `relationshipTraitsData` awkward, as though the argument is saying "These are the traits datas that belong to, that define the relationship, but they are not the relationship", in the same way you might talk about "BookPages" and "Book" as different but essentially the same thing.

Long story short, maybe this ?

```suggestion
        `entityReferences` along with a list of entity references for
        entities that have the relationship specified by \p relationshipTraitsData
        If there are no relations, an empty list will be supplied. The callback will be
```

src/openassetio-python/package/openassetio/hostApi/Manager.py
```diff
@@ -430,6 +436,23 @@ def getWithRelationship(
...
+        called on the same thread that initiated the call to
+        `getWithRelationship`.
+
+        @param errorCallback Callback that will be called for each
+        failed relationship query. It will be given the
+        corresponding index of the entity reference in `entityReferences`
```
elliotcmorris, line 450: [Consider][Minor]
Suggesting that we do this to document parameters. It will have no real changes, as all `\p` is is an alias for `\c`, which just displays typewriter font. However this does document that this intends to refer to a parameter rather then just an abstract variable.
```suggestion
        corresponding index of the entity reference in \p entityReferences
```

src/openassetio-python/package/openassetio/hostApi/Manager.py
```diff
@@ -480,6 +511,23 @@ def getWithRelationships(
 
         @param context Context The calling context.
 
+        @param successCallback Callback that will be called for each
+        successful relationship query. It will be
+        given the corresponding index of the relationship in
+        `relationshipTraitsDatas` along with a list of entity references
+        for entities that have the specified relationship. If there
+        are no relations, an empty list will be supplied. The callback
```
elliotcmorris, line 519: [Consider]
`an empty list will be supplied` seems ambiguous, you could mean the `relationshipTraitsDatas` arg has a default argument that is an empty list, or that the callback will receive an empty list as the argument.

src/openassetio-python/package/openassetio/managerApi/ManagerInterface.py
```diff
@@ -513,6 +520,24 @@ def getWithRelationship(
         logging and provides access to the openassetio.managerApi.Host
         object representing the process that initiated the API session.
 
+        @param successCallback Callback that must be called for each
+        successful relationship query for an input entity reference. It
+        should be given the corresponding index of the entity reference
+        in `entityReferences` along with a list of entity references for
+        entities that are related by the input entity by the specified
+        relationship. If there are no relations, an empty list should be
+        used. The callback must be called on the same thread that
```
elliotcmorris, line 529: [Consider]
Similarly ambiguous to the docs in manager. Perhaps?

```suggestion
        relationship. If there are no relations, an empty list should be
        provided to the callback. The callback must be called on the same thread that
```

src/openassetio-python/package/openassetio/managerApi/ManagerInterface.py
```diff
@@ -527,13 +552,21 @@ def getWithRelationship(
...
+        hostSession,
+        successCallback,
+        errorCallback,
+        resultTraitSet=None,
     ):
         """
-        Returns entity references that are related to the input
+        Queries entity references that are related to the input
         reference by the relationships defined by a set of traits and
         their properties.
```
elliotcmorris, line 571: [Consider]
Sort of harping on the same point, but I wonder if defining the term `relationship` here to map to `relationshipTraitsDatas would help some readers apply the concept to the signature in the docs that come later in each method.
```suggestion
        Queries entity references that are related to the input
        reference by the relationships defined by a set of traits
        (\p relationshipTraitsDatas) and their properties.
```

N.B : is `and their properties` saying anything of value here?

src/openassetio-python/pyproject.toml
```diff
@@ -84,6 +84,18 @@ disable = [
...
+# Presently it's not possible to disable the `duplicate-code` check for
+# specific files, so it thinks that symlinks are duplicated
+# implementation. We also have issues where multi-line calls are
+# classed as duplicates.
+#
+# See:
+#     https://github.com/PyCQA/pylint/issues/214
+#
+# As we're not going to have a lot of python
+# inmplementation anyway, use this trick to disable the similarity
```
elliotcmorris, line 96: [fix][minor]
```suggestion
# implementation anyway, use this trick to disable the similarity
```

elliotcmorris (changes requested):
Entirely doc stuff.

I wonder if we should have some of these tests in the api-compliance suite now? Even though related references isn't a required function, if it is implemented stuff like equal/empty lists getting routed to the callback can be checked. Maybe a task for later.

src/openassetio-python/package/openassetio/hostApi/Manager.py
feltech: Like it, but will `@p` work instead?  Bit more consistent

feltech (approved):
LGTM

src/openassetio-python/package/openassetio/managerApi/ManagerInterface.py
foundrytom: It was attempting to highlight that its a TraitsData - that can hold property values (vs a trait set), and that those property values influence the result of the query.

src/openassetio-python/package/openassetio/hostApi/Manager.py
foundrytom: Ill update these more generally (as we use `name` in other methods) once we have sign-off on the copy.

src/openassetio-python/pyproject.toml
foundrytom: Doh! [thanks](https://github.com/OpenAssetIO/OpenAssetIO/pull/935/commits/db3f6b446cac89d2da26ec450c2fed162ec4c8d2).

src/openassetio-python/package/openassetio/hostApi/Manager.py
foundrytom: Added, along with a more general reshuffle [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/935/commits/d575d22518fb6a9167edd5f72ba3c57a77b8e4ea).

src/openassetio-python/package/openassetio/managerApi/ManagerInterface.py
foundrytom: Expanded on the topic [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/935/commits/d575d22518fb6a9167edd5f72ba3c57a77b8e4ea).

src/openassetio-python/package/openassetio/hostApi/Manager.py
foundrytom: Applied as diff, but then reworded as part of [this](https://github.com/OpenAssetIO/OpenAssetIO/pull/935/commits/d575d22518fb6a9167edd5f72ba3c57a77b8e4ea).

elliotcmorris: :+1:

src/openassetio-python/package/openassetio/managerApi/ManagerInterface.py
elliotcmorris: Ty. I think just mentioning what argument we're talking about will help a lot.

elliotcmorris (approved):

src/openassetio-python/package/openassetio/hostApi/Manager.py
elliotcmorris: Yeah, it's minor, (it would be better if it linked,) using `@p` is additional documentation language that makes it clearer that the following token is meant to refer to a parameter of the method in question, which I think is worth doing.

feltech: Yeah I agree, assuming the read knows what @p means, then it's really useful to know that a term is a parameter.

foundrytom: Changed all param references in `Manager` and `ManagerInterface` docs. We may have others in other files. https://github.com/OpenAssetIO/OpenAssetIO/pull/935/commits/2934bbaa991bf1c5f6a7c929799af7443b81c121

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -249,7 +249,7 @@ class OPENASSETIO_CORE_EXPORT Manager {
    *
    * @param context The calling context.
    *
-   * @return a `TraitsData` for each element in `traitSets`.
+   * @return a `TraitsData` for each element in @p traitSets.
```
feltech, line 252: [question] Is it only the Manager docs that need `@p`-ing?  Or is it just that this change is deliberately limited to Manager docs?  I looked through a few others and didn't see anywhere that needs `@p`s.

feltech (approved):
Wow, gone all-in on `@p`, at least for Manager docs, nice.  Checked a sample in the rendered output, LGTM.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Ah OK, already answered, https://github.com/OpenAssetIO/OpenAssetIO/pull/935#discussion_r1205769432

> We may have others in other files.

foundrytom: I was just trying to avoid scope-creeping this PR. It felt odd to have  a mix of conventions within these files, so hoped sorting out the manager classes to at least be consistent was a reasonable compromise for now?

PR #938, opened 2023-05-24
[Core] Add optional `configPath` argument to `defaultManagerForInterface`

### LINKED ISSUE #937: Allow custom config path in `defaultManagerForInterface`
## What

Allow a host to specify the TOML file used when constructing the default manager via `ManagerFactory`

## Why

There are many scenarios where end-users do not have ready means of configuring the process environment (eg: double-clicking on an icon), and a host may way to provide alternate means of determining the path to the default config file.

## Acceptance Criteria

- `deafultManagerForInterface(configPath, hostInterface, managerImplementationFactory, logger)`
  - The host supplies the path to the toml file
  - The supplied file is missing (or path is empty), then a `runtime_error` is raised.
  - Logging is otherwise as-per the existing implementation.

feltech:
## Description

Closes #937.

Add a new overload of `defaultManagerForInterface` that uses a `configPath` argument, rather than an environment variable, to determine the location of the TOML config file.

### DIFF
--- a/src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
@@ -3,6 +3,7 @@
 #pragma once
 
 #include <memory>
+#include <string_view>
 #include <unordered_map>
 #include <vector>
 
@@ -183,24 +184,16 @@ class OPENASSETIO_CORE_EXPORT ManagerFactory final {
    * the TOML configuration file referenced by the
    * @ref default_config_var.
    *
-   * This allows deployments to centralize OpenAssetIO manager settings,
-   * and for hosts to instantiate this manager without the need for their
-   * own settings and persistence mechanism.
-   *
    * @note This mechanism should be the default approach for a host to
-   * initialize the API. Extended functionality to override this configuration
-   * can optionally be provided, but the ability to use the shared, default
-   * configuration is always required.
+   * initialize the API. Extended functionality to override this
+   * configuration can optionally be provided, but the ability to use
+   * the shared, default configuration is always required.
    *
-   * The referenced TOML file should have the following structure.
-   *
-   * @code{.toml}
-   * [manager]
-   * identifier = "some.identifier"
-   *
-   * [manager.settings]  # Optional
-   * some_setting = "value"
-   * @endcode
+   * @see @ref defaultManagerForInterface(std::string_view, <!--
+   * -->const HostInterfacePtr&,<!--
+   * -->const ManagerImplementationFactoryInterfacePtr&,<!--
+   * -->const log::LoggerInterfacePtr&) "Alternative direct signature"
+   * for more details.
    *
    * @envvar **OPENASSETIO_DEFAULT_CONFIG** *str* The path to a
    * TOML file containing configuration information for the default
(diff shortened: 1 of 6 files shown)


src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
```diff
@@ -219,6 +220,55 @@ class OPENASSETIO_CORE_EXPORT ManagerFactory final {
...
+   * the given TOML configuration file.
+   *
+   * This allows deployments to centralize OpenAssetIO manager settings,
+   * and for hosts to instantiate this manager without the need for
+   * their own settings and persistence mechanism.
+   *
+   * @note This mechanism should be the default approach for a host to
+   * initialize the API. Extended functionality to override this
+   * configuration can optionally be provided, but the ability to use
+   * the shared, default configuration is always required.
```
foundrytom, line 234: [fix] Be good to remove this from this overload - as the intention of the env var one being the default  mechanism means that no additional configuration is required in supported applications. As this requires an explicit path, that may well be different between tools, it creates more potential for divergence/tool-specific setup.

Can we also de-dupe the formatting info, remove it from the env var one, and replace with a reference to this, otherwise I recon we'll end up with it being out of sync over time.

src/openassetio-core/src/hostApi/ManagerFactory.cpp
```diff
@@ -91,6 +91,14 @@ ManagerPtr ManagerFactory::defaultManagerForInterface(
...
   }
 
+  return defaultManagerForInterface(configPath, hostInterface, managerImplementationFactory,
+                                    logger);
+}
+
+ManagerPtr ManagerFactory::defaultManagerForInterface(
+    const std::string_view configPath, const HostInterfacePtr& hostInterface,
+    const ManagerImplementationFactoryInterfacePtr& managerImplementationFactory,
+    const log::LoggerInterfacePtr& logger) {
```
foundrytom, line 101: [fix] Can't anchor a comment below, but we need to remove:

```
    msg += "' [" + kDefaultManagerConfigEnvVarName + "]";
```
As its no longer always from the env var. Maybe add something to the env var method to indicate that it was used?

foundrytom (changes requested):
Need to give it a proper test, but in the interim

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
feltech: De-deuped and `@ref`'d in 36b139a2

src/openassetio-core/src/hostApi/ManagerFactory.cpp
feltech: Whoops. Fixed and tweaked the logging in 08f2b1f0

foundrytom:
Not sure when the overflow started, but raised this issue to fix how hard the docs for this class has become to read: https://github.com/OpenAssetIO/OpenAssetIO/issues/940 No need to fix here.

elliotcmorris (approved):
Sorry this took forever, i got distracted trying to solve local environment bumps when testing it.

src/openassetio-python/tests/package/hostApi/test_managerfactory.py
```diff
@@ -404,19 +478,33 @@ def resources_dir():
...
-    monkeypatch.setenv(ManagerFactory.kDefaultManagerConfigEnvVarName, __file__)
+def invalid_manager_config(monkeypatch, use_env_var_for_config_file):
+    toml_path = __file__
+    if use_env_var_for_config_file:
+        monkeypatch.setenv(ManagerFactory.kDefaultManagerConfigEnvVarName, toml_path)
+    return toml_path
+
+
+@pytest.fixture
+def use_env_var_for_config_file():
```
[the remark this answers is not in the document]
feltech: Good shout, appears to be unnecessary here, since we're always overriding it anyway in the actual tests.

src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
```diff
@@ -219,6 +212,50 @@ class OPENASSETIO_CORE_EXPORT ManagerFactory final {
...
+   * @code{.toml}
+   * [manager]
+   * identifier = "some.identifier"
+   *
+   * [manager.settings]  # Optional
+   * some_setting = "value"
+   * @endcode
+   *
+   * @param configPath Path to the TOML config file, compatible with
+   * `std::ifstream::open`. Relative paths resolve to a
```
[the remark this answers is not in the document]
feltech: Added in f5c9b06e

src/openassetio-python/tests/package/hostApi/test_managerfactory.py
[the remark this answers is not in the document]
feltech: Removed in 332ad0bf

foundrytom (approved):

PR #939, opened 2023-05-24
[Core] Add $config_dir config TOML interpolation

### LINKED ISSUE #804: Config-relative paths in `$OPENASSETIO_DEFAULT_CONFIG`
## What

Allow manager settings in the `$OPENASSETIO_DEFAULT_CONFIG` TOML file to reference other files relative to the path of the TOML file itself.

## Why
It is common in test scenarios to create [portable, self-contained manager states using BAL](https://github.com/OpenAssetIO/OpenAssetIO/tree/main/examples/host/simpleResolver). 

At the moment, this requires relative paths in the TOML file, and so becomes working directory sensitive. This makes it fragile.

As a user, I would like to be able to specify the config in such a way it is not sensitive to the cwd or absolute path.

## Acceptance Criteria

- When present in string-typed TOML values, the following strings are replaced with the the TOML file's directory path:
    - `${config_dir}`
    - `$config_dir`
- The substituted directory path should not contain a trailing `/`(or `\`).
- Docs update with note that only these vars are supported.

## Out of scope

Env var expansion my be useful, but is out of scope for this work, which solely aims to facilitate config relative configurations.

feltech:
## Description

Closes #804.

It is common in test scenarios to create portable, self-contained manager states using BAL.

At the moment, this requires relative paths in the TOML file, and so becomes working directory sensitive. This makes it fragile.

So add support for the substitution in TOML string values of `$config_dir` or `${config_dir}` tokens with the absolute directory to the TOML config file.

This is currently done in the most basic way possible, by direct string find-and-replace, rather than a more complete solution supporting e.g. environment variable interpolation. Such improvement is for future work if/when the need ever arises.

### DIFF
--- a/src/openassetio-core/src/hostApi/ManagerFactory.cpp
+++ b/src/openassetio-core/src/hostApi/ManagerFactory.cpp
@@ -2,6 +2,7 @@
 // Copyright 2022 The Foundry Visionmongers Ltd
 #include <cstdlib>
 #include <filesystem>
+#include <string_view>
 
 #include <toml++/toml.h>
 
@@ -15,6 +16,10 @@
 #include <openassetio/managerApi/ManagerInterface.hpp>
 #include <openassetio/typedefs.hpp>
 #include "openassetio/InfoDictionary.hpp"
+
+namespace {
+constexpr std::string_view kConfigDirVar = "${config_dir}";
+}  // namespace
 
 namespace openassetio {
 inline namespace OPENASSETIO_CORE_ABI_VERSION {
@@ -123,6 +128,20 @@ ManagerPtr ManagerFactory::defaultManagerForInterface(
 
   const std::string_view identifier = config["manager"]["identifier"].value_or("");
 
+  // Function to substitute ${config_dir} with the absolute,
+  // canonicalised directory of the TOML config file.
+  const auto substituteConfigDir =
+      [configDir =
+           std::filesystem::canonical(configPath).parent_path().string()](std::string str) {
+        // Adapted from https://en.cppreference.com/w/cpp/string/basic_string/replace
+        for (std::string::size_type pos{};
+             (pos = str.find(kConfigDirVar, pos)) != std::string::npos;
+             pos += configDir.length()) {
+          str.replace(pos, kConfigDirVar.length(), configDir);
+        }
+        return str;
+      };
+
   InfoDictionary settings;
   if (toml::table* settingsTable = config["manager"]["settings"].as_table()) {
     // It'd be nice to use settingsTable::for_each, a lambda and
@@ -135,7 +154,7 @@ ManagerPtr ManagerFactory::defaultManagerForInterface(
       } else if (val.is_floating_point()) {
         settings.insert({Str{key}, val.as_floating_point()->get()});
       } else if (val.is_string()) {
-        settings.insert({Str{key}, val.as_string()->get()});
+        settings.insert({Str{key}, substituteConfigDir(val.as_string()->get())});
       } else if (val.is_boolean()) {
         settings.insert({Str{key}, val.as_boolean()->get()});
       } else {
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -11,6 +11,12 @@ v1.0.0-alpha.xx
   environment variable.
   [#937](https://github.com/OpenAssetIO/OpenAssetIO/issues/937)
 
+- Added support for `${config_dir}` interpolation within manager string
+  settings retrieved from the TOML config file used by
+  `defaultManagerForInterface`. This token expands to the absolute
+  directory of the TOML config file.
+  [#804](https://github.com/OpenAssetIO/OpenAssetIO/issues/804)
+
 v1.0.0-alpha.11
 ---------------
 
--- a/src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
@@ -195,6 +195,10 @@ class OPENASSETIO_CORE_EXPORT ManagerFactory final {
    * -->const log::LoggerInterfacePtr&) "Alternative direct signature"
    * for more details.
    *
+   * Any occurrences of `${config_dir}` within TOML string values will
+   * be substituted with the absolute path to the directory containing
+   * the TOML file, before being passed on to the manager settings.
+   *
    * @envvar **OPENASSETIO_DEFAULT_CONFIG** *str* The path to a
    * TOML file containing configuration information for the default
    * manager.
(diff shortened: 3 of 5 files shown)


foundrytom (approved):
Need to give it a proper test, but otherwise LGTM

src/openassetio-core/src/hostApi/ManagerFactory.cpp
```diff
@@ -109,6 +115,21 @@ ManagerPtr ManagerFactory::defaultManagerForInterface(
 
   const std::string_view identifier = config["manager"]["identifier"].value_or("");
 
+  // Function to substitute $config_dir or ${config_dir} with the
+  // absolute, canonicalised directory of the TOML config file.
+  const auto substituteConfigDir =
+      [configDir = std::filesystem::canonical(configPath).parent_path().string()](
+          std::string& str) -> std::string& {
```
elliotcmorris: [fix] As discussed offline, this by-ref return is actually a pessimisation, causing an additional copy.

I'd love to link the quick-bench benchmark i did on this, but as usual the online servers timeout on anything significant.
Therefore i'll put the test code below, as well as a screenshot of the result from my local instance.
<details>
  <summary>Test Comparisons</summary>
  
```cpp
#include <string>
#include <unordered_map>
#include <variant>

std::string& substituteConfigDirRef(std::string& str)
{
    return str;
}

std::string substituteConfigDirVal(std::string str)
{
  return str;
}


static void StringRefReturn(benchmark::State& state) {
  std::unordered_map<int, std::variant<bool, int, float, std::string>> infoDict;
  int counter = 0;
  std::string setting = "a string that isn't sso optimised";
  for (auto _ : state) {
    infoDict.insert({counter, substituteConfigDirRef(setting)});
    counter++;
    benchmark::DoNotOptimize(infoDict);
    benchmark::DoNotOptimize(substituteConfigDirRef);
  }
}
// Register the function as a benchmark
BENCHMARK(StringRefReturn);

static void StringRefReturnWithMove(benchmark::State& state) {
  std::unordered_map<int, std::variant<bool, int, float, std::string>> infoDict;
  int counter = 0;
  std::string setting = "a string that isn't sso optimised";
  for (auto _ : state) {
    infoDict.insert({counter, std::move(substituteConfigDirRef(setting))});
    counter++;
    benchmark::DoNotOptimize(infoDict);
    benchmark::DoNotOptimize(substituteConfigDirRef);
  }
}
// Register the function as a benchmark
BENCHMARK(StringRefReturnWithMove);

static void StringValueReturn(benchmark::State& state) {
  std::unordered_map<int, std::variant<bool, int, float, std::string>> infoDict;
  int counter = 0;
  std::string setting = "a string that isn't sso optimised";
  for (auto _ : state) {
    infoDict.insert({counter, substituteConfigDirVal(std::move(setting))});
    counter++;
    benchmark::DoNotOptimize(infoDict);
    benchmark::DoNotOptimize(substituteConfigDirVal);
  }
}
BENCHMARK(StringValueReturn);
```

</details>

image

feltech: Great find, tweaked in e465f308

elliotcmorris: :+1: 

(That .get() to r-value is super weird and uncomfortable, at a glance you'd think you were forgetting to move.)

src/openassetio-core/src/hostApi/ManagerFactory.cpp
```diff
@@ -16,6 +17,11 @@
 #include <openassetio/typedefs.hpp>
 #include "openassetio/InfoDictionary.hpp"
 
+namespace {
+constexpr std::string_view kConfigDirVarNonBraced = "$config_dir";
+constexpr std::string_view kConfigDirVarBraced = "${config_dir}";
```
elliotcmorris, line 22: [Consider] If it's as simple as just adding 2 more constexpr's, I'd argue for adding the capitalised versions of these too, it can be frustrating when you want to keep to your style and can't.

src/openassetio-core/src/hostApi/ManagerFactory.cpp
```diff
@@ -16,6 +17,11 @@
 #include <openassetio/typedefs.hpp>
 #include "openassetio/InfoDictionary.hpp"
 
+namespace {
+constexpr std::string_view kConfigDirVarNonBraced = "$config_dir";
```
elliotcmorris, line 21: [Consider] I wonder if having an unbraced version is wise with this simplistic approach. It's unlikely other folks will be running substitutions themselves over this file, as some sort of preprocessing step, and even more unlikely that they would be using a conflicting token, (like $config_director_file_path or something), but it's possible.

This is probably fine, just thought i'd mention.

elliotcmorris (approved):

src/openassetio-core/src/hostApi/ManagerFactory.cpp
feltech: I agree the prefix issue is real, and would require a more complete `find` to ensure the token is surrounded by a word boundary - probably at that point `std::regex` would be the way forward.

We did discuss the drawbacks of such a simple solution, with a view to fixing any issues only if/when someone cares enough.

To be fair, we didn't really consider the prefix issue, but it still seems like a very unlikely case.

I'm happy to look into to a more robust `std::regex` solution, though, if we think it's worthwhile?

foundrytom: > Don't follow what you mean with the bracing not being unique to us,

I may have misinterpreted that you were also concerned about the general use of other var expansion methods, as well as the prefix-collision issue

elliotcmorris: >concerned about the general use of other var expansion methods 

I was, but just in that it's a reason I could think of why folks might be putting `$token` formatted strings in this file.

> is it worth leaving the implementation as-is though

Emphatically yes, I regret mentioning this at all as I think it's just not worth the time of day we've given it.

foundrytom: Just to be on the safe side in terms of 'as is' I was thinking of keeping the implementation structure, just dropping the brace-less input (vs rewriting to be simpler). But thats probably code smell 🙃

feltech: Yeah I think we should get rid of the loop
```
for (const std::string_view& what : {kConfigDirVarNonBraced, kConfigDirVarBraced}) {
```
if there's only one element.

feltech: Remove support for `$config_var` in 6d1ac38b

feltech:
Rebased and squashed as e65bd2d4 - there were conflicts in the test module, in case anyone wants to double-check.

foundrytom (approved):

elliotcmorris (approved):

PR #942, opened 2023-05-25
Use ${config_dir} in simpleResolver test

### LINKED ISSUE #941: Update `simpleResolver` example to use `${config_dir}`
## What

Update the [config file](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/examples/host/simpleResolver/bal_animals_openassetio_config.toml) included in the simpleResolver example to make use of `${config_dir}` as provided by #939.

## Why

It is frequently used as a reference, so this will help raise awareness.

feltech:
## Description

Closes #941. 

Dogfood the new `${config_dir}` substitution in the simpleResolver test example config file.

### DIFF
--- a/examples/host/simpleResolver/test_simpleResolver.py
+++ b/examples/host/simpleResolver/test_simpleResolver.py
@@ -23,9 +23,9 @@ import pathlib
 import pytest
 import sys
 
-
 from openassetio import BatchElementError
 from openassetio.hostApi import ManagerFactory
+
 
 # pylint: disable=no-self-use
 # pylint: disable=invalid-name,redefined-outer-name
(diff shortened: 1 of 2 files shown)


examples/host/simpleResolver/test_simpleResolver.py
```diff
@@ -95,7 +95,14 @@ def execute_cli(*args):
...
     # We explicitly don't want an exception to be raised.
     # pylint: disable=subprocess-run-check
-    return subprocess.run(all_args, capture_output=True, encoding="utf-8", cwd=this_file.parent)
+    return subprocess.run(
+        all_args,
+        capture_output=True,
+        encoding="utf-8",
+        # Deliberately not same directory as the library JSON file, so
+        # relies on ${config_dir} interpolation.
+        cwd=this_file.parent.parent,
```
foundrytom, line 104: [question] It's a long shot, but do we definitely want to make assumptions that we're still in a structure that has the `simpleResolver` dir?  Eg, if someone had copied the contents of `examples/host/simpleResolver` to `E:\`

foundrytom: Or actually, do we even need to set this, as I think in CI, this is running from the repo root?

feltech: Good idea, switched to tmpdir in bcd89b5f.  I think its worth explicitly setting, since we don't want to risk false negative failures, and it's good to be explicit that we're testing this.

foundrytom (approved):
👌 TVMFFTSAMAYLWTCID

elliotcmorris (approved):
LGTM :+1:

PR #949, opened 2023-05-31
Re-enable BAL integration tests on rrr feature branch

foundrytom:
Targets the BAL feature branch with callback support to re-instate BAL integration tests.

### DIFF
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -66,13 +66,12 @@ jobs:
         with:
           repository: OpenAssetIO/OpenAssetIO-Manager-BAL
           path: external/BAL
+          ref: feature/relatedReferencesRefactor
 
-      # Disabled pending https://github.com/OpenAssetIO/OpenAssetIO-Manager-BAL/issues/34
-      #
-      # - name: Test BAL
-      #   run: |
-      #     python -m pip install ./external/BAL
-      #     python -m pytest -v external/BAL/tests
+      - name: Test BAL
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/integrations.yml
```diff
@@ -66,13 +66,12 @@ jobs:
         with:
           repository: OpenAssetIO/OpenAssetIO-Manager-BAL
           path: external/BAL
+          ref: feature/relatedReferencesRefactor
```
feltech, line 69: [consider] If I understand correctly, this `drop!` commit will not be dropped immediately, but will be dropped when the feature branch is merged to `main`?

In which case, when this commit is dropped, it will revert to disabling BAL tests.  

Shouldn't the commit be split into two, a normal commit to re-enable the tests, and a `drop!` commit to add the `ref:`?

feltech (approved):

.github/workflows/integrations.yml
foundrytom: Yeah - a great Q - I kind of left it as that to make sure we schemed on it. The one that disables CI is also a `drop!`.

feltech: The BAL integration test was disabled in f2947172, which isn't a drop! commit.

foundrytom:
Supersceeded by #951

.github/workflows/integrations.yml
foundrytom: Ahh yeah- sorry I thought it was - I might have combined it before after all

PR #952, opened 2023-06-05
[Core] Add apiComplianceSuite coverage for `getWithRelationship[s]`

### LINKED ISSUE #914: Add `getRelationship[s]` coverage to `apiComplianceSuite`

foundrytom:
Adds basic coverage for the relationship query methods. Note: this has deliberately been left in it's somewhat long-winded form, as it hints that we may well want to look at a bit of a refactor of the harness, now its being used at scale.

Closes #914

On-hold prending #968.

### DIFF
--- a/src/openassetio-python/package/openassetio/test/manager/apiComplianceSuite.py
+++ b/src/openassetio-python/package/openassetio/test/manager/apiComplianceSuite.py
@@ -513,6 +513,201 @@ class Test_register(FixtureAugmentedTestCase):
         self.assertEqual(actual_error.message, expected_error.message)
 
 
+class Test_getWithRelationship(FixtureAugmentedTestCase):
+    """
+    Check plugin's implementation of
+    managerApi.ManagerInterface.Test_getWithRelationship.
+    """
+
+    def test_when_relation_unknown_then_empty_list_returned(self):
+        a_ref = self.requireEntityReferenceFixture("a_reference", skipTestIfMissing=True)
+        an_unknown_rel = TraitsData({"🐠🐟🐠🐟"})
+        self.__testGetWithRelationship([a_ref], an_unknown_rel, [[]])
+
+    def test_when_multiple_references_then_same_number_of_returned_relationships(self):
+        a_ref = self.requireEntityReferenceFixture("a_reference", skipTestIfMissing=True)
+        a_rel = TraitsData(self.requireFixture("a_relationship_trait_set"))
+        expected = self.requireEntityReferencesFixture("expected_related_entity_references")
+        self.__testGetWithRelationship([a_ref] * 5, a_rel, [expected] * 5)
+
+    def test_when_relationship_trait_set_known_then_all_with_trait_set_returned(self):
+        a_ref = self.requireEntityReferenceFixture("a_reference", skipTestIfMissing=True)
+        a_rel = TraitsData(self.requireFixture("a_relationship_trait_set"))
+        expected = self.requireEntityReferencesFixture("expected_related_entity_references")
+        self.__testGetWithRelationship([a_ref], a_rel, [expected])
+
+    def test_when_relationship_trait_set_known_and_props_set_then_filtered_refs_returned(self):
+        a_ref = self.requireEntityReferenceFixture("a_reference", skipTestIfMissing=True)
+        a_rel = self.requireFixture("a_relationship_traits_data_with_props")
+        expected = self.requireEntityReferencesFixture("expected_related_entity_references")
+        self.__testGetWithRelationship([a_ref], a_rel, [expected])
+
+    def test_when_result_trait_set_supplied_then_filtered_refs_returned(self):
+        a_ref = self.requireEntityReferenceFixture("a_reference", skipTestIfMissing=True)
+        a_rel = TraitsData(self.requireFixture("a_relationship_trait_set"))
+        result_trait_set = self.requireFixture("an_entity_trait_set_to_filter_by")
+        expected = self.requireEntityReferencesFixture("expected_related_entity_references")
+        self.__testGetWithRelationship([a_ref], a_rel, [expected], result_trait_set)
+
+    def test_when_querying_missing_reference_then_resolution_error_is_returned(self):
+        self.__testGetWithRelationshipError("a_reference_to_a_missing_entity")
+
+    def test_when_querying_malformed_reference_then_malformed_reference_error_is_returned(self):
+        self.__testGetWithRelationshipError(
+            "a_malformed_reference",
+            errorCode=BatchElementError.ErrorCode.kMalformedEntityReference,
+        )
+
+    def __testGetWithRelationship(
+        self, references, relationship, expected_relations, resultTraitSet=None
+    ):
+        if resultTraitSet is None:
+            resultTraitSet = set()
+        context = self.createTestContext(access=Context.Access.kRead)
+        results = []
+
+        self._manager.getWithRelationship(
+            references,
+            relationship,
+            context,
+            lambda _idx, traits_data: results.append(traits_data),
+            lambda idx, batch_element_error: self.fail(
+                f"getWithRelationship should not error for: '{references[idx].toString()}': "
+                f"{batch_element_error.message}"
+            ),
+            resultTraitSet,
+        )
+
+        self.assertEqual(results, expected_relations)
+
+    def __testGetWithRelationshipError(
+        self,
+        fixture_name,
+        access=Context.Access.kRead,
+        errorCode=BatchElementError.ErrorCode.kEntityResolutionError,
+    ):
+        reference = self.requireEntityReferenceFixture(fixture_name, skipTestIfMissing=True)
+
+        expected_msg = self.requireFixture("expected_error_message")
+        expected_error = BatchElementError(errorCode, expected_msg)
+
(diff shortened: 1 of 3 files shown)


src/openassetio-python/package/openassetio/test/manager/apiComplianceSuite.py
```diff
@@ -513,6 +513,270 @@ def __testRegisterError(self, fixture_name, errorCode):
...
 
+class Test_getWithRelationship(FixtureAugmentedTestCase):
+    """
+    Check plugin's implementation of
+    managerApi.ManagerInterface.Test_getWithRelationship.
+    """
+
+    def setUp(self):
+        self.a_reference_to_a_readable_entity_with_relations = self._manager.createEntityReference(
+            self.requireFixture(
```
feltech, line 524: [question] I take it that the sheer volume of fixtures to cover the whole matrix is what you meant by being a bit long-winded?

src/openassetio-python/package/openassetio/test/manager/apiComplianceSuite.py
```diff
@@ -513,6 +513,270 @@ def __testRegisterError(self, fixture_name, errorCode):
...
+
+        results = []
+        self._manager.getWithRelationships(
+            [a_rel],
+            reference,
+            context,
+            lambda _idx, _refs: self.fail("Unexpected success callback"),
+            lambda _idx, batch_element_error: results.append(batch_element_error),
+        )
+        [actual_error] = results  # pylint: disable=unbalanced-tuple-unpacking
```
feltech, line 775: [consider] The exception raised if no callback was ever called isn't very informative
```
ValueError: not enough values to unpack (expected 1, got 0)
```
One idea: initialise to `results = ["callback not called"]`, so the error log generated by the `assertEqual` has some hint as to what happened?

Another idea: `self.assertEqual(len(results), 1, msg="Callback should have been called once")`

:shrug:

feltech: Hmm, quick experimentation finds the output isn't great, would take some experimentation to find the nicest assertion function.  E.g. output of `assertEqual` with a custom `msg`
```python
>>> class test(unittest.TestCase):
...   def test(self):
...     self.assertEqual(["a"], [], msg="Callback should have been called once")
... 
>>> test().test()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in test
  File "/usr/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib/python3.8/unittest/case.py", line 1118, in assertListEqual
    self.assertSequenceEqual(list1, list2, msg, seq_type=list)
  File "/usr/lib/python3.8/unittest/case.py", line 1100, in assertSequenceEqual
    self.fail(msg)
  File "/usr/lib/python3.8/unittest/case.py", line 753, in fail
    raise self.failureException(msg)
AssertionError: Lists differ: ['a'] != []

First list contains 1 additional elements.
First extra element 0:
'a'

- ['a']
+ [] : Callback should have been called once
```

foundrytom: Capturing off-line discussion - keeping these things out-of-scope at this point as it also affects other existing tests.

src/openassetio-python/package/openassetio/test/manager/apiComplianceSuite.py
foundrytom: Yeah, and the fragility of string literals etc...

foundrytom:
@elliotcmorris @feltech how about [this] for improved fixture names as per our discussion (didn't feel the need to split them into specific cases, the suite name seemed enough)?

foundrytom:
@feltech this should updated with new args order, but I didn't have the updated BAL to test against yet

foundrytom:
@feltech this has WIP BAL test updates to use these for e2e dev https://github.com/foundrytom/OpenAssetIO-Manager-BAL/commits/work/getWithRelationshipTests

feltech:
Tweaked the fixtures based on our meeting, in which, IIRC, we decided to try to remove `"shared"` fixtures where possible, to ease readability/discoverability.  See https://github.com/OpenAssetIO/OpenAssetIO-Manager-BAL/pull/50 for implementation.

src/openassetio-python/package/openassetio/test/manager/apiComplianceSuite.py
```diff
@@ -729,6 +731,8 @@ def test_when_querying_malformed_reference_then_malformed_reference_error_is_ret
     def __testGetWithRelationships(
         self, reference, relationships, expected_relations, resultTraitSet=None
     ):
+        if resultTraitSet is None:
+            resultTraitSet = set()
```
foundrytom, line 661: [question] Is this a pybind thing? ie. setting a kwarg to `None` is treated differently to not setting it at all?

foundrytom: [plan] I knew there was something in there somewhere. Not related to this PR per-se, but we should double check`py::arg("resultTraitSet") = trait::TraitSet{})` doesn't risk the same problems as if you did that in a `def`, and fix if so.

src/openassetio-python/package/openassetio/test/manager/harness.py
```diff
@@ -249,6 +249,32 @@ def assertValuesOfType(self, valueList, valueType, allowNone=False):
...
+        A convenience companion to requireFixture, to construct an
+        entity reference object around a raw string fixture.
+        """
+        return self._manager.createEntityReference(
+            self.requireFixture(
+                fixtureName,
+                skipTestIfMissing,
+            )
+        )
+
```
foundrytom, line 277: [fix] Good thinking, had this on the list for a while, but can we split this out into its own commit, add tests and release notes (as this is public API and people use these in their own suites).

foundrytom (commented):
Good call on the split our fixtures, hopefully it doesn't make the fixtures.py file too intimidating.

elliotcmorris (approved):
You're right about the labour of the compliance suite, these tests do have a certain eye-glazing effect.
Worked locally, testing against Davids BAL branch.

src/openassetio-python/package/openassetio/test/manager/apiComplianceSuite.py
feltech: Added https://github.com/OpenAssetIO/OpenAssetIO/issues/987 - but, as noted in the ticket, it looks like there's no problem.

src/openassetio-python/package/openassetio/test/manager/harness.py
feltech: I took the liberty of squashing and rebasing, so inserted a commit for this as 529df33c

foundrytom:
> You're right about the labour of the compliance suite, these tests do have a certain eye-glazing effect.
> Worked locally, testing against Davids BAL branch.

That has gotten a lot worse since we aren't using shared fixtures, but the benefit is the names make more sense.

feltech:
> > You're right about the labour of the compliance suite, these tests do have a certain eye-glazing effect.
> > Worked locally, testing against Davids BAL branch.
> 
> That has gotten a lot worse since we aren't using shared fixtures, but the benefit is the names make more sense.

I think expecting manager plugin authors to understand what the fixtures mean based on their name isn't likely to work. They'd end up looking in the test case code anyway.  So tying the fixtures more explicitly to the test case, so it's clear which values are used where, will help readability.

I'm not best placed to judge though, since I wrote this.  So if you two think it's easier for plugin authors to understand the shared fixtures, then maybe we should rethink this PR.

foundrytom:
> > > You're right about the labour of the compliance suite, these tests do have a certain eye-glazing effect.
> > > Worked locally, testing against Davids BAL branch.
> > 
> > 
> > That has gotten a lot worse since we aren't using shared fixtures, but the benefit is the names make more sense.
> 
> I think expecting manager plugin authors to understand what the fixtures mean based on their name isn't likely to work. They'd end up looking in the test case code anyway. So tying the fixtures more explicitly to the test case, so it's clear which values are used where, will help readability.
> 
> I'm not best placed to judge though, since I wrote this. So if you two think it's easier for plugin authors to understand the shared fixtures, then maybe we should rethink this PR.

Im happy with it as-is for now, and we can use the fixtures as they stand to start a discussion with @mattdaw @antirotor.

src/openassetio-python/tests/package/test/manager/test_harness.py
```diff
@@ -220,6 +221,93 @@ def test_when_called_with_iterable_then_passes(self, a_test_case):
...
+        mock_manager.createEntityReference.return_value = expected
+
+        try:
+            assert a_test_case.requireEntityReferenceFixture(required) == expected
+        except SkipTest:
+            pytest.fail("Test should not be skipped")
+
+        mock_manager.createEntityReference.assert_called_once_with("a")
+
+    def test_when_invalid_reference_then_exception_propagates(self, a_test_case, mock_manager):
```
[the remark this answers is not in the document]
foundrytom: Perfect, thanks!

foundrytom:
I can't add a review as its my PR. But LGTM 🤣

Issue #913, opened 2023-05-03
Migrate `getWithRelationship[s]` to C++


foundrytom:
# What

Add `Manger` /`ManagerInterface` methods `getWithRelationship` / `getWithRelationships` to C++ and replace pure Python implementation with C++ bindings.

# Why

This is required so that C++ hosts can call these methods, and so that C++ managers can implement and override them.

# ACs

* Polymorphic methods are declared and their default implementation defined in C++ rather than Python.
* C++/Python hosts can call method implementations provided by Python/C++ manager plugins.

# Implementation notes

* Existing Python unit tests should still pass with minimal modification. Explicit C++ tests not required.
* BAL should continue to work without modification - issues added for any minor tweaks required.

feltech:
With https://github.com/OpenAssetIO/OpenAssetIO-Manager-BAL/pull/48 now merged, I think this can be closed.

PR #968, opened 2023-06-08
Migrate getWithRelationship[s] to C++

feltech:
## Description

Part of #913

## Reviewer Notes

Split into individual commits. Should they be squashed?

### DIFF
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -826,11 +826,11 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * will be returned for that entity, and no errors will be
    * raised.
    *
-   * @param relationshipTraitsData The traits of the relationship to
-   * query.
-   *
    * @param entityReferences A list of @ref entity_reference to query
    * the specified relationship for.
+   *
+   * @param relationshipTraitsData The traits of the relationship to
+   * query.
    *
    * @param context The calling context.
    *
@@ -858,8 +858,8 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * to @ref managementPolicy in order to determine if the manager
    * handles relationships of that type.
    */
-  void getWithRelationship(const TraitsDataPtr& relationshipTraitsData,
-                           const EntityReferences& entityReferences,
+  void getWithRelationship(const EntityReferences& entityReferences,
+                           const TraitsDataPtr& relationshipTraitsData,
                            const ContextConstPtr& context,
                            const RelationshipSuccessCallback& successCallback,
                            const BatchElementErrorCallback& errorCallback,
@@ -880,11 +880,11 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * If any relationship definition is unknown, then an empty list will
    * be returned for that relationship, and no errors will be raised.
    *
-   * @param relationshipTraitsDatas The traits of the relationships to
-   * query.
-   *
    * @param entityReference The @ref entity_reference to query the
    * specified relationships for.
+   *
+   * @param relationshipTraitsDatas The traits of the relationships to
+   * query.
    *
    * @param context The calling context.
    *
@@ -911,8 +911,9 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * to @ref managementPolicy in order to determine if the manager
    * handles relationships of that type.
    */
-  void getWithRelationships(const trait::TraitsDatas& relationshipTraitsDatas,
-                            const EntityReference& entityReference, const ContextConstPtr& context,
+  void getWithRelationships(const EntityReference& entityReference,
+                            const trait::TraitsDatas& relationshipTraitsDatas,
+                            const ContextConstPtr& context,
                             const RelationshipSuccessCallback& successCallback,
                             const BatchElementErrorCallback& errorCallback,
                             const trait::TraitSet& resultTraitSet = {});
--- a/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
+++ b/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
@@ -755,11 +755,14 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
    * default implementation returns an empty list for all
    * relationships.
    *
+   * @param entityReferences A list of @ref entity_reference
+   * "entity references" to query the specified relationship for.
+   *
    * @param relationshipTraitsData The traits of the relationship to
    * query.
    *
-   * @param entityReferences A list of @ref entity_reference
-   * "entity references" to query the specified relationship for.
+   * @param resultTraitSet A hint as to what traits the returned
+   * entities should have. May be empty.
    *
    * @param context The calling context.
    *
@@ -786,21 +789,18 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
    * must be called on the same thread that initiated the call to
    * `getWithRelationship`.
    *
-   * @param resultTraitSet A hint as to what traits the returned
-   * entities should have. May be empty.
-   *
    * @note Ensure that your implementation of @ref managementPolicy
    * responds appropriately when queried with relationship trait sets
    * and a read policy to communicate to the host whether or not you are
    * capable of handling queries for those relationships in this method.
    */
-  virtual void getWithRelationship(const TraitsDataPtr& relationshipTraitsData,
-                                   const EntityReferences& entityReferences,
+  virtual void getWithRelationship(const EntityReferences& entityReferences,
+                                   const TraitsDataPtr& relationshipTraitsData,
+                                   const trait::TraitSet& resultTraitSet,
                                    const ContextConstPtr& context,
                                    const HostSessionPtr& hostSession,
                                    const RelationshipSuccessCallback& successCallback,
-                                   const BatchElementErrorCallback& errorCallback,
-                                   const trait::TraitSet& resultTraitSet);
+                                   const BatchElementErrorCallback& errorCallback);
 
   /**
    * Queries entity references that are related to the input
@@ -820,11 +820,14 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
    * default implementation returns an empty list for all
    * relationships.
    *
+   * @param entityReference The @ref entity_reference to query the
+   * specified relationships for.
+   *
    * @param relationshipTraitsDatas The traits of the relationships to
    * query.
    *
-   * @param entityReference The @ref entity_reference to query the
-   * specified relationships for.
+   * @param resultTraitSet A hint as to what traits the returned
+   * entities should have. May be empty.
    *
    * @param hostSession The host session that maps to the caller, this
    * should be used for all logging and provides access to the Host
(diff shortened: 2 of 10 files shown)


foundrytom:
> Split into individual commits. Should they be squashed?

Do they make sense as atomic changes, that you'd want to roll-back to in isolation and potentially release at that point?

elliotcmorris (approved):
As Tom said, Feltell standard work, no notes. 

I would probably squash the commits though.

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -479,8 +479,8 @@ def call_callbacks(*_args):
             two_refs,
             a_context,
             a_host_session,
-            success_callback,
-            error_callback,
+            mock.ANY,
+            mock.ANY,
```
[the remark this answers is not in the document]
feltech: Exactly, yeah, we've had to do similar in other tests.

src/openassetio-core/src/managerApi/ManagerInterface.cpp
```diff
@@ -48,6 +48,18 @@ ManagerStateBasePtr ManagerInterface::stateFromPersistenceToken(
...
 }
+
+void ManagerInterface::getWithRelationship(
+    [[maybe_unused]] const TraitsDataPtr& relationshipTraitsData,
+    const EntityReferences& entityReferences, [[maybe_unused]] const ContextConstPtr& context,
+    [[maybe_unused]] const HostSessionPtr& hostSession,
+    const ManagerInterface::RelationshipSuccessCallback& successCallback,
+    [[maybe_unused]] const ManagerInterface::BatchElementErrorCallback& errorCallback,
+    [[maybe_unused]] const trait::TraitSet& resultTraitSet) {
+  for (std::size_t idx = 0; idx < entityReferences.size(); ++idx) {
```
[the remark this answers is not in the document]
feltech: Good shout. Some interesting [reading](https://stackoverflow.com/a/58553484/535103) and [experimentation](https://godbolt.org/z/b96Wcfc1f) led me to conclude:

In terms of macro-optimisation: `.size()` is almost definitely inlined and for `std::vector` is just a direct memory access. If we could swap out the container type (e.g. if the function was templated on container type) then it would be more of a concern.

In terms of micro-optimisation: in our case the compiler cannot _prove_ that `.size()` always returns the same value at each iteration, because we call out to another function, which could conceivably modify the vector.  So the compiler must _always_ fetch the latest value of `.size()` from memory, rather than a CPU register.

Good learnings.  I'll do the change.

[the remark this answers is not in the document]
elliotcmorris: Interesting. I did some [quick-bench](https://quick-bench.com/q/CziHGLS6Weuo1pk6sVejRRzbJ1k) benchmarks. Over 1000 elements. Clang 15.0
| InlineSizeCheck | HoistedSizeCheck | Normalized Form |
| ----------- | ----------- |-----------|
| 252.4 | 250.8 | 1 : 1.0063 |
| 251.3 | 249.9 | 1 : 1.0056 |
| 247.5 | 247.7 | 1 : 0.9992 |
| 250.8 | 251.6 | 1 : 0.9968 |
| 250.9 | 248.4 | 1 : 1.0101 |

As data ... this is worthless! :medal_sports: Smells like the difference being small enough that random variance in testing it overrides any real information.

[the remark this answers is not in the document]
feltech: I [tweaked quick-bench](https://quick-bench.com/q/PJuXiiWIUU6oS_gPxnKaR3ItMDg) code to be a bit more representative, and do see a small performance difference:

image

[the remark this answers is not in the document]
elliotcmorris: Explicitly setting the function to not be optimised smells a tad unfair to me a bit, but certainly this is quite interesting. I might quibble, especially in the presence of LTO, but there's certainly an argument.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -697,6 +697,117 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+   * A 'related' entity could take many forms. For example:
+   *
+   *  @li In 3D CGI, Multiple AOVs or layers may be related to a
+   *  'beauty' render. @li In Compositing, an image sequence may be
+   *  related to the script that created it. @li An asset may be related
+   *  to a task that specifies work to be done. @li Parent/child
+   *  relationships are also (semantically) covered by these
+   *  relationships.
+   *
+   * In the this API, these relationships are represented by trait data.
```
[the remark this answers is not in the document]
feltech: Fixed in e60c3d11

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -471,23 +471,23 @@ def call_callbacks(*_args):
...
-            error_callback,
-            resultTraitSet=an_entity_trait_set,
+            mock.ANY,
+            mock.ANY,
+            an_entity_trait_set,
         )
 
 
 class Test_Manager_getWithRelationships:
     def test_method_defined_in_python(self, method_introspector):
```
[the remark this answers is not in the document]
feltech: Fixed in 783cec82

src/openassetio-core/src/managerApi/ManagerInterface.cpp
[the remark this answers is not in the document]
feltech: Micro-optimised in bc2b0f5c and 9562c994.

Tempted to drop though favouring slightly better readability over imperceptibly better performance.  The work done in the the loop will dominate any real-world timings _by far_.

I guess the key thing is - will contributors/adopters of OpenAssetIO feel more confident in a project that prioritises micro-optimisation vs. maximising readability?  I suspect, this day and age, readability has higher weighting.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -761,6 +761,170 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
+   * relationship query. It will be given the corresponding index of the
+   * entity reference in @p entityReferences along with a populated
+   * BatchElementError (see @ref BatchElementError.ErrorCode
+   * "ErrorCodes"). The callback will be called on the same thread that
+   * initiated the call to `getWithRelationship`.
+   *
+   * @param resultTraitSet A hint as to what traits the returned
+   * entities should have.
+   *
+   * @return A list of references to related entities, for each input
```
feltech, line 857: [fix] Here and in the other functions, `@return` in incorrect

feltech: Fixed in f422832a 25f7a80e c816563f and 58ef5c0d

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -761,6 +761,164 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
+   * initiated the call to `getWithRelationship`.
+   *
+   * @param resultTraitSet A hint as to what traits the returned
+   * entities should have.
+   *
+   * @note The @ref trait_set of the queried relationship can be passed
+   * to @ref managementPolicy in order to determine if the manager
+   * handles relationships of that type.
+   */
+  void getWithRelationship(const TraitsDataPtr& relationshipTraitsData,
```
feltech, line 861: [fix] Documenting offline chat. For consistency, `entityReferences` should come first.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -697,6 +697,178 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+   * and a read policy to communicate to the host whether or not you are
+   * capable of handling queries for those relationships in this method.
+   */
+  virtual void getWithRelationship(const TraitsDataPtr& relationshipTraitsData,
+                                   const EntityReferences& entityReferences,
+                                   const ContextConstPtr& context,
+                                   const HostSessionPtr& hostSession,
+                                   const RelationshipSuccessCallback& successCallback,
+                                   const BatchElementErrorCallback& errorCallback,
+                                   const trait::TraitSet& resultTraitSet);
```
feltech, line 807: [fix] As per @foundrytom's comment and subsequent offline chat, since `resultTraitSet` is no longer optional in `ManagerInterface`, it should be moved alongside the other function-specific parameter group at the beginning.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Rearranged in 4165c819

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Rearranged as part of 4165c819

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
elliotcmorris: [fix] `getWithRelationships` still reads 

> Returns entity references that are related to the input reference

Should be changed to "queries".

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -858,8 +858,8 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * to @ref managementPolicy in order to determine if the manager
    * handles relationships of that type.
    */
-  void getWithRelationship(const TraitsDataPtr& relationshipTraitsData,
-                           const EntityReferences& entityReferences,
+  void getWithRelationship(const EntityReferences& entityReferences,
```
foundrytom, line 861: [fix] Needs a breaking change in `RELEASE_NOTES.md` for these changes.

foundrytom (changes requested):

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
foundrytom: Sorry, scratch that - I forgot we haven't released since we refactored the method in the first place.

RELEASE_NOTES.md
```diff
@@ -21,6 +21,10 @@ v1.0.0-alpha.xx
   opt-in for manager implementations.
   [#163](https://github.com/OpenAssetIO/OpenAssetIO/issues/163)
 
+- Migrated `getWithRelationship` and `getWithRelationships` to C++
+  with Python bindings.
+  [#913](https://github.com/OpenAssetIO/OpenAssetIO/issues/913)
```
foundrytom, line 26: [fix] As these methods didn't exist before this release, can we just mention that they're now in C++ in the breaking change note above?

src/openassetio-core/src/managerApi/ManagerInterface.cpp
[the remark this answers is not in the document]
feltech: After asking around internally, there is no clear consensus.  So I'll make an executive decision to leave as is (i.e. with the micro-optimisation in place). It's slightly more future-proof against swapping out the container, and shows any inquisitive engineers that we're considerate of C++ quirks.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Fixed in 8717747f

RELEASE_NOTES.md
feltech: Combined in c77a8e35.  I've assumed that the four "Migrate" commits will be squashed together (otherwise will need a few more fixup commits).

foundrytom: Spot on 👍 thanks.

elliotcmorris (approved):

feltech:
Squashed commits, in case anyone wants to check.

foundrytom:
> Squashed commits, in case anyone wants to check.

Noice.

foundrytom (approved):

Issue #980, opened 2023-06-15
Remove Versions API


foundrytom:
Remove version query methods from the core API and replace with lightweight note in the documentation for related references (docstrings + for managers + for hosts as required) that directs them to industry specific traits.

PR #982, opened 2023-06-16
[Core] Remove version query API

foundrytom:
DR017 describes the rationale, this commit removes the concrete API surface area for interacting with versions. This will be replaced by calls to `resolve` or `getWithRelationship` and domain specific traits.

Part of #980.

Note: There are plans to update some of the documentation to direct people towards their domain-specific trait packages, just waiting on #968 and friends. This PR gets the main deletion out the way.


doc/doxygen/src/ForManagers.dox
```diff
@@ -165,7 +165,8 @@
  * @subsection manager_todo_related_entities Supporting Relationships
  *
  * Relationship support is not a pre-requisite, but allows for advanced
- * functionality in many hosts.
+ * functionality in many hosts, and to handle concepts such as
+ * versioning.
```
elliotcmorris, line 169: [Minor][Consider] Grammar.

```suggestion
 * functionality in many hosts, such as handling concepts like
 * versioning.
```

doc/doxygen/src/Glossary.dox
```diff
@@ -332,10 +332,11 @@
  * It is common for am @ref asset_management_system to allow an @ref entity
  * version managed. Some systems also allow the concept of what we call a
  * 'meta-version'. For example "latest" or "latestApprovedForLighting".  These
- * are version strings that need to be further resolved using @ref
- * openassetio.hostApi.Manager.Manager.finalizedEntityVersion
- * "finalizedEntityVersion()" before they can be assumed to be
- * pointing to a specific 'concrete' version of an entity.
+ * are version strings that can't be assumed to be stable. Traits may
+ * exists within domain specific uses of OpenAssetIO (such as in
+ * OpenAssetIO-MediaCreation) that allow the
```
elliotcmorris, line 337: [Consider] Should this link to the mediacreation repo?

elliotcmorris (commented):
Removals looks good. A few notes on doc.

doc/doxygen/src/Glossary.dox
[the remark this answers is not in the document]
foundrytom: A good Q - the reason we have it is that we bring up it up in the context of the Manager State Token as a "reason you may need to use this mechanism", so we'd need to remove it from there too. I think it is perhaps helpful to keep that as a use case for "why we have that at all" at the mo?

RELEASE_NOTES.md
feltech, line 1: [consider][minor] Some docs I came across, used for exposition, which my stray too far into considering versioning as a first-class thing:

`BatchElementError.hpp` docs for `kEntityResolutionError` says
```
     * This code is also used during version finalisation and any other
     * entity-based operations on a valid @ref entity_reference that
     * fail for some reason.
```

`Manager.hpp`/`ManagerInterface.hpp` docs for `@name Publishing` says
```
   * reference to the Asset itself, so that the system can determine the
   * 'next' version number at the time of publish. It may also make
   * sense to reference a specific version of this asset to implicitly
   * state which version it will be written to. Other entity types may
   * not have this flexibility.
```

`EntityTraitsSpecifications.dox`
```
 * At its core OpenAssetIO has no detailed understanding of the
 * specifics of any given entity, only that:
...
 *  - They may (or may not) have multiple versions.
```

feltech (approved):
LGTM, just a couple of grep results to possibly consider

doc/doxygen/src/Glossary.dox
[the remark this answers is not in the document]
feltech: Yeah when I was grepping through for usages of the word "version", there's quite a few places where versioning is used as a motivating example, even if not a first-class part of the API any more. I was tempted to flag this Glossary section as being not relevant, but figured it's still useful for examples.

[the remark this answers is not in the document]
foundrytom: Its a subtle one - as we _do_ have to describe how publishing should logically map to versions within the manager, without being specific as to exactly what qualities a version has.

foundrytom: Added link [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/982/commits/99e7cbf5a694157fe5b33fec5775ebc3907b3c0a).

RELEASE_NOTES.md
foundrytom: Re-phrased a few of these [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/982/commits/810af81faeb48f7cebbdb8d43a209c5afb55e086), left the others are they felt more relevant to how we have to describe publishing working.

foundrytom:
@elliotcmorris @feltech sorry I had to rebase as there were merge conflicts preventing CI from running.

doc/doxygen/src/Glossary.dox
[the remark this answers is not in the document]
foundrytom: How about this for a perspective:

 “OpenAssetIO talks about the concept of versions, but its up to domain specific applications to define exactly what properties they have.”

elliotcmorris (approved):

feltech (approved):

PR #986, opened 2023-06-20
Paged GetWithRelationship

### LINKED ISSUE #971: C++ paged `getWithRelationship[s]`
## What

Add paged version of `getWithRelationship[s]` using the design featured in #966.

The sketch in https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/pull/21 illustrates a case where the list of relations may be large (the Proxy relation). An entity versions replacement would also result in paged-scale results.

## Why

Both methods are needed to allow for optimisations of input validation.

## ACs

- Public interface exposes concrete pager interface for the specific methods that require it as per #966:
  - `EntityReferencesPager` (where the result was `vector<EntityReference>`).
- ManagerInterfaces and Pager methods documented including failure modes and performance profile.
- Method names TBC (get)
- Python bindings.
- Abstraction layer between manager implementation/host to allow auditing/etc.. future changes to be managed.
- Default implementations as-per un-paged versions.

## Notes

Use a specific concrete implementation to avoid documentation/visibility issues, as we currently have no other paged methods slated.

## Out of scope

Detached Host/Manager copy covering the strategy around paged API methods is covered in #974.

elliotcmorris:
## Description
Adds paged GetWithRelationship methods.

Closes #971 

- [ ] I have updated all relevant user documentation. (I'm not sure, does this need guide doc?)

## Reviewer Notes
Split into 4 main commits, but will be squashed into 2 (maybe just 1)

This PR introduces a new pattern for PyRetainingSharedPtr in the ManagerInterfaceBinding. This needs to be factored out a bit, or at least docc'd.

We did some C++ testing for the pager object as part of a test-loop workflow. In theory these can be deleted now to fit with our python test philosophy, but I see no need to.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -17,6 +17,15 @@ v1.0.0-alpha.xx
 
 ### Improvements
 
+- Added paged implementations of `getWithRelationship` and
+  `getWithRelationships`, called `getWithRelationshipPaged` and
+  `getWithRelationshipsPaged`
+  These methods are the equivalent of the non-paged versions, but
+  provide an `EntityReferencePager` object, rather than a direct list
+  of results, allowing for correct handling of extremely large/unbounded
+  data sets.
+  [#971](https://github.com/OpenAssetIO/OpenAssetIO/issues/971)
+
 - Added default implementations of `getWithRelationship` and
   `getWithRelationships` that return empty lists, making these methods
   opt-in for manager implementations.
--- a/src/openassetio-core/include/openassetio/hostApi/EntityReferencePager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/EntityReferencePager.hpp
@@ -1,0 +1,127 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2023 The Foundry Visionmongers Ltd
+
+#pragma once
+
+#include <openassetio/EntityReference.hpp>
+#include <openassetio/typedefs.hpp>
+
+OPENASSETIO_FWD_DECLARE(managerApi, HostSession)
+OPENASSETIO_FWD_DECLARE(managerApi, EntityReferencePagerInterface)
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace hostApi {
+
+OPENASSETIO_DECLARE_PTR(EntityReferencePager)
+
+/**
+ * The EntityReferencePager allows for the retrieval and traversal of
+ * large datasets in a paginated manner.
+ *
+ * @note Instances of this class should not be constructed directly by
+ * the host.
+ *
+ * @see @ref Manager.getWithRelationshipPaged
+ * @see @ref Manager.getWithRelationshipsPaged
+ *
+ * None of the functions of this class should be considered thread-safe.
+ * Hosts should add their own synchronization around concurrent usage.
+ *
+ * Due to the variance of backends, construction, `hasNext`, `get` and
+ * `next` may all reasonably need to perform non-trivial, networked
+ * operations, and thus performance characteristics should not be
+ * assumed.
+ *
+ * Destruction of this object is a signal to the manager that the
+ * connection query is finished. For this reason you should avoid
+ * keeping hold of this object for longer than necessary.
+ */
+class OPENASSETIO_CORE_EXPORT EntityReferencePager final {
+ public:
+  OPENASSETIO_ALIAS_PTR(EntityReferencePager)
+  using Page = EntityReferences;
+
+  /**
+   * Constructs a new EntityReferencePager wrapping a @ref manager
+   * plugin's implementation.
+   *
+   * @note Instances of this class should not be constructed
+   * directly by the host.
+   *
+   * @param pagerInterface Implementation of the underlying pager.
+   * @param hostSession The API session.
+   * @return Newly created instance wrapped in a `std::shared_ptr`.
+   */
+  [[nodiscard]] static EntityReferencePager::Ptr make(
+      managerApi::EntityReferencePagerInterfacePtr pagerInterface,
+      managerApi::HostSessionPtr hostSession);
+
+  /**
+   * Deleted copy constructor.
+   *
+   * EntityReferencePager cannot be copied, as each object represents a
+   * single paginated query.
+   */
+  EntityReferencePager(const EntityReferencePager&) = delete;
+
+  /**
+   * Deleted copy assignment operator.
+   *
+   * EntityReferencePager cannot be copied, as each object represents a
+   * single paginated query.
+   */
+  EntityReferencePager& operator=(const EntityReferencePager&) = delete;
+
+  /// Defaulted move constructor.
+  EntityReferencePager(EntityReferencePager&&) noexcept = default;
+
+  /// Defaulted move assignment operator.
+  EntityReferencePager& operator=(EntityReferencePager&&) noexcept = default;
+
+  /**
+   *  Destruction of this object is tantamount to closing the query.
+   */
+  ~EntityReferencePager() = default;
+
+  /**
+   * Return whether or not there is more data accessible by advancing
+   * the page.
+   *
+   * @return `true` if another page is available, `false` otherwise.
+   */
+  bool hasNext();
+
+  /**
+   * Return the current page of data.
+   *
+   * If the current page has advanced beyond the last page, an empty
+   * list will be returned.
+   *
+   * @return The current page's list of entity references.
+   */
+  Page get();
+
+  /**
+   * Advance the page.
+   *
+   * Advancing beyond the last page is not an error, but will result in
+   * all subsequent calls to @ref get to return an empty page, whilst
+   * @ref hasNext will continue to return `false`.
+   */
+  void next();
+
+ private:
+  EntityReferencePager(managerApi::EntityReferencePagerInterfacePtr pagerInterface,
+                       managerApi::HostSessionPtr hostSession);
+
+  managerApi::EntityReferencePagerInterfacePtr pagerInterface_;
+  managerApi::HostSessionPtr hostSession_;
+};
+static_assert(!std::is_default_constructible_v<EntityReferencePager>);
+static_assert(!std::is_copy_constructible_v<EntityReferencePager>);
+static_assert(!std::is_copy_assignable_v<EntityReferencePager>);
+
+}  // namespace hostApi
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -12,6 +12,7 @@
 #include <openassetio/BatchElementError.hpp>
 #include <openassetio/EntityReference.hpp>
 #include <openassetio/InfoDictionary.hpp>
+#include <openassetio/hostApi/EntityReferencePager.hpp>
 #include <openassetio/trait/collection.hpp>
 #include <openassetio/typedefs.hpp>
 
@@ -835,13 +836,12 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * @param context The calling context.
    *
    * @param successCallback Callback that will be called for each
-   * successful relationship query. It will be given the
-   * corresponding index of the entity reference in
-   * @p entityReferences along with a list of entity references for
-   * entities that have the relationship specified by
-   * @p relationshipTraitsData. If there are no relations, the
-   * callback will receive an empty list. The callback will be called
-   * on the same thread that initiated the call to
+   * successful relationship query. It will be given the corresponding
+   * index of the entity reference in @p entityReferences along with a
+   * list of entity references for entities that have the relationship
+   * specified by @p relationshipTraitsData. If there are no relations,
+   * the callback will receive an empty list. The callback will be
+   * called on the same thread that initiated the call to
    * `getWithRelationship`.
    *
    * @param errorCallback Callback that will be called for each failed
@@ -918,6 +918,108 @@ class OPENASSETIO_CORE_EXPORT Manager {
                             const BatchElementErrorCallback& errorCallback,
                             const trait::TraitSet& resultTraitSet = {});
 
+  /**
+   * Callback signature used for a successful paged entity relationship
+   * query.
+   */
+  using PagedRelationshipSuccessCallback =
+      std::function<void(std::size_t, EntityReferencePagerPtr)>;
+
+  /**
+   * Paged version of getWithRelationship. See non-paged
+   * @ref getWithRelationship for further documentation.
+   *
+   * @param relationshipTraitsData The traits of the relationship to
+   * query.
+   *
+   * @param entityReferences A list of @ref entity_reference to query
+   * the specified relationship for.
+   *
+   * @param pageSize The size of each page of data. The page size is
+   * fixed for the lifetime of pager object given to the @p
+   * successCallback. Must be greater than zero.
+   *
+   * @param context The calling context.
+   *
+   * @param successCallback Callback that will be called for each
+   * successful relationship query. It will be given the corresponding
+   * index of the entity reference in @p entityReferences as well as a
+   * pager capable of returning pages of entities that have the
+   * relationship to the entity at the corresponding index, specified by
+   * @p relationshipTraitsData. If there are no relations, the pager
+   * will have no pages. The callback will be called on the same thread
+   * that initiated the call to `getWithRelationshipPaged`. To access
+   * the data, retrieve the @ref EntityReferencePager from the callback,
+   * and use its interface to traverse pages.
+   *
+   * @param errorCallback Callback that will be called for each failed
+   * relationship query. It will be given the corresponding index of the
+   * entity reference in @p entityReferences along with a populated
+   * BatchElementError (see @ref BatchElementError.ErrorCode
+   * "ErrorCodes"). The callback will be called on the same thread that
+   * initiated the call to `getWithRelationshipPaged`.
+   *
+   * @param resultTraitSet A hint as to what traits the returned
+   * entities should have.
+   *
+   * @throws std::out_of_range if @p pageSize is zero.
+   */
+  void getWithRelationshipPaged(const EntityReferences& entityReferences,
+                                const TraitsDataPtr& relationshipTraitsData, size_t pageSize,
+                                const ContextConstPtr& context,
+                                const PagedRelationshipSuccessCallback& successCallback,
+                                const BatchElementErrorCallback& errorCallback,
+                                const trait::TraitSet& resultTraitSet = {});
+
+  /**
+   * Paged version of getWithRelationships. See non-paged
+   * @ref getWithRelationships for further documentation.
+   *
+   * @param relationshipTraitsDatas The traits of the relationships to
+   * query.
+   *
+   * @param entityReference The @ref entity_reference to query the
+   * specified relationships for.
+   *
+   * @param pageSize The size of each page of data. The page size is
+   * fixed for the lifetime of pager object given to the @p
+   * successCallback. Must be greater than zero.
+   *
+   * @param context The calling context.
+   *
+   * @param successCallback Callback that will be called for each
+   * successful relationship query. It will be given the corresponding
+   * index of the relationship in @p relationshipTraitsDatas as well as
+   * a pager capable of returning pages of entities related to @p
+   * entityReference by the relationship at that corresponding index. If
+   * there are no relations, the pager will have no pages. The callback
+   * will be called on the same thread that initiated the call to
+   * `getWithRelationshipsPaged`. To access the data, retrieve the
+   * @ref EntityReferencePager from the callback, and use its interface
+   * to traverse pages.
+   *
+   * @param errorCallback Callback that will be called for each failed
+   * relationship query. It will be given the corresponding index of the
+   * relationship in @p relationshipTraitsDatas along with a populated
+   * BatchElementError (see @fqref{BatchElementError.ErrorCode}
+   * "ErrorCodes"). The callback will be called on the same thread that
+   * initiated the call to `getWithRelationshipsPaged`.
+   *
+   * @param resultTraitSet A hint as to what traits the returned
+   * entities should have.
+   *
+   * @note The @ref trait_set of any queried relationship can be passed
+   * to @ref managementPolicy in order to determine if the manager
+   * handles relationships of that type.
+   *
+   * @throws std::out_of_range if @p pageSize is zero.
+   */
+  void getWithRelationshipsPaged(const EntityReference& entityReference,
+                                 const trait::TraitsDatas& relationshipTraitsDatas,
+                                 size_t pageSize, const ContextConstPtr& context,
+                                 const PagedRelationshipSuccessCallback& successCallback,
+                                 const BatchElementErrorCallback& errorCallback,
+                                 const trait::TraitSet& resultTraitSet = {});
   /// @}
 
   /**
--- a/src/openassetio-core/include/openassetio/managerApi/EntityReferencePagerInterface.hpp
+++ b/src/openassetio-core/include/openassetio/managerApi/EntityReferencePagerInterface.hpp
@@ -1,0 +1,122 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2023 The Foundry Visionmongers Ltd
+
+#pragma once
+
+#include <vector>
+
+#include <openassetio/export.h>
+#include <openassetio/EntityReference.hpp>
+#include <openassetio/typedefs.hpp>
+
+OPENASSETIO_FWD_DECLARE(managerApi, HostSession)
+OPENASSETIO_FWD_DECLARE(Context)
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace managerApi {
+
+OPENASSETIO_DECLARE_PTR(EntityReferencePagerInterface)
+
+/**
+ * Deals with the retrieval of paginated data from the backend at the
+ * behest of the host.
+ *
+ * The manager is expected to extend this type, and store data necessary
+ * to perform the paging operations on the extended object, utilizing
+ * caching when possible to reduce redundant queries.
+ *
+ * Thread-safety of operations is not expected. Hosts will synchronize
+ * calls themselves, if required.
+ *
+ * This is a non-copyable object that will be held in a `shared_ptr`,
+ * meaning multiple references may be held, but only to a single
+ * instance of the ongoing query, whose destructor will be called when
+ * all references are released. As such, the destructor of this class is
+ * a good place to place any complex cleanup operations (e.g. closing
+ * open connections).
+ *
+ * To support as wide array of possible backends as possible,
+ * OpenAssetIO places no restraints on the behaviour of this type
+ * concerning performance, however, it is considered friendly to
+ * document the performance characteristics of your Pager
+ * implementation.
+ */
+class OPENASSETIO_CORE_EXPORT EntityReferencePagerInterface {
+ public:
+  OPENASSETIO_ALIAS_PTR(EntityReferencePagerInterface)
+  using Page = std::vector<EntityReference>;
+
+  /// Defaulted default constructor.
+  EntityReferencePagerInterface() = default;
+
+  /**
+   * Deleted copy constructor.
+   *
+   * Copying of pagers is disallowed, for convenience of implementation.
+   *
+   * However, be aware that multiple references to the pager may be
+   * held.
+   */
+  explicit EntityReferencePagerInterface(const EntityReferencePagerInterface&) = delete;
+
+  /**
+   * Deleted copy assignment operator.
+   *
+   * Copying of pagers is disallowed, for convenience of implementation.
+   *
+   * However, be aware that multiple references to the pager may be
+   * held.
+   */
+  EntityReferencePagerInterface& operator=(const EntityReferencePagerInterface&) = delete;
+
+  /// Defaulted move constructor.
+  EntityReferencePagerInterface(EntityReferencePagerInterface&&) noexcept = default;
+
+  /// Defaulted move assignment operator.
+  EntityReferencePagerInterface& operator=(EntityReferencePagerInterface&&) noexcept = default;
+
+  /**
+   * Manager should override destructor to be notified when query has
+   * finished.
+   */
+  virtual ~EntityReferencePagerInterface() = default;
+
+  /**
+   * Returns whether or not there is more data accessible by advancing
+   * the page.
+   *
+   * The mechanism to acquire this information is variable, and left up
+   * to the specifics of the backend implementation.
+   *
+   * @return `true` if another page is available, `false` otherwise.
+   */
+  virtual bool hasNext(const HostSessionPtr&) = 0;
+
+  /**
+   * Return the current page of data.
+   *
+   * If the current page has advanced beyond the last page, an empty
+   * list should be returned.
+   *
+   * @return The current page's list of entity references.
+   */
+  virtual Page get(const HostSessionPtr&) = 0;
+
+  /**
+   * Advance the page.
+   *
+   * If currently on the last page of results, calling `next` should
+   * logically advance to the page after the last page, in analogy with
+   * `std::end`. Subsequent calls should then be a no-op. In this state,
+   * @ref hasNext should continue to return `false` and @ref get should
+   * return an empty page.
+   */
+  virtual void next(const HostSessionPtr&) = 0;
+};
+static_assert(!std::is_copy_constructible_v<EntityReferencePagerInterface>);
+static_assert(!std::is_copy_assignable_v<EntityReferencePagerInterface>);
+
+}  // namespace managerApi
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
--- a/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
+++ b/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
@@ -15,6 +15,7 @@
 #include <openassetio/trait/collection.hpp>
 #include <openassetio/typedefs.hpp>
 
+OPENASSETIO_FWD_DECLARE(managerApi, EntityReferencePagerInterface)
 OPENASSETIO_FWD_DECLARE(managerApi, ManagerStateBase)
 OPENASSETIO_FWD_DECLARE(managerApi, HostSession)
 OPENASSETIO_FWD_DECLARE(Context)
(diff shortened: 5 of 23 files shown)


src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -567,6 +567,180 @@ def test_when_relationship_is_None_then_exception_raised(
...
+            a_context,
+            a_host_session,
+            mock.ANY,
+            mock.ANY,
+        )
+        success_callback.assert_called_once_with(0, mock.ANY)
+
+        # Additional assert to check cpp constructed Pager object is as
+        # expected.
+        pager = success_callback.call_args[0][1]
```
foundrytom, line 647: [fix] Can we also test with `ininstance(pager, EntityReferencePager)` to make sure the pager is definitely wrapper?

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -567,6 +567,180 @@ def test_when_relationship_is_None_then_exception_raised(
...
+            a_host_session,
+            mock.ANY,  # success
+            mock.ANY,  # error
+        )
+
+        success_callback.assert_called_once_with(0, mock.ANY)
+
+        # Additional assert to check cpp constructed Pager object is as
+        # expected.
+        pager = success_callback.call_args[0][1]
```
foundrytom, line 779: [fix] as above re isinstance.

foundrytom (changes requested):
Great to see the way this has worked out- nice one @elliotcmorris. Sorry some WIP comments as @feltech is tweaking as we speak.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -918,6 +918,104 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
+   * @param context The calling context.
+   *
+   * @param successCallback Callback that will be called for each
+   * successful relationship query. It will be given the corresponding
+   * index of the entity reference in @p entityReferences as well as a
+   * pager capable of returning pages of entities that have the
+   * relationship to the entity at the corresponding index, specified by
+   * @p relationshipTraitsData. If there are no relations, the pager
+   * will have no pages. The callback will be called on the same thread
+   * that initiated the call to `getWithRelationship`. To access the
```
foundrytom, line 951: [fix]
```suggestion
   * that initiated the call to `getWithRelationshipPaged`. To access the
```

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -918,6 +918,104 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
+   * data, retrieve the @ref EntityReferencePager from the callback, and
+   * use its interface to traverse pages.
+   *
+   * @param errorCallback Callback that will be called for each failed
+   * relationship query. It will be given the corresponding index of the
+   * entity reference in @p entityReferences along with a populated
+   * BatchElementError (see @ref BatchElementError.ErrorCode
+   * "ErrorCodes"). The callback will be called on the same thread that
+   * initiated the call to `getWithRelationship`.
```
foundrytom, line 960: [fix]
```suggestion
   * initiated the call to `getWithRelationshipPaged`.
```

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -918,6 +918,104 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
+   *
+   * @param relationshipTraitsData The traits of the relationship to
+   * query.
+   *
+   * @param entityReferences A list of @ref entity_reference to query
+   * the specified relationship for.
+   *
+   * @param pageSize The size of each page of data. The page size is
+   * fixed for the lifetime of pager object given to the @p
+   * successCallback.
```
foundrytom, line 940: [question] Should we define this should be > 0 and/or introduce any validation?

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -918,6 +918,104 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
+   * `getWithRelationship`. To access the data, retrieve the
+   * @ref EntityReferencePager from the callback, and use its interface
+   * to traverse pages.
+   *
+   * @param errorCallback Callback that will be called for each failed
+   * relationship query. It will be given the corresponding index of the
+   * relationship in @p relationshipTraitsDatas along with a populated
+   * BatchElementError (see @fqref{BatchElementError.ErrorCode}
+   * "ErrorCodes"). The callback will be called on the same thread that
+   * initiated the call to `getWithRelationships`.
```
foundrytom, line 1004: [fix] 
```suggestion
   * initiated the call to `getWithRelationshipsPaged`.
```

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -918,6 +918,104 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
+   * @param context The calling context.
+   *
+   * @param successCallback Callback that will be called for each
+   * successful relationship query. It will be given the corresponding
+   * index of the relationship in @p relationshipTraitsDatas as well as
+   * a pager capable of returning pages of entities related to @p
+   * entityReference by the relationship at that corresponding index. If
+   * there are no relations, the pager will have no pages. The callback
+   * will be called on the same thread that initiated the call to
+   * `getWithRelationship`. To access the data, retrieve the
```
foundrytom, line 995: [fix]
```suggestion
   * `getWithRelationshipPaged`. To access the data, retrieve the
```

src/openassetio-core/src/managerApi/ManagerInterface.cpp
```diff
@@ -74,6 +77,52 @@ void ManagerInterface::getWithRelationships(
     successCallback(idx, {});
   }
 }
+
+namespace {
+/*
+ * A dummy pager interface that acts as if it has no data. For use in
+ * getWithRelationship[s] default implementation.
+ */
+class IdentityEntityReferencePagerInterface : public managerApi::EntityReferencePagerInterface {
```
foundrytom, line 86: [consider] Just as 'identity' is quite loaded in 3D graphics/maths. `EmptyEntityReferencePagerInterface`.

src/openassetio-core/include/openassetio/hostApi/EntityReferencePager.hpp
```diff
@@ -0,0 +1,83 @@
...
+ * Due to the variance of backends, construction, `hasNext`, `get` and
+ * `next` may all reasonably need to perform non trivial, networked
+ * operations, and thus performance characteristics should not be
+ * assumed.singleEntityReferencePager<Elem>
+ *
+ * Destruction of this object is a signal to the manager that the
+ * connection query is finished. For this reason you should avoid
+ * keeping hold of this object for longer than necessary.
+ */
+class OPENASSETIO_CORE_EXPORT EntityReferencePager {
```
foundrytom, line 36: [consider] `final`?

src/openassetio-python/cmodule/src/hostApi/ManagerBinding.cpp
```diff
@@ -136,9 +136,13 @@ void registerManager(const py::module& mod) {
...
-      .def("getWithRelationship", &Manager::getWithRelationship, py::arg("entityReferences"),
-           py::arg("relationshipTraitsData").none(false), py::arg("context").none(false),
-           py::arg("successCallback"), py::arg("errorCallback"),
+      .def("getWithRelationship",
+           py::overload_cast<const EntityReferences&, const TraitsDataPtr&, const ContextConstPtr&,
+                             const Manager::RelationshipSuccessCallback&,
+                             const Manager::BatchElementErrorCallback&, const trait::TraitSet&>(
+               &Manager::getWithRelationship),
+           py::arg("entityReferences"), py::arg("relationshipTraitsData").none(false),
+           py::arg("context").none(false), py::arg("successCallback"), py::arg("errorCallback"),
```
foundrytom, line 145: [question] Is this still needed now were not overloading methods? Is it needed below too?

src/openassetio-python/cmodule/src/hostApi/ManagerBinding.cpp
```diff
@@ -154,6 +158,33 @@ void registerManager(const py::module& mod) {
...
           py::arg("context").none(false), py::arg("successCallback"), py::arg("errorCallback"),
           py::arg("resultTraitSet") = trait::TraitSet{})
+      .def("getWithRelationshipPaged",
+           py::overload_cast<const EntityReferences&, const TraitsDataPtr&, std::size_t,
+                             const ContextConstPtr&,
+                             const Manager::PagedRelationshipSuccessCallback&,
+                             const Manager::BatchElementErrorCallback&, const trait::TraitSet&>(
+               &Manager::getWithRelationshipPaged),
+           py::arg("entityReferences"), py::arg("relationshipTraitsData").none(false),
+           py::arg("pageSize").none(false), py::arg("context").none(false),
```
foundrytom, line 168: [question] Is `non(false)` relevant for `pageSize` as its an int? Same Q below

src/openassetio-core/include/openassetio/hostApi/EntityReferencePager.hpp
```diff
@@ -0,0 +1,83 @@
...
+ public:
+  OPENASSETIO_ALIAS_PTR(EntityReferencePager)
+  using Page = EntityReferences;
+
+  [[nodiscard]] static EntityReferencePager::Ptr make(
+      managerApi::EntityReferencePagerInterfacePtr pagerInterface,
+      managerApi::HostSessionPtr hostSession);
```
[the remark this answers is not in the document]
feltech: :thinking: This would mean we could no longer test it in isolation, so would take a little bit of work to refactor the C++ tests.

`friend`ing also means exposing _all_ `private` members.  Though I'm not against it in this case. 

I think if we're really limiting construction of `EntityReferencePager` objects to the `Manager`, then a better option would be to make `EntityReferencePager` a child class of the `Manager`.

Then `EntityReferencePagerInterface` should probably be a child class of `ManagerInterface` - but that's slightly ugly since manager plugin authors need to inherit from it.

I think there's an argument somewhere that probably applies, about loose coupling of components etc. So I'm happy to leave as-is too.

Will ponder some more.

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
```diff
@@ -112,6 +114,48 @@ struct PyManagerInterface : ManagerInterface {
...
+      const ManagerInterface::PagedRelationshipSuccessCallback& successCallback,
+      const ManagerInterface::BatchElementErrorCallback& errorCallback) override {
+    auto successRetainingWrapper =
+        [&successCallback](
+            size_t idx,
+            const PyRetainingSharedPtr<EntityReferencePagerInterface>& pagerInterface) {
+          successCallback(idx, pagerInterface);
+        };
+
+    PYBIND11_OVERRIDE_IMPL(void, ManagerInterface, "getWithRelationshipPaged", entityReferences,
```
foundrytom, line 130: [fix] Can we document what's going on here and below as its not using public pybind API and looks like we always call the base class method (which would be v bad).

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -567,6 +568,26 @@ def test_when_relationship_is_None_then_exception_raised(
         assert not mock_manager_interface.mock.getWithRelationships.called
 
 
+class DyingPagerInterface(EntityReferencePagerInterface):
+    """
+    Throwaway pager interface def, so we can create a temporary
+    interface intended to fall out of scope.
+    Any interface would do, but this avoids ugly imports.
```
foundrytom, line 575: [question] Is this true, as it goes through C++ then it needs to inherit from `EntityReferencePagerInterface` right?

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -654,6 +675,51 @@ def call_callbacks(*_args):
...
+            a_context,
+            lambda _idx, pager: pagers.append(pager),
+            error_callback,
+            resultTraitSet=an_entity_trait_set,
+        )
+
+        """
+        Without PyRetainingSharedPtr, will raise
+        "Tried to call pure virtual function "EntityReferencePagerInterface::get"
+        """
```
foundrytom, line 720: [consider] We generally use comments vs string literals. Same below.

src/openassetio-core/tests/hostApi/EntityReferencePagerTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+      managerApi::Host::make(std::make_shared<openassetio::MockHostInterface>()),
+      std::make_shared<openassetio::MockLoggerInterface>());
+};
+
+}  // namespace
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
+
+namespace {
+struct MockEntityReferencePagerInterface
```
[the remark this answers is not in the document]
feltech: I don't think there was any reason.  The other mocks will have been copy-pasted from elsewhere, whereas the pager mock is new.  I'd say use whatever is convenient - wrapping in our namespace means that individual type names don't have to be fully-qualified, that's the only reason I can think of.

Anyway, as mentioned in https://github.com/OpenAssetIO/OpenAssetIO/pull/986#discussion_r1237018871 none of the mock classes other than `MockEntityReferencePagerInterface` are actually needed, so can just be removed.  Problem solved.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -888,6 +888,8 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
    *
    * @param context The calling context.
    *
+   * @param hostSession The API session.
+   *
```
foundrytom, line 892: [fix] Can we use the standard desc for this as it highlights its use for logging?

foundrytom (changes requested):
Sorry some more generally minor/doc related comments.

src/openassetio-core/src/hostApi/EntityReferencePager.cpp
```diff
@@ -0,0 +1,40 @@
...
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace hostApi {
+
+typename EntityReferencePager::Ptr EntityReferencePager::make(
+    managerApi::EntityReferencePagerInterfacePtr pagerInterface,
+    managerApi::HostSessionPtr hostSession) {
+  // Trick to give private constructor access to std::make_shared.
+  struct Constructible : EntityReferencePager {
+    Constructible(managerApi::EntityReferencePagerInterfacePtr pagerInterface,
```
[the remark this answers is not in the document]
foundrytom: Perfecto - thanks, sorry that had slipped my memory.

src/openassetio-core/tests/hostApi/EntityReferencePagerTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+    const openassetio::ManagerFixture fixture;
+
+    const std::shared_ptr<MockEntityReferencePagerInterface> mockEntityReferencePagerInterface =
+        std::make_shared<MockEntityReferencePagerInterface>();
+
+    const openassetio::hostApi::EntityReferencePager::Ptr pager =
+        openassetio::hostApi::EntityReferencePager::make(mockEntityReferencePagerInterface,
+                                                         fixture.hostSession);
+
+    AND_GIVEN("plugin's hasMore expects to be called and returns false") {
```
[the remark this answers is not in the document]
feltech: Ah interesting, yeah, I pondered about the correct nomenclature to use.  

`interface` to me is closer to meaning "the contract" than "particular implementation of the contract".  Though I know in various other places we refer to "the interface" as shorthand for a particular implementation of the abstract `ManagerInterface` class.

So maybe `interface's` isn't so bad, from a consistency point of view, at least.

How about `implementation's`?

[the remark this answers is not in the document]
foundrytom: Thats a way better suggestion

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: :thinking: `size_t` is unsigned, so negative values are not a problem.  But `== 0` is still an issue.

I guess we should validate, since zero is potentially quite bad, given that some implementations may well use the page size in arithmetic (i.e. division by zero).

foundrytom: We have (or maybe had) prior art where `Manager` used to validate array lengths etc. when multiple args needed to be coherent.

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: I guess it's meant to read as "any EntityReferencePagerInterface would do"

src/openassetio-core/include/openassetio/hostApi/EntityReferencePager.hpp
foundrytom: For me, `final` conveys intent in the public interface, and could perhaps avoid a manager implementation attempting to derive the wrong class, the `make_shared` trick feels to be a private implementation detail that makes us feel better.

src/openassetio-python/tests/package/hostApi/test_manager.py
foundrytom: Just feels like the comment isn't really adding any clarity to the test setup/purpose?

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: There are lots of other methods that simply say "The API session".  But I see the text you mean, will switch.

src/openassetio-core/include/openassetio/managerApi/EntityReferencePagerInterface.hpp
```diff
@@ -0,0 +1,80 @@
...
+   * Returns whether or not there is more data accessible by advancing
+   * the page.
+   *
+   * The mechanism to acquire this information is variable, and left up
+   * to the specifics of the backend implementation.
+   */
+  virtual bool hasNext(const HostSessionPtr&) = 0;
+
+  /**
+   * Return the current page data.
```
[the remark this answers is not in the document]
feltech: Added as part of 15f65f71

src/openassetio-core/include/openassetio/managerApi/EntityReferencePagerInterface.hpp
```diff
@@ -0,0 +1,80 @@
...
+   */
+  virtual Page get(const HostSessionPtr&) = 0;
+
+  /**
+   * Advance the page.
```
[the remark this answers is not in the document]
feltech: Added as part of 15f65f71

src/openassetio-core/include/openassetio/hostApi/EntityReferencePager.hpp
```diff
@@ -0,0 +1,83 @@
...
+  EntityReferencePager(EntityReferencePager&&) noexcept = default;
+  EntityReferencePager& operator=(EntityReferencePager&&) noexcept = default;
+  /**
+   *  Destruction of this object is tantamount to closing the query.
+   */
+  ~EntityReferencePager() = default;
+
+  /**
+   * Return whether or not there is more data accessible by advancing
+   * the page.
```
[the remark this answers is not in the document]
feltech: Fleshed out as part of 15f65f71

src/openassetio-core/include/openassetio/hostApi/EntityReferencePager.hpp
```diff
@@ -0,0 +1,83 @@
...
+   */
+  bool hasNext();
+
+  /**
+   * Return the current page data.
+   */
+  Page get();
+
+  /**
+   * Advance the page.
```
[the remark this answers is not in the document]
feltech: Fleshed out as part of 15f65f71

src/openassetio-core/include/openassetio/hostApi/EntityReferencePager.hpp
```diff
@@ -0,0 +1,83 @@
...
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace hostApi {
+
+OPENASSETIO_DECLARE_PTR(EntityReferencePager)
+
+/**
+ * The EntityReferencePager is the Host facing representation of a
+ * @fqref{managerApi.EntityReferencePagerInterface} implementation.
+ * The Manager class shouldn't be directly constructed by the host.
```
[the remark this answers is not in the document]
foundrytom: LGTM 👍 thanks.

src/openassetio-core/tests/hostApi/EntityReferencePagerTest.cpp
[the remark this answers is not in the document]
feltech: Removed as part of 35973829.

src/openassetio-core/tests/hostApi/EntityReferencePagerTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+  IMPLEMENT_MOCK1(next);
+};
+
+}  // namespace
+
+SCENARIO("Using an EntityReferencePager") {
+  using trompeloeil::_;
+
+  GIVEN("a configured EntityReferencePager") {
+    const openassetio::trait::TraitSet traits = {"fakeTrait", "secondFakeTrait"};
```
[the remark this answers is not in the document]
feltech: Removed as part of 35973829

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: Added in 891e7eb4

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
```diff
@@ -249,6 +249,96 @@ def test_returns_empty_list_for_each_input(self, manager_interface, a_host_sessi
...
+                success_callback,
+                error_callback,
+            )
+
+            error_callback.assert_not_called()
+
+            # The default pager behaviour is to return no data and
+            # report no new pages.
+            for idx, x in enumerate(refs):
+                print(idx)
```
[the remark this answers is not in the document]
feltech: Removed in da03791f

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
```diff
@@ -249,6 +249,96 @@ def test_returns_empty_list_for_each_input(self, manager_interface, a_host_sessi
...
+                pager = success_callback.call_args_list[idx][0][1]
```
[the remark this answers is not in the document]
feltech: Added in db43024a

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
```diff
@@ -249,6 +249,96 @@ def test_returns_empty_list_for_each_input(self, manager_interface, a_host_sessi
...
+            # The default pager behaviour is to return no data and
+            # report no new pages.
+            for idx, x in enumerate(rels):
+                print(idx)
+                pager = success_callback.call_args_list[idx][0][1]
+                assert pager.hasNext(a_host_session) == False
+                assert pager.get(a_host_session) == []
+                pager.next(a_host_session)
+                assert pager.hasNext(a_host_session) == False
+                assert pager.get(a_host_session) == []
```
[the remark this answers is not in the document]
feltech: Good shout, refactored in b09ab751

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Fixed in 94111234

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Fixed in 94111234 / 2799773e

src/openassetio-core/src/managerApi/ManagerInterface.cpp
feltech: Renamed as suggested in 90a04e42

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -567,6 +568,26 @@ def test_when_relationship_is_None_then_exception_raised(
         assert not mock_manager_interface.mock.getWithRelationships.called
 
 
+class DyingPagerInterface(EntityReferencePagerInterface):
```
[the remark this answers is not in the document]
feltech: Renamed to `FakeEntityReferencePagerInterface` in 5cdb06d8, since that seems to fit well into automated testing nomenclature.

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: Dodged by simply removing that sentence as part of cf155b54

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: Switched in c341414d

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Used the more descriptive form in 3e1807f0

src/openassetio-core/include/openassetio/managerApi/EntityReferencePagerInterface.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * behest of the host.
+ *
+ * The manager is expected to extend this type, and store data
+ * data necessary to perform the paging operations on the extended
+ * object, utilizing caching when possible to reduce redundant
+ * queries.
+ *
+ * This object does not time out until the host gives up ownership. A
+ * manager should implement the destructor if they wish to close any
+ * open connections in response to this.
```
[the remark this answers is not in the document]
feltech: Rephrased in 4f9cccf8, how's that?

src/openassetio-python/tests/conftest.py
```diff
@@ -295,6 +309,72 @@ def getWithRelationships(
...
+        pageSize,
+        context,
+        hostSession,
+        successCallback,
+        errorCallback,
+    ):
+        assert isinstance(relationshipTraitsData, TraitsData)
+        self.__assertIsIterableOf(entityReferences, EntityReference)
+        self.__assertCallingContext(context, hostSession)
+        if resultTraitSet is not None:
```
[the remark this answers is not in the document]
feltech: Fixed, also for the non-paged signatures (whoops) in 1069b2a5

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
feltech: Added and documented a macro to override this in e7293ce1

RELEASE_NOTES.md
```diff
@@ -17,6 +17,19 @@ v1.0.0-alpha.xx
 
 ### Improvements
 
+- Added paged implementations of `getWithRelationship` and
```
feltech, line 20: [consider][minor] Super minor, but this seems a bit verbose for release notes, and covers some stuff I'd expect to see in API docs rather than release notes.

But like I say, super minor subjective opinion.

feltech (commented):
A few minor things to add  :slightly_smiling_face:

src/openassetio-core/tests/hostApi/EntityReferencePagerTest.cpp
```diff
@@ -0,0 +1,136 @@
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
elliotcmorris: Aye, i mentioned this in the PR description as well 
> In theory these can be deleted now to fit with our python test philosophy.

It is only the intuitive distaste of deleting perfectly good tests that grasps me here. Looking at it later, analytically, unless  I feel we should stick to our guns and actually delete this test file, it served is purpose and has been replaced. Exceptions to rules are expensive just in their very existence.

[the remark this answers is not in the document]
feltech: We do indeed have good coverage in Python too, so they can be removed.

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -567,6 +568,288 @@ def test_when_relationship_is_None_then_exception_raised(
...
+class FakeEntityReferencePagerInterface(EntityReferencePagerInterface):
+    """
+    Throwaway pager interface def, so we can create a temporary
+    interface intended to fall out of scope.
+
+    See `test_pager_kept_alive_by_retaining_shared_ptr`.
+    """
+
+    def __init__(self):
+        super().__init__()
```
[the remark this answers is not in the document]
elliotcmorris: Might as well do it right, patterns leak. https://github.com/OpenAssetIO/OpenAssetIO/pull/986/commits/7acba4391dfe86277bec5d0fffd862ba46163750

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
```diff
@@ -165,14 +201,37 @@ void registerManagerInterface(const py::module& mod) {
       .def("resolve", &ManagerInterface::resolve, py::arg("entityReferences"), py::arg("traitSet"),
            py::arg("context").none(false), py::arg("hostSession").none(false),
            py::arg("successCallback"), py::arg("errorCallback"))
-      .def("getWithRelationship", &ManagerInterface::getWithRelationship,
+      .def("getWithRelationshipPaged",
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/986/commits/5f594e23461a169f0fc80867eabcd2ad7e231aa9

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
elliotcmorris: Added a an index error for when the size is zero.
https://github.com/OpenAssetIO/OpenAssetIO/pull/986/commits/29e07320989366f23bb3886f525e8e437bee2dec

and test : 
https://github.com/OpenAssetIO/OpenAssetIO/pull/986/commits/db6f4c2e9f4548f33365d9f68cfc440547a75ba1

src/openassetio-core/include/openassetio/hostApi/EntityReferencePager.hpp
elliotcmorris: Don't personally mind using the raw shared ptr constructor, even without final it seems like the constructable thing is more prone to confusion.

Added final, removed constructible trick : https://github.com/OpenAssetIO/OpenAssetIO/pull/986/commits/1b8d6f0affb98a575470bfc0082c3e9cc947b001

src/openassetio-core/tests/hostApi/EntityReferencePagerTest.cpp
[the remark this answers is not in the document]
elliotcmorris: Removed the cpp test in : https://github.com/OpenAssetIO/OpenAssetIO/pull/986/commits/690daacd9e47015bb792104fd526c2f365bdb941

src/openassetio-core/tests/hostApi/EntityReferencePagerTest.cpp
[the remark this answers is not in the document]
elliotcmorris: Not so relevant anymore given that the cpp test is removed in : https://github.com/OpenAssetIO/OpenAssetIO/pull/986/commits/690daacd9e47015bb792104fd526c2f365bdb941

src/openassetio-core/tests/hostApi/EntityReferencePagerTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+    AND_GIVEN("plugin's get expects to be called and returns some entity references") {
+      const std::vector<openassetio::EntityReference> testEntRefs = {
+          openassetio::EntityReference("One!"), openassetio::EntityReference("Two!")};
+      REQUIRE_CALL(*mockEntityReferencePagerInterface, get(fixture.hostSession))
+          .RETURN(testEntRefs);
+      WHEN("host wrapper's get is called") {
+        const openassetio::EntityReferences getReturn = pager->get();
+        THEN("the returned entity references match those provided by the plugin") {
+          CHECK(getReturn == testEntRefs);
+        }
```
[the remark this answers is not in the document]
elliotcmorris: Even with the cpp tests deleted, we were still thinking about this.
Added a test to validate that the Pager does indeed hold the PagerInterface, and will trigger its destruction when released, even in Python : 40033a20275d559213db0f899b97cf0c2de5d8db

However, @feltech and I were discussion the pythonic implications of using the destructor as a signalling mechanism, and we realise that there is an implicit assumption that you construct your PagerInterface in the same scope as your Pager, and don't (even, perhaps primarily, accidentally,) have any other references to it.

I wonder if this is a bit of a footgun for the python developer. We should discuss this together, potentially the thing to do is to add a `finalize` or `destroy` or `finished` method and call that in the C++ destructor instead, and untie the concept of object and query lifetime.

src/openassetio-python/tests/package/hostApi/test_entityreferencepager.py
```diff
@@ -0,0 +1,109 @@
...
+            r".+The following argument types are supported:[^(]+"
+            r"EntityReferencePager\([^,]+managerApi.EntityReferencePagerInterface,[^,]+managerApi.HostSession.+"
+        )
+
+        with pytest.raises(TypeError, match=matchExpr):
+            EntityReferencePager(None, a_host_session)
+
+
+class Test_EntityReferencePager_next:
+    def test_method_defined_in_cpp(self, method_introspector):
```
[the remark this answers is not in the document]
feltech: Ah, sorry, that's not what I meant. I mean the various `test_method_defined_in_cpp` tests that have been added, which are not necessary for the entirely new C++ methods that were written from the ground up in C++, i.e. they were not prototyped in Python and ported to C++.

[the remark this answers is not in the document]
elliotcmorris: Ah! No, my bad! Just assumed there's was a lot of comments about the same thing. Reading comprehension fail.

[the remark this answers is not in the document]
elliotcmorris: Removed the method_defined_in_cpp_tests https://github.com/OpenAssetIO/OpenAssetIO/pull/986/commits/a76cd5c55fc1b8528bc7da4ab71fc777dcae2589

src/openassetio-core/tests/hostApi/EntityReferencePagerTest.cpp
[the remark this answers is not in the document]
elliotcmorris: Craeted issue : https://github.com/OpenAssetIO/OpenAssetIO/issues/990 as discussed.

foundrytom (approved):
Thanks both, please may we squash this down ready for merge once @feltech is happy with his comments, so we can take one last look in final form?

PR #1001, opened 2023-07-03
Remove old constants

foundrytom:
Preparatory work for #998, removing some very old and redundant constants.

### DIFF
--- a/src/openassetio-python/package/openassetio/constants.py
+++ b/src/openassetio-python/package/openassetio/constants.py
@@ -21,23 +21,15 @@ Constants used throughout the OpenAssetIO API.
 See @ref openassetio.log.LoggerInterface for example.
 """
 
-kSupportedAttributeTypes = (str, int, float, bool, type(None))
-
-## @name Field Names
-## Field names are to be used whenever data is get or set from a dictionary by
-## key, rather than through an accessor in some wrapper class (eg:
-## Item/Specification etc...). Bare strings should never be used to help
-## protect against inconsistencies and future changes.
+## @name Info dict field Names
 ##
-## These fields should also be used, if necessary, when declaring derived
-## classes of a @ref Specification.
+## Bare strings should never be used to help protect against
+## inconsistencies and future changes.
+##
 ## @{
-
-kField_ItemType = "fnItemType"
 
 # General
 
-kField_DisplayName = "displayName"
 kField_SmallIcon = "smallIcon"
 kField_Icon = "icon"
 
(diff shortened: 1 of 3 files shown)


src/openassetio-python/package/openassetio/constants.py
```diff
@@ -23,18 +23,13 @@
...
 ##
-## These fields should also be used, if necessary, when declaring derived
-## classes of a @ref Specification.
 ## @{
 
-kField_ItemType = "fnItemType"
-
 # General
 
 kField_DisplayName = "displayName"
```
feltech, line 35: [question] What is the purpose of the `"displayName"` field? 

Is it just a convenient alternative to calling `manager.displayName()` if you already happen to have the `.info()` dictionary?  

Although in that case, the host would always have to code defensively in case `"displayName"` wasn't in `.info()`, and so fall back to calling `.displayName()` anyway, kinda defeating the point.

src/openassetio-python/package/openassetio/constants.py
```diff
@@ -49,64 +44,4 @@
 ## called from multi-threaded C++).
 kField_EntityReferencesMatchPrefix = "entityReferencesMatchPrefix"
```
feltech, line 45: [question] As far as I can tell, this is the only constant that we _require_.  If that's likely to remain the case, then perhaps the entity prefix should have special status, e.g. a `ManagerInterface::entityReferencePrefix() -> std::optional<std::string>` method.

Alternatively...

The entity reference prefix allows us to to help manager plugins simplify, very slightly, their code; as well as optimise away a single virtual function call, at the cost of an `if-else` branch.  

So maybe for such small gains, and assuming this constant is the only one we require, we could get rid of it?

src/openassetio-python/package/openassetio/constants.py
```diff
@@ -49,64 +44,4 @@
...
-kField_MediaFrameOut = "mediaOutFrame"
-
-## @}
-
-
-## @name Setting Key Names
-## Name for keys in the manager settings dict, where applicable.
-## @{
-
-kSetting_ManagerIdentifier = "OpenAssetIO:managerIdentifier"
```
feltech, line 99: [question] Is the idea to drop common `.settings()` keys entirely, or just these ones because they're not useful?

elliotcmorris (approved):
Nothing further.

src/openassetio-python/package/openassetio/constants.py
foundrytom: Yeah, this one one of the "use needs validating" ones - I _think_ it was as before we refactord [this to c++](https://github.com/OpenAssetIO/OpenAssetIO/blob/db3ddc2dc5bee475dfa9c4388bf25854f479174d/src/openassetio-core/src/hostApi/ManagerFactory.cpp#L62), the Python version used to set that key with the result of `displayName` in the returned dict, before we made it a proper struct. This can most likely go.

src/openassetio-python/package/openassetio/constants.py
foundrytom: Its not for code simplification, its about not acquiring the GIL in a python implementation that has a static prefix. There is no guarantee that this is ever the case (some managers could consume file paths as entity references if they are under a certain filesystem root for example).

A first class method might be a nice way to formalize it if we prefer that. The reason it was in there, is there are also a bunch of other things we used to put in there eg: `Icon`, `Smallcon` etc... that were used by integrations. So I guess we don't want to snowball into too many methods.

src/openassetio-python/package/openassetio/constants.py
foundrytom: This was a legacy from the old `Session` API that used to have a `getSettings` mechanism, that actually added the current manager identifier into the settings dict itself, so that you could in theory store/reacall the whole thing in a host. We binned all that off [ages ago](https://github.com/OpenAssetIO/OpenAssetIO/pull/525/files#diff-fc4ebba0bbc6ffa1aaac5024dcb536722e303c6c695f452372a7ccb474aa6411L192).

src/openassetio-python/package/openassetio/constants.py
foundrytom: Display name? [What display name?](https://github.com/OpenAssetIO/OpenAssetIO/pull/1001/commits/a4f10c1b4169ccbc3d86a2c240effdbbc47a6cd3)

src/openassetio-python/package/openassetio/constants.py
feltech: Ah, that damned GIL.  Makes sense.

feltech (approved):
LGTM

feltech (approved):

PR #1005, opened 2023-07-05
Migrate `flushCaches` to C++

### LINKED ISSUE #994: Migrate `flushCaches` to C+
## ACs

- Default noop implementation
- Manager test harness flushes caches between each test case.

feltech:
## Description

Closes #994

### DIFF
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -171,6 +171,15 @@ class OPENASSETIO_CORE_EXPORT Manager {
   void initialize(InfoDictionary managerSettings);
 
   /**
+   * Clears any internal caches.
+   *
+   * Only applicable if the manager makes use of any caching, otherwise
+   * it is a no-op.  In caching interfaces, this should cause any
+   * retained data to be discarded to ensure future queries are fresh.
+   */
+  void flushCaches();
+
+  /**
    * @}
    */
 
--- a/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
+++ b/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
@@ -305,6 +305,17 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
   virtual void initialize(InfoDictionary managerSettings, const HostSessionPtr& hostSession) = 0;
 
   /**
+   * Clears any internal caches.
+   *
+   * Only applicable if the implementation makes use of any caching,
+   * otherwise it is a no-op. In caching interfaces, this will cause any
+   * retained data to be discarded to ensure future queries are fresh.
+   *
+   * @param hostSession The API session.
+   */
+  virtual void flushCaches(const HostSessionPtr& hostSession);
+
+  /**
    * @}
    */
 
--- a/src/openassetio-python/tests/package/hostApi/test_manager.py
+++ b/src/openassetio-python/tests/package/hostApi/test_manager.py
@@ -203,15 +203,15 @@ class Test_Manager_initialize:
 
 
 class Test_Manager_flushCaches:
-    def test_method_defined_in_python(self, method_introspector):
-        assert method_introspector.is_defined_in_python(Manager.flushCaches)
+    def test_method_defined_in_cpp(self, method_introspector):
+        assert not method_introspector.is_defined_in_python(Manager.flushCaches)
         assert method_introspector.is_implemented_once(Manager, "flushCaches")
 
     def test_wraps_the_corresponding_method_of_the_held_interface(
         self, manager, mock_manager_interface, a_host_session
     ):
         method = mock_manager_interface.mock.flushCaches
-        assert manager.flushCaches() == method.return_value
+        manager.flushCaches()
         method.assert_called_once_with(a_host_session)
 
 
--- a/src/openassetio-python/tests/package/managerApi/test_managerinterface.py
+++ b/src/openassetio-python/tests/package/managerApi/test_managerinterface.py
@@ -71,6 +71,15 @@ class Test_ManagerInterface_info:
         assert info == {}
 
 
+class Test_ManagerInterface_flushCaches:
+    def test_method_defined_in_cpp(self, method_introspector):
+        assert not method_introspector.is_defined_in_python(ManagerInterface.flushCaches)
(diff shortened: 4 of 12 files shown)


src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -170,6 +170,17 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
   void initialize(InfoDictionary managerSettings);
 
+  /**
+   * Clears any internal caches.
+   *
+   * Only applicable if the manager makes use of any caching, otherwise
+   * it is a no-op.  In caching interfaces, this should cause any
+   * retained data to be discarded to ensure future queries are fresh.
+   *
+   * @unstable
```
feltech, line 180: [question] Is `flushCaches` still `@unstable`?

foundrytom: I don't think so now we've decided to keep it.

foundrytom (approved):
Lovely. LGTM once unstable is removed

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Removed and force-pushed: https://github.com/OpenAssetIO/OpenAssetIO/compare/5c68a0cadcf4b23ee00a9a826675c7580c194e21..15e96ff36b2b862da190a3e7dbcfb913ae007d92

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -304,6 +304,18 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
    */
   virtual void initialize(InfoDictionary managerSettings, const HostSessionPtr& hostSession) = 0;
 
+  /**
+   * Clears any internal caches.
+   *
+   * Only applicable if the implementation makes use of any caching,
+   * otherwise it is a no-op. In caching interfaces, this should cause
```
elliotcmorris, line 311: [consider][minor] "Should" is a bit of a weasel word, i'd prefer "May" if this is an optional behaviour, or "Will" if it's guaranteed.

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -203,15 +203,15 @@ def test_wraps_the_corresponding_method_of_the_held_interface(
...
+    def test_method_defined_in_cpp(self, method_introspector):
+        assert not method_introspector.is_defined_in_python(Manager.flushCaches)
         assert method_introspector.is_implemented_once(Manager, "flushCaches")
 
     def test_wraps_the_corresponding_method_of_the_held_interface(
         self, manager, mock_manager_interface, a_host_session
     ):
         method = mock_manager_interface.mock.flushCaches
-        assert manager.flushCaches() == method.return_value
+        assert manager.flushCaches() is None
```
elliotcmorris, line 214: [Consider] Is it necessary to test for `None` at all now? Not sure we do this for other void returns.

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
```diff
@@ -71,6 +71,15 @@ def test_when_not_overridden_then_returns_empty_dict(self):
         assert info == {}
 
 
+class Test_ManagerInterface_flushCaches:
+    def test_method_defined_in_cpp(self, method_introspector):
+        assert not method_introspector.is_defined_in_python(ManagerInterface.flushCaches)
+        assert method_introspector.is_implemented_once(ManagerInterface, "flushCaches")
+
+    def test_is_noop(self, a_host_session):
+        assert ManagerInterface().flushCaches(a_host_session) is None
```
elliotcmorris, line 80: [fix] Similarly, is this testing anything? I think the name of this test at least needs to change, as it would still pass even if `flushCaches` here wasn't a noop, since it always returns void (and thus evaluates to `None`)

elliotcmorris (changes requested):
Minor wording stuff.

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: It was just a way that I could still use the word `assert` in the test case, rather than just calling the function. You're right though, we otherwise don't test void returns.  I'll get rid of it.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Copy-pasted docstring, but reasonable argument, I'll update

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
feltech: It was just a way that I could still use the word assert in the test case, rather than just calling the function. You're right though, we otherwise don't test void returns. I'll get rid of it.

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: Updated as part of d0c780cb

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
feltech: Updated as part of d0c780cb

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Updated in ca036121

elliotcmorris: > Copy-pasted docstring, but reasonable argument, I'll update

We should maybe actually discuss this in retro, it keeps coming up.

elliotcmorris (approved):

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
foundrytom: I think "must" was the original intention of the API contract. If a manager has caches, it has to clear them when this method is called to be compliant. The only reason cache might not be cleared is if there aren't any...

feltech: Updated to "will" in d0e8282f

PR #1011, opened 2023-07-06
`isEntityReferenceString` prefix optimisation

### LINKED ISSUE #566: [Core] Support `kField_EntityReferencesMatchPrefix`
## What

Add support in the C++ implementation of `isEntityReferenceString` for a prefix-based default via the `kField_EntityReferencesMatchPrefix` property of the manager's `info()` dictionary; such that when this key is provided, the `Manager` wrapper doesn't invoke the corresponding method of the held interface.

## Why

When a Python based manager implementation is used, the cost of acquiring the GIL and entering Python in order to invoke the a simple string prefix check can be prohibitive.

We have several docstrings that refer to the `kField_EntityReferencesMatchPrefix` optimisation/convenience of `isEntityReferenceString` . However, it is not currently supported, making the docs invalid.

## ACs

- When `isEntityReferenceString` is called on `Manager`, if the interfaces info provides the aforementioned key then the implementations corresponding method is not called, and the method returns if the input string begins with the specified prefix.
-  `info` should only be called once per `Manager` instantiation`

## Notes

Caused by work on #549.

A prerequisite is actually migrating the constant to C++, i.e. #354

feltech:
## Description

Closes #566

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -77,6 +77,16 @@ v1.0.0-alpha.xx
   chance to clean up between tests.
   [#994](https://github.com/OpenAssetIO/OpenAssetIO/issues/994)
 
+### Bug fixes
+
+- Reintroduced the optional optimized entity reference prefix check
+  for `isEntityReferenceString`, allowing the plugin's implementation to
+  be short-circuited. In particular, if the plugin's implementation is
+  written in Python, then a prefix check short-circuits the need for a
+  costly Python function call for this hot code path.
+  [#566](https://github.com/OpenAssetIO/OpenAssetIO/issues/566)
+
+
 v1.0.0-alpha.12
 ---------------
 
--- a/src/openassetio-core/src/hostApi/Manager.cpp
+++ b/src/openassetio-core/src/hostApi/Manager.cpp
@@ -5,8 +5,10 @@
 
 #include <openassetio/Context.hpp>
 #include <openassetio/TraitsData.hpp>
+#include <openassetio/constants.hpp>
 #include <openassetio/hostApi/EntityReferencePager.hpp>
 #include <openassetio/hostApi/Manager.hpp>
+#include <openassetio/log/LoggerInterface.hpp>
 #include <openassetio/managerApi/EntityReferencePagerInterface.hpp>
 #include <openassetio/managerApi/HostSession.hpp>
 #include <openassetio/managerApi/ManagerInterface.hpp>
@@ -37,7 +39,34 @@ void throwFromBatchElementError(std::size_t index, BatchElementError error) {
   }
 }
 
+/**
+ * Extract the entity reference prefix from a manager plugin's info
+ * dictionary, if available.
+ */
+std::optional<Str> entityReferencePrefixFromInfo(const log::LoggerInterfacePtr &logger,
+                                                 const InfoDictionary &info) {
+  // Check if the info dict has the prefix key.
+  if (auto iter = info.find(Str{constants::kInfoKey_EntityReferencesMatchPrefix});
+      iter != info.end()) {
+    if (const auto *prefixPtr = std::get_if<openassetio::Str>(&iter->second)) {
+      std::string msg = "Entity reference prefix '";
+      msg += *prefixPtr;
+      msg +=
+          "' provided by manager's info() dict. Subsequent calls to isEntityReferenceString will"
+          " use this prefix rather than call the manager's implementation.";
+      logger->debugApi(msg);
+
+      return *prefixPtr;
+    }
+
+    logger->warning("Entity reference prefix given but is an invalid type: should be a string.");
+  }
+
+  // Prefix string not found, so return unset optional.
+  return {};
+}
 }  // namespace
+
 namespace hostApi {
 
 ManagerPtr Manager::make(managerApi::ManagerInterfacePtr managerInterface,
@@ -60,6 +89,9 @@ InfoDictionary Manager::settings() const { return managerInterface_->settings(ho
 
 void Manager::initialize(InfoDictionary managerSettings) {
   managerInterface_->initialize(std::move(managerSettings), hostSession_);
+
+  entityReferencePrefix_ =
+      entityReferencePrefixFromInfo(hostSession_->logger(), managerInterface_->info());
 }
 
 void Manager::flushCaches() { managerInterface_->flushCaches(hostSession_); }
@@ -104,7 +136,11 @@ ContextPtr Manager::contextFromPersistenceToken(const Str &token) {
 }
 
 bool Manager::isEntityReferenceString(const Str &someString) const {
-  return managerInterface_->isEntityReferenceString(someString, hostSession_);
+  if (!entityReferencePrefix_) {
+    return managerInterface_->isEntityReferenceString(someString, hostSession_);
+  }
+
+  return someString.rfind(*entityReferencePrefix_, 0) != Str::npos;
 }
 
 const Str kCreateEntityReferenceErrorMessage = "Invalid entity reference: ";
--- a/src/openassetio-python/tests/package/hostApi/test_manager.py
+++ b/src/openassetio-python/tests/package/hostApi/test_manager.py
@@ -35,6 +35,7 @@ from openassetio import (
     EntityReference,
     TraitsData,
     managerApi,
+    constants,
 )
 from openassetio.hostApi import Manager, EntityReferencePager
 from openassetio.managerApi import EntityReferencePagerInterface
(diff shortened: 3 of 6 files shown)


src/openassetio-core/src/hostApi/Manager.cpp
```diff
@@ -46,9 +47,32 @@ ManagerPtr Manager::make(managerApi::ManagerInterfacePtr managerInterface,
...
+  return {};
+}
+}  // namespace
+
 Manager::Manager(managerApi::ManagerInterfacePtr managerInterface,
                  managerApi::HostSessionPtr hostSession)
-    : managerInterface_{std::move(managerInterface)}, hostSession_{std::move(hostSession)} {}
+    : managerInterface_{std::move(managerInterface)},
+      hostSession_{std::move(hostSession)},
+      entityReferencePrefix_{entityReferencePrefixFromInfo(managerInterface_->info())} {}
```
foundrytom, line 75: [fix] Sorry should have thought of this when we wrote it up - do you think we could log a `kDebugApi` message to state when the shortcut is activated and note that the managers own method wont be called?

src/openassetio-core/src/hostApi/Manager.cpp
```diff
@@ -46,9 +47,32 @@ ManagerPtr Manager::make(managerApi::ManagerInterfacePtr managerInterface,
...
+namespace {
+/**
+ * Extract the entity reference prefix from a manager plugin's info
+ * dictionary, if available.
+ */
+std::optional<Str> entityReferencePrefixFromInfo(const InfoDictionary &info) {
+  // Check if the info dict has the prefix key.
+  if (auto iter = info.find(Str{constants::kInfoKey_EntityReferencesMatchPrefix});
+      iter != info.end()) {
+    // Extract prefix string, and fail silently if the wrong type is
```
foundrytom, line 59: [consider] Worth a `kDebugApi` method if the type is wrong?

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -230,6 +231,32 @@ def test_wraps_the_corresponding_method_of_the_held_interface(
         assert manager.isEntityReferenceString(a_ref_string) == expected
         method.assert_called_once_with(a_ref_string, a_host_session)
 
+    @pytest.mark.parametrize(
+        # Override `manager_info_dict` fixture here, so that it is used
+        # when the `manager` fixture is constructed for the test case.
+        "manager_info_dict,expected",
+        (
+            ({constants.kInfoKey_EntityReferencesMatchPrefix: "asset://"}, True),
```
foundrytom, line 239: [fix] May we sneak some unicode in there?

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -230,6 +231,32 @@ def test_wraps_the_corresponding_method_of_the_held_interface(
...
+            ({constants.kInfoKey_EntityReferencesMatchPrefix: "a"}, True),
+            ({constants.kInfoKey_EntityReferencesMatchPrefix: "s"}, False),
+            (
+                {
+                    constants.kInfoKey_EntityReferencesMatchPrefix: (
+                        "some_really_long_prefix_thats_longer_than_the_ref"
```
foundrytom, line 245: ❤️

foundrytom (changes requested):
How elegant is that - in all ways - nice one!

src/openassetio-core/src/hostApi/Manager.cpp
feltech: Added (force-pushed) - though did a `warning` rather than debug log.

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: Added (force-pushed). 

As an aside, I tried to find a UTF-8 character such that when a byte was truncated it would become another valid character -  obviously a problem for a prefix match based on `char` bytes.  That led me to discover a little bit about UTF-8 encoding and the fact that such an issue is impossible (assuming valid UTF-8 codepoints in both strings).

src/openassetio-core/src/hostApi/Manager.cpp
```diff
@@ -37,7 +39,34 @@ void throwFromBatchElementError(std::size_t index, BatchElementError error) {
...
+ * Extract the entity reference prefix from a manager plugin's info
+ * dictionary, if available.
+ */
+std::optional<Str> entityReferencePrefixFromInfo(const log::LoggerInterfacePtr &logger,
+                                                 const InfoDictionary &info) {
+  // Check if the info dict has the prefix key.
+  if (auto iter = info.find(Str{constants::kInfoKey_EntityReferencesMatchPrefix});
+      iter != info.end()) {
+    if (const auto *prefixPtr = std::get_if<openassetio::Str>(&iter->second)) {
+      std::string msg = "Entity reference prefix '";
```
feltech, line 52: [question] It would be good to avoid constructing this string if the logger severity is such that it won't get logged.

We could
* Merge `SeverityFilter` with `LoggerInterface` so that we can check logger severity on any  `LoggerInterface`.
* Have `HostSession` require a `SeverityFilter` as its logger rather than the more generic `LoggerInterface`.
* `dynamic_cast<SeverityFilter>` and log regardless if the cast fails, but use the severity level if the cast succeeds.

I'm not too concerned about this particular function, since it's very rarely called.  But could add the `dynamic_cast<SeverityFilter>` option if it seems worthwhile?

feltech: That third option should really be factored into a common utility, e.g. `static` function on `SeverityFilter`.

```
bool SeverityFilter::isSeverityLogged(const LoggerInterfacePtr& logger, const Severity severity) {
  if (auto* severityFilteredLogger = dynamic_cast<SeverityFilter*>(logger.get())) {
    return severity >= severityFilteredLogger->getSeverity();
  }

  return true;
}
```

Which would then need its own tests and Python bindings etc etc, so I don't propose to do as part of this PR.

foundrytom: `SeverityFilter`/`ConsoleLogger` are just conveniences. Most real-world applications bridge these methods into their own logging systems, which have their own build-in severity controls, so I think you'd have to go with the something like the first option, and also add a `Logger` wrapper that anyone logging a message uses. Plus ask host authors to update its severity when the user changes the presented severity?

RELEASE_NOTES.md
```diff
@@ -68,6 +68,13 @@ v1.0.0-alpha.xx
   objects, respectively.
   [#914](https://github.com/OpenAssetIO/OpenAssetIO/issues/914)
 
+### Bug fixes
+
+- Reintroduced the optional optimized entity reference prefix check
+  for `isEntityReferenceString`.
```
foundrytom, line 74: [fix] Can we elaborate that it will short-circuit python?

src/openassetio-core/src/hostApi/Manager.cpp
feltech: > also add a Logger wrapper that anyone logging a message uses. Plus ask host authors to update its severity when the user changes the presented severity?

I'm not clear on why these are required?

Would adding a single method to `LoggerInterface` do the trick? 
```
virtual bool isSeverityLogged(Severity severity) { return true; }
```
which should work since the `Severity` enum is a child of the `LoggerInterface`.

foundrytom: Oh, sorry yes, I see what you mean, yeah if we're happy that calling back into the host to ask this needs to be as cheap as possible. I guess we'd at least want to wrap it up into macros (or something) that get used instead of the host session methods then too, as for it to work properly at saving the string overheads, no code should ever call any of the log methods directly? Just wondering if in that case, we don't want to leave them lying around in such a prominent place?

feltech: As discussed offline, we should add a convenience macro too, but the limitation of the macro is it's a one-liner, whereas constructing a log string can span multiple lines, so in general both patterns will be required.

foundrytom: [plan] Would you mind making an issue please? And we can pick this up in another PR.

feltech: Added as https://github.com/OpenAssetIO/OpenAssetIO/issues/1014

RELEASE_NOTES.md
feltech: Elaborated in ab9b2c84, how's that?

foundrytom: Epic, thanks!

foundrytom (approved):

PR #1013, opened 2023-07-08
[Core] Add close method to PagerInterface

### LINKED ISSUE #990: Add 'close' method to EntityReferencePagerInterface
In the PagerInterface, we currently say that the descruction of the object is the mechanism that the host signals that the query is over.

This has some problems that we should consider :
 - Does this encourage throwing in constructors?
 - Is overriding `del` pythonic
 - Is python destruction too unpredictable/non standard to rely on this.
 - Is it fine assuming that the manager implementer understands the lifetime implications of creating a pagerInterface, and having it be held alive by the pager object.
 
 The obvious alternate suggestion is to add a "closed" or "finished" or some other finalizing method to the pagerinterface, and call it in the destructor of pager.

_Triggered from a comment posted by @elliotcmorris in https://github.com/OpenAssetIO/OpenAssetIO/pull/986#discussion_r1238794887_

## ACs

- `close` method added to `PagerInterface`, called from destructor of `Pager` in try catch + error.
- Docs in interface state only called once and other methods wont be called, method must not throw.
- Default noop implementation

elliotcmorris:
## Description
Add `close` method to pager interface, called upon destruction of pager

There were various subtle issues with tying the notification of the end of the page query to the destruction of the pager interface. Aside from relying on lifetime like that not being very pythonic (having to override del,) it creates an implicit usage assumption that the PagerInterface is not declared in a greater scope than the pager. Not to mention that the prior design encourages throwing in destructors moreso than this new one.

Adding a close method binds the cleanup entry hook directly to the lifetime of the Pager, rather than the PagerInterface.
Closes #990

~~I have updated all relevant user documentation.~~

## Reviewer Notes

This comes with a single additional test that just checks the method is called after a `del`. Tbh I couldn't think of any further surface that might need testing. Let me know if anything pops to mind.

### DIFF
--- a/src/openassetio-core/include/openassetio/managerApi/EntityReferencePagerInterface.hpp
+++ b/src/openassetio-core/include/openassetio/managerApi/EntityReferencePagerInterface.hpp
@@ -113,6 +113,20 @@ class OPENASSETIO_CORE_EXPORT EntityReferencePagerInterface {
    * return an empty page.
    */
   virtual void next(const HostSessionPtr&) = 0;
+
+  /**
+   * Close the paging query.
+   *
+   * Signals that the host is finished with the paging query,
+   * allowing for any potential cleanup that may need to be performed.
+   * This method is guaranteed to be called only once, and no other
+   * interface methods will be called by the host thereafter.
+   *
+   * This method is called from a destructor. Thrown exceptions will be
+   * caught and logged as errors if possible. Despite that, throwing
+   * from this function is nonetheless discouraged.
+   */
+  virtual void close(const HostSessionPtr& hostSession);
 };
 static_assert(!std::is_copy_constructible_v<EntityReferencePagerInterface>);
 static_assert(!std::is_copy_assignable_v<EntityReferencePagerInterface>);
--- a/src/openassetio-core/src/hostApi/EntityReferencePager.cpp
+++ b/src/openassetio-core/src/hostApi/EntityReferencePager.cpp
@@ -3,7 +3,9 @@
 
 #include <openassetio/EntityReference.hpp>
 #include <openassetio/hostApi/EntityReferencePager.hpp>
+#include <openassetio/log/LoggerInterface.hpp>
 #include <openassetio/managerApi/EntityReferencePagerInterface.hpp>
+#include <openassetio/managerApi/HostSession.hpp>
 
 namespace openassetio {
 inline namespace OPENASSETIO_CORE_ABI_VERSION {
@@ -20,6 +22,17 @@ EntityReferencePager::EntityReferencePager(
     managerApi::EntityReferencePagerInterfacePtr pagerInterface,
     managerApi::HostSessionPtr hostSession)
     : pagerInterface_(std::move(pagerInterface)), hostSession_(std::move(hostSession)) {}
+
+EntityReferencePager::~EntityReferencePager() {
+  try {
+    pagerInterface_->close(hostSession_);
+  } catch (const std::exception& ex) {
+    hostSession_->logger()->error(ex.what());
+  } catch (...) {
+    hostSession_->logger()->error(
+        "Unknown non-exception object caught during destruction of EntityReferencePager");
+  }
+}
 
 bool EntityReferencePager::hasNext() { return pagerInterface_->hasNext(hostSession_); }
 
--- a/src/openassetio-core/src/managerApi/EntityReferencePagerInterface.cpp
+++ b/src/openassetio-core/src/managerApi/EntityReferencePagerInterface.cpp
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2023 The Foundry Visionmongers Ltd
+#include <openassetio/managerApi/EntityReferencePagerInterface.hpp>
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace managerApi {
+
+void EntityReferencePagerInterface::close([[maybe_unused]] const HostSessionPtr& hostSession) {}
+}  // namespace managerApi
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
--- a/src/openassetio-python/tests/package/hostApi/test_entityreferencepager.py
+++ b/src/openassetio-python/tests/package/hostApi/test_entityreferencepager.py
@@ -27,6 +27,7 @@ import weakref
 
 from openassetio import EntityReference
 from openassetio.hostApi import Manager, EntityReferencePager
+from openassetio.log import LoggerInterface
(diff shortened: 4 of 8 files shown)


RELEASE_NOTES.md
```diff
@@ -77,6 +77,13 @@ v1.0.0-alpha.xx
   chance to clean up between tests.
   [#994](https://github.com/OpenAssetIO/OpenAssetIO/issues/994)
 
+- `close` method added to `EntityReferencePagerInterface` for managers
+  to override in order to respond to hosts closing paging queries.
+  This method is now called upon `EntityReferencePager` destruction,
+  supplanting the prior recommended method of reacting in the
+  destructor/del method of `EntityReferencePagerInterface`
+  [#990](https://github.com/OpenAssetIO/OpenAssetIO/issues/990)
```
foundrytom, line 85: [fix] We can remove this, as we never released with the old approach (see above).

foundrytom (approved):
LGTM - thanks - just a minor release notes tweak.

src/openassetio-core/src/managerApi/EntityReferencePagerInterface.cpp
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2023 The Foundry Visionmongers Ltd
+#include <openassetio/managerApi/EntityReferencePagerInterface.hpp>
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace managerApi {
+
+void EntityReferencePagerInterface::close([[maybe_unused]] const HostSessionPtr& hsp) {}
```
feltech, line 9: [consider][minor] Consistency
```suggestion
void EntityReferencePagerInterface::close([[maybe_unused]] const HostSessionPtr& hostSession) {}
```

src/openassetio-python/tests/package/hostApi/test_entityreferencepager.py
```diff
@@ -137,6 +137,13 @@ def makePagerInScope():
         # destroyed.
         assert weak_interface_ref() is None
 
+    def test_when_EntityReferencePager_destructed_close_is_called(
```
feltech, line 140: [fix] Could do with a test for when an exception is thrown in `close`.

I don't think it's possible to test the non-`exception` exception case from Python, but we could at least test the `std::exception` case.

feltech (changes requested):

src/openassetio-core/src/hostApi/EntityReferencePager.cpp
```diff
@@ -21,6 +23,17 @@ EntityReferencePager::EntityReferencePager(
...
     : pagerInterface_(std::move(pagerInterface)), hostSession_(std::move(hostSession)) {}
 
+EntityReferencePager::~EntityReferencePager() {
+  try {
+    pagerInterface_->close(hostSession_);
+  } catch (const std::exception& ex) {
+    hostSession_->logger()->error(ex.what());
+  } catch (...) {
+    hostSession_->logger()->error(
+        "Unknown exception caught from EntityReferencePagerInterface::close");
```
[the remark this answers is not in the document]
elliotcmorris: I'll go with that for consistency, although it's worth remembering that just cause something dosen't inherit from `std::exception` dosen't necessarily mean it's not an exception object. (Although, practically I expect 99% of things that are intended to be exceptions do in fact inherit from std::exception)

src/openassetio-python/tests/package/hostApi/test_entityreferencepager.py
elliotcmorris: Yeah good point.

https://github.com/OpenAssetIO/OpenAssetIO/pull/1013/commits/5562006bb300f4ea094be3dda8dc339d86114e4f

src/openassetio-core/src/hostApi/EntityReferencePager.cpp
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1013/commits/4f541a78dd1505ef737a45c2317def4a023b83c3

src/openassetio-core/include/openassetio/managerApi/EntityReferencePagerInterface.hpp
```diff
@@ -113,6 +113,20 @@ class OPENASSETIO_CORE_EXPORT EntityReferencePagerInterface {
...
+   * Signals that the host is finished with the paging query,
+   * allowing for any potential cleanup that may need to be performed.
+   * This method is guaranteed to be called only once, and no other
+   * interface methods will be called by the host thereafter.
+   *
+   * This method is called from a destructor. Thrown exceptions will be
+   * caught and logged as errors if possible. Despite that, throwing
+   * from this function is nonetheless discouraged.
+   */
+  virtual void close(const HostSessionPtr& hsp);
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1013/commits/a3c932e63ee863be8661670c3f4803068972c224

RELEASE_NOTES.md
elliotcmorris: Oh right yeah, nice catch.

https://github.com/OpenAssetIO/OpenAssetIO/pull/1013/commits/1dc2d9f3c77226dfd026ed2c40055c829787eeb2

src/openassetio-core/src/managerApi/EntityReferencePagerInterface.cpp
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/commit/a3c932e63ee863be8661670c3f4803068972c224

src/openassetio-python/tests/package/hostApi/test_entityreferencepager.py
elliotcmorris: I tried doing a `raise 1` to see if I could throw an int but you're right, python complains that all exceptions must be exception types.

RELEASE_NOTES.md
feltech: Resolving on behalf of Tom since he's away

src/openassetio-core/src/hostApi/EntityReferencePager.cpp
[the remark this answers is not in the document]
feltech: > just cause something dosen't inherit from std::exception dosen't necessarily mean it's not an exception object

That's exactly the ambiguity I was referring to and trying to avoid, it's tricky to word succinctly though.  I'm pretty sure if a C++ dev sees "non-exception object caught" they'll understand what that means, so this LGTM.

feltech (approved):

PR #1030, opened 2023-08-02
Investigate pybind11 default args

### LINKED ISSUE #987: Investigate if Python defaulted `resultTraitSet` mutations persist in `getWithRelationship[s]` methods
## What

Ensure the default empty `resultTraitSet` argument of the `Manager.getWithRelationship[s]` methods cannot be modified by the manager plugin, such that subsequent calls see the modified data.

## Why

This is a common pitfall of Python default arguments, but does it also apply to pybind11 default arguments?

feltech:
## Description

Closes #987.  Add a couple of tests to assert the finding that `unordered_set` is safe to use as a default argument value in pybind11 bindings.  Add a short explanation of pybind11 default argument pitfalls in CODING_STANDARDS.

## Background

TL;DR: the `unordered_set` we supply as a default value to `py::arg(...)` is copied into a Python `set` and stored by pybind11, then recalled when needed, where it is copied back to an `unordered_set` to pass to C++.

A common pitfall in pure Python is setting a default value for an argument in a function to a mutable value, e.g.

```python 
def someFunction(*args, resultTraitSet=set()): ... 
```

In the above, any mutation to the default-provided `resultTraitSet` will persist in the next invocation.

In our pybind11 bindings we have a similar pattern, i.e.

```c++ 
py::arg("resultTraitSet") = trait::TraitSet{}
```

where `TraitSet` is an alias for `std::unordered_set`.

It turns out that this is not a problem, since we're relying on pybind11's automatic casting between a Python `set` and `unordered_set`, which necessitates a copy of the default value into a new C++ value on every invocation.

Stepping through in the debugger reveals that in pybind11 default arguments are first converted to a Python object before being squirrelled away in a `function_record`.

When calling the C++ function from Python, and supplying an insufficient number of arguments, the squirrelled-away default Python value is recalled (in `cpp_function::dispatcher`) then `cast`ed from the Python object back to a C++ object.

This is the bit that surprised me the most: to reiterate, the default value is stored in the function record as a _Python object_ converted from the input C++ object.  So subsequent use of the default value must obey whatever Python->C++ conversion is appropriate to that type.

For `unordered_set`, this means a Python `set` is stored on the `function_record`. `casti`ng is then done via an explicit `set_caster`, which constructs a new `std::unordered_set` by copying entries across from the Python `set`. This new `unordered_set` is then given to the C++ function.

On the other hand, a default value to an argument of a _generic_ user-defined type is stored in the `function_record` as a Python object pointing to a heap-allocated C++ object that was in turn move-constructed from the default value given in `py::arg(...) = ...`.  This is `cast`ed via a `type_caster_base`, which simply extracts the C++ pointer stored in the Python object, so no copies are made.

## Previous incarnation

This PR previously contained additional tests/logging to explore the inner workings of pybind11 default argument values. 

The original commits are still available under [a branch on my fork](https://github.com/feltech/OpenAssetIO/tree/prototype/987-investigatePybindDefaultArgs).

The following sections relate to that work.

### Reviewer Notes

Probably not worth merging any of this.  Certainly the `drop!` commit just pollutes things and is only useful for gdb explorations.  The first commit may have some value, though?

The test in the `drop!` commit intentionally fails.

### Test Instructions

Run the tests as normal, but don't capture any `stdout` output, in order to see the logs (from the work in the `drop!` commit).

For gdb exploration, build and install, then debug
```
export PYTHONPATH=build/dist/lib/python3.9/site-packages 
python -m pytest -s src/openassetio-python/tests/cmodule
```
specifically for gdb command-line
```
gdb --args build/test-venv/bin/python -m pytest -s src/openassetio-python/tests/cmodule
```

### DIFF
--- a/doc/contributing/CODING_STANDARDS.md
+++ b/doc/contributing/CODING_STANDARDS.md
@@ -293,6 +293,85 @@ will be converted to `PyRetainingSharedPtr<MyArg1>` or
 `PyRetainingSharedPtr<MyArg2>`, respectively. This is equivalent to
 manually wrapping in a lambda, as shown above.
 
+### Default arguments
+
+pybind11 allows bound methods to have keyword arguments with default
+values, for example
+```c++
+.def("myMethod",
+     &MyClass::myMethod, py::arg("myArg") = myDefaultValue, ...
+```
+
+In general, it is not safe to set the default value to an arbitrary
+user-defined type, since mutations persist to the next invocation.
+
+However, it is safe to set the default value to a primitive type, or
+C++ types that are auto-converted to/from Python native types.
+
+In particular, many STL types are [auto-converted by
+pybind11](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#automatic-conversion)
+to/from Python native types, necessitating a copy, and are therefore
+safe to use as default values.
+
+These types include `std::string`, `std::vector<>`, `std::deque<>`,
+`std::list<>`, `std::array<>`, `std::valarray<>`, `std::set<>`,
+`std::unordered_set<>`, `std::map<>`, and `std::unordered_map<>`.
+
+Note that any values within the container are subject to the same
+conditions.
+
+A more detailed explanation follows.
+
+#### Explanation
+
+We have to be careful with what types we use when setting default
+values. This is analogous to the well-known common pitfall in pure
+Python, where a keyword argument's default value is set to a mutable
+type, for example
+
+```python
+def myMethod(self, myArg=set()):
+  myArg.add("myValue")
+```
+
+Here, subsequent calls to `myMethod` will find that `myArg` is no longer
+empty - the default value has been mutated.
+
+A similar problem exists with pybind11 default argument values, but it
+is subtly different.
+
+When setting a default value for an argument, pybind11 will
+
+1. Convert the default value to a Python object
+2. Store the Python object in the function record.
+
+When a C++ function is called from Python and a defaulted argument not
+provided, then
+
+1. The default value is recalled from the function record and inserted
+   into the input Python arguments.
+2. The standard conversions from Python arguments to C++ arguments then
+   takes place.
+
+When the default value is a generic user-defined type, the stored Python
+object will contain a reference. In this case, the same problem occurs
+with pybind11 as with pure Python - default values can be mutated during
+a C++ function call and those mutations persist to the next invocation.
+
+However, some C++ types are converted to/from native Python types,
+including primitives such as `int` and `float`, but also (by default)
+`std::string` and container types like `std::vector` (converted to
+Python `list`) and `std::unordered_set` (converted to Python `set`).
+
+In these cases, the initial conversion of the C++ default value to a
+Python object (to be stored in the function record) necessitates a copy.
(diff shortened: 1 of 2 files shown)


feltech:
Removed `drop!` commit.  The original commits are still available under [a branch on my fork](https://github.com/feltech/OpenAssetIO/tree/prototype/987-investigatePybindDefaultArgs).

After offline conversation, added some notes to the CODING_STANDARDS to warn future devs of the pitfalls.

doc/contributing/CODING_STANDARDS.md
```diff
@@ -293,6 +293,63 @@ will be converted to `PyRetainingSharedPtr<MyArg1>` or
...
+Python `list`) and `std::unordered_set` (converted to Python `set`).
+
+In these cases, the initial conversion of the C++ default value to a
+Python object (to be stored in the function record) necessitates a copy.
+Similarly, the conversion back to a C++ object when the function is
+called also necessitates a copy.
+
+So for types that are converted to/from native Python types, it is safe
+to assign them as default argument values.  Note that this includes
+e.g. `std::vector` (Python `list`), which is _not_ safe in pure Python.
```
elliotcmorris, line 351: [Question] So is what we're saying here is that we can use default or library types (ie, types that are converted to python types,) as default arguments, but not our own api types?

If that is the case, is it possible to create a list of these types, there can't be too many, any put it in this document as "acceptable default argument types"?

elliotcmorris (commented):
Even in the current formulation, i'm not sure if the tests give us much, you mentioned that they might just be testing pybind and I think I agree with that. I don't mind them going in, but personally I'd drop em and just merge the standards change.

doc/contributing/CODING_STANDARDS.md
feltech: Good point, how about a link to the [pybind11 docs](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#automatic-conversion)?

elliotcmorris: Sound good, although the main point is to have a clear directive. The explanation is fine and good, but what I want to see at the end is a clear instruction as to what I can and can't do, rather than having to read and understand the paragraphs above to deduce the standard.

feltech: I've started with a summary bit in d9308e2d, how does that look?

feltech:
> Do you still think we should just drop them? I'm in two minds, so will go with what you think.

Capturing offline conversation, we tentatively decided to keep the tests.

doc/contributing/CODING_STANDARDS.md
elliotcmorris: Aye looks good.
It's a wee shame we can't put `shared_ptr` there, especially as you've proven it's working at least in the cases where we use it, but tbh if that means sinking more time into this i'd just leave it as is.

elliotcmorris (approved):

doc/contributing/CODING_STANDARDS.md
feltech: I think since `shared_ptr` is the holder type, not the bound type itself, its a whole other kettle of fish to try to explain that.

Issue #1016, opened 2023-07-12
Refine `Context.access` modes to better convey intent


foundrytom:
## What

Refine the `Context.Access` enum to better convey the intent of the host.

## Why

The current options leave ambiguity in the handling of publishing operations (see notes).

## Acceptance Criteria

- Updatet the `Context.Access` modes:
  - See proposal below for details of new modes
  - `isForWrite` method update to include the new `kCreateRelated`
  - `IsMultiple` removed
- Clear documentation of when to use each mode are provided in the OpenAssetIO documentation.
   - Ie: use Write when you have a 1:1 reference, use CreateRelated when you are programmatically creating entities against a reference to another entity.  Note that this means after `preflight`, as you then have a specific working ref, you a host should use `kWrite` for the ensuing `register`. If you're not calling preflight, then `kCreateRelated` should be used still.
   - Elaborate `managementPolicyDocs`
   - `apiComplianceSuite` coverage (spike to determine exact checks - eg, if you dont provide the fixtures, is there a default check for error cases, checks that preflight and register don't return the input ref when its not from preflight)

## Notes

The behavior when publishing is currently defined solely in terms of how the published entities trait set relates to that of the target reference:

1. When the reference points to a non-existant entity, that entity should be created.
2. When the reference points to an existing entity:
   a. When the trait set of the existing entity matches that of the new entity, a new version should be created if allowed, or error.
   b. When the trait set of the existing entity does not match, the new entity as a child or other meaningful relation to the existing entity if permitted, or error.

The fact that most references are provided by the manager (or the manager's user) means that in many situations, this works well, and the intended result is easily discerned.

There are however edge cases that can arise with certain trait sets, or when more programatic operations are undertaken.

### Making a "folder in a folder".

The common file system operation of making a folder inside a folder highlights the case for us in terms that are perhaps more accessible than OpenAssetIO abstract terminology.

As a host, if I wanted to make a folder, inside a folder for which I know the entity reference, using the publishing rules of thumb, I "publish the new folder to its parent", which follows 2b above. There is an issue however, the trait set of the new folder is the same as the parent (as it's a 'folder in a folder'). This means that the manager will assume we are intending 2a. It has now way of knowing that this is meant to be a new folder, rather than an update to the one targeted by the entity reference.

We need a way to tell the manager that the entities being published are always intended to be new.

[...]

PR #1036, opened 2023-08-10
[Core] Refine Context.access modes

elliotcmorris:
## Description
https://github.com/OpenAssetIO/OpenAssetIO/issues/1016

Adds the `kCreateRelated` context access mode.
This is an access mode necessary to signal intent to create related entities, rather than mutations/versions of an entity. This comes up particularly in the "folder in a folder" case, where it's very likely that the regular method of determining version/new, (checking that the traitsets match,) will not be enough.

Also remove the idea of multiple writes, due to ther being no coherent use for these within the core API methods themselves, and so this behavior could be more accurately and coherently expressed as part of the UI layer itself.

Dosen't close the linked issue, as workflow tests havn't been explored yet. OpenAssetIO api surface has been prioritised, see https://github.com/OpenAssetIO/OpenAssetIO/issues/1016#issuecomment-1670967369

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,20 @@ v1.0.0-alpha.x
 
 ### Breaking changes
 
+- Removed `Context.Access.kWriteMultiple` and
+ `Context.Access.kReadMultiple`access patterns, due to not having
+ coherent use cases.
+ [1016](https://github.com/OpenAssetIO/OpenAssetIO/issues/1016)
+
+### New Features
+
+- Added `Context.Access.kCreateRelated` access pattern, to  indicate
+  when a workflow specifically creates a new entity as a relation to an
+  existing one.
+  [1016](https://github.com/OpenAssetIO/OpenAssetIO/issues/1016)
+
+### Breaking changes
+
 - Removed `cpython` dependency from `conanfile.py`. When building
   OpenAssetIO, it is now expected that a development install of the
   appropriate Python version is discoverable on the system.
--- a/src/openassetio-core/include/openassetio/Context.hpp
+++ b/src/openassetio-core/include/openassetio/Context.hpp
@@ -41,10 +41,38 @@ class OPENASSETIO_CORE_EXPORT Context final {
   /**
    * @name Access Pattern
    */
-  enum class Access { kRead, kReadMultiple, kWrite, kWriteMultiple, kUnknown };
+  enum class Access {
+    /**
+     * Host intends to read data. For example, trait property values
+     * obtained via `resolve` may be used to control the loading of data
+     * from a resource, and its subsequent interpretation.
+     */
+    kRead,
+    /**
+     * Host intends to write data.  For example, trait property values
+     * obtained via `resolve` may be used to control the writing of data
+     * to a resource, and specifics of its format or content. During
+     * publishing this must be used whenever the entity reference
+     * explicitly targets the specific entity whose data is being
+     * written. For example creating or updating a simple, unstructured
+     * asset such as an image or other file-based data.
+     */
+    kWrite,
+    /**
+     * Hosts intends to write data for a new entity in relation to
+     * another. During publishing this must be used whenever the entity
+     * reference points to an existing entity, and the intention is to
+     * create a  new, related entity instead of updating the target. For
+     * example, when programmatically creating new  entities under an
+     * existing parent collection, or the publishing of the components
+     * of a structured asset based on a single root entity reference.
+     */
+    kCreateRelated,
+    /// Unknown Access Pattern
+    kUnknown
+  };
 
-  static constexpr std::array kAccessNames{"read", "readMultiple", "write", "writeMultiple",
-                                           "unknown"};
+  static constexpr std::array kAccessNames{"read", "write", "createRelated", "unknown"};
   /// @}
 
   /**
@@ -67,10 +95,12 @@ class OPENASSETIO_CORE_EXPORT Context final {
   /**
    * Describes what the @ref host is intending to do with the data.
    *
-   * For example, when passed to resolve, it specifies if the @ref
-   * host is about to read or write. When configuring a BrowserWidget,
-   * then it will hint as to whether the Host is wanting to choose a
-   * new file name to save, or open an existing one.
+   * For example, when passed to resolve, it specifies if the @ref host
+   * is about to read or write. When configuring a BrowserWidget, then
+   * it will hint as to whether the Host is wanting to choose a new file
+   * name to save, or open an existing one.
+   *
+   * See also @ref create_related "Create related glossary entry".
    */
   Access access;
 
@@ -133,30 +163,17 @@ class OPENASSETIO_CORE_EXPORT Context final {
                                        TraitsDataPtr locale = nullptr,
                                        managerApi::ManagerStateBasePtr managerState = nullptr);
   /**
-   * @return `true` if the context is any of the 'Read' based access
-   * patterns. If the access is unknown (Access::kUnknown), then `false`
-   * is returned.
+   * @return `true` if the context is a 'Read' based access pattern. If
+   * the access is unknown (Access::kUnknown), then `false` is returned.
    */
-  [[nodiscard]] inline bool isForRead() const {
-    return access == Access::kRead || access == Access::kReadMultiple;
-  }
+  [[nodiscard]] inline bool isForRead() const { return access == Access::kRead; }
 
   /**
-   * @return `true` if the context is any of the 'Write' based access
-   * patterns. If the access is unknown (Access::kUnknown), then `false`
-   * is returned.
+   * @return `true` if the context is a 'Write' based access pattern. If
+   * the access is unknown (Access::kUnknown), then `false` is returned.
    */
   [[nodiscard]] inline bool isForWrite() const {
-    return access == Access::kWrite || access == Access::kWriteMultiple;
-  }
-
-  /**
-   * @return `true` if the context is any of the 'Multiple' based access
-   * patterns. If the access is unknown (Access::kUnknown), then `false`
-   * is returned.
-   */
-  [[nodiscard]] inline bool isForMultiple() const {
-    return access == Access::kReadMultiple || access == Access::kWriteMultiple;
+    return access == Access::kWrite || access == Access::kCreateRelated;
   }
 
  private:
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -215,10 +215,12 @@ class OPENASSETIO_CORE_EXPORT Manager {
    *
    *  - The trait set of the entity type in question. This is usually
    *    obtained from the relevant @ref Specification.
-   *  - For read contexts, any additional traits with properties that
-   *    you wish to resolve for that type of entity.
-   *  - For write contexts, any additional traits with properties that
-   *    you wish to publish for that type of entity.
+   *  - For @fqref{Context.Access} "read" contexts, any additional
+   *    traits with properties that you wish to resolve for that type of
+   *    entity.
+   *  - For @fqref{Context.Access} "write" contexts, any additional
+   *    traits with properties that you wish to publish for that type of
+   *    entity.
    *
    * Along with the traits that describe the manager's desired
    * interaction pattern (ones with the `managementPolicy` usage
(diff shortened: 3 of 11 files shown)


RELEASE_NOTES.md
```diff
@@ -3,6 +3,19 @@ Release Notes
...
+- Removed `Context.Access.kWriteMultiple` and
+ `Context.Access.kReadMultiple`access patterns, due to not having
+ coherent use cases.
+ [1016](https://github.com/OpenAssetIO/OpenAssetIO/issues/1016)
+
+### New Features
+
+- Added `Context.Access.kCreateRelated` access pattern, to support use
+  cases where host wants to specify a new related entity is created,
+  but differentiation based on traitset comparison is not possible.
```
foundrytom, line 17: [fix]
The trait-set comparison rules were our implicit way of stating that the entity should be new before, which was un clear - so we added this access to be explicit. Instead of considering whether trait set matches when publishing, a Host should ask "is the reference I have for the specific thing i'm creating, or some other entity that I wish to create a new thing in relation to" and set this accordingly.  How about something like:

```suggestion
- Added `Context.Access.kCreateRelated` access pattern, to  indicate
  when a workflow specifically creates a new entity as a relation to an
  existing one.
```

src/openassetio-core/include/openassetio/Context.hpp
```diff
@@ -41,10 +41,32 @@ class OPENASSETIO_CORE_EXPORT Context final {
   /**
    * @name Access Pattern
    */
-  enum class Access { kRead, kReadMultiple, kWrite, kWriteMultiple, kUnknown };
+  enum class Access {
+    /**
+     * Host intends to read data
```
foundrytom, line 46: [consider] 
```suggestion
     * Host intends to read data. For example, trait property values obtained 
       via `resolve` may be used to control the loading of data from a resource, 
       and its subsequent interpretation.
```

src/openassetio-core/include/openassetio/Context.hpp
```diff
@@ -41,10 +41,32 @@ class OPENASSETIO_CORE_EXPORT Context final {
...
+     */
+    kRead,
+    /**
+     * Host intends to write data. This should be the default choice for
+     * write, unless the conditions for @ref Access.kCreateRelated are
+     * met.
```
foundrytom, line 52: [fix] Attempting to capture the "what is the reference for" aspect:
```suggestion
     * Host intends to write data.  For example, trait property values obtained 
       via `resolve` may be used to control the writing of data to a resource, 
       and specifics of its format or content.
       During publishing this must be used whenever the entity reference 
       explicity targets the specific entity whose data is being written. 
       For example creating or updating a simple, unstructured asset such
       as an image or other file-based data.
```

src/openassetio-core/include/openassetio/Context.hpp
```diff
@@ -41,10 +41,32 @@ class OPENASSETIO_CORE_EXPORT Context final {
...
+     */
+    kWrite,
+    /**
+     * Hosts intends to write related data against a reference to
+     * another entity. This is a specialization of `kWrite`, and should
+     * be used when the host knows up front that it wishes to to publish
+     * a new related entity, and not an update to an existing entity.
+     * The canonical motivating example for this is that host may know
+     * it wishes to publish a new subfolder inside an existing folder,
+     * and not an update to said existing folder.
```
foundrytom, line 62: [fix] Attempting to align with the `kWrite` suggestion

```suggestion
     * Hosts intends to write data for a new entity in relation to another.
       During publishing this must be used whenever the entity reference 
       points to an existing entity, and the intention is to create a  new, 
       related entity instead of updating the target. For example, when 
       programmatically creating new  entities under an existing parent
       collection, or the publishing of the components of a structured
       asset based on a single root entity reference.
```

(I was just trying to avoid 'folder' due to the file-system connotations).

src/openassetio-core/include/openassetio/Context.hpp
```diff
@@ -67,10 +89,29 @@ class OPENASSETIO_CORE_EXPORT Context final {
...
+   *      logical (eg: a child), or error is emitted.
+   *
+   *    * b. When the access is @ref Access.kWrite :
+   *        * i. When the trait set of the existing entity matches that
+   *          of the new entity, the entity should be updated (possibly
+   *          by versioning up). If this is not permitted, an error
+   *          should be emitted.
+   *        * ii. When the trait set of the existing entity does not
+   *          match, behave as per 2a.
    */
```
foundrytom, line 115: [consider] Is it worth refing this to the publishing section in the interface docs, to avoid getting out of sync?

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -1379,13 +1379,12 @@ class OPENASSETIO_CORE_EXPORT Manager {
...
-   * particular trait set in relation to this reference. For example,
-   * if you received this reference in response to browsing for a
-   * target to `kWriteMultiple` and the traits of a
-   * `ShotSpecification`s, then the Manager should have returned you
-   * a reference that you can then register multiple
-   * `ShotSpecification` entities to without error. Each resulting
-   * entity reference should then reference the newly created Shot.
+   * particular trait set in relation to this reference. For example, if
+   * you received this reference in response to browsing for a target to
+   * `kWrite` and the traits of a `ShotSpecification`s, then the Manager
```
foundrytom, line 1384: [fix] ```suggestion
   * `kWrite` and the traits of a `ShotSpecification`, then the Manager
```

**Review by foundrytom — CHANGES_REQUESTED** [review]
Grand, thanks. Haven't checked the code, just some thoughts on docs;

**Review by feltech — CHANGES_REQUESTED** [review]
Looks good, just concerned about tricky API compliance suite coverage.

File: src/openassetio-core/include/openassetio/Context.hpp (hunk 5 above)
Comment (foundrytom) [reply src/openassetio-core/include/openassetio/Context.hpp:115]: Maybe a glossary link then? It feels like it is more important to describe this with all the other information about publishing etc? As these access modes don't just affect publishing etc (eg: getWithRelationship too)

File: src/openassetio-python/package/openassetio/test/manager/apiComplianceSuite.py [hunk 7]
```diff
@@ -183,14 +183,9 @@ def test_calling_with_write_context(self):
         context.access = context.Access.kWrite
         self.__assertPolicyResults(1, context)
 
-    def test_calling_with_read_multiple_context(self):
+    def test_calling_with_createRelated_context(self):
```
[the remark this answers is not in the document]
foundrytom: Lets not get into a massive apiCompliance rabbit hole here. Make an issue and we can work it out a bit more holisticly?

RELEASE_NOTES.md
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1036/commits/0df5a207aa7be7a37f52b73c31078fb1bc575a0c

src/openassetio-core/include/openassetio/Context.hpp
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1036/commits/951242c2cae37bd67fd164f410b8bc8c9e7c53b8

src/openassetio-core/include/openassetio/Context.hpp
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1036/commits/dbf716a78430f6d2f622684718ee844dfdc0435d

Gone with this cause you're the source of truth here, but these docs, (the above one for kRead as well,) are more confusing than illuminating to me personally.

Specifically,
> For example, trait property values obtained via `resolve` may be used to control the writing of data to a resource.

I don't see how this is an example of a host intending to write data, nor do I understand what a trait-property-value being used as a control means.

>  During publishing this must be used whenever the entity reference explicity targets the specific entity whose data is being written. For example creating or updating a simple, unstructured asset such as an image or other file-based data

The first part of that reads _okay_ to me, although it does seem to imply that there's an alternate you would consider as an equal partner to `kWrite`, but the second part, "creating or updating a simple, unstructured asset", I feel strongly implies for more complicated cases there is an alternate you should be looking at, but as far as I know, `kWrite` is the correct applicable context for the vast majority of publish workflows.

As ever, this is likely my fault for being slow, and we should just go with what you've written to get this through.

src/openassetio-core/include/openassetio/Context.hpp
elliotcmorris: Yeah, avoiding folder is wise
https://github.com/OpenAssetIO/OpenAssetIO/pull/1036/commits/08505a52d4e9ac7232f457e4b21c853841ffb416

src/openassetio-core/include/openassetio/Context.hpp
elliotcmorris: Maybe being so specific as to manager side implementation flow in the api docs is folly.

What do you think of this, just a section in the `register` documentation expanding on the previously existing section about lack of opinion?

https://github.com/OpenAssetIO/OpenAssetIO/pull/1036/commits/fffffcbc6fb529ceec9c17a5144fae6ef8df7377

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1036/commits/544d73d7e4a65adf0b6f28e95d9150bdbd8f8341

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -215,9 +215,9 @@ class OPENASSETIO_CORE_EXPORT Manager {
    *
    *  - The trait set of the entity type in question. This is usually
    *    obtained from the relevant @ref Specification.
-   *  - For read contexts, any additional traits with properties that
+   *  - For @fqref{Context.Access} "read" contexts, any additional traits with properties that
    *    you wish to resolve for that type of entity.
-   *  - For write contexts, any additional traits with properties that
+   *  - For @fqref{Context.Access} "write" contexts, any additional traits with properties that
    *    you wish to publish for that type of entity.
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1036/commits/d1aac51259f1344b424c15c955867a406dcfb4b1

src/openassetio-core/include/openassetio/Context.hpp
```diff
@@ -133,30 +174,19 @@ class OPENASSETIO_CORE_EXPORT Context final {
                                        TraitsDataPtr locale = nullptr,
                                        managerApi::ManagerStateBasePtr managerState = nullptr);
   /**
-   * @return `true` if the context is any of the 'Read' based access
-   * patterns. If the access is unknown (Access::kUnknown), then `false`
+   * @return `true` if the context is a 'Read' based access
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1036/commits/d77b3014c620276f0d92d3ff7e3ab8a89c56f202

src/openassetio-python/package/openassetio/test/manager/apiComplianceSuite.py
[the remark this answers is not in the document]
elliotcmorris: [Plan]ned here https://github.com/OpenAssetIO/OpenAssetIO/issues/1040

src/openassetio-core/include/openassetio/Context.hpp
foundrytom: > I don't see how this is an example of a host intending to write data, nor did I really understand what a trait-property-value being used as a control means without thinking about it for a hot minute. I'm guessing you're referring to the idea that a trait value might allow the manager to branch their storage mechanism or reaction based on the value stored? 

e.g The Host uses the resolved value of the LocatableContent trait's location property to determine where to write the file to. Many pipelines put working data in a different place to its final resting place. 

> I feel strongly implies for more complicated cases there is an alternate you should be looking at, but as far as I know, kWrite is the correct applicable context for the vast majority of publish workflows.

kCreateRelated is what you use when you are creating components of a structured asset based on some anchor. So for anything where a host is making more than one thing for against a single reference.

Not being slow at all! Holidays etc have not been kind to us of late in terms of overlap!

feltech (approved):

src/openassetio-core/include/openassetio/Context.hpp
foundrytom: Perfec 👍 🙏

foundrytom:
> LGTM. I do note there are quite a few docs remaining that talk about `kWrite` without mentioning that `kCreateRelated` is also an option and may be more appropriate, e.g. `ForHosts.dox` and `ForManagers.dox`, along with various examples littered throughout the docs.

Good call - added https://github.com/OpenAssetIO/OpenAssetIO/issues/1043, so we can take a step back once all these related PRs are in.

RELEASE_NOTES.md
```diff
@@ -3,6 +3,19 @@ Release Notes
 
 v1.0.0-alpha.x
 ---------------
+### Breaking changes
```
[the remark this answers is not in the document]
elliotcmorris: [036ada2](https://github.com/OpenAssetIO/OpenAssetIO/pull/1036/commits/036ada2ca85d253b6170ced4df030dccc53fc3dc)

PR #1037, opened 2023-08-10
Preflight `TraitsData` and relax `register` trait set homogeneity

### LINKED ISSUE #1028: Refactor `preflight`
## What

Refactor `preflight` considering #1015 and #1003. 

As a manager plugin
I want to know all the information that the host already has at preflight time
Because I can better prepare my backend system for eventual `register` calls
Because I can tailor my subsequent `resolve`-for-write responses more appropriately

As a host
I want to know early if a publish is likely to fail
Because then I can avoid expensive operations that are not supported

## Why

We cannot know what combination of workflows and host and manager capabilities are going to crop up.

A well-behaved host should give the manager as much information as it can at `preflight` time to have the best chance of the manager supporting intermediate `resolve`-for-write and the final `register` call.

That is, information that the host "owns" should be provided to the manager on `preflight` if it's available.

Information that the host needs in order to perform an operation should be `resolve`d with `kWrite` access from the entity returned by `preflight`.

Either of these cases may be true for a publish, depending on workflow.  

E.g. the host may know beforehand the MIME type of the file it is about to create, and so should communicate that to the manager; alternatively, the host may want to ask the manager what the MIME type of the file should be, so that it can choose an output codec before creating the file.  In either of these cases, retrieving the location to write the file to, via a subsequent `resolve` of the entity returned from `preflight`, will then have the required file extension information.

The same host operating on the same entity type could behave differently depending on the workflow.

## Acceptance Criteria

- `preflight` accepts a `TraitsData` hint per batch element, rather than a per-batch trait set
  * No checks are made to constrain the trait set across the supplied TraitsData
- A new `BatchElementError` type is added to indicate more information is needed in `preflight`
- Supporting documentation for publishing is updated to describe
   * That a host should always provide known data up front.
   * A manager can error if required preflight data is missing.

## Tasks

- [ ] Update BAL to new signature
- [ ] Update `apiComplianceSuite`
- [ ] Update `Manager`/`ManagerInterface`
- [ ] Update examples
- [ ] Add BatchElementError type `kInvalidPreflightHints`

## Examples

Given manager cannot resolve {property} with `kWrite`
When `resolve` is called with `kWrite` to retrieve {property}
Then trait property is unset

Given manager can resolve {property} with `kWrite` to {resolved}
When `resolve` is called with `kWrite` to retrieve {property}
Then trait property is {resolved}

Given manager requires {prelight_property} in `preflight`
When `preflight` is called without {preflight_property}
Then `BatchElementError` is returned with `kInvalidPreflightHint` error code

Given manager will resolve {property} with `kWrite` to {value} only if provided {prelight_property} in `preflight`
And `preflight` is called with {preflight_property}
When `resolve` is called with `kWrite` to retrieve {property}
Then trait property is {value}

Given manager will resolve {property} with `kWrite` to {value} if provided {prelight_property} in `preflight` or otherwise resolves to {other_value}
And `preflight` is called with {preflight_property}
When `resolve` is called with `kWrite` to retrieve {property}
Then trait property is {value}

Given manager will resolve {property} with `kWrite` to {value} if provided {prelight_property} in `preflight` or otherwise resolves to {other_value}
And `preflight` is called without {preflight_property}
When `resolve` is called with `kWrite` to retrieve {property}
Then trait property is {other_value}

### LINKED ISSUE #1029: Remove contraint on unique trait sets in `register`
## What

Remove the assertion in `register` that the trait sets of the supplied `TraitsData` instances should be the same.

## Why

See #1015. 

## Acceptance Criteria

- `register` can be used with divergent trait sets in the supplied `TraitsData` instances.
- A manager can determine if the `TraitsData` instances supplied to the `register` method share a common trait set with a single line of code.
- Docs that describe publishing updated as needed.

## Notes

- No need for `apiComplianceSuite` check (we don't validate batch variance for other methods)

## Tasks

- [x] #1037

feltech:
## Description

Closes #1028
Closes #1029

Changes the signature of `preflight` to accept a `TraitsData` per reference, rather than a single trait set applied to all references. Existing docs and examples updated.

Simultaneously relaxes the restriction that all batch published entities must be of the same type.

## Reviewer Notes

Combined two issues into one PR here, since it avoids adding an artificial restriction to `preflight` to ensure trait set homogeneity, just to remove it again in the next issue.

Commits split into roughly related chunks of work, intended to be squashed into a single commit once approved.

Minimal changes to BAL https://github.com/OpenAssetIO/OpenAssetIO-Manager-BAL/pull/63 to get these API changes in as soon as possible.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,17 @@ v1.0.0-alpha.x
 
 ### Breaking changes
 
+- Changed signature of `preflight` to accept a `TraitsData` per
+  reference, rather than a single trait set. The host is now expected to
+  communicate any relevant information that it owns and is known at
+  `preflight` time.
+  [1028](https://github.com/OpenAssetIO/OpenAssetIO/issues/1028)
+
+- Relaxed the restriction on `register` that all trait sets of the
+  provided `TraitsData` elements must match. This allows batched
+  publishing of heterogeneous entity types.
+  [1029](https://github.com/OpenAssetIO/OpenAssetIO/issues/1029)
+
 - Removed `Context.Access.kWriteMultiple` and
  `Context.Access.kReadMultiple`access patterns, due to not having
  coherent use cases.
--- a/doc/doxygen/src/Examples.dox
+++ b/doc/doxygen/src/Examples.dox
@@ -277,9 +277,13 @@
  *
  * # Whenever we make new data, we always tell the manager first,
  * # This allows it to create a placeholder version or similar.
+ * # We must provide the manager with any relevant information that the
+ * # host owns (i.e. won't be queried from the manager during
+ * # publishing) and can be provided up-front.
+ * file_spec = TextFileSpecification.create()
+ * file_spec.markupTrait().setMarkupType("plain")
  * # NOTE: It is critical to always use the working_ref from now on.
- * working_ref = manager.preflight(
- *         entity_ref, TextFileSpecification.kTraitSet, context)
+ * working_ref = manager.preflight(entity_ref, file_spec, context)
  *
  * # We then check if the manager can tell us where to save the file.
  * if ResolvesFutureEntitiesTrait.isImbuedTo(policy):
@@ -297,7 +301,6 @@
  *
  * # Prepare the entity specification to register, with the data about
  * # where we actually wrote the data to, and with what encoding.
- * file_spec = TextFileSpecification.create()
  * file_spec.locatableContentTrait().setLocation(pathToURL(save_path))
  * file_spec.textEncodingTrait().setEncoding(encoding)
  *
--- a/doc/doxygen/src/Glossary.dox
+++ b/doc/doxygen/src/Glossary.dox
@@ -381,6 +381,38 @@
  * returned @ref entity_reference to determine the manager's chosen
  * location for the data and use this for any write operations.
  *
+ * A host should always @ref publish the full set of @ref trait "traits"
+ * of an appropriate @ref Specification "specification" for the type of
+ * entity being published.
+ *
+ * Some of a specification's traits may hold properties, and some of
+ * those properties may be required by
+ * @fqref{managerApi.ManagerInterface.preflight} "preflight" in order
+ * for subsequent operations in the publishing workflow to succeed.
+ * Other properties may be optional.
+ *
+ * The host should fill in and pass as a "hint" to
+ * @fqref{hostApi.Manager.preflight} "preflight"  any trait properties
+ * that are owned by the host and known at the time `preflight` is
+ * called.
+ *
+ * In this context, a property is "owned" by the host if it will not be
+ * dictated by the manager during the publishing workflow (i.e. by @ref
+ * glossary_resolve "resolving" with a @fqref{Context.Access.kWrite}
+ * "write" access flag). Note that this definition of host ownership
+ * might include properties that have been resolved from the manager at
+ * some point in the past.
+ *
+ * If a property is not already known or understood by the host, and is
+ * optional, then it may be omitted.
+ *
+ * Providing the preflight hint allows the manager to prepare the
+ * backend and provide the most appropriate responses to subsequent @ref
+ * glossary_resolve and @ref glossary_register requests; or
+ * @fqref{BatchElementError.ErrorCode.kInvalidPreflightHint} "error"
+ * early if publishing is likely to fail or insufficient information is
+ * given in the hint.
+ *
  * The `preflight` method is also allowed to change the entity
  * reference as required. This mechanism serves several purposes:
  *
--- a/doc/doxygen/src/Glossary.dox
+++ b/doc/doxygen/src/Glossary.dox
@@ -366,7 +366,7 @@
  * indirection allows sundry house keeping and state management.
  *
  *
- * @section preflight preflight
+ * @section glossary_preflight preflight
  *
  * @ref asset_management_system "Asset management systems" are roughly
  * divided into two camps - those that can tell a @ref host in advance
@@ -401,9 +401,9 @@
  * - When the data already exists at some target location, such as when
  *   assetizing an existing file.
  *
- * In these situations, only @ref register will be called by the host.
- * In the first situation, calling `preflight` would be pointless
- * as the manager has no way to provide a meaningful answer.
+ * In these situations, only @ref glossary_register will be called by
+ * the host. In the first situation, calling `preflight` would be
+ * pointless as the manager has no way to provide a meaningful answer.
  *
  * In the second situation it removes the responsibility for the host to
  * transfer the data from its current location to that specified by
@@ -433,13 +433,13 @@
  *   @ref entity "entity's" @ref trait "traits", via the manager's
  *   @fqref{hostApi.Manager.managementPolicy} "managementPolicy", and
  *   what traits it is capable of persisting property data for.
- * - Call @ref preflight to allow any prerequisite housekeeping
+ * - Call @ref glossary_preflight to allow any prerequisite housekeeping
  *   to be performed, and a working @ref entity_reference returned.
  * - Resolve the working reference to determine the URL for new data to
  *   be written to.
  * - Write the data accordingly.
- * - Call @ref register to finalize the publish and confirm to the
- *   manager that the data has been written.
+ * - Call @ref glossary_register to finalize the publish and confirm to
+ *   the manager that the data has been written.
  *
  * @note Some managers may not be able to determine trait properties in
  * advance, and act more as librarians. This is indicated in the
@@ -447,7 +447,7 @@
  *
  * When publishing existing data, the preflight step is omitted.
  *
- * @section register register
+ * @section glossary_register register
  *
  * The registration mechanism allows for a @ref host to create or update
  * an @ref entity. This is the sole mechanism by which data can be
@@ -472,7 +472,7 @@
  * @fqref{Context.Access.kCreateRelated} "Context.Access.kCreateRelated"
  * in the context is used as a flag to indicate this intent.
  *
- * @section resolve resolve
+ * @section glossary_resolve resolve
  *
  * 'Resolving' an @ref entity_reference is the process of turning it
  * back into the data for one or more @ref trait "traits". Resolution is
@@ -513,9 +513,9 @@
  * A trait may optionally define one or more properties that hold
  * simple-typed values.
  *
- * When an entity is @ref publish "published" or @ref resolve "resolved"
- * through the API, these trait properties are used to hold the entity's
- * data.
+ * When an entity is @ref publish "published" or @ref glossary_resolve
+ * "resolved" through the API, these trait properties are used to hold
+ * the entity's data.
  *
  * @see @ref entities_traits_and_specifications
  *
--- a/doc/doxygen/src/Thumbnails.dox
+++ b/doc/doxygen/src/Thumbnails.dox
@@ -29,13 +29,14 @@
  * "managementPolicy" for any given set of entity @ref trait "traits".
  *
  * If supported by the host, it will then:
- * - Call @ref preflight with the target entity and the traits from the
- *   @needsref ThumbnailSpecification, using the
-     @fqref{Context.Access.kCreateRelated} "Access.kCreateRelated"
-     access mode.
- * - Call @needsref resolve on the returned reference to determine the
- *   desired width/height/format for the thumbnail.
- * - Generate a thumbnail and @ref register it to the same reference.
+ * - Call @ref glossary_preflight with the target entity and the traits
+ *   from the @needsref ThumbnailSpecification, using the
+ *   @fqref{Context.Access.kCreateRelated} "Access.kCreateRelated"
+ *   access mode.
+ * - Call @ref glossary_resolve on the returned reference to determine
+ *   the desired width/height/format for the thumbnail.
+ * - Generate a thumbnail and @ref glossary_register it to the same
+ *   reference.
  *
  * @note The thumbnail generation process may be carried out
  * asynchronously by the host and registration may occur at an arbitrary
--- a/doc/doxygen/src/Thumbnails.dox
+++ b/doc/doxygen/src/Thumbnails.dox
@@ -30,7 +30,7 @@
  *
  * If supported by the host, it will then:
  * - Call @ref glossary_preflight with the target entity and the traits
- *   from the @needsref ThumbnailSpecification, using the
+ *   data from the @needsref ThumbnailSpecification, using the
  *   @fqref{Context.Access.kCreateRelated} "Access.kCreateRelated"
  *   access mode.
  * - Call @ref glossary_resolve on the returned reference to determine
--- a/src/openassetio-core-c/src/managerApi/CManagerInterfaceAdapter.cpp
+++ b/src/openassetio-core-c/src/managerApi/CManagerInterfaceAdapter.cpp
@@ -110,7 +110,7 @@ void CManagerInterfaceAdapter::resolve(
 
 void CManagerInterfaceAdapter::preflight(
     [[maybe_unused]] const EntityReferences& entityReferences,
-    [[maybe_unused]] const trait::TraitSet& traitSet,
+    [[maybe_unused]] const trait::TraitsDatas& traitsDatas,
     [[maybe_unused]] const ContextConstPtr& context,
     [[maybe_unused]] const HostSessionPtr& hostSession,
     [[maybe_unused]] const PreflightSuccessCallback& successCallback,
--- a/src/openassetio-core/include/openassetio/BatchElementError.hpp
+++ b/src/openassetio-core/include/openassetio/BatchElementError.hpp
@@ -84,16 +84,16 @@ class BatchElementError final {
      * Error code used when the reference is valid, but the supplied
      * @ref Context access is invalid for the operation. A common
      * example of this would be resolving a read-only entity with a
-     * write access Context, or during @ref preflight or @ref register
-     * when the target entity s read-only and does not support
-     * updating.
+     * write access Context, or during @ref glossary_preflight or @ref
+     * glossary_register when the target entity s read-only and does not
+     * support updating.
      */
     kEntityAccessError = OPENASSETIO_BatchErrorCode_kEntityAccessError,
 
     /**
-     * Error code used during @ref resolve "entity resolution" when the
-     * reference itself is valid, but it is not possible to retrieve
-     * data for the referenced @ref entity.
+     * Error code used during @ref glossary_resolve "entity resolution"
+     * when the reference itself is valid, but it is not possible to
+     * retrieve data for the referenced @ref entity.
      *
      * This could be because it does not exist, or some other
      * entity-specific reason that this data cannot be resolved for a
--- a/src/openassetio-core/include/openassetio/BatchElementError.hpp
+++ b/src/openassetio-core/include/openassetio/BatchElementError.hpp
@@ -107,7 +107,18 @@ class BatchElementError final {
      * entity-based operations on a valid @ref entity_reference that
      * fail for some reason.
      */
-    kEntityResolutionError = OPENASSETIO_BatchErrorCode_kEntityResolutionError
+    kEntityResolutionError = OPENASSETIO_BatchErrorCode_kEntityResolutionError,
+
+    /**
+     * Error code response from @ref glossary_preflight if the provided
+     * @fqref{TraitsData} "traits data" hint holds insufficient or
+     * invalid information.
+     *
+     * This will occur when the manager requires information that the
+     * host owns to be passed to `preflight`, but the host did not
+     * provide it.
+     */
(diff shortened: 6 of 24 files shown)


src/openassetio-python/cmodule/src/hostApi/ManagerBinding.cpp
```diff
@@ -181,50 +181,52 @@ void registerManager(const py::module& mod) {
...
-               const Manager::BatchElementErrorCallback&)>(&Manager::preflight),
-           py::arg("entityReferences"), py::arg("traitSet"), py::arg("context").none(false),
-           py::arg("successCallback"), py::arg("errorCallback"))
+           py::overload_cast<const EntityReferences&, const trait::TraitsDatas&,
+                             const ContextConstPtr&, const Manager::PreflightSuccessCallback&,
+                             const Manager::BatchElementErrorCallback&>(&Manager::preflight),
+           py::arg("entityReferences"), py::arg("entityTraitsDatas"),
+           py::arg("context").none(false), py::arg("successCallback"), py::arg("errorCallback"))
       .def("preflight",
-           static_cast<EntityReference (Manager::*)(
```
elliotcmorris, line 191: [Question] Is the change from `static_cast` to `py::overload_cast` a neccesary change or do you just prefer the style? I guess I wouldn't have expected a neccesary change here changing one arg type to another.

src/openassetio-core/src/hostApi/Manager.cpp
```diff
@@ -323,6 +323,9 @@ void Manager::preflight(const EntityReferences &entityReferences,
                         const ContextConstPtr &context,
                         const PreflightSuccessCallback &successCallback,
                         const BatchElementErrorCallback &errorCallback) {
+  if (entityReferences.size() != entityTraitsDatas.size()) {
+    throw std::out_of_range{"Parameter lists must be of the same length"};
```
elliotcmorris, line 327: [Consider] It might be friendly to have the sizes of the lists in the exception message.

src/openassetio-core/include/openassetio/BatchElementError.hpp
```diff
@@ -107,7 +107,21 @@ class BatchElementError final {
...
+     * Error code response from @ref preflight if the provided
+     * @fqref{TraitsData} "traits data" hint holds insufficient
+     * information.
+     *
+     * This will occur when a workflow requires information up-front, so
+     * that the manager can prepare for upcoming @ref resolve "resolves"
+     * to inform @fqref{Context.Access.kWrite} "write operations" and
+     * eventual @ref register "registration" of the final data, but
+     * insufficient information was given to service those expected
+     * upcoming operations.
```
elliotcmorris, line 122: [Consider][Minor] I wonder if spelling out the resolves->write operations->register flow here is more confusing than helpful.

Perhaps just something like?
```
... so that the manager can service upcoming write-workflow operations.
```

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -1136,9 +1136,9 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * the Manager may even use this opportunity to switch to some
    * temporary working path or some such.
    *
-   * @warning If the supplied @ref trait_set is missing traits required
-   * by the manager for any input entity reference, then that element
-   * will error.
+   * @warning If the supplied @fqref{TraitsData} "trait data" is missing
```
elliotcmorris, line 1139: [Consider] `@fqref{TraitsDatas} "traits datas"` ?

doc/doxygen/src/Examples.dox
```diff
@@ -277,9 +277,14 @@
  *
  * # Whenever we make new data, we always tell the manager first,
  * # This allows it to create a placeholder version or similar.
+ * # We must provide the manager with as much information as possible
+ * # up-front, to help it prepare the backend and determine the most
+ * # appropriate response to our subsequent `resolve` and `register`
+ * # calls, or error early if publishing is likely to fail.
```
elliotcmorris, line 283: [Question] "As much information as possible up front" ... I wonder if this is correct. Do we want to say something more like, "We must provide the manager with _all_ the information known about the entity"? I know they technically say the same thing, but the latter implies less wiggle room.

doc/doxygen/src/Examples.dox
```diff
@@ -277,9 +277,14 @@
...
+ * preflight_hint = TextFileSpecification.create()
+ * preflight_hint.markupTrait().setMarkupType("plain")
```
elliotcmorris, line 285: [Consider] Perhaps another trait here would be good for the example, to convey that you should provide all the data you have, (if we do want to convey that,) rather than just the data you "know" your specific manager cares about. (Which i guess you could have worked out from trial and error.)

doc/doxygen/src/Glossary.dox
```diff
@@ -382,6 +382,12 @@
  * returned @ref entity_reference to determine the manager's chosen
  * location for the data and use this for any write operations.
  *
+ * As much relevant available information as possible should be provided
+ * ahead of time to `preflight`, so that the manager can prepare the
+ * backend and provide the most appropriate responses to subsequent
+ * `resolve` and `register` requests, or error early if publishing is
+ * likely to fail.
```
elliotcmorris, line 389: [Consider] Same thing here. Is "All available information" what we want to be saying?

src/openassetio-core-c/src/managerApi/CManagerInterfaceAdapter.cpp
```diff
@@ -110,7 +110,7 @@ void CManagerInterfaceAdapter::resolve(
 
 void CManagerInterfaceAdapter::preflight(
     [[maybe_unused]] const EntityReferences& entityReferences,
-    [[maybe_unused]] const trait::TraitSet& traitSet,
+    [[maybe_unused]] const trait::TraitsDatas& traitsDatas,
```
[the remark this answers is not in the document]
foundrytom: IIRC we were going to make sure it had 'hints' in the name, so it lines up with the preflight hint error?

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -1153,8 +1153,9 @@ class OPENASSETIO_CORE_EXPORT Manager {
    * @param entityReferences The entity references to preflight prior
    * to registration.
    *
-   * @param traitSet The @ref trait_set of the entities that are being
-   * published.
+   * @param entityTraitsDatas @ref trait_set for each entity,
+   * determining the type of entity to publish, complete with any
+   * properties that can be provided at this time.
```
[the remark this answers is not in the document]
foundrytom: [fix] The key thing to call out here is that the host should set any properties that it _owns_. Ie. any properties that are omitted here may well be `resolved` during the publishing process, before `register`. We should also note that managers may error if this is missing prerequisite data etc... (see the description on the issue for more on this).

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -1111,8 +1111,13 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
    *
-   * @param traitSet The @ref trait_set of the entities that are being
-   * published.
+   * @param entityTraitsDatas  @ref trait_set for each entity,
+   * determining the type of entity that the host will publish to this
+   * reference, complete with any property values that the host can
+   * provide ahead of time. This information should be used as a hint to
+   * help prepare for upcoming @ref resolve calls for
+   * @fqref{Context.Access.kWrite} "write access" and eventual @ref
+   * register to the associated entity in @p entityReferences.
```
foundrytom, line 1123: [consider] Might need a tweak to match what we update the `Manager` side to.

Can we also add notes on erroring when required properties are missing to the error callback section?

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
```diff
@@ -146,12 +146,12 @@ struct PyManagerInterface : ManagerInterface {
         RetainCommonPyArgs::forFn(successCallback), errorCallback);
   }
 
-  void preflight(const EntityReferences& entityReferences, const trait::TraitSet& traitSet,
+  void preflight(const EntityReferences& entityReferences, const trait::TraitsDatas& traitsDatas,
```
foundrytom, line 149: [fix] Arg names need updating in the bindings below

doc/doxygen/src/Examples.dox
foundrytom: [fix] I think the key wording we worked out before was about property values that the host _owns_ vs ones it will try to resolve from the manager as part of publish (more on this in the issue).

doc/doxygen/src/Examples.dox
foundrytom: The "owning" part is the way you decide, which feels to remove the ambiguity.

Also we need to make sure we set the markupTrait for `register` too. It might be worth just assembling the final data at the beginning, vs making these a separate traits data? (Ie move `file_spec` up)

doc/doxygen/src/Glossary.dox
foundrytom: Same re: Its the data owned by the host vs stuff it wants to query.

src/openassetio-core/src/hostApi/Manager.cpp
```diff
@@ -410,21 +410,6 @@ void Manager::register_(const EntityReferences &entityReferences,
...
-      const trait::TraitSet currentTraitSet = entityTraitsDatas[idx]->traitSet();
-      if (currentTraitSet != firstTraitSet) {
-        Str msg = "Mismatched traits at index ";
-        msg += std::to_string(idx);
-        // TODO(DF): expand on error message to include actual trait
-        //  set.
-        throw std::invalid_argument{msg};
-      }
-    }
-  }
```
foundrytom, line 422: [fix] Don't forget we need to add something so

> A manager can determine if the TraitsData instances supplied to the register method share a common trait set with a single line of code.

RELEASE_NOTES.md
```diff
@@ -6,6 +6,16 @@ v1.0.0-alpha.x
 
 ### Breaking changes
 
+- Changed signature of `preflight` to accept a `TraitsData` per
+  reference, rather than a single trait set. The host is now expected to
+  communicate any information that is already known at `preflight` time.
```
foundrytom, line 11: [fix] "already known" -> "it owns"

foundrytom (changes requested):
Great stuff, thanks! Hope all the search + replace wasn't too long winded.

Mostly docs notes as my brain definitely isnt capable of checking code at the moment 🙃 ✈️

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: The sentence is referring to the `TraitsData` for a particular reference, so I _think_ the grammar is correct.  Tricky with all these plurals of plurals flying around.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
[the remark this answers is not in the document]
feltech: Might it be good to simplify the method docstrings and instead update the glossary section for `preflight`, then cross-reference it in `Manager` and `ManagerInterface`.  We need to define what "owns" means, along with he error modes.  Probably better to centralise that doc than burying it under the doc of a parameter.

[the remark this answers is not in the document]
feltech: The acceptance criteria in the issue does simply say "Supporting documentation ... That a host should always provide known data up front." - but we're saying known isn't really the point, it's "owned"  that's important.

So if a host has previously queried some data for from the manager for the entity reference to be `preflight`ed, then there's no need to provide it to `preflight`, since that information is "owned" by the manager.  Makes sense.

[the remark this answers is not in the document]
feltech: Shortened, used the word "owns", and linked to an expanded Glossary entry, in daa8a1ac

src/openassetio-core-c/src/managerApi/CManagerInterfaceAdapter.cpp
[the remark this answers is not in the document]
feltech: Renamed to `traitsHints` in d9e42e84

src/openassetio-core/include/openassetio/BatchElementError.hpp
feltech: Shortened in daa8a1ac, relying on Glossary entry link to provide more info if required.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Not specifically relevant for this concern, but updated that paragraph to have a link to the glossary in daa8a1ac.

doc/doxygen/src/Examples.dox
feltech: Shortened and used the word "owns" in daa8a1ac

doc/doxygen/src/Examples.dox
feltech: Moved `file_spec` up in daa8a1ac

doc/doxygen/src/Glossary.dox
feltech: Reworded and expanded the glossary entry in daa8a1ac.  Does this get the heart of the difference between various sources of information during the publishing process?

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Left as is for now, let me know if this needs to change with all the changes from daa8a1ac. Also added notes to the `errorCallback` docs as part of daa8a1ac.

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
feltech: Fixed as part of d9e42e84 (then fixed the one you actually highlighted in 772c0670 :roll_eyes: )

RELEASE_NOTES.md
feltech: Reworded in daa8a1ac

feltech:
Note: 67aaa575 `[Docs] Disambiguate glossary from methods`  added to fix some ambiguity between `@ref` to the glossary vs. methods.

src/openassetio-core/src/hostApi/Manager.cpp
feltech: Updated error messages in f3fe18d4

src/openassetio-core/src/hostApi/Manager.cpp
feltech: Ah whoops, yup missed this.  I recall that from a PO perspective you weren't too worried what the shape of such a function should be.  But would be good to at least get together with @elliotcmorris to sketch a quick design.

If time is an issue, this seems like the sort of thing that could be a separate PR(?)

src/openassetio-core/src/hostApi/Manager.cpp
elliotcmorris: My kindom for `std::format`!. :)

doc/doxygen/src/Examples.dox
```diff
@@ -277,9 +277,13 @@
  *
  * # Whenever we make new data, we always tell the manager first,
  * # This allows it to create a placeholder version or similar.
+ * # We must provide the manager with any relevant information that the
+ * # host owns (i.e. that the manager is unaware of) and can be provided
```
foundrytom, line 281: [consider] 
```suggestion
 * # host owns (i.e. that won't be queried from the manager during publishing) and can be provided
```

doc/doxygen/src/Glossary.dox
feltech: Reworded yet again in e9b9f191, based on offline discussion.

src/openassetio-core/include/openassetio/BatchElementError.hpp
```diff
@@ -107,7 +107,18 @@ class BatchElementError final {
      * entity-based operations on a valid @ref entity_reference that
      * fail for some reason.
      */
-    kEntityResolutionError = OPENASSETIO_BatchErrorCode_kEntityResolutionError
+    kEntityResolutionError = OPENASSETIO_BatchErrorCode_kEntityResolutionError,
+
+    /**
+     * Error code response from @ref glossary_preflight if the provided
+     * @fqref{TraitsData} "traits data" hint holds insufficient
+     * information.
```
foundrytom, line 115: [consider]
```suggestion
     * @fqref{TraitsData} "traits data" hint holds insufficient
     * information, or the property values are invalid.
```

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -1102,17 +1102,22 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
    * thread that called `preflight`
    *
-   * @warning If the supplied @ref trait_set is missing required traits
-   * for any of the provided references (maybe they are mismatched with
-   * the target entity, or missing essential data) then error that
-   * element with an appropriate @fqref{BatchElementError.ErrorCode}.
+   * @warning If the supplied @fqref{TraitsData} "trait data" is missing
+   * required traits for any of the provided references (maybe they are
+   * mismatched with the target entity), or the populated properties are
+   * insufficient to service upcoming @ref resolve for
```
foundrytom, line 1108: [consider] insufficient or invalid?

src/openassetio-core/src/hostApi/Manager.cpp
foundrytom: Capturing offline discussion - bump to a separate PR.

foundrytom (commented):
Not looked into the implementation in much detail, but the docs changes (aside from glossary, which I left for now as requested) look good 👍. Maybe just including that the preflight traits are "insufficient or invalid"?

doc/doxygen/src/Glossary.dox
foundrytom: LGTM 👍

src/openassetio-core/src/hostApi/Manager.cpp
feltech: Made an issue: #1042

doc/doxygen/src/Glossary.dox
feltech: One last tweak, mostly adding more cross-references: d691db4d

doc/doxygen/src/Examples.dox
feltech: Updated as suggested in aa443b74

doc/doxygen/src/Thumbnails.dox
```diff
@@ -29,11 +29,12 @@
  * "managementPolicy" for any given set of entity @ref trait "traits".
  *
  * If supported by the host, it will then:
- * - Call @ref preflight with the target entity and the traits from the
- *   @needsref ThumbnailSpecification.
- * - Call @needsref resolve on the returned reference to determine the
- *   desired width/height/format for the thumbnail.
- * - Generate a thumbnail and @ref register it to the same reference.
+ * - Call @ref glossary_preflight with the target entity and the traits
```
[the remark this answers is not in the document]
feltech: Updated as suggested in 8f782aab

src/openassetio-core/include/openassetio/BatchElementError.hpp
feltech: Updated in 06b5b240

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Updated as part of 7931b1db

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -1141,16 +1146,21 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
    * initiated the call to `preflight`. A
    * @fqref{BatchElementError.ErrorCode.kEntityAccessError}
    * "kEntityAccessError" should be used for any target references that
-   * are conceptually read-only.
+   * are conceptually read-only. A
+   * @fqref{BatchElementError.ErrorCode.kInvalidPreflightHint}
+   * "kInvalidPreflightHint" should be used for any target references
+   * who's corresponding @p traitsHints entry holds insufficient
```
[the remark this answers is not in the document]
feltech: Updated as part of 7931b1db

elliotcmorris (approved):

feltech:
Squashed down, in case anyone wants to check the commit message before merge.

foundrytom (approved):
LGTM 🙏

elliotcmorris (approved):

PR #1038, opened 2023-08-10
[CI] Remove `cpython` dependency

### LINKED ISSUE #977: Reconsider Python build dependency
## What

Reconsider our strategy for using the `cpython` ConanCenter package to provide our Python build dependency.

## Why

Cross-platform cross-Python version builds have caused us many headaches and many issues remain unresolved.

E.g.
- #722 
- #528

Whatever we choose has knock-on effects to other issues, e.g.
- #845 
- #666
- #623

A central problem we have is the assumption that we use CMake's built-in `FindPython.cmake` module to provide support for `find_package(Python)`, yet we (try to) use ConanCenter's `cpython` package to provide the artifacts.  In many ways the two are simply not compatible.  For example
- `FindPython.cmake` assumes Python's external dependencies are available on the system, whereas `cpython` assumes they're provided by the Conan cache. This leads to either libraries not being found, or the wrong library being found (we've been fairly lucky so far in this regard).
- Windows debug builds simply don't work due to the directory layout of the `cpython` package not matching the expected layout by `FindPython.cmake`
- Trying to get `FindPython.cmake` to locate the Python environment in the Conan cache (i.e. the `cpython` package) is a pain on non-Linux platforms.  E.g. our public ConanCenter recipe for OpenAssetIO has to force the `Python_BINARY` and `Python_LIBRARY` variables, which should be auto-discovered by `FindPython.cmake`.

### Options

Brainstorming some options:
- Abandon use of the `cpython` package, and use whatever `FindPython.cmake` happens to find on the system. It is up to the developer/devops to ensure that the expected Python version is discovered and has all required components (e.g. debug libraries).
- Shim `cpython` with a custom `PythonConfig.cmake` that delegates to the `cpython` package if discovered, but falls back on `FindPython.cmake` if not.

feltech:
## Description

Closes #977 

Conflicting libuuid pacakges are in the build dependency tree for cpython at this time, causing our CI to fail. So just remove cpython.

See failure here: https://github.com/elliotcmorris/OpenAssetIO/actions/runs/5820687754/job/15781579400?pr=29
To be fixed in the `cpython` recipe upstream (eventually) here: https://github.com/conan-io/conan-center-index/pull/18558

## Reviewer Notes

Need to check dependent projects that build OpenAssetIO from source.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -46,10 +46,6 @@ jobs:
 
       - name: Bootstrap
         uses: ./.github/bootstrap_platform
-        env:
-          # As we're building wheels inside a managed environment,
-          # (cibuildwheels) there is no need to install cpython.
-          OPENASSETIO_CONAN_SKIP_CPYTHON: "True"
 
       - name: Build wheels
         run: |
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,13 @@ Release Notes
 v1.0.0-alpha.x
 ---------------
 
+### Breaking changes
+
+- Removed `cpython` dependency from `conanfile.py`. When building
+  OpenAssetIO, it is now expected that a development install of the
+  appropriate Python version is discoverable on the system.
+  [1038](https://github.com/OpenAssetIO/OpenAssetIO/pull/1038)
+
 ### Improvements
 
 - Increased verbosity when running the `openassetio.test.manager` API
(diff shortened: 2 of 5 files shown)


elliotcmorris (approved):
Since dependent repos are happy, and this seems to have no downsides (and a lot of upsides, especially in relation to conan center.) I say merge this.

foundrytom:
HAPPY DAYS ❤️ 

Great solution.

elliotcmorris:
Worth keeping in mind that we might be able to make a pain-free conan center recipe now too.

PR #1046, opened 2023-08-17
[Test] Alter which fixture skips relation tests

### LINKED ISSUE #1022: Relationship compliancetest uses malformed reference as its skipped parameter
## What
Found this when updating the python manager template. The compliance suite tests for a malformed reference in `Test_getWithRelationship_All` uses the existence of a malformed entity reference fixture as the deciding factor as to whether it should be skipped. I propose it should be including whether the `a_relationship_trait_set` fixture exists either instead or in addition to this.

https://github.com/OpenAssetIO/OpenAssetIO/blob/3323a23ff57e086f48fb81d5e21ab593344b20db/src/openassetio-python/package/openassetio/test/manager/apiComplianceSuite.py#L686
## Why
The `a_malformed_reference` fixture is common to other non-optional flows, and is likely to be defined as a globally shared fixture, (which is what BAL does, really the only other example available, and bound to be copy-pasted,) This may give the wrong impression the the relationship functions must be implemented to a new implementer, and could bounce the tentative explorer.

## Acceptance Criteria

- Tests skipped based on relationship specific fixtures where possible.

## Notes

- Changes to BAL out of scope

elliotcmorris:
Closes #1022
## Description
Alter the skip fixture in certain getWithRelation tests away from common shared ones  to the `a_relationship_trait_set`, which is less likely to be defined in a shared context.

This is helpful because relationships are an optional feature, so it is likely managers will want to skip all the relationship tests as a group, and this makes that more or less the default behaviour without thinking about it.

~~- [ ] I have updated all relevant user documentation.~~

## Test Instructions

Tested manually with BAL, checking that removing the fixtures causes tests to skip.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -53,6 +53,10 @@ v1.0.0-alpha.x
 - Added some aliases to the Doxygen API documentation. In
   particular, `Ptr` and `ConstPtr` aliases are now cross-referencable.
 
+- Altered the fixture that causes a skip in relationship based API
+  compliance tests, to be the more specific relationship traitset.
+  [#1022](https://github.com/OpenAssetIO/OpenAssetIO/issues/1022)
+
 v1.0.0-alpha.13
 ---------------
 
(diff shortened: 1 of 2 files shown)


foundrytom (approved):
🙏

feltech (approved):
LGTM. Double-checked with BAL and all updated tests still run and pass.

PR #1051, opened 2023-08-22
[Docs] Add DR for the removal of `access` from `Context`

foundrytom:
TL;DR - We want to break some things (well most things) to Make Everything Better™. We need to fix the ambiguity around how to use `Context.access` for both hosts and managers. Now is the last chance we have, before too many people have done a lot of harder-than-it-should-be typing.

Rendered version for 👀 [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/1051/files?short_path=d32e611#diff-d32e611102451d160a468398b9f514d423484731dc41f3ac42183d2d4fc9dd6b)

We need to ratify this decision by e.o.d Thurs 24th August.

### DIFF
--- a/doc/decisions/DR019-Remove-access-from-the-context.md
+++ b/doc/decisions/DR019-Remove-access-from-the-context.md
@@ -1,0 +1,202 @@
+# DR019 Remove access from the Context
+
+- **Status:** Decided
+- **Impact:** High
+- **Driver:** @foundrytom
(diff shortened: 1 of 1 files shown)


feltech (approved):
Seems to capture what we've discussed. Will keep pondering, but for now (bar tiny typos) LGTM.

foundrytom:
@feltech I've updated this to be Decided based on the feedback weve had on/off line to date.

PR #1056, opened 2023-08-24
Move `Access` out of `Context`

### LINKED ISSUE #1054: Remove `access` from `Context`
## What

👆

## Why

https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/decisions/DR019-Remove-access-from-the-context.md

## Acceptance Criteria

- `access` removed from `Context`
- New specific enums added to the `openassetio::access` namespace for specific workflows:
  - **`PolicyAccess` (`managementPolicy`)**: `kRead`, `kWrite`, `kCreateRelated`
  - **`RelationsAccess` (`getWithRelationship*`)**: `kRead`, `kWrite`, `kCreateRelated`
  - **`ResolveAccess` (`resolve`)**: `kRead`, `kWrite`
  - **`PublishingAccess` (`preflight`/`register`)**: `kWrite`, `kCreateRelated`
  - Matching enum members should have the same value across the definitions.
  - `kAccessNames` array persisted to allow bi-directional string conversions.
 - Methods update with the above as a[nother] positional argument `access` at the end of the batch arguments, before `context` or `hostSession`, callbacks.
 - Validate existing docs that refer to `access` are still meaningful, no additional docs required.
 - `Examples.dox` updated
 - `simpleResolver` updated
 - Significant note in `RELEASE_NOTES` via block quote with `**Warning:**` heading.
- Python bindings hoist contents of `access` in python `access/__init__.py` to allow `from openassetio.access import ResolveAccess`

feltech:
## Description

Closes #1054 
- #1054 

## Reviewer Notes

Rebased on https://github.com/OpenAssetIO/OpenAssetIO/pull/1058 and implemented `access` argument for `defaultEntityReference`.  This method isn't mentioned in the ACs so may need further discussion.  I struggled to think of a use-case (==docstring) for `kWrite` so left it as `kRead` and `kCreateRelated` only.

Not gone into depth with docs/docstring updates, given that docs cause the most review churn.

BAL is retargetted to a corresponding branch: `foundrytom/work/67-access`.  Once confirmed working and PR accepted, this commit can be dropped and the `integrations.yml` tests disabled to allow this PR to be merged. Subsequent work can then get dependent projects up to date and re-enable the integration tests.

### DIFF
--- a/doc/doxygen/src/Examples.dox
+++ b/doc/doxygen/src/Examples.dox
@@ -144,7 +144,7 @@
  * `kInfoKey_EntityReferencesMatchPrefix` key.
  *
  * @code{.py}
- * from openassetio import Context
+ * from openassetio.access import ResolveAccess
  * from openassetio_mediacreation.traits.content import LocatableContentTrait
  *
  * # Note: this will raise an exception if given a string that is not
@@ -154,19 +154,18 @@
  * entity_reference = manager.createEntityReference(some_string)
  *
  * # All calls to the manager must have a Context, these should always
- * # be created by the target manager. The Context expresses the host's
- * # intent, and ensure that any manager state is properly managed
- * # between API calls.
+ * # be created by the target manager. The Context ensures that any
+ * # manager state is properly managed between API calls.
  * context = manager.createContext()
- *
- * # We describe what we want to do with the asset
- * context.access = context.access.kRead
  *
  * # We can now resolve a token we may have if it is a reference. In
  * # this example, we'll attempt to resolve the LocatableContent trait
- * # for the entity.
+ * # for the entity. We inform the manager that we just want to read the
+ * # current data for the entity (kRead). See the publishing examples
+ * # for use of the alternative `kWrite` access mode.
  * resolved_asset = manager.resolve(
- *         entity_reference, {LocatableContentTrait.kId}, context)
+ *         entity_reference, {LocatableContentTrait.kId},
+ *         ResolveAccess.kRead, context)
  * url = LocatableContentTrait(resolved_asset).getLocation()  # May be None
  * @endcode
  *
@@ -196,6 +195,8 @@
  *
  * @code{.py}
  *
+ * from openassetio.access import PolicyAccess
+ *
  * # Commented imports are illustrative and may not exist yet
  * from openassetio_mediacreation.traits.managementPolicy import (
  *     ManagedTrait,
@@ -207,13 +208,12 @@
  *     # TextFileSpecification
  * )
  *
- * # Make sure we have the correct context
- * context.access = context.kRead
- *
  * # We use the well-known specification for a text file to determine
  * # the correct trait set to query. Using the standard definition
- * # ensures consistent behaviour across managers/hosts.
- * [policy] = manager.managementPolicy([TextFileSpecification.kTraitSet], context)
+ * # ensures consistent behaviour across managers/hosts. We request
+ * # the manager's policy for read access to this entity type.
+ * [policy] = manager.managementPolicy(
+ *  [TextFileSpecification.kTraitSet], PolicyAccess.kRead, context)
  *
  * # We can now check which traits were imbued in the policy, the
  * # absence of a trait means it is unsupported.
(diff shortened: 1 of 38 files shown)


src/openassetio-core/include/openassetio/access.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * Common constant values for strong enumerations.
+ *
+ * We must ensure correspondence between the values of the various
+ * workflows' strong enums. This allows for comparison, string lookup,
+ * simplifies serialisation, and supports language (especially C)
+ * bindings.
+ */
+enum Access { kRead = 0, kWrite, kCreateRelated };
```
feltech, line 28: [question] In `errors.h` we use `#define`s for constants to make them useable in C.  But we could/should have used a C enum instead, like here, I think(?).

Given the C interface is on the back burner, I didn't bother extracting this enum out to a C-compatible `.h` file.  Is that worth doing now?

foundrytom (approved):
🎉

src/openassetio-core/include/openassetio/Context.hpp
```diff
@@ -16,77 +16,32 @@ inline namespace OPENASSETIO_CORE_ABI_VERSION {
...
+ * The Context object is used to convey information about the calling
+ * environment to a @ref manager.
  *
- *  A Manager will also use this information to ensure it presents the
- *  correct UI, or behavior.
+ * It holds additional information about the @ref host that may be
+ * useful to the @ref manager. Optionally, it holds arbitrary, opaque
+ * (to the host), serializable manager state, which can be used by the
+ * manager plugin to tie together related calls whilst remaining
+ * stateless itself.
```
elliotcmorris, line 26: [Consider] "whilst remaining stateless itself." sees like redundant information as far as the implementor is concerned.

doc/doxygen/src/Examples.dox
```diff
@@ -144,7 +144,7 @@
  * `kInfoKey_EntityReferencesMatchPrefix` key.
  *
  * @code{.py}
- * from openassetio import Context
+ * from openassetio.access import ResolveAccess
```
elliotcmorris, line 147: [Question] Don't we still need to context import for this example?

doc/doxygen/src/Examples.dox
```diff
@@ -278,8 +281,13 @@
  * # publishing) and can be provided up-front.
  * file_spec = TextFileSpecification.create()
  * file_spec.markupTrait().setMarkupType("plain")
+ *
+ * # We're not, as far as we know, creating a new related entity. We are
+ * # most likely mutating an existing entity. So we use kWrite access
+ * # mode.
```
elliotcmorris, line 288: [Consider] Weird take, but to me this sentance sort of gives the vibe that the host should/wants to know about the method of storage on the back end. I wonder if a better phrasing might be

"Our intent is to write data to this entity, not to create a new related entity (kCreateRelated,) so we use the kWrite access mode."

elliotcmorris (approved):

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -1291,9 +1291,7 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * reference, then that element will error. See @ref
    * glossary_preflight "glossary entry" for details.
    *
-   * @note It's vital that the @ref Context is well configured here,
-   * in particular the @fqref{Context.access}
-   * "Context.access".
+   * @note It's vital that the @ref Context is well configured here.
```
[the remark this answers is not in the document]
feltech: Removed in 76eaba43

doc/doxygen/src/Examples.dox
feltech: The `Context` class is not referenced directly (only indirectly by `createContext`) so doesn't need to be explicitly imported.

src/openassetio-core/include/openassetio/Context.hpp
feltech: Removed in a01dec12

doc/doxygen/src/Examples.dox
feltech: Like it. Used your wording in d13f14ca

PR #1081, opened 2023-09-05
[Core] Relax checks in `TraitsData::getTraitProperty`

### LINKED ISSUE #970: Relax `TraitsData::getTraitProperty` exceptions
## What

Return `false` instead of throwing if the requested trait is missing.

## Why

There are many host situations where they wish to use a default value when attempting to check for a trait property. Splitting the failure mode between missing traits and missing properties makes for cumbersome host code. It forces a host to always check if a trait is imbued first, before then potentially looking up the value. Traits being missing isn't any more exceptional than their properties in many real-world applications, forcing an `isImbued` check makes code more complex and potentially doubles the number of map lookups for that trait.

If a host wants to explicitly differentiate the two failure modes, then the current approach can still be used.

eg:
```python
data = resolve(..., {LocatableContentTrait.kId}, ...)
if LocatableContentTrait.isImbuedTo(data):
  value = LocatableContentTrait(data).getLocation(someDefault)
```
Becomes

```python
data = resolve(...)
value = LocatableContentTrait(data).getLocation(someDefault)
```

## Acceptance Criteria

- `getTraitProperty` returns `false` if the trait is not imbued _or_ the property is not set.

foundrytom:
Raising if the trait was missing added significant boilerplate to many common access patterns.

Closes #970

Note: This felt too easy. Im sure I've missed something.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,18 @@
 Release Notes
 =============
 
+v1.0.0-alpha.xx
+---------------
+
+### Breaking changes
+
+- Attempting to retrieve a trait property with
+  `TraitsData.getTraitProperty` or using trait view classes no longer
+  raises an exception if the trait itself is missing. This simplifies
+  many common access patterns, and can remove the need for
+  `hasTrait`/`isImbuedTo` checks.
+  [#970](https://github.com/OpenAssetIO/OpenAssetIO/issues/970)
+
 v1.0.0-alpha.14
 ---------------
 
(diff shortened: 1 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -1,6 +1,18 @@
 Release Notes
 =============
 
+v1.0.0-alpha.xx
+---------------
+
+### Improvements
+
+- Attempting to retrieve a trait property with
```
feltech, line 9: [consider] Could view this as a breaking change, if people are already relying on exceptions to tell them if a trait is not imbued.

feltech (approved):
LGTM - I think it really is that simple.

RELEASE_NOTES.md
foundrytom: Good call - updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/1081/commits/bd811520f1a4d09b0d3cb6e288d0372448ddcbe9)

elliotcmorris (approved):
Couldn't find any reason for more complexity. Traitgen is written to always expect to pass existing traits so dosen't depend on any exception behaviour.

PR #1082, opened 2023-09-06
Remove `objects.py`

### LINKED ISSUE #1068: Remove objects.py
## What
Delete [objects.py](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/src/openassetio-python/package/openassetio/_core/objects.py)
## Why
Does anybody know what it does? It's types don't seem to be referenced outside of itself.

foundrytom:
Closes #1068


elliotcmorris (approved):
Great! Was certain this would be a hidden dependency that I hadn't seen.

feltech (approved):

PR #1086, opened 2023-09-07
[Core] Remove backdoor access to Manager internals

### LINKED ISSUE #26: Conform public/protected methods in `Manager` and `Host`
## What/Why

[This](https://github.com/TheFoundryVisionmongers/OpenAssetIO/pull/23#discussion_r718576482) comment raised a good point. Some methods that we see as being "only used internally by the API middleware" are just marked `@protected` in Doxygen. Others are explicitly protected via [single underscore](https://github.com/OpenAssetIO/OpenAssetIO/blob/5a5fa8b2dbdd81c1e4ab82c23f4ca03d1be06067/src/openassetio-core/include/openassetio/hostApi/Manager.hpp#L1834).

We should be consistent about this, and use single underscore for any api-internal methods to avoid confusion.

## Acceptance Criteria

- The public API does not leak any methods/classes that are for middleware use only. Anything public can be used at will by host/manager integrators.
- Internal methods are excluded from the public documentation build.

foundrytom:
See commit messages, closes #26

### DIFF
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -185,15 +185,14 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * applicable) once initialized.
    *
    * The behavior of calling initialize() on an already initialized
-   * Manager should be a no-op, but if an error was raised
-   * previously, then initialization will be re-attempted.
+   * Manager is to re-initialize the manager with any updated settings
+   * that are provided. If an error was raised previously, then
+   * initialization will be re-attempted.
    *
    * @note This must be called prior to any entity-related calls or
    * an Exception will be raised.
    *
    * @note This method may block for extended periods of time.
-   *
-   * @protected
    */
   void initialize(InfoDictionary managerSettings);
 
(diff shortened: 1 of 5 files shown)


elliotcmorris (approved):

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -185,15 +185,14 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * applicable) once initialized.
    *
    * The behavior of calling initialize() on an already initialized
-   * Manager should be a no-op, but if an error was raised
-   * previously, then initialization will be re-attempted.
+   * Manager is to re-initialize the manager with any updated settings
```
feltech, line 188: [plan] Same problem in `ManagerInterface`

feltech (approved):

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Was in two minds whether to just fix it as part of this issue, but instead made https://github.com/OpenAssetIO/OpenAssetIO/issues/1090 as a follow-up.

PR #1087, opened 2023-09-07
Conform use of `const` with Interface methods.

### LINKED ISSUE #518: Decide on `const` vs. `non-const` member functions
# What

Go through existing C++ implementation and decide if member functions and/or their parameters should be `const`.

# Why

Using `const` member functions wherever possible is good practice. However, it is obviously constraining. This is especially true for an abstract interface, where the underlying (host/plugin) implementation might want to mutate some internal state.

Similarly, `const` function arguments are good practice where possible. However, again we must be careful in assuming that mutation (including non-`const` member function calls) will never be required.

# Notes

One heuristic worth considering is "should this function be super cheap?" - enforcing `const`ness is a way to signal this.

Another consideration is whether and where an object is expected to be passed as a `const` value/reference, meaning only the `const` subset of member functions can be called.

Perhaps worth considering [`std::experimental::propagate_const`](https://en.cppreference.com/w/cpp/experimental/propagate_const), depending on compiler support.

## ACs

- `ManagerInterface`: only `identifier` `displayName`  and `createEntityReference` remain const.
- `Manager` should mirror the interface
- `Host` + interface follows suit

foundrytom:
Closes #518, see commit message/CODING_STANDARDS.md

### DIFF
--- a/doc/contributing/CODING_STANDARDS.md
+++ b/doc/contributing/CODING_STANDARDS.md
@@ -236,6 +236,7 @@ This causes problems in particular when we use Pybind "trampoline"
 classes to call a Python method override from a C++ virtual member
 function. For example, if the Python instance is destroyed, then in the
 case of pure virtual member functions, `pybind` may throw an exception
+
 ```
 RuntimeError: Tried to call pure virtual function
 ```
@@ -257,19 +258,23 @@ be kept alive) in `Args` to instead be `PyRetainingSharedPtr` types.
 To handle instances returned from a Python method to C++ we must modify
 the `PYBIND11_OVERRIDE` macro arguments in the Pybind "trampoline" class
 member functions from, for example
+
 ```c++
 std::shared_ptr<MyReturnType> PyMyClass::myMethod(myArg) override {
   PYBIND11_OVERRIDE(
       std::shared_ptr<MyReturnType>, MyClass, myMethod, myArg);
 }
 ```
+
 to
+
 ```c++
 std::shared_ptr<MyReturnType> PyMyClass::myMethod(myArg) override {
   PYBIND11_OVERRIDE(
       PyRetainingSharedPtr<MyReturnType>, MyClass, myMethod, myArg);
 }
 ```
+
 Note that the return type of the C++ member function is unchanged.
 
 This ensures that the `shared_ptr` returned by the member function call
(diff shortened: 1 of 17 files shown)


doc/contributing/CODING_STANDARDS.md
```diff
@@ -176,6 +176,16 @@ when binding a C++ `enum class` using pybind, use
...
 
+### Const
+
+Methods of interface classes that are conceptually stateless, and their
+corresponding wrapper class methods should not be marked `const`. For
+example, `ManagerInterface` + `Manager`, `HostInterface` and `Host`.
+This may seem to contradict the "stateless" definition, but it allows
+the implementation to maintain private state to service the API requests
+efficiently, such as server connections or caches. Exceptions to this
+are the identification methods that should be compile-time constant.
```
feltech, line 187: [consider] "should be compile-time constant" might be read more literally than intended. Perhaps
```suggestion
are the identification methods that should be (effectively) compile-time constant.
```
or maybe there's a better qualification

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -530,7 +530,7 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    *
    * @todo Use a custom exception type rather than std::domain_error.
    */
-  [[nodiscard]] EntityReference createEntityReference(Str entityReferenceString) const;
+  [[nodiscard]] EntityReference createEntityReference(Str entityReferenceString);
```
feltech, line 533: [question] Presumably the `const` was removed here just because internally it needs to use `isEntityReference`?  Just checking since it's against the ACs.

feltech (approved):
LGTM

doc/contributing/CODING_STANDARDS.md
elliotcmorris: Heck we should make them `constexpr` if that were the case. (Although i doubt we can for most if not all, due to constexpr not being available on virtual methods pre c++20.)
I think keeping this simple may be best, don't bring up the messy idea of compile-time constness at all, simply:
```
Exceptions to this are the identification methods that should be constants.
```

elliotcmorris (approved):
LGTM nothing further.

doc/contributing/CODING_STANDARDS.md
feltech: Updated - (rebased and) force-pushed.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: I think this is pretty obviously the reason, will resolve (Tom is away, I'm taking over).

PR #1089, opened 2023-09-08
Add new exception types.

### LINKED ISSUE #1071: Add new exception hierarchy
## What 

Redesign exception hierarchy to incorporate/replace `exceptions.py` in C++ with Python pseudo-bindings.

## Why

It is useful for hosts to be able to branch on specific types of exception in order to customise their mitigation. In particular, it is useful for hosts to know if an exception comes from the OpenAssetIO ecosystem.

The first step in enabling this is to add the exceptions, and use them wherever we're currently relying on standard C++/Python exceptions.

## ACs

* The following exceptions are added under an `errors` namespace / submodule

```c++
OpenAssetIOException : std::runtime_error
```
```c++
UnhandledException : OpenAssetIOException
```
```c++
InputValidationException : OpenAssetIOException
```
```c++
ConfigurationException : OpenAssetIOException
```
```c++
NotImplementedException : OpenAssetIOException
```
  These exceptions simply forward the constructor of the base class.
* Python bindings using current strategy (i.e. C++->Python translation only).
* Existing `raise` or `throw` invocations migrated to use the most appropriate exception from the above.
* Issues raised to update dependent projects if required.

elliotcmorris:
## Description
Adds new exception types based in C++ for both C++ and Python workflows.
These types are based on `OpenAssetIOException`, with the intent that all OpenAssetIO exceptions share a common base, and can be caught in client code on that basis.

Basing these types in C++ allows us to remove the python exceptions also, which will be done in subsequent commit.

Closes #1071 


src/openassetio-core/include/openassetio/errors/exceptions.hpp
```diff
@@ -28,8 +28,8 @@ struct OPENASSETIO_CORE_EXPORT OpenAssetIOException : std::runtime_error {
 };
 
 /**
- * Thrown whenever a procedure must abort due to an invalid input being
- * provided as an input state.
+ * Thrown whenever the input to a public API function is invalid for the
```
elliotcmorris, line 31: [Consider] We've used the term `procedure` in all the other docstrings, whereas now this one uses `function`. I went with `procedure` for ... I dunno, silly reasons, I think it's even less technically correct in this context. Should we just change them all over to `function` for consistency?

src/openassetio-python/cmodule/src/_openassetio.hpp
```diff
@@ -116,7 +116,7 @@ void registerEntityReference(const py::module& mod);
 void registerBatchElementError(py::module& mod);
 
 // Register errors, linking exceptions to/from Python.
-void registerErrors(py::module& mod);
+void registerErrors(const py::module& mod);
```
elliotcmorris: [consider] If we're going for consistency, should we do the same to `BatchElementError` whilst we're here?

src/openassetio-core/src/hostApi/Manager.cpp
```diff
@@ -39,7 +39,8 @@ void throwFromBatchElementError(std::size_t index, BatchElementError error) {
       exceptionMessage += std::to_string(static_cast<int>(error.code));
       exceptionMessage += " Message: ";
       exceptionMessage += error.message;
-      throw errors::InputValidationException{exceptionMessage};
+      error.message = std::move(exceptionMessage);
+      throw UnknownBatchElementException{index, std::move(error)};
```
elliotcmorris, line 43: [Consider] I guess this was the subject of discussions, so we don't need to re-legislate, but I'll register that this seems off to me, we already have an `unknown` code that converts to the unknown exception. ... sometimes i wonder if cases like this should just be `assert(false)`s to indicate programming error.

src/openassetio-python/cmodule/src/errorsBinding.cpp
```diff
@@ -2,59 +2,32 @@
...
+  // cause C++ exceptions to be translated. See
+  // `register_exception_translator` below.
+  const auto registerPyException = [&mod](const char *pyTypeName, const py::handle &base) {
+    return py::exception<void /* unused */>{mod, pyTypeName, base};
   };
 
-  // Register our OpenAssetIO exception types
+  // Register our OpenAssetIO exception types. Since these exceptions
+  // simply take a string message and no other data, we can make use
+  // of pybind11's limited custom exception support.
```
elliotcmorris, line 27: Oh! Nice. :+1:

src/openassetio-core/include/openassetio/errors/exceptions.hpp
feltech: I think in this case it was more to be super clear that this is meant for validation of API function arguments, rather than  problems that happen during processing

src/openassetio-python/cmodule/src/_openassetio.hpp
feltech: Yeah if possible :+1: - though should that be done as part of the other issue?

src/openassetio-core/src/hostApi/Manager.cpp
feltech: True it is a programming error, `assert` might be good but obviously won't work in Release builds.

The change here was because the exception propagates to the host, when it was the manager that made the error. So from the host's perspective it really is an "unknown" error.  Should we have another exception type, just for this case, perhaps?  The message and index might still be useful to hosts.

src/openassetio-python/cmodule/src/_openassetio.hpp
elliotcmorris: Might as well, it's already done on that branch.

src/openassetio-core/src/hostApi/Manager.cpp
elliotcmorris: Na, a new exception type isn't worth it, this isn't even a real case, i'm convinced it can't happen with our clang checks enable, just leave it.

src/openassetio-core/include/openassetio/errors/exceptions.hpp
elliotcmorris: Fair enough.

feltech:
I've added some more commits. 

In particular, I wrapped the tomlplusplus `parse_file` command in a try/catch to wrap its error in our `ConfigurationException`. 

After going through all the `catch`/`except`/`pytest.raises` in the codebase, the tomlplusplus error was the only third-party exception I detected.  There may be others that we're not testing for.

But in the meantime I thought since that's the only case I may as well roll it into this PR, rather than create a new ticket.

@elliotcmorris

elliotcmorris:
"Pull request authors can't approve their own pull request" :face_exhaling: 

Looked through your new commits, good catches, especially the changing to more specific types in the test, sort of thing that always slips my mind. I think we're good to rebase and merge this.

feltech (approved):
Make it so.

src/openassetio-core/src/hostApi/Manager.cpp
```diff
@@ -300,7 +302,7 @@ void Manager::getWithRelationshipPaged(
     const Manager::BatchElementErrorCallback &errorCallback,
     const trait::TraitSet &resultTraitSet) {
   if (pageSize == 0) {
-    throw std::out_of_range{"pageSize must be greater than zero."};
+    throw errors::InputValidationException{"pageSize must be greater than zero."};
```
foundrytom, line 305: [plan] Docs need updating (and for below).

src/openassetio-core/src/hostApi/ManagerFactory.cpp
```diff
@@ -121,11 +122,17 @@ ManagerPtr ManagerFactory::defaultManagerForInterface(
     Str msg = "Could not load default manager config from '";
     msg += configPath;
     msg += "', file does not exist.";
-    throw std::runtime_error(msg);
+    throw errors::InputValidationException(msg);
```
foundrytom, line 125: [plan] Docs need updating (and for below).

src/openassetio-python/bridge/src/python/converter.cpp
```diff
@@ -44,7 +45,7 @@ PyObject* castToPyObject(const T& objectPtr) {
   const py::error_scope castErrorState;
 
   if (castErrorState.value != nullptr) {
-    throw std::runtime_error(py::str(castErrorState.value));
+    throw errors::InputValidationException(py::str(castErrorState.value));
```
foundrytom, line 48: [plan] Docs need updating (and for below).

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
```diff
@@ -176,13 +173,13 @@ def plugin(self, identifier):
         @return @ref openassetio.pluginSystem.PythonPluginSystemPlugin
         "PythonPluginSystemPlugin"
 
-        @exception openassetio.exceptions.PluginError Raised if no
-        plugin provides the specified identifier.
+        @exception openassetio.errors.ConfigurationException Raised if
```
foundrytom, line 176: [plan] Docstring needs updating.

src/openassetio-python/tests/cmodule/resources/_openassetio_test/errorsTest.cpp
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Bindings used for testing errors behaviour.
+ * Specifically, the conversion from cpp -> python.
+ */
+
+#include <pybind11/pybind11.h>
+
+#include <openassetio/errors/exceptions.hpp>
+namespace py = pybind11;
+
```
foundrytom, line 12: [plan] Can we add some documentation on the who/what/when/where/why

src/openassetio-python/tests/package/hostApi/test_managerfactory.py
```diff
@@ -174,15 +174,21 @@ def test_when_non_existent_path_then_runtime_error_raised(
         )
 
     @pytest.mark.parametrize("use_env_var_for_config_file", [True, False])
-    def test_when_invalid_content_then_runtime_error_raised(
+    def test_when_invalid_format_then_RuntimeError_raised(
```
foundrytom, line 177: [plan] Test name needs updating

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
```diff
@@ -148,7 +148,7 @@ def test_when_none_is_supplied_then_TypeError_is_raised(self, a_host_session):
             ManagerInterface().createChildState(None, a_host_session)
 
     def test_default_implementation_raises_RuntimeError(self, a_host_session):
```
foundrytom, line 150: [plan] Test name needs updating

src/openassetio-python/tests/package/pluginSystem/test_pythonpluginsystem.py
```diff
@@ -198,6 +199,15 @@ def test_when_plugins_broken_then_skipped_with_expected_errors(
         )
 
 
+class Test_PythonPluginSystem_plugin:
+    def test_when_plugin_not_found_then_raises_ConfigurationException(self, a_plugin_system):
```
foundrytom, line 203: [plan] Test name needs updating

foundrytom (commented):
Great stuff - thanks. Just some out of date docs/names/etc...

src/openassetio-core/src/hostApi/Manager.cpp
feltech: Addressed as part of #1104

src/openassetio-core/src/hostApi/ManagerFactory.cpp
feltech: Addressed as part of #1104

src/openassetio-python/bridge/src/python/converter.cpp
feltech: Addressed as part of #1104

src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
feltech: Addressed as part of #1104

src/openassetio-python/tests/cmodule/resources/_openassetio_test/errorsTest.cpp
feltech: Addressed as part of #1104

src/openassetio-python/tests/package/hostApi/test_managerfactory.py
feltech: Addressed as part of #1104

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
feltech: Addressed as part of #1104

src/openassetio-python/tests/package/pluginSystem/test_pythonpluginsystem.py
feltech: Addressed as part of #1104

PR #1092, opened 2023-09-08
[Docs] Fix ManagerInterface reinitialize behaviour

### LINKED ISSUE #1090: `ManagerInterface.initialize` docstring implies no-op rather than update settings
This was fixed in `Manager` as part of the below PR, but needs also fixing in `ManagerInterface`

_Originally posted by @feltech in https://github.com/OpenAssetIO/OpenAssetIO/pull/1086#discussion_r1319749027_

feltech:
## Description

Closes #1090. The manager should re-initialize with updated settings, rather than no-op. The corresponding docs in `Manager` were recently fixed. So add the fix also to the `ManagerInterface.initialize` member function.

### DIFF
--- a/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
+++ b/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
@@ -310,8 +310,8 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
    * responsibility to deal with transient connection errors (if
    * applicable) once initialized.
    *
-   * The behavior of calling initialize() on an already initialized
-   * instance should be a no-op, but if an error was raised
+   * If called on an already initialized instance, re-initialize with
+   * any updated settings that are provided. If an error was raised
    * previously, then initialization should be re-attempted.
    *
    * @note This will always be called prior to any Entity-related


src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -311,8 +311,8 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
    * responsibility to deal with transient connection errors (if
    * applicable) once initialized.
    *
-   * The behavior of calling initialize() on an already initialized
-   * instance should be a no-op, but if an error was raised
+   * If called on an already initialized instance, re-initialize with
+   * any updated settings that are provided. If an error was raised
    * previously, then initialization should be re-attempted.
```
elliotcmorris, line 316: [Question] Does mentioning that we redo initialization if an error was raised previously still parse? Seems like this paragraph is saying "Reinitialize, if an error was raised ... then reinitialize." Which seems like a redundant statement.

feltech: I think we need to disambiguate between re-initializing with updated settings, vs. retrying a failed initialization even if no settings have changed.  Can you think of a better wording for that?

The corresponding `Manager` docs say
```
   * The behavior of calling initialize() on an already initialized
   * Manager is to re-initialize the manager with any updated settings
   * that are provided. If an error was raised previously, then
   * initialization will be re-attempted.
```

elliotcmorris: I see, I guess the current wording parses then.

PR #1096, opened 2023-09-13
[Docs] Add coding style DR

foundrytom:
Better late than never... hopefully this captures our thinking _at the time_ (we shouldn't include any new motivations we have thought of in the interim, to allow fair criticism).

PR #1097, opened 2023-09-13
[Docs] Add DR for language consistency

foundrytom:
Better late than never... hopefully this captures our thinking at the time (we shouldn't include any new motivations we have thought of in the interim, to allow fair criticism).

NB, needs #1096 for cross-link.


feltech (approved):
LGTM, I think captures the reasoning at the time.

PR #1105, opened 2023-09-19
[Core] `optional` defaultEntityReference return

### LINKED ISSUE #1100: Use `optional<EntityReference>` in `defaultEntityReference`
## What

Change returns type of `defaultEntityReference` to optional

## Why

Because having no data is not an error state

## ACs

- Change the `DefaultEntityReferenceSuccessCallback` to take `std::optional<EntityReference>`.
- Update documentation to not instruct the use of `kEntityAccessError` when one isn't available, and to supply a null optional instead.

feltech:
## Description

Closes #1100. Having no data for a default entity reference is not an error state.


src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -888,20 +889,20 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    *
    * @param successCallback Callback that will be called for each
    * successful default retrieved for each of the given sets in @p
-   * traitSets. It will be given the corresponding index of the traitSet
-   * in @p traitSets along with the default entity reference. The
-   * callback will be called on the same thread that initiated the call
-   * to `defaultEntityReference`.
+   * traitSets. It will be given the corresponding index of the trait
+   * set in @p traitSets along with the default entity reference. If the
+   * query is well-formed, but there simply is no default entity
```
elliotcmorris, line 894: [consider][minor]
```suggestion
   * query is well-formed, but there is no available default entity
```
avoid editorialising.

elliotcmorris (approved):

feltech:
> Thanks! Should we update the default implementation to return a null optional for now, as the `kEntityAccessError` is no longer documented.

Although it'll change with the result of https://github.com/OpenAssetIO/OpenAssetIO/pull/1101, it's a good point that it's currently inconsistent and undocumented.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Updated as given in 7bc92d80

feltech:
> > Thanks! Should we update the default implementation to return a null optional for now, as the `kEntityAccessError` is no longer documented.
> 
> Although it'll change with the result of #1101, it's a good point that it's currently inconsistent and undocumented.

Modified default behaviour in 42ca5099

foundrytom (approved):
LGTM, thanks, FFTSAMAYL

PR #1106, opened 2023-09-19
[Docs] Fix placement of ManagerFactory var expansion docs

foundrytom:
Was accidentally with the env var version rather than the one with the file spec.


feltech (approved):
Good spot.

PR #1109, opened 2023-09-19
Add BatchElementException messages

### LINKED ISSUE #1073: Update `BatchElementException` hierarchy
## What

Update the `BatchElementException` hierarchy to include additional useful information for hosts, including an intermediate base class allowing entity reference -specific errors to be filtered.

## Why

The current `BatchElementException`s currently include the element index as a data member, but simply regurgitate the underlying `BatchElementError` message. Given the location where the exception is caught could be far from the call site, additional information in the error object and message would be valuable to the host.

## ACs

* All `BatchElementError` and `BatchElementException` types moved under an `errors` namespace.
* ~~Aliases added in place of previous location, with `[[deprecated]]` annotation (and equivalent for Python, if possible).~~ Decided to allow breaking changes based on [TSC]()
* `BatchElementException` updated to inherit from (recently added) `OpenAssetIOException`
* All `BatchElementException` subclasses removed.
* `BatchElementException` constructor takes the index, the BEE and an optional `suffix` / `additionalIInfo` (or similar) string parameter.  
* The exception message:
  - should be all on one line, for nicer concurrency (flush-on-newline) and greppability
  - should begin with the `BatchElementError` type name (i.e. translate `code` to string)
  - if either the `BatchElementError.message` or `suffix` is provided, then follow with a `: `
  - then should include the `BatchElementError.message`, if non-blank 
  - then should include the suffix, if provided.
* `BatchElementException`-throwing methods updated to construct a `suffix` argument from available relevant information eg:
  - Entity reference as `[<entity reference>]`
  - Access mode as `[access=<mode>]`
  E.g. 
  - `"InvalidPreflightHint: manager-provided error [index=2][access=read][my://entity/ref]"`
  - `"EntityAccessError: [index=2][access=write][my://entity/ref]"`
  - `"Unknown: manager-specific error [index=2]"`
* `CODING_STANDARDS` updated with guidelines for the order of extra data
* log an issue for any future message enhancements

```[tasklist]
### Tasks
- [ ] Implement the feature
- [ ] Re-enable intregrations
```

elliotcmorris:
## Description
Adds friendly messages to batch exceptions when thrown/
Also finalises some errors reorganisation started in the prior non-batch commit, as well as removing the BatchException heirarchy, favouring a single, simple exception for the exceptional batch conveniences.

Closes #1073 

## Reviewer Notes
This PR is in two commits.
 - The first is structural changes, groundwork. Namely, this is moving the batch error/exceptions into the errors namespace and directory. There's lots of mechanical namespace changes in this. **This also includes moving the exception subtypes, which are removed in the following commit.**
 - The second commit adds the messaging functionality to BatchElementException, as well as removing the Batch exception hierarchy following discussion.


src/openassetio-core/include/openassetio/errors/exceptions.hpp
```diff
@@ -4,16 +4,36 @@
...
+#include <utility>
 
 #include <openassetio/export.h>
+#include <openassetio/errors/BatchElementError.hpp>
 #include <openassetio/typedefs.hpp>
 
 namespace openassetio {
 inline namespace OPENASSETIO_CORE_ABI_VERSION {
+/**
+ * This namespace contains types related to error handling.
```
foundrytom, line 16: [consider] Is this now "Exception types" vs general types related to error handling?

src/openassetio-core/src/errors/exceptionMessages.cpp
```diff
@@ -0,0 +1,58 @@
...
+  return fmt::format(
+      "{}{}{}{}{}", fmt::format("{}:", errorCodeName(err.code)),
+      err.message.empty() ? "" : fmt::format(" {}", err.message),
+      fmt::format(" [index={}]", std::to_string(index)),
+      (access.has_value() ? fmt::format(" [access={}]", access::kAccessNames[access.value()])
+                          : ""),
+      (entityReference.has_value()
+           ? fmt::format(" [entity={}]", entityReference.value().toString())
+           : ""));
+}
```
foundrytom, line 55: [consider] Factoring this out from nested `fmt` calls for readability - its quite hard to grok.

src/openassetio-core/src/errors/exceptionMessages.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include <openassetio/typedefs.hpp>
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+namespace errors {
+/**Get an error code name as a printable string*/
+Str errorCodeName(BatchElementError::ErrorCode code);
+
+/**Construct a full message to place into a convenience exception.*/
+std::string createBatchElementExceptionMessage(const BatchElementError &err, size_t index,
```
foundrytom, line 20: [fix] Sorry if I missed them, but looks like this is missing test coverage?

foundrytom (dismissed):
Great stuff, nice to see the hierarchy simplified. What do you think about squashing these together at the end of the day?

src/openassetio-core/tests/hostApi/ManagerTest.cpp
```diff
@@ -243,27 +247,38 @@ SCENARIO("Resolving entities") {
 
       WHEN("singular resolve is called with default errorPolicyTag") {
         THEN("an exception is thrown") {
-          CHECK_THROWS_MATCHES(manager->resolve(ref, traits, resolveAccess, context),
-                               openassetio::MalformedEntityReferenceBatchElementException,
-                               Catch::Message("Error Message"));
+          CHECK_THROWS_MATCHES(
+              manager->resolve(ref, traits, resolveAccess, context),
+              openassetio::errors::BatchElementException,
+              Catch::Predicate<openassetio::errors::BatchElementException>(
```
feltech, line 253: [consider] Factor out this handy `Catch::Predicate` object (or factory for it), since it's duplicated quite a bit.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -74,3 +76,71 @@ void registerErrors(const py::module &mod) {
...
+  // A way forward is provided by the sketch in:
+  // https://github.com/pybind/pybind11/issues/1281#issuecomment-1375950333
+  //
+  // We execute a Python string literal to create our exception
+  // type. The `globals` and `locals` dict parameters dictate the scope
+  // of execution, so we use this to ensure the definition is scoped to
+  // our `_openassetio` module.
+
+  py::exec(R"pybind(
+class BatchElementException(RuntimeError):
```
feltech, line 96: [fix] Inherit from `OpenAssetIOException`.  

Might be worth reviving some of the reverted tests of exception hierarchy.

src/openassetio-core/src/errors/exceptionMessages.hpp
feltech: The message is tested in `test_manager.py` per API function.  Doesn't look like this function is tested in isolation, but many branches in this function are (currently) unused.  I've added a comment re. possibly removing the `optional`ity.

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -3366,6 +2776,246 @@ def test_when_empty_then_no_state_and_stateFromPersistenceToken_is_not_called(
...
+                    ensure_singular(singular_or_multi_ref),
+                )
+            ),
+        ) as exc:
+            manager.resolve(
+                singular_or_multi_ref,
+                an_entity_trait_set,
+                access,
+                a_context,
+                Manager.BatchElementErrorPolicyTag.kException,
```
feltech, line 2856: [fix] We've lost coverage of the fact that the default signature is the exception signature.  

Perhaps yet another parametrization like `"tag",  [[], [kException]]` then passed as `*tag`?

src/openassetio-python/tests/package/test_batchelementerror.py
```diff
@@ -118,44 +108,28 @@ def test_when_comparing_with_different_message_then_are_not_equal(
...
-        expected_code = BatchElementError.ErrorCode.kUnknown
+    def test_when_thrown_then_correct_data_set(self, exception_code):
+        expected_message = "Message"
+        expected_code = exception_code
 
         a_batch_element_error = BatchElementError(expected_code, expected_message)
 
         with pytest.raises(BatchElementException, match=expected_message) as exc:
-            raise exception_type(123, a_batch_element_error)
+            raise BatchElementException(123, a_batch_element_error, expected_message)
```
feltech, line 130: [consider] Differentiate expected message for exception and `BatchElementError`

src/openassetio-python/package/openassetio/errors.py
```diff
@@ -20,9 +20,10 @@
 
 from . import _openassetio  # pylint: disable=no-name-in-module
 
-
 OpenAssetIOException = _openassetio.errors.OpenAssetIOException
 InputValidationException = _openassetio.errors.InputValidationException
 ConfigurationException = _openassetio.errors.ConfigurationException
 NotImplementedException = _openassetio.errors.NotImplementedException
 UnhandledException = _openassetio.errors.UnhandledException
+BatchElementError = _openassetio.errors.BatchElementError
```
feltech, line 28: [fix] Add to list in `cmodule/test_errors.py`

feltech (changes requested):
A few minor niggles, but looking good.

src/openassetio-core/src/errors/exceptionMessages.hpp
foundrytom: Sorry I'd missed [this bit](https://github.com/OpenAssetIO/OpenAssetIO/pull/1109/files#diff-b6b4a676c1e9833a98eec7842db81decc905f6d5562d8b4bc2785b692a431c6eR2794)

src/openassetio-core/src/errors/exceptionMessages.cpp
elliotcmorris: Factored out the substrings, I think it makes the construction clearer.

https://github.com/OpenAssetIO/OpenAssetIO/pull/1109/commits/8760330c5c96feb127885c4703c227c54b18bd4c

Alas the hard wrapping limit forces inconsistency and messiness (imo) here, looks very clean when the ternerys can all get a line to themself, even prior to factoring them out.

src/openassetio-core/include/openassetio/errors/exceptions.hpp
```diff
@@ -65,6 +85,28 @@ struct OPENASSETIO_CORE_EXPORT UnhandledException : OpenAssetIOException {
   using OpenAssetIOException::OpenAssetIOException;
 };
 
+/**
+ * Exception thrown when a @ref BatchElementError is converted.
+ *
+ * Not a type that a manager should throw, exclusively thrown via the
+ * middleware when the user is calling an exceptional convenience and a
+ * @ref BatchElementError is emitted by the manager.
```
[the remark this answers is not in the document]
elliotcmorris: Has potential overlap with Qt thinking but I couldn't think of anything better, ty.

src/openassetio-core/src/errors/exceptionMessages.cpp
```diff
@@ -0,0 +1,58 @@
...
+   * 2. If existing, the message inside the BatchElementError.
+   * 3. The index that the batch error relates to.
+   * 4. If existing, the access.
+   * 5. If existing, the entity reference.
+   *
+   * Ends up looking something like : "entityAccessError: Could not
+   * access Entity [index=2] [access=read] [entity=bal:///entityRef]"
+   */
+  return fmt::format(
+      "{}{}{}{}{}", fmt::format("{}:", errorCodeName(err.code)),
```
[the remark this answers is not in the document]
elliotcmorris: I see the point, but I'd prefer not to do this, The idea that the message is made of a set of independent textual chunks I think is helpful.

src/openassetio-core/src/errors/exceptionMessages.cpp
```diff
@@ -0,0 +1,58 @@
...
+    case BatchElementError::ErrorCode::kInvalidTraitSet:
+      return "invalidTraitSet";
+  }
+
+  return "Unknown ErrorCode";
+}
+
+std::string createBatchElementExceptionMessage(const BatchElementError &err, size_t index,
+                                               std::optional<EntityReference> entityReference,
+                                               std::optional<access::Access> access) {
```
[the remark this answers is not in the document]
elliotcmorris: Aye, that'd work, but i'm not sure worth the template overhead, gonna leave as is.

src/openassetio-python/tests/package/hostApi/test_manager.py
elliotcmorris: Oh right! That's why we were parameterizing over those tags, I thought some prior tags had just made a mistake, took it out on purpose.

src/openassetio-core/src/errors/exceptionMessages.cpp
foundrytom: Def an improvement, thanks. 👍  If the inconsistency is bugging you, what about something along the lines of the following (not sure if it's what @feltech was alluding to)?

```cpp
std::string result = fmt::format("{}:", errorCodeName(err.code));
if( !err.message.empty() ) {
 result += fmt::format(" {}", err.message);
}

result += fmt::format(" [index={}]", index);
if( access.has_value() ) {
    result += fmt::format(" [access={}]", access::kAccessNames[access.value()]);
}
if( entityReference.has_value() ) {
    result += fmt::format(" [entity={}]", entityReference.value().toString())
}

return result;
```

Ironicly, it kind of feels like stringstream could be easier to read here 🙃

src/openassetio-python/package/openassetio/errors.py
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1109/commits/e65bc0d7f63d8d4dc044a8ba031135dfe211c9a1

src/openassetio-python/tests/package/test_batchelementerror.py
elliotcmorris: Yeah, good idea.
https://github.com/OpenAssetIO/OpenAssetIO/pull/1109/commits/d27629a7633fa84a7596f5f56d975896af31ec45

src/openassetio-python/tests/package/hostApi/test_manager.py
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1109/commits/7a0d88222e3288bbe2b7e70777ae718975a351de

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
elliotcmorris: Can't believe i made this mistake _twice_.

https://github.com/OpenAssetIO/OpenAssetIO/pull/1109/commits/84f9425d425e891cbd8985483fe0564d893fb0cf

I've added just a simple test for this, i'd hope we wouldn't need anything more than that for the single type.

src/openassetio-core/src/errors/exceptionMessages.cpp
```diff
@@ -0,0 +1,58 @@
...
+      return "invalidPreflightHint";
+    case BatchElementError::ErrorCode::kInvalidTraitSet:
+      return "invalidTraitSet";
+  }
+
+  return "Unknown ErrorCode";
+}
+
+std::string createBatchElementExceptionMessage(const BatchElementError &err, size_t index,
+                                               std::optional<EntityReference> entityReference,
```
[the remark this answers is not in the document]
elliotcmorris: Yeah, that's a good point, let's YAGNI it. Removes the test overhead too. https://github.com/OpenAssetIO/OpenAssetIO/pull/1109/commits/77f47a7b6b3f42f605a01e7dec51436ad11ac99f

src/openassetio-core/src/errors/exceptionMessages.cpp
```diff
@@ -0,0 +1,58 @@
...
+      return "entityAccessError";
+    case BatchElementError::ErrorCode::kEntityResolutionError:
+      return "entityResolutionError";
+    case BatchElementError::ErrorCode::kInvalidPreflightHint:
+      return "invalidPreflightHint";
+    case BatchElementError::ErrorCode::kInvalidTraitSet:
+      return "invalidTraitSet";
+  }
+
+  return "Unknown ErrorCode";
```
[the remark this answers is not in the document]
elliotcmorris: Honestly not sure on best practice here anymore. Do we trust our external tools or double up... 
Added it nonetheless https://github.com/OpenAssetIO/OpenAssetIO/pull/1109/commits/9e187dacaccf734bb44f16466b2f426144b60fb4

src/openassetio-core/src/errors/exceptionMessages.cpp
elliotcmorris: SS would still have the inconsistent wrapping, that's just to do with line length.
I don't think the new version is unreadable at all tbh, and it's even better now that we've YAGNI's the optionals. Resolving this.

src/openassetio-core/src/errors/exceptionMessages.cpp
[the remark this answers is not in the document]
feltech: Roll on `std::unreachable`. Is it worth using the trick of `assert(false && "some message")`?

src/openassetio-core/src/errors/exceptionMessages.cpp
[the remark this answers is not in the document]
feltech: OK, new version looks good. Personally I would try to avoid intermediate strings, but that's a micro-optimisation not important in this case.

src/openassetio-core/tests/hostApi/ManagerTest.cpp
feltech: Nice

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: Could we add the same parametrization to the other tests too?

elliotcmorris: I'm a moron ... yes.

elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1109/commits/d90d160c0fc9b33a78f83a47eca6fd0fe207e93e

src/openassetio-core/src/errors/exceptionMessages.cpp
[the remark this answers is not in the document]
elliotcmorris: I don't really think so, the assert is just a marker that this should be impossible, I think it's obvious enough as it stands.

feltech (approved):
LGTM! Second commit message could do with the issue ref.

elliotcmorris:
Merged this into the one commit as Tom suggested, in reflection, the migration and then removal of the extended type heirarchy just adds confusion to the history imo.

Was gonna merge this but can't until the changes requested flag by @foundrytom is removed.

Issue #838, opened 2023-03-20
Add test of `dist` structure


foundrytom:
## What

Ensure the dist tree matches expected contents (no additions either!)

## Why

https://github.com/OpenAssetIO/OpenAssetIO-Test-CMake/issues/8 removed explicit exhaustitive header checks.

foundrytom:
NB: consider all files including stray additions

PR #1111, opened 2023-09-19
Add Python `.dist-info`

### LINKED ISSUE #1088: Create `dist-info` with CMake built Python packages
## What

Provide a skeleton `dist-info` along side the `openassetio` Python package in `dist`.

## Why

`openassetio` is an "interesting" dependency. It should generally be provide by the Host (as they're the ones initializing/using the API). However it is also a dependency of middleware like traits packages, and manager plugins.

In a simple, pure Python world, all components could list it as a dependency, and something like `pip` would install the right version, and flag any issues.

However, we are not in a simple pure Python world. Many hosts are C++, and so manage the dependency in a different way.

This poses a problem, as a C++ host, that provides a Python interpreter needs to single-source `openassetio` between C++ and Python using the dynamic library build (to avoid multiple copies of `openassetio` being loaded simultaneously, with potentially conflicting versions).

Presently, we don't provide a `dist-info` with our CMake build Python package. If we add a hard dependency on `openassetio` in a manager, when the manager is installed, the package manager can't see that `openassetio` is already satisfied, and instead installs its own (static linked) version, with out any regard to the version pin imposed by the host. This would mean that data objects returned by the manager may not be runtime compatible with the host.

Long-term, hosts may opt to use the C isolation layer, which would mitigate this issue (at the expense of a copy). We're not there yet, and so we need to ensure that an dependency check as part of a package install is able to 'see' the existing `openassetio` installation provided by a host with its embedded interpreter.

By providing a `dist-info` directory with our manual builds, it will allow package managers to determine if the fixed host provided package is compatible with the versioning requirements of the other packages being installed, and stop a second, potentially conflicting version being installed.

The same is true for MediaCreation.

## Acceptance Criteria

- CMake builds of the Python package include a `dist-info` directory with:
  - A minimal `METADATA` required by downstream tooling (or more if needed). 
  - An `INSTALLER` entry that lists `cmake`
  - No `RECORD` to [prevent other package managers from trying to uninstall it](https://packaging.python.org/en/latest/specifications/recording-installed-packages/#intentionally-preventing-changes-to-installed-packages) as part of a dependency upgrade.
  - A CMake option to disable `dist-info` generation.
- When this build is manually installed to a python interpreter, then`-m pip` installing a package with:
  - ... a compatible `openassetio` dependency, using that interpreter, sees the existing installation and considers the dependency satisfied.
  - ... an incompatible `openassetio` dependency errors, and does not attempt to uninstall/update the package
  - Tests that `pip`can see the package and version, and un-installation fails.

feltech:
## Description

Closes #1088
- #1088

Add a `.dist-info` directory when `cmake --install`ing the project with Python enabled. This effectively prevents a `pip install` from overwriting it, with caveats (see below).

## Reviewer Notes

If this looks good, then I propose to simply duplicate it in OpenAssetIO-MediaCreation, rather than make it generic and portable and figure out a way for projects to share CMake libraries.  If we had more target projects, or there was expectation this feature would change frequently, then putting the effort in would be worthwhile.  But not for the sake of two projects.

The CI is structured differently than the rest of our workflows. For simplicity, it uses `bash` on all platforms, with the knock on effect of using the default CMake generator for the platform (see commit), and is roughly what I propose in #1114 .

~~CI only tests the effect we want, rather than deep tests of the `dist-info` structure. I propose to do that as part of #838, if needed.~~

## Test Instructions

~~As noted above, CI tests are limited to checking that `pip install` fails as expected. You may wish to manually inspect the contents of the installed files under `dist-info`.~~

`pip` will warn when installing `openassetio` into an environment where the `PYTHONPATH` contains another copy of `openassetio`, e.g.

> Not uninstalling openassetio at [...], outside environment

but will still install a competing copy.

However, if `openassetio` is installed into the Python environment directly, the we get an error, e.g.

> ERROR: Cannot uninstall openassetio 1.0.0a14, RECORD file not found. Hint: The package was installed by cmake.

and pip fails to overwrite the CMake installed `openassetio`.


.github/workflows/build-instructions.yml
```diff
@@ -116,3 +116,40 @@ jobs:
 
     - name: Run Python tests
       run: pytest src/openassetio-python/tests/package
+
```
foundrytom, line 119: [fix] Capturing offline chat:

Be great to have local coverage of this. So add `ctest` checks that:
 - `python -m pip list` contains `openassetio` with the correct version (can use the cmake version)
 - no `RECORD` file exists, as this is the ["public API"](https://packaging.python.org/en/latest/specifications/recording-installed-packages/#intentionally-preventing-changes-to-installed-packages) for preventing removal.

foundrytom (changes requested):
Works a charm - nice one! Good call on the extra file.

foundrytom:
Oh,[`BUILDING.md`](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/BUILDING.md#cmake-options) needs updating too

feltech:
> Oh,[`BUILDING.md`](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/BUILDING.md#cmake-options) needs updating too

What's the specific issue with `BUILDING.md?`  The new CMake option was added.

foundrytom:
> > Oh,[`BUILDING.md`](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/BUILDING.md#cmake-options) needs updating too
> 
> What's the specific issue with `BUILDING.md?` The new CMake option was added.

Lol, sorry don't know how I missed that! I may have been inadvertently looking at the tree for just one commnit 🙃 sorry about that. Ignore me.

src/openassetio-python/tests/CMakeLists.txt
```diff
@@ -19,6 +19,72 @@ add_subdirectory(cmodule)
...
+# I.e. that `pip install` will error when attempting to overwrite a
+# CMake installed package.
+
+add_custom_target(
+    openassetio.internal.python-package-no-record-test
+    COMMAND cmake -E echo -- "Check for a files under dist-info"
+    # Abuse md5sum to check existence of METADATA file.
+    COMMAND
+    ${CMAKE_COMMAND} -E md5sum
+    "${CMAKE_INSTALL_PREFIX}/${OPENASSETIO_PYTHON_SITEDIR}/openassetio-${PROJECT_VERSION}*.dist-info/METADATA"
```
foundrytom, line 65: [question] Whilst were at it, is it worth also checking `INSTALLER` as that ensures we have the friendly message?

feltech: We could, though this is implicitly tested in CI. Shall I go ahead and test it?  Do we need a test of its contents too?

foundrytom: Up to you - if this is more a 'test the structure' test vs a 'test no record' test...

.github/workflows/build-instructions.yml
feltech: Ended up with 82eacd40 - using pytest and `importlib`, since testing cross-platform using shell commands got rather unwieldly.

src/openassetio-python/tests/CMakeLists.txt
feltech: Incorporated into pytest tests in dd0e3e6b

CMakeLists.txt
```diff
@@ -120,6 +120,9 @@ if (OPENASSETIO_ENABLE_PYTHON)
         ""
         CACHE STRING
         "Override default Python module install directory, relative to CMAKE_INSTALL_PREFIX")
+
+    option(OPENASSETIO_ENABLE_PYTHON_INSTALL_DIST_INFO
+        "Create a dist-info metadata directory alongside Python installation" ON)
```
elliotcmorris, line 125: [Consider][Minor]
Just trying to be helpful to folks who might not understand the point of this.
```suggestion
    option(OPENASSETIO_ENABLE_PYTHON_INSTALL_DIST_INFO
        "Create a dist-info metadata directory alongside Python installation. Allows pip to understand that OpenAssetIO is installed." ON)
```

cmake/packaging/python.dist-info/INSTALLER
```diff
@@ -0,0 +1 @@
+cmake
```
elliotcmorris, line 1: [Consider][minor] I wonder if a README file for the .dist-info dir, (or maybe the packaging dir?) might be helpful. I imagine it would be difficult to understand what these files are for when coming across them. And a simple "These files are templates for generating .dist-info when creating a python package via cmake" in a readme or something could help.

I understand a README file might seem a bit overkill here ... I wish directories themselves had metadata you could add to them which could serve as comments, or something like that.

doc/BUILDING.md
```diff
@@ -177,26 +177,27 @@ about [here](SANDBOXED_BUILDS.md).
...
-| `OPENASSETIO_ENABLE_SANITIZER_UNDEFINED_BEHAVIOR`   | Enable undefined behavior sanitizer | `OFF` |
-| `OPENASSETIO_ENABLE_SANITIZER_THREAD` | Enable thread sanitizer | `OFF` |
-| `OPENASSETIO_ENABLE_SANITIZER_MEMORY` | Enable memory sanitizer | `OFF` |
-| `OPENASSETIO_ENABLE_CLANG_TIDY` | Enable clang-tidy analysis during build | `OFF` |
-| `OPENASSETIO_ENABLE_CPPLINT` |Enable cpplint linter during build | `OFF` |
+| Option                                            | Description                                                           | Default |
+|---------------------------------------------------|-----------------------------------------------------------------------|---------|
+| `BUILD_SHARED_LIBS`                               | `ON` to build shared libraries. `OFF` to build static libraries.      | `ON`    |
+| `OPENASSETIO_ENABLE_PYTHON`                       | Additionally build python bindings                                    | `ON`    |
+| `OPENASSETIO_ENABLE_PYTHON_INSTALL_DIST_INFO`     | Create a dist-info metadata directory alongside Python installation   | `ON`    |
```
elliotcmorris, line 184: [Consider][Minor] Same as before, maybe gesture at the point of dist-info rather than just saying what it does.

elliotcmorris (approved):
LGTM. Just a couple of minor considerations around conveying intent.

CMakeLists.txt
feltech: Augmented the description in 198c7f25

cmake/packaging/python.dist-info/INSTALLER
feltech: Added a README in 2c654bfd (for first commit that creates the directory) then updated the text in c718188f (as part of adding the dist-info).

doc/BUILDING.md
feltech: I'm reluctant here, since this table is already really wide in the plain-text view.  If we could come up with better wording that isn't any longer, I'd be happy to change it.

I note that we have several other options that are explained in techie terms.  Perhaps it's worth a [plan] to have some dedicated sub-headings for options that are not immediately obvious from their short description?

foundrytom (approved):

doc/BUILDING.md
elliotcmorris: Fair doo's

I wouldn't bother planning something like that ... I don't think we'd ever come back to it :P

elliotcmorris (approved):

PR #1120, opened 2023-09-21
Minor docs tweaks to aid host development

foundrytom:
A few tweaks based on observations of real-world implementations.

### DIFF
--- a/doc/doxygen/src/ForHosts.dox
+++ b/doc/doxygen/src/ForHosts.dox
@@ -98,21 +98,18 @@
  *   user with an additional manual configuration mechanism for your
  *   specific API session if you so wish.
  *
- * - Create (and configure) a @ref Context as appropriate for
- *   all calls to the API. The Context instance should be kept alive
- *   and re-used for API calls that are related to the same user
- *   session, or logical set of interactions. For example, one per
- *   document. Use @fqref{hostApi.Manager.createChildContext}
- *   "createChildContext" to split off a new Context should you need to
- *   re-configure it for any specific series of actions, whilst
- *   maintaining the association.
+ * - Create (and configure) a @ref Context as appropriate for calls to
+ *   the API. Re-use context objects for all API calls that are related
+ *   to the same user session or logical set of interactions. For
+ *   example, one per document or working environment.
  *
  * - Ask the @ref manager if they support resolve for the relevant
  *   data by supplying its @ref trait_set to
  *   @fqref{hostApi.Manager.managementPolicy} "Manager.managementPolicy"
  *   with a @ref Context with `kRead` access, respecting the returned
  *   traits - this may be a subset if the manager is only capable of
- *   resolving some of an entity's traits.
+ *   resolving some of an entity's traits. Note that management policy
+ *   is runtime invariant for any given context/access/trait set.
  *
  * - Always check any suspected entity references with
  *   @fqref{hostApi.Manager.isEntityReferenceString}
--- a/doc/doxygen/src/ForHosts.dox
+++ b/doc/doxygen/src/ForHosts.dox
@@ -98,8 +98,14 @@
  *   user with an additional manual configuration mechanism for your
  *   specific API session if you so wish.
  *
- * - Create, persist and configure a @ref Context as appropriate for
- *   all calls to the API.
+ * - Create (and configure) a @ref Context as appropriate for
+ *   all calls to the API. The Context instance should be kept alive
+ *   and re-used for API calls that are related to the same user
+ *   session, or logical set of interactions. For example, one per
+ *   document. Use @fqref{hostApi.Manager.createChildContext}
+ *   "createChildContext" to split off a new Context should you need to
+ *   re-configure it for any specific series of actions, whilst
+ *   maintaining the association.
  *
  * - Ask the @ref manager if they support resolve for the relevant
  *   data by supplying its @ref trait_set to
@@ -112,6 +118,10 @@
  *   @fqref{hostApi.Manager.isEntityReferenceString}
  *   "Manager.isEntityReferenceString" before passing to any other API
  *   calls.
+ *
+ * - If you support distributed processing provide a mechanism to
+ *   @fqref{hostApi.Manager.persistenceTokenForContext} "persist" a @ref
+ *   Context across processes.
  *
  * - Provide means for users to 'de-assetize' documents to facilitate
  *   sharing outside of the assetized environment. This usually consists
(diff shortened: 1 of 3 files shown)


doc/doxygen/src/ForHosts.dox
```diff
@@ -98,21 +98,32 @@
...
+ *   runtime invariant for any given context/access/trait set.
  *
  * - Always check any suspected entity references with
  *   @fqref{hostApi.Manager.isEntityReferenceString}
  *   "Manager.isEntityReferenceString" before passing to any other API
  *   calls.
  *
+ * - If you support distributed processing provide a mechanism to
+ *   @fqref{hostApi.Manager.persistenceTokenForContext} "persist" a @ref
+ *   Context across processes.
```
elliotcmorris, line 125: [Question] Curious where this comes from? Seems like a non-simple thing to call out in the requirements for simple resolution section.

doc/doxygen/src/ForHosts.dox
```diff
@@ -98,21 +98,32 @@
...
- * - Create, persist and configure a @ref Context as appropriate for
- *   all calls to the API.
+ * - Create (and configure) a @ref Context as appropriate for
+ *   all calls to the API. The Context instance should be kept alive
+ *   and re-used for API calls that are related to the same user
+ *   session, or logical set of interactions. For example, one per
+ *   document. Use @fqref{hostApi.Manager.createChildContext}
+ *   "createChildContext" to split off a new Context should you need to
+ *   re-configure it for any specific series of actions, whilst
+ *   maintaining the association.
```
elliotcmorris, line 108: [Consider] I wonder if this is too much prose for the simple checklist section. Perhaps consider
```suggestion
 * - Create (and configure) a @ref Context as appropriate for
 *   all calls to the API. Maintain the same context object for
 *   all subsequent calls related to the same logical set of interactions.
 *   (Link to somewhere explaining/exampling context more deeply.)
```

doc/doxygen/src/ForHosts.dox
```diff
@@ -98,21 +98,32 @@
...
  *
  * - Ask the @ref manager if they support resolve for the relevant
  *   data by supplying its @ref trait_set to
  *   @fqref{hostApi.Manager.managementPolicy} "Manager.managementPolicy"
  *   with a @ref Context with `kRead` access, respecting the returned
  *   traits - this may be a subset if the manager is only capable of
- *   resolving some of an entity's traits.
+ *   resolving some of an entity's traits. Note that policy is
+ *   runtime invariant for any given context/access/trait set.
```
elliotcmorris, line 116: [Question] What's the action that a implementor should take in response to this note? I suppose you saw this come up somewhere, but I'm struggling to see what the outcome is, is it like, you don't need to test the management policy every time for the same trait set?

Edit : Ah yes, i see you've outlined that below, I wonder if the docs from there would also fit here. This is a checklist and the things in it should be actionable.

elliotcmorris: [consider][minor] Trying to keep phrasing consistent.
```suggestion
 *   resolving some of an entity's traits. Note that managementPolicy is
 *   runtime invariant for any given set of inputs.
```

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -354,6 +356,10 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    *  distributed between processes to associate subsequent API usage
    *  with the supplied context.
    *
+   *  @note This should not be used within the same process to store
+   *  a context for use between subsequent API calls or threads. Retain
+   *  the Context object directly in this situation.
+   *
```
elliotcmorris, line 362: [Consider] Feeling here is that the prior framing prescriptively implies it is forbidden and there's a special reason things will go wrong if you do, (which I don't _think_ is the case,) whereas this framing does a better job at informing the user why they wouldn't want to use it.
```suggestion
   *  @note Using this within the same process to store a context
   *  for use between subsequent API calls or threads is redundant
   *  Retain the Context object directly in this situation.
   *
```

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -357,13 +357,14 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
    *
-   * This method is usually called early on by a host to determine
-   * whether to enable OpenAssetIO related functionality when handling
-   * specific kinds of data. The host will often adapt its subsequent
-   * behaviour to minimise unsupported interactions with the manager. In
-   * high call volume scenarios (such as CG rendering), this can
-   * potentially save hundreds of thousands of redundant calls into the
-   * API for unmanaged entity traits.
+   * A manager's policy must be runtime invariant for any given set of
+   * inputs. This method is usually called early on by a host to
```
elliotcmorris, line 361: [Consider] I think this is an innapropriate place for this, it's taking the oxygen away from the description of how management policy is used.

I'd recommend either adding it below in it's own paragraph (or as a `@note`), or appending it to the end of the initial paragraph, which is more definitional already.

doc/doxygen/src/ForHosts.dox
foundrytom: Rendering different frames across different machines is very common in post production. This mechanism is essential to allow any kind of temporal stability.

doc/doxygen/src/ForHosts.dox
foundrytom: The action is "Ask the @ref manager if they support resolve...", the note was just providing more context whilst we had their attention, in case they didn't read the rest of the docs.

Re details vs 'inputs', as we dont describe the signature in this para, it felt like it helped give a more comprehensive picture of the invariance, that didn't need a follow up investigation. So the reader may hopefully come away with a deeper understanding of the underlying concept if this is all `they.`

doc/doxygen/src/ForHosts.dox
foundrytom: Had opted to add a little more here as the short version inevitably seems to confuse people without examples.

foundrytom: Hows [this](https://github.com/OpenAssetIO/OpenAssetIO/pull/1120/commits/15b27a40230dd7a8573be1f0185be81f9d7899ba) as an alternative?

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
foundrytom: _Much_ better, thanks. Updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/1120/commits/3616d47df4d640590c7cc75dd9a90dbeed7404e4).

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
foundrytom: Hows [this](https://github.com/OpenAssetIO/OpenAssetIO/pull/1120/commits/748bb5445f1c5e26472091159050853b770ce1f4)?

elliotcmorris (approved):

feltech (approved):

PR #1121, opened 2023-09-21
Don't scan Python hoists

### LINKED ISSUE #1093: Deprioritise hoist links in doxygen output
## What
In at least one view in the docs, only one class of the same name can be displayed. (Specifically i'm thinking of the `Namespace Members->All` page.)
A consequence of this is that when we have python hoists of the same name as the c++ types, it looks to be selecting the python hoist as the item to link to, rather than the useful c++ page.

This came up when doing errors, see screenshot : 
image

There is no duplicate `BatchElementError` in the page that links to `v1`, that's the only reference.

We should find a way to prioritise linking the the c++ docs, perhaps by configuring doxygen to not scan the hoists.
## Why
The C++ should be our main docs in all views, linking to the hoists in unhelpful.

foundrytom:
We were accidentally `@ref`ing Python hoists. This updates those refs, and then removes the hoists from scanning, which cleans up the class index.

Closes #1093

### DIFF
--- a/doc/doxygen/Doxyfile
+++ b/doc/doxygen/Doxyfile
@@ -796,10 +796,15 @@ WARN_LOGFILE           =
 # spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = ../../src/openassetio-python/package \
-                         ./src \
+ # In the Python world, only include the specific paths to pure-python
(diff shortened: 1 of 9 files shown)


elliotcmorris (approved):
LTGM.

Only `openassetio.test` and `openassetio.hostApi.terminology` remain in the all namespace lists as python elements, which seems correct for the moment.

PR #1126, opened 2023-09-26
`hasCapability` Method.

### LINKED ISSUE #1113: Capability introspection API
## What

Add a mechanism to introspect the capabilities of a manager (and later, perhaps a host).

## Why

#1101 defines that default implementations will throw a `NotImplementedError`, as such, we need a simple, cheap way to determine in advance whether a manager implements any specific capability to avoid onerous error handling boilerplate at method call sites.

We use the term 'capability' as this is not a method-wise concept. For example, `preflight` and `register` must both be implemented to support publishing. Similarly, the `getWithRelationsip*` methods must also be implemented together.

## Acceptance Criteria

- A host can trivially query the supported capabilities of a Manager.
- The method must be:
  - Invariant post `initialize`
  - Guaranteed cheap to call for the caller
  - Simple to use in conditionals without elaborate value unpacking
  - Customizable by the `ManagerInterface` implementation to support wrapper interfaces such as gRPC, fan-out or multiplex.
  - Easy as possible for a manager author to implement correctly/consistently.
- `apiComplianceSuite` validation that the manager honors its response from this method.
- If practical, default implementations/helpers provided that reduce programming error for the manager implementor.
- stringification name lookup table as per `Access`.
- Update `forManagers.dox`

```[tasklist]
### Tasks
- [ ] https://github.com/OpenAssetIO/OpenAssetIO/issues/1119
- [ ] https://github.com/OpenAssetIO/OpenAssetIO/issues/1128
- [ ] https://github.com/OpenAssetIO/OpenAssetIO/issues/1124
- [ ] Create issues for update of dependent projects
```

## Notes

Sketch capturing preliminary discussions

```c++
namespace capability {

enum class ManagerCapability {
   kCustomState = 0  // createState, createChildState, persistenceTokenForState, stateFromPersistenceToken
   kCustomTerminology,  // updateTerminology
   // Put 'verbs' later to help group as we add methods like entityTraitSet etc
   kResolution,  // resolve
   kPublishing,  // preflight, register
   kRelationshipQueries,  // getWithRelationship*
   kDefaultEntityReferences,  // defaultEntityReference
}

static std::string managerCapabilityNames = { "resolution", ... };
} 

virtual bool ManagerInterface::hasCapability( ManagerCapability c ) const = 0;
```
Methods not covered above default to nops/empty data as they are effectively already introspection methods.

### LINKED ISSUE #1119: Core `Manager.hasCapability` API method
See #1113

Out of scope for this task:
- Default implementation/helpers
- Dependent project updates
- apiCompliance 
- Host doc updates

elliotcmorris:
## Description
Add `hasCapability` method, along with the supporting enum.
Documentation added to the enum in particular to explain the concepts.

Closes #1119 

paired implementation with @foundrytom 

## Reviewer Notes

This implementation needed some C-api changes because it's pure virtual, so that's in there.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -11,6 +11,11 @@ v1.0.0-alpha.xx
   namespace will be removed in a future release.
   [#1127](https://github.com/OpenAssetIO/OpenAssetIO/issues/1127)
 
+- Add the `Manager`/`ManagerInterface` `hasCapability` method. A
+  mechanism for the host to query the manager about which particular
+  optional method-sets, or capabilities, it implements.
+  [#1113](https://github.com/OpenAssetIO/OpenAssetIO/issues/1113)
+
 ### Breaking changes
 
 - Renamed the `Manager`/`ManagerInterface`
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -1118,12 +1118,13 @@ class OPENASSETIO_CORE_EXPORT Manager final {
   /**
    * @name Publishing
    *
-   * The publishing functions allow the host to create an @ref entity
-   * within the @ref asset_management_system represented by the Manager.
-   * The API is designed to accommodate the broad variety of roles that
-   * different asset managers embody. Some are 'librarians' that simply
-   * catalog the locations of existing media. Others take an active role
-   * in both the temporary and long-term paths to items they manage.
+   * The publishing functions allow the host to create or update an @ref
+   * entity within the @ref asset_management_system represented by the
+   * Manager. The API is designed to accommodate the broad variety of
+   * roles that different asset managers embody. Some are 'librarians'
+   * that simply catalog the locations of existing media. Others take an
+   * active role in both the temporary and long-term paths to items they
+   * manage.
    *
    * There are two key components to publishing within this API.
    *
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -6,6 +6,7 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <type_traits>
 #include <vector>
 
 #include <openassetio/export.h>
@@ -14,11 +15,12 @@
 #include <openassetio/access.hpp>
 #include <openassetio/errors/BatchElementError.hpp>
 #include <openassetio/hostApi/EntityReferencePager.hpp>
+#include <openassetio/internal.hpp>
 #include <openassetio/trait/collection.hpp>
 #include <openassetio/typedefs.hpp>
 
-OPENASSETIO_FWD_DECLARE(managerApi, HostSession)
 OPENASSETIO_FWD_DECLARE(managerApi, ManagerInterface)
+OPENASSETIO_FWD_DECLARE(managerApi, HostSession)
 OPENASSETIO_FWD_DECLARE(Context)
 
 namespace openassetio {
@@ -111,6 +113,112 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * asset_management_system itself.
    * @{
    */
+
+  /**
+   * Capabilities that the manager implements.
+   *
+   * Many OpenAssetIO methods are optional. This enum is used with the
+   * introspection mechanism @ref hasCapability to provide a means of
+   * querying which sets of methods the manager provides.
+   */
+  enum class Capability : std::underlying_type_t<internal::capability::manager::Capability> {
+    /**
+     * Manager makes use of the context to persist custom state for
+     * performance reasons or otherwise.
+     *
+     * @note If this capability is true, then the host must reuse the
+     * same context across related API calls (including the use of @ref
+     * persistenceTokenForContext when the calls are distributed
+     * cross-process).
+     *
+     * @see @ref stable_resolution
+     */
+    kStatefulContexts = internal::capability::manager::Capability::kStatefulContexts,
+    /**
+     * Manager customizes certain human-readable strings that the host
+     * might want to use in UI/messages.
+     *
+     * This capability means the manager implements the following
+     * methods:
+     * - @ref updateTerminology
+     */
+    kCustomTerminology = internal::capability::manager::Capability::kCustomTerminology,
+    /**
+     * Manager is capable of resolving @ref entity_reference into the
+     * data for one or more @ref trait "traits".
+     *
+     * This capability means the manager implements the following
+     * methods:
+     * - @ref resolve
+     */
+    kResolution = internal::capability::manager::Capability::kResolution,
+    /**
+     * Manager allows the host to create or update an @ref entity within
+     * the @ref asset_management_system.
+     *
+     * This capability means the manager implements the following
+     * methods:
+     * - @ref preflight
+     * - @ref register_
+     */
+    kPublishing = internal::capability::manager::Capability::kPublishing,
+    /**
+     * Manager is capable of querying entity references that are related
+     * to the input references by the relationship defined by a set of
+     * traits and their properties.
+     *
+     * This capability means the manager implements the following
+     * methods:
+     * - @ref getWithRelationship
+     * - @ref getWithRelationships
+     */
+    kRelationshipQueries = internal::capability::manager::Capability::kRelationshipQueries,
+    /**
+     * Manager is capable of confirming the existence of entities.
+     *
+     * This capability means the manager implements the following
+     * methods:
+     * - @ref entityExists
+     */
+    kExistenceQueries = internal::capability::manager::Capability::kExistenceQueries,
+
+    /**
+     * Manager may be capable of a providing an @ref EntityReference
+     * considered to be a sensible default for a particular @ref
+     * trait_set "trait set".
+     *
+     * This capability means the manager implements the following
+     * methods:
+     * - @ref defaultEntityReference
+     */
+    kDefaultEntityReferences = internal::capability::manager::Capability::kDefaultEntityReferences
+  };
+
+  /**
+   * Query the manager as to which capabilities it implements.
+   *
+   * API methods are grouped into "capabilities", which are independent
+   * groupings of functionality. For example, @ref
+   * Capability.kPublishing "publishing" or @ref Capability.kResolution
+   * "resolution".
+   *
+   * Support for each of these capabilities is optional for the manger,
+   * and the default implementation will throw a
+   * @fqref{errors.NotImplementedException} "NotImplementedException".
+   *
+   * This method can be called after @ref initialize to determine
+   * whether a manager supports a given capability. It's a low-overhead
+   * call, whose return value remains constant once the manager has been
+   * initialized.
+   *
+   * For information on what methods belong to which capability set,
+   * @see @ref Capability.
+   *
+   * @param capability The capability to check.
+   *
+   * @return Whether the manager has the capability in question.
+   */
+  [[nodiscard]] bool hasCapability(Capability capability);
 
   /**
    * Returns other information that may be useful about this @ref
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -70,9 +70,9 @@ class OPENASSETIO_CORE_EXPORT Manager final {
                                        managerApi::HostSessionPtr hostSession);
 
   /**
-   * @name Asset Management System Information
+   * @name Asset Management System Identification
    *
-   * These functions provide general information about the @ref
+   * These functions provide general identity information about the @ref
    * asset_management_system itself. These can all be called before
    * @ref initialize has been called.
    *
@@ -100,6 +100,18 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    *     "OpenAssetIO Test Manager"
    */
   [[nodiscard]] Str displayName() const;
+
+  /**
+   * @}
+   */
+
+  /**
+   * @name Asset Management System Information
+   *
+   * These functions provide general information about the @ref
+   * asset_management_system itself.
+   * @{
+   */
 
   /**
    * Returns other information that may be useful about this @ref
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -13,7 +13,6 @@
 #include <openassetio/InfoDictionary.hpp>
 #include <openassetio/access.hpp>
 #include <openassetio/errors/BatchElementError.hpp>
-#include <openassetio/errors/exceptions.hpp>
 #include <openassetio/hostApi/EntityReferencePager.hpp>
 #include <openassetio/trait/collection.hpp>
 #include <openassetio/typedefs.hpp>
--- a/src/openassetio-core/include/openassetio/internal.hpp
+++ b/src/openassetio-core/include/openassetio/internal.hpp
@@ -6,7 +6,8 @@ inline namespace OPENASSETIO_CORE_ABI_VERSION {
 /**
  * Contains private internal implementation details.
  */
-namespace internal::access {
+namespace internal {
+namespace access {
 /**
  * Common constant values for strong enumerations of access pattern.
  *
@@ -16,6 +17,30 @@ namespace internal::access {
  * bindings.
  */
 enum Access : std::size_t { kRead = 0, kWrite, kCreateRelated };
-}  // namespace internal::access
+}  // namespace access
+
+namespace capability::manager {
+/**
+ * Common constant values for strong enumerations of capability sets.
+ *
+ * We must ensure correspondence between the values of the middleware
+ * and manager implementation's strong enums. This allows for
+ * comparison, string lookup, simplifies serialisation, and supports
+ * language (especially C) bindings.
+ */
+enum Capability : std::size_t {
+  kEntityReferenceIdentification = 0,
+  kManagementPolicyQueries,
+  kStatefulContexts,
+  kCustomTerminology,
+  kResolution,
+  kPublishing,
+  kRelationshipQueries,
+  kExistenceQueries,
+  kDefaultEntityReferences
+};
+}  // namespace capability::manager
+
+}  // namespace internal
 }  // namespace OPENASSETIO_CORE_ABI_VERSION
 }  // namespace openassetio
--- a/src/openassetio-core/include/openassetio/internal.hpp
+++ b/src/openassetio-core/include/openassetio/internal.hpp
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2023 The Foundry Visionmongers Ltd
+#pragma once
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+/**
+ * Contains private internal implementation details.
+ */
+namespace internal::access {
+/**
+ * Common constant values for strong enumerations of access pattern.
+ *
+ * We must ensure correspondence between the values of the various
+ * workflows' strong enums. This allows for comparison, string lookup,
+ * simplifies serialisation, and supports language (especially C)
+ * bindings.
+ */
+enum Access : std::size_t { kRead = 0, kWrite, kCreateRelated };
+}  // namespace internal::access
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
--- a/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
+++ b/src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
@@ -7,6 +7,7 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <type_traits>
 #include <vector>
 
 #include <openassetio/export.h>
@@ -14,6 +15,7 @@
 #include <openassetio/InfoDictionary.hpp>
 #include <openassetio/access.hpp>
 #include <openassetio/errors/BatchElementError.hpp>
+#include <openassetio/internal.hpp>
 #include <openassetio/trait/collection.hpp>
 #include <openassetio/typedefs.hpp>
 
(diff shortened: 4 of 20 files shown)


foundrytom:
Draft as we need to:
- Split between `Manager` and `ManagerInterface`
- Add `kEntityReferenceIdentification` and `kManagementPolicyQueries`

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -101,6 +101,128 @@ class OPENASSETIO_CORE_EXPORT Manager final {
...
+   * Query the manager as to which capabilities it implements.
+   *
+   * API methods are grouped into "capabilities", which are independent
+   * groupings of functionality. For example, @ref
+   * Capability.kPublishing "publishing" or @ref Capability.kResolution
+   * "resolution".
+   *
+   * Support for each of these capabilities is optional for the manger,
+   * and the default implementation will throw
+   * @fqref{errors.NotImplementedException} "NotImplementedException"
```
foundrytom, line 212: [fix]
```suggestion
   * a @fqref{errors.NotImplementedException} "NotImplementedException".
```

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -101,6 +101,128 @@ class OPENASSETIO_CORE_EXPORT Manager final {
...
+     * methods:
+     * - @ref entityExists
+     */
+    kExistenceQueries =
+        static_cast<int>(managerApi::ManagerInterface::Capability::kExistenceQueries),
+
+    /**
+     * Manager may be capable of a providing an @ref EntityReference
+     * considered to be a sensible default for each of the given entity
+     * @ref trait_set "trait sets"
```
feltech, line 192: [fix] Misleading to say "each of the given entity trait sets", since none are given as part of this function. Suggest just removing that bit.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -216,6 +217,171 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+                                                                "resolution",
+                                                                "publishing",
+                                                                "relationshipQueries",
+                                                                "existenceQueries",
+                                                                "defaultEntityReferences"};
+
+  /**
+   * Query the manager as to which capabilities it implements.
+   *
+   * API methods are grouped into "capabilities", which are independent
```
foundrytom, line 365: [fix]

```suggestion
   * This method will only be called post-initialize, but must be cheap to
   * evaluate, and always return the same values.
   * 
   * API methods are grouped into "capabilities", which are independent
```

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -240,6 +240,54 @@ def test_when_entity_ref_prefix_type_invalid_then_debug_log_printed(
...
+        "manager_capability,managerinterface_capability", manager_capabilities
+    )
+    def test_values_match_managerinterface(self, manager_capability, managerinterface_capability):
+        assert manager_capability.value == managerinterface_capability.value
+
+
+class Test_Manager_hasCapability:
+    def test_method_defined_in_cpp(self, method_introspector):
+        assert not method_introspector.is_defined_in_python(Manager.hasCapability)
+        assert method_introspector.is_implemented_once(Manager, "hasCapability")
```
foundrytom, line 268: [question] Were we going to keep this pattern going?

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -101,6 +101,128 @@ class OPENASSETIO_CORE_EXPORT Manager final {
...
+     *
+     * @note If this capability is true, then the host must reuse the
+     * same context across related API calls (including the use of @ref
+     * persistenceTokenForContext when the calls are distributed
+     * cross-process).
+     *
+     * @see @ref stable_resolution
+     */
+    kStatefulContexts =
+        static_cast<int>(managerApi::ManagerInterface::Capability::kStatefulContexts),
```
feltech, line 136: [question] It still bothers me that we got rid of the forward declaration of `ManagerInterface` in order to support this conforming of enum values.

Alternatives:

* The pattern established for `Access`, where independent `enum class`es  have a common `enum` dependency.  But that requires additional boilerplate (i.e. an extra public header).
* IIRC the key reason to conform the values was for indexing into `kCapabilityNames`, so another solution would be to duplicate and customise that. Not very DRY, runs the risk of accidentally indexing into the wrong `kCapabilityNames` , and makes the forwarding of enum value  from `Manager` to `ManagerInterface` more complex.

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -240,6 +240,54 @@ def test_when_entity_ref_prefix_type_invalid_then_debug_log_printed(
...
+
+    @pytest.mark.parametrize(
+        "manager_capability,managerinterface_capability", manager_capabilities
+    )
+    @pytest.mark.parametrize("return_value", (True, False))
+    def test_wraps_the_corresponding_method_of_the_held_interface(
```
foundrytom, line 274: [question] Did we need to check this has converted the enum properly?

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -240,6 +240,54 @@ def test_when_entity_ref_prefix_type_invalid_then_debug_log_printed(
...
+]
+
+
+class Test_Manager_Capability:
+    @pytest.mark.parametrize(
+        "manager_capability,managerinterface_capability", manager_capabilities
+    )
+    def test_values_match_managerinterface(self, manager_capability, managerinterface_capability):
+        assert manager_capability.value == managerinterface_capability.value
+
```
foundrytom, line 263: [fix] Needs a num values test

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
```diff
@@ -97,6 +97,10 @@ struct PyManagerInterface : ManagerInterface {
                            hostSession, successCallback, errorCallback);
   }
 
+  [[nodiscard]] bool hasCapability(ManagerInterface::Capability managerCapability) override {
+    PYBIND11_OVERRIDE_PURE(bool, ManagerInterface, hasCapability, managerCapability);
```
foundrytom, line 101: [fix] SAM needs an implementation

foundrytom (changes requested):
Noice, mostly minor things.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -66,7 +66,7 @@ OPENASSETIO_DECLARE_PTR(ManagerInterface)
  *   hosts/batch process etc...
  *
  *   @li You generally don't need to call the superclass implementation
- *   of any methods in this interface, unless you are deriving from your
+ *   of any methods in this interface, unless you are deriving from
```
feltech, line 69: [fix] Accidental deletion?

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -216,6 +217,171 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+     *
+     * This capability means the manager implements the following
+     * methods:
+     * - @ref entityExists
+     */
+    kExistenceQueries,
+
+    /**
+     * Manager may be capable of a providing an @ref EntityReference
+     * considered to be a sensible default for each of the given entity
```
feltech, line 341: [fix] As in other comment, "for each of the given" is misleading since the capability API doesn't take traits as arguments.

feltech (changes requested):
Minor niggles. Still unsure about coupling `Manager` and `ManagerInterface`...

RELEASE_NOTES.md
```diff
@@ -15,6 +15,11 @@ v1.0.0-alpha.xx
   of the exception.
   [#1073](https://github.com/OpenAssetIO/OpenAssetIO/issues/1073)
 
+- Add `hasCapability` method. A mechanism for the host to query the
```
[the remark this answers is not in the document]
feltech: Is `ManagerInterface` implied?  In other cases we've used `Manager`/`ManagerInterface.method` or `ManagerInterface`/`Manager.method`. I'm happy either way (would be good to establish this for future reference).

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: Ah yes good point, not necessary for non-migrated functions, we should remove this.

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: That's handled by the `hasCapability(manager_capability)` vs `assert_called_once_with(managerinterface_capability)`, I think.

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: Do you mean to check the numeric values of all the enums, like we do for `ManagerInterface.Capability`?  

I'm not sure that's necessary, since the only guarantee (currently) is that the numeric values match the `ManagerInterface.Capability` values, which are then tested independently...?

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Snap: https://github.com/OpenAssetIO/OpenAssetIO/pull/1126#discussion_r1342490005

src/openassetio-python/tests/package/hostApi/test_manager.py
foundrytom: Sorry, just to check how many there are, we do [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/1126/files#diff-05562a713140e71a048205b438edf81769892ade3d32b2d004c652281fdc6395R77)?

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
foundrytom: For me, the main reason was for the avoidance of confusion ie making it very clear that `Manager::kPublishing` _is and always must be_ the same as `ManagerInterface::kPublishing`. We couldn't think of any logical reason to diverge the manager constants from interface constants. The only real scenario was the one we have where the Manager ones are a sub-set of Interface ones. 

We did talk about the potential need to add new ones to `Manager` as a "collection of capabilities", but then realized that this would most likely be done via some other mechanism to a) include `managementPolicy` and b) ensure the capabilities in the set were discoverable (which they wouldn't be if we added them here).

 I'm not to worried about how it is right now as to me, the `ManagerInterface` enum _is_ the base enum, and unlike access, there is only _one_ case these are queried (access has several logically different operations that share similar concepts). 

Is the removal of the forward declare a deal breaker?

foundrytom: Sorry, forgot - if we factor it into a base enum (and this is probably true for access), it its potentially more confusing than the access one for consumers, as we don't have a lot of scope to make the names more meaningful for the "enums you use"? And we don't have a `private` public header space yet do we?

feltech: Thinking more generally: we could perhaps do with a consistent policy re. enums, especially where shared between `Manager` and `ManagerInterface`.  E.g. I'm wondering if `Access` enums should be duplicated by `Manager` and `ManagerInterface`, allowing us to doc them independently, and decouple them so they can to vary independently (though we don't have a use case for that, it seems good practice).

Perhaps having a shared header with a root enum, which `Manager` and `ManagerInterface` then duplicate a subset from, would be a good consistent universal pattern?

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: Ah yes, good idea

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
feltech: I wonder if we should remove SAM. We have the template manager, and BAL. SAM doesn't even implement `isEntityReferenceString` so is already a non-conforming manager (though at least we can indicate that via `hasCapability`).

Though, since SAM does set `kInfoKey_EntityReferencesMatchPrefix`, I think it will still work, despite not implementing `isEntityReferenceString`, since that'll be short-circuited anyway. So should its `kEntityReferenceIdentification` capability be `true`?

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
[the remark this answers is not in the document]
feltech: Reverted in 410a6d49

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -101,6 +101,128 @@ class OPENASSETIO_CORE_EXPORT Manager final {
...
+     * - @ref updateTerminology
+     */
+    kCustomTerminology =
+        static_cast<int>(managerApi::ManagerInterface::Capability::kCustomTerminology),
+    /**
+     * Manager is capable of resolving @ref entity_reference into
+     * the data for one or more @ref trait "traits."
+     *
+     * This capability means the
+     * manager implements the following methods:
```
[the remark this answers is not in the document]
feltech: Fixed in f6d175e5 and 5db68a22

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Updated as suggested in b6420aa6

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -101,6 +101,128 @@ class OPENASSETIO_CORE_EXPORT Manager final {
...
+   * groupings of functionality. For example, @ref
+   * Capability.kPublishing "publishing" or @ref Capability.kResolution
+   * "resolution".
+   *
+   * Support for each of these capabilities is optional for the manger,
+   * and the default implementation will throw
+   * @fqref{errors.NotImplementedException} "NotImplementedException"
+   *
+   * Call this method ahead of time to determine whether a manager
+   * supports a given capability.
```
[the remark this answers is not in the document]
feltech: Updated as suggested in fd4396c0

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Updated as suggested in 1989ca38

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -216,6 +217,171 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+   * that you have implemented.
+   *
+   * For information on what methods belong to which capability set,
+   * @see @ref Capability.
+   *
+   * @param managerCapability The capability to check.
+   *
+   * @return Whether the manager has the capability in question.
+   */
+  [[nodiscard]] virtual bool hasCapability(Capability managerCapability) = 0;
```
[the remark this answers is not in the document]
feltech: Renamed in 85c933d7

src/openassetio-python/tests/package/hostApi/test_manager.py
feltech: Added in c6ed7dbf

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
feltech: Added method in f8961807, returning `False` for everything except `kManagementPolicyQueries`, i.e. assuming that `kInfoKey_EntityReferencesMatchPrefix` is not sufficient for `kEntityReferenceIdentification` capability.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Reworded in fe470196

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: [Capturing some of offline conversation] It turns out we don't (yet) have a general problem of consistency between enums, there really is only `Access` and `Capability` that are manager-specific, and they are both quite different use-cases.  So I'm less concerned with consistency now.

I'm still wary of losing the forward-declaration of `ManagerInterface.hpp`, but I do wonder if avoiding it this time is just postponing the inevitable...

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
foundrytom: Yeah - you should def implement the method too I think 👍

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
foundrytom: lovely 👍

RELEASE_NOTES.md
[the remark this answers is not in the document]
foundrytom: Updated for consistency [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/1126/commits/b35fcc905ec9bd49a091fbb857083160df431629#diff-68d24fa81558aae3d8c59e2aa57a4fa719ea3b04d7fa14beff45f16f00858f50)

foundrytom (approved):

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: After more offline discussion, added c6ed7dbf, which creates a new `internal` namespace with both the `Access` and `Capability` root enum values.  How is that?

src/openassetio-core/include/openassetio/internal.hpp
```diff
@@ -0,0 +1,45 @@
...
+namespace capability {
+/**
+ * Common constant values for strong enumerations of capability sets.
+ *
+ * We must ensure correspondence between the values of the middleware
+ * and manager implementation's strong enums. This allows for
+ * comparison, string lookup, simplifies serialisation, and supports
+ * language (especially C) bindings.
+ */
+enum Capability : std::size_t {
```
foundrytom, line 31: [fix] Needs to be 'manager' qualified somehow - as we'll pretty certainly have 'host' capabilities at some point.

feltech: OK good point, put in an addtional `manager` sub-namespace in 25bfd4c1.  

Can't simply change the name of the `enum`, since the values are exported to the namespace, so could potentially conflict with a future host enum.  Could switch to `enum class`, but that would mean adding lots more `static_cast`s everywhere.

foundrytom: cool of the beans 👍 nice one.

foundrytom (approved):

feltech (approved):

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -112,6 +114,114 @@ class OPENASSETIO_CORE_EXPORT Manager final {
...
+     * Manager is capable of confirming the existence of entities.
+     *
+     * This capability means the manager implements the following
+     * methods:
+     * - @ref entityExists
+     */
+    kExistenceQueries = internal::capability::manager::Capability::kExistenceQueries,
+
+    /**
+     * Manager may be capable of a providing an @ref EntityReference
```
elliotcmorris, line 188: [Consider] Just "may" is too ambiguous here. The host needs advice on what that means specifically. I think the note here from before achieved that.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -229,6 +231,163 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+   */
+  enum class Capability : std::underlying_type_t<internal::capability::manager::Capability> {
+    /**
+     * Manager can inform the host whether a given string matches the
+     * pattern of a valid entity reference.
+     *
+     * Support for this capability is required by all managers. In
+     * situations where plugins are implemented as multiple component
+     * plugins (e.g. Python and C++) at least one of the component
+     * plugins must implement this capability.
```
elliotcmorris, line 252: [Consider][Minor] Perhaps this is so significant it should be in a warning block.

elliotcmorris (commented):
(This isn't a non-approval, just PR authors can't approve their own work.)

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Ah I think I found the text you're referring to

```
     * @note The fact that this capability is implemented does not
     * necessarily mean that a default entity reference can be provided
     * for every traitSet, you may get still get an empty value.
```

IIRC we removed this because it's covered by the method docstring itself, not to mention the method signature.

Perhaps something like "Manager is capable of responding appropriately to queries for an @ref EntityReference considered to be a sensible default"?

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Good shout, added in 68a48346

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
elliotcmorris: Hmm, I guess that makes sense, although i'd personally favour over-specifying, even if it means repeating ourselves, especially as this type is becoming a sort of reference index.

Nonetheless though, the reasoning is valid, so resolving.

Issue #1139, opened 2023-10-04
Remove `method_defined_in` tests


elliotcmorris:
## What

Remove tests that assert methods are implemented in C++ rather than Python

## Why

These were added whilst we migrated functionality from pure Python to C++ and are no longer necessary.

_Originally posted by @feltech in https://github.com/OpenAssetIO/OpenAssetIO/pull/1138#discussion_r1345517442_

PR #1141, opened 2023-10-04
Finalize abstract methods in ManagerInterface

### LINKED ISSUE #163: Finalize abstract methods in `ManagerInterface`
## What

Update `ManagerInterface` methods to provide default implementations (as per #1101) or mark as pure virtual.

## Why

The current implementation is inconsistent with "valid behaviour"

## Acceptance Criteria

- The following methods should be pure virtual:
  - `identifier`
  - `displayName`
  - `hasCapability`
- The following methods should have no-op/return empty default implementations:
  - `initialize` (throw if any settings provided)
  - `info`
  - `settings`
  - `flushCaches`
- The following methods should throw a `NotImplementedException` (with a message instructing how to call `hasCapability` to avoid the exception and which enum to use):
  - `managementPolicy`
  - `isEntityReferenceString`,
  - `entityExists` (new `kExistenceQueries` capability)
  - `updateTerminology`
  - `resolve`
  - `preflight`
  - `register`
  - `getWithRelationship*`
  - `defaultEntityReference`
  - `createState`, `createChildState` 
  - `persistenceTokenForState`, `stateFromPersistenceToken`
- Wrappers/conveniences updated
   - `createContext` must check the `kStatefullContexts` and not throw if `createState` isn't implemented.
   - update host docs
   - Update Manager method docs to denote  optional + throws and see `hasCapabilty`
   - Update ManagerInterface method docs to denote optional and that default implementation doesn't not need to be called.
   - Check for other middleware methods that may potentially call into default method implementations that shouldn't throw.

feltech:
## Description

Closes #163

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -72,6 +72,12 @@ v1.0.0-alpha.xx
   exceptional conveniences.
   [#1073](https://github.com/OpenAssetIO/OpenAssetIO/issues/1073)
 
+- Most non-required methods (all that must be checked with
+  `hasCapability`), now have default implementations that throw
+  `NotImplementedException` when called without a manager provided
+  implementation.
+  [#163](https://github.com/OpenAssetIO/OpenAssetIO/issues/163)
+
 ### New Features
 
 - Added new exception types, allowing all OpenAssetIO exceptions to
@@ -95,6 +101,12 @@ v1.0.0-alpha.xx
   variable `OPENASSETIO_ENABLE_PYTHON_INSTALL_DIST_INFO` to disable this
   feature.
   [#1088](https://github.com/OpenAssetIO/OpenAssetIO/issues/1088)
+
+- `Manager.createContext` does not call `ManagerInterface.createState`
+  if manager is not capable of `kStatefulContexts`, in which case the
+  `Context` is returned without a `managerState`.
+  [#163](https://github.com/OpenAssetIO/OpenAssetIO/issues/163)
+
 
 v1.0.0-alpha.14
 ---------------
--- a/doc/doxygen/src/Examples.dox
+++ b/doc/doxygen/src/Examples.dox
@@ -163,6 +163,11 @@
  * # for the entity. We inform the manager that we just want to read the
  * # current data for the entity (kRead). See the publishing examples
  * # for use of the alternative `kWrite` access mode.
+ *
+ * # First check that the manager has implemented resolution capability.
+ * if not manager.hasCapability(Manager.Capability.kResolution):
+ *   return
+
  * resolved_asset = manager.resolve(
  *         entity_reference, {LocatableContentTrait.kId},
  *         ResolveAccess.kRead, context)
@@ -175,23 +180,51 @@
  * an entity. In a real-world scenario, it is important to remember
  * that:
  *
+ * - Managers may not support a particular workflow at all.
  * - Managers may not handle certain types of entity (determined by
  *   their @ref trait_set).
  * - Managers may not be able to provide data for all traits.
  *
+ * @subsection Manager Capabilities
+ *
+ * The @fqref{hostApi.Manager.hasCapability} "hasCapability" method
+ * allows a host to determine which workflows that a manager has
+ * implemented.
+ *
+ * This is achieved by grouping sets of methods potentially implemented
+ * by the manager into "capabilities". These capabilities are defined
+ * in @fqref{hostApi.Manager.Capability} "Capability".
+ *
+ * Calling a method not implemented by the manager will result in an
+ * exception. Therefore hosts should check
+ * @fqref{hostApi.Manager.hasCapability} "hasCapability" before calling
+ * into any of these optional methods. The return value of
+ * @fqref{hostApi.Manager.hasCapability} "hasCapability"  is runtime
+ * invariant, post-@fqref{hostApi.Manager.initialize} "initialize",
+ * therefore the check only needs to be made once per manager.
+ *
+ * @subsection Management Policy
+ *
  * The @fqref{hostApi.Manager.managementPolicy} "managementPolicy"
- * method allows a host to query a manager's capabilities and its
- * intentions towards different types of entity.
+ * method allows a host to query a manager's behaviours and intentions
+ * towards different types of entity.
  *
  * This method should be used wherever possible to adapt the host's
  * behaviour so that:
  *
  * - The manager is not invoked in relation to unsupported entity types.
- * - User workflows are appropriate to the supported capabilities.
+ * - User workflows are appropriate to the supported behaviour for a
+ *   given entity type.
+ *
+ * Failing to check the policy leaves a host vulnerable to recieving
+ * empty, null or invalid data by making queries outside of the managers
+ * realm of understanding.
+ *
+ * @subsection Example
  *
  * This example demonstrates how to query a manager in relation to a
- * specific entity type - in this case a simple text file, and
- * inspect its capabilities.
+ * specific entity type - in this case a simple text file, and inspect
+ * its capabilities and the data it may be able to resolve for it.
  *
  * @code{.py}
  *
@@ -207,6 +240,10 @@
  * from openassetio_mediacreation.specifications.files import (
  *     # TextFileSpecification
  * )
+ *
+ * # Ensure that the manager is capable of the resolution workflow.
+ * if not manager.hasCapability(Manager.Capability.kResolution):
+ *   return
  *
  * # We use the well-known specification for a text file to determine
  * # the correct trait set to query. Using the standard definition
@@ -235,7 +272,6 @@
  *
  * if TextEncodingTrait.isImbuedTo(policy):
  *   print("The manager can provide the text encoding used")
- *
  * @endcode
  *
  *
@@ -255,6 +291,9 @@
  *     # TextFileSpecification
  * )
  *
+ * # Ensure that the manager supports publishing.
+ * if not manager.hasCapability(Manager.Capability.kPublishing):
+ *   return
  *
  * # As ever, an appropriately configured context is required
  * context = manager.createContext()
@@ -291,6 +330,11 @@
  *
  * # We then check if the manager can tell us where to save the file.
  * if ResolvesFutureEntitiesTrait.isImbuedTo(policy):
+ *
+ *     # Note we haven't used  `hasCapability` to check the resolution
+ *     # capability, as `ResolvedFutureEntitiesTrait` being imbued
+ *     # assures us the functionality is available.
+ *
  *     working_data = manager.resolve(
  *             working_ref, TextFileSpecification.kTraitSet,
  *             ResolveAccess.kWrite, context)
@@ -331,6 +375,10 @@
  * specifications that are yet to be finalized.
  *
  * @code{.py}
+ *
+ * # Ensure that the manager supports publishing.
+ * if not manager.hasCapability(Manager.Capability.kPublishing):
+ *   return
  *
  * # See if the manager wants a thumbnail
  * if not WantsThumbnailTrait.isImbuedTo(policy):
--- a/doc/doxygen/src/ForHosts.dox
+++ b/doc/doxygen/src/ForHosts.dox
@@ -98,6 +98,11 @@
  *   user with an additional manual configuration mechanism for your
  *   specific API session if you so wish.
  *
+ * - Ask the manager if it supports the resolution workflow by calling
+ *   @fqref{hostApi.Manager.hasCapability} "Manager.hasCapability",
+ *   passing @fqref{hostApi.Manager.Capability.kResolution}
+ *   "kResolution".
+ *
  * - Create (and configure) a @ref Context as appropriate for calls to
  *   the API. Re-use context objects for all API calls that are related
  *   to the same user session or logical set of interactions. For
@@ -130,6 +135,11 @@
  * @see @ref examples_resolving_a_reference
  *
  * @subsection host_todo_required_publishing Required for Publishing
+ *
+ * - Ask the manager if it supports the publishing workflow by calling
+ *   @fqref{hostApi.Manager.hasCapability} "Manager.hasCapability",
+ *   passing @fqref{hostApi.Manager.Capability.kPublishing}
+ *   "kPublishing".
  *
  * - Ask the @ref manager if they support publishing for the relevant
  *   data by supplying its @ref trait_set to
--- a/doc/doxygen/src/ForManagers.dox
+++ b/doc/doxygen/src/ForManagers.dox
@@ -106,8 +106,10 @@
  * - Implement the @fqref{managerApi.ManagerInterface}
  *   "ManagerInterface" class methods
  *   @fqref{managerApi.ManagerInterface.identifier} "identifier",
- *   @fqref{managerApi.ManagerInterface.displayName} "displayName" and
- *   @fqref{managerApi.ManagerInterface.hasCapability} "hasCapability".
+ *   @fqref{managerApi.ManagerInterface.displayName} "displayName",
+ *   @fqref{managerApi.ManagerInterface.hasCapability} "hasCapability"
+ *   and @fqref{managerApi.ManagerInterface.isEntityReferenceString}
+ *   "isEntityReferenceString".
  *
  * - Implement @fqref{managerApi.ManagerInterface.managementPolicy}
  *   "managementPolicy" to control which Host-side data types you wish
@@ -130,14 +132,19 @@
  * @subsection manager_todo_publishing Required for Publishing
  *
  * - Update the implementation of
+ *   @fqref{managerApi.ManagerInterface.hasCapability} "hasCapability"
+ *   to return true when queried with
+ *   @fqref{managerApi.ManagerInterface.Capability.kPublishing}
+ *   "kPublishing".
+ *
+ * - Update the implementation of
  *   @fqref{managerApi.ManagerInterface.managementPolicy}
  *   "managementPolicy" to return suitable traits for contexts with a
  *   `kWrite` access.
  *
- * - Implement the methods grouped under the "Publishing" section, and
- *   the remaining methods in the "Entity Information" section of
- *   @fqref{managerApi.ManagerInterface} "ManagerInterface", to support
- *   the registration of new assets.
+ * - Implement the methods listed under the
+ *   @fqref{managerApi.ManagerInterface.Capability.kPublishing}
+ *   "publishing" capability to support the registration of new assets.
  *
  * - Map supported OpenAssetIO @ref trait "traits" to internal asset
  *   types where possible.
@@ -204,6 +211,11 @@
  *   simplify common pipeline integration tasks. For example, loading
  *   multiple AOVs for a render, or determining data dependencies when
  *   transferring assets.
+ * - Update your implementation of
+ *   @fqref{managerApi.ManagerInterface.hasCapability} "hasCapability"
+ *   to report that you are now implement
+ *   @fqref{managerApi.ManagerInterface.Capability.kRelationshipQueries}
+ *   "relationship queries".
  * - Update your implementation of @fqref{managerApi.ManagerInterface.managementPolicy}
  *   "managementPolicy" to cover the relationship trait sets you support.
  *
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -260,6 +260,13 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * @param terms Map of terms to be substituted by the manager.
    *
    * @return Substituted map of terms.
+   *
+   * @throws errors.NotImplementedException Thrown when this method is
+   * not implemented by the manager. Check that this method is
+   * implemented before use by calling @ref hasCapability with @ref
+   * Capability.kCustomTerminology.
+   *
+   * @see @ref Capability.kCustomTerminology
    */
   StrMap updateTerminology(StrMap terms);
 
@@ -498,7 +505,13 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    *  supplied here. It does not encode the current locale or other
    *  propeties.
    *
-   *  @see @ref stable_resolution
+   * @throws errors.NotImplementedException Thrown when this method is
+   * not implemented by the manager. Check that this method is
+   * implemented before use by calling @ref hasCapability with @ref
+   * Capability.kStatefulContexts.
+   *
+   * @see @ref Capability.kStatefulContexts
+   * @see @ref stable_resolution
    */
   Str persistenceTokenForContext(const ContextPtr& context);
 
@@ -520,6 +533,13 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * restored by this action.
    *
    * @see @ref stable_resolution
+   *
+   * @throws errors.NotImplementedException Thrown when this method is
+   * not implemented by the manager. Check that this method is
+   * implemented before use by calling @ref hasCapability with @ref
+   * Capability.kStatefulContexts.
+   *
+   * @see @ref Capability.kStatefulContexts
    *
    * @todo Should we concatenate the manager id in
    * persistenceTokenForContext so we can verify that they match?
@@ -723,6 +743,13 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * "BatchElementError" (see @fqref{errors.BatchElementError.ErrorCode}
    * "ErrorCodes"). The callback will be called on the same thread
    * that initiated the call to `entityExists`.
+   *
+   * @throws errors.NotImplementedException Thrown when this method is
+   * not implemented by the manager. Check that this method is
+   * implemented before use by calling @ref hasCapability with @ref
+   * Capability.kExistenceQueries.
+   *
+   * @see @ref Capability.kExistenceQueries
    */
   void entityExists(const EntityReferences& entityReferences, const ContextConstPtr& context,
                     const ExistsSuccessCallback& successCallback,
@@ -819,6 +846,13 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * "BatchElementError" (see @fqref{errors.BatchElementError.ErrorCode}
    * "ErrorCodes"). The callback will be called on the same thread
    * that initiated the call to `resolve`.
+   *
+   * @throws errors.NotImplementedException Thrown when this method is
+   * not implemented by the manager. Check that this method is
+   * implemented before use by calling @ref hasCapability with @ref
+   * Capability.kResolution.
+   *
+   * @see @ref Capability.kResolution
    */
   void resolve(const EntityReferences& entityReferences, const trait::TraitSet& traitSet,
                access::ResolveAccess resolveAccess, const ContextConstPtr& context,
@@ -857,6 +891,13 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * BatchElementErrorPolicyTag::Exception.
    *
    * @return Populated data.
+   *
+   * @throws errors.NotImplementedException Thrown when this method is
+   * not implemented by the manager. Check that this method is
+   * implemented before use by calling @ref hasCapability with @ref
+   * Capability.kResolution.
+   *
+   * @see @ref Capability.kResolution
    */
   trait::TraitsDataPtr resolve(const EntityReference& entityReference,
                                const trait::TraitSet& traitSet,
@@ -900,6 +941,13 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    *
    * @return Object containing either the populated data or an error
    * object.
+   *
+   * @throws errors.NotImplementedException Thrown when this method is
+   * not implemented by the manager. Check that this method is
+   * implemented before use by calling @ref hasCapability with @ref
+   * Capability.kResolution.
+   *
+   * @see @ref Capability.kResolution
    */
   std::variant<errors::BatchElementError, trait::TraitsDataPtr> resolve(
       const EntityReference& entityReference, const trait::TraitSet& traitSet,
@@ -939,6 +987,13 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * BatchElementErrorPolicyTag::Exception.
    *
    * @return List of populated data objects.
+   *
+   * @throws errors.NotImplementedException Thrown when this method is
+   * not implemented by the manager. Check that this method is
+   * implemented before use by calling @ref hasCapability with @ref
+   * Capability.kResolution.
+   *
+   * @see @ref Capability.kResolution
    */
   std::vector<trait::TraitsDataPtr> resolve(
       const EntityReferences& entityReferences, const trait::TraitSet& traitSet,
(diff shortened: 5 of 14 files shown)


feltech (commented):
A few docs things missed

RELEASE_NOTES.md
```diff
@@ -96,6 +102,12 @@ v1.0.0-alpha.xx
   feature.
   [#1088](https://github.com/OpenAssetIO/OpenAssetIO/issues/1088)
 
+- `createContext` does not call `createState` if manager is not capable
+  of `kStatefulContexts`. Context is returned without a set
+  managerState.
```
feltech, line 107: [consider][minor]
```suggestion
  of `kStatefulContexts`, the `Context` is returned without a set
  `managerState`.
```

or perhaps

```suggestion
  of `kStatefulContexts`, the context is returned without any
  manager state.
```

doc/doxygen/src/ForHosts.dox
```diff
@@ -131,6 +136,14 @@
  *
  * @subsection host_todo_required_publishing Required for Publishing
  *
+ * - Ask the manager if it supports the publishing workflow by calling
+ *   @fqref{hostApi.Manager.hasCapability} "Manager.hasCapability",
+ *   passing @fqref{hostApi.Manager.Capability.kPublishing}
+ *   "kPublishing", along with any other capabilities required during
+ *   the publishing procedure (e.g. you may need to
+ *   @fqref{hostApi.Manager.Capability.kResolution} "resolve"
+ *   information during publishing.
```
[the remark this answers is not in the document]
feltech: Fixed in 066111ad

doc/doxygen/src/ForManagers.dox
```diff
@@ -106,8 +106,10 @@
  * - Implement the @fqref{managerApi.ManagerInterface}
  *   "ManagerInterface" class methods
  *   @fqref{managerApi.ManagerInterface.identifier} "identifier",
- *   @fqref{managerApi.ManagerInterface.displayName} "displayName" and
- *   @fqref{managerApi.ManagerInterface.hasCapability} "hasCapability".
+ *   @fqref{managerApi.ManagerInterface.displayName} "displayName",
+ *   @fqref{managerApi.ManagerInterface.hasCapability} "hasCapability",
```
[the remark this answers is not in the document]
feltech: Fixed in 066111ad

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -497,7 +502,12 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    *  resulting Context will be logically associated with the one
    *  supplied here. It does not encode the current access or locale.
    *
-   *  @see @ref stable_resolution
+   *
```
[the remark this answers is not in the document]
feltech: Fixed in 066111ad

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -520,6 +530,12 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    *
    * @see @ref stable_resolution
    *
+   *
```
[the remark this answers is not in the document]
feltech: Fixed in 066111ad

RELEASE_NOTES.md
feltech: Updated in 066111ad, how is that?

doc/doxygen/src/Examples.dox
```diff
@@ -255,6 +287,10 @@
  *     # TextFileSpecification
  * )
  *
+ * # Ensure that the manager is capable of the required workflows.
+ * if not (manager.hasCapability(Manager.Capability.kResolution) and
```
foundrytom, line 291: [Fix] The resolve check needs to move to the conditional with `ResolvesFutureEntitiesTrait.isImbuedTo`, as we only resolve in that case. You could argue that check is redundant as a manager shouldn't imbue that trait if it doesnt implement resolve, so that check could even be skipped.

doc/doxygen/src/Examples.dox
```diff
@@ -255,6 +287,10 @@
...
+ *         manager.hasCapability(Manager.Capability.kPublishing)):
+ *   return
```
foundrytom, line 293: [Fix] The thumbnails sketch needs updating similarly.

doc/doxygen/src/ForHosts.dox
foundrytom: [Fix] Resolving during publishing is predicated on the management policy imbuing the `ResolvesFutureEntitiesTrait`, so unless you absolutely require write resolves then you don't want to disable all publishing without it. As noted above too, that check is technically redundant too.

doc/doxygen/src/ForManagers.dox
```diff
@@ -106,8 +106,10 @@
  * - Implement the @fqref{managerApi.ManagerInterface}
  *   "ManagerInterface" class methods
  *   @fqref{managerApi.ManagerInterface.identifier} "identifier",
- *   @fqref{managerApi.ManagerInterface.displayName} "displayName" and
- *   @fqref{managerApi.ManagerInterface.hasCapability} "hasCapability".
+ *   @fqref{managerApi.ManagerInterface.displayName} "displayName",
```
foundrytom, line 109: [Fix] We should add the same additions to this doc as the hosts one, eg in "Required for publishing" - "update the implementation of `managementPolicy` _and `hasCapability`_...",

RELEASE_NOTES.md
```diff
@@ -96,6 +102,12 @@ v1.0.0-alpha.xx
   feature.
   [#1088](https://github.com/OpenAssetIO/OpenAssetIO/issues/1088)
 
+- `createContext` does not call `createState` if manager is not capable
```
foundrytom, line 105: [Fix]
```suggestion
- `Manager.createContext` does not call `ManagerInterface.createState` if manager is not capable
```

foundrytom: [Consider] Is this more an improvement?

foundrytom (changes requested):
Great to see this milestone in! Sorry there may have been some stale ACs after the hybrid plugins chat.

doc/doxygen/src/Examples.dox
```diff
@@ -175,19 +175,43 @@
  * an entity. In a real-world scenario, it is important to remember
  * that:
  *
+ * - Managers may not support a particular workflow at all.
  * - Managers may not handle certain types of entity (determined by
  *   their @ref trait_set).
  * - Managers may not be able to provide data for all traits.
  *
+ * @subsection Manager Capabilities
```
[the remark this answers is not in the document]
elliotcmorris: The prior example omits management policy checks as well, leading into the section about a real-world scenario, where managementPolicy (and now capability) is introduced. Seems to be the intent is not to muddy the waters in the first example. 

I agree that folks will copy/paste, but do you think we need to add management policy as well to it then? Personally I think the simple->then expand flow is better here.

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
```diff
@@ -134,16 +117,66 @@ def test_names_indices_match_constants(self):
         )
 
 
-class Test_ManagerInterface_updateTerminology:
-    def test_method_defined_in_cpp(self, method_introspector):
-        assert not method_introspector.is_defined_in_python(ManagerInterface.updateTerminology)
-        assert method_introspector.is_implemented_once(ManagerInterface, "updateTerminology")
+class Test_ManagerInterface_hasCapability:
```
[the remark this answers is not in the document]
feltech: We can just reintroduce them pending https://github.com/OpenAssetIO/OpenAssetIO/issues/1139

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -260,6 +260,11 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * @param terms Map of terms to be substituted by the manager.
    *
    * @return Substituted map of terms.
+   *
+   * @throws errors.NotImplementedException when this method is not
+   * implemented by the manager.
```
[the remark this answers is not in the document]
feltech: There is the `@see` below, and the enum links back to the method. Could we just move the `@see` to be in the same sentence, perhaps?

doc/doxygen/src/Examples.dox
```diff
@@ -236,6 +264,10 @@
  * if TextEncodingTrait.isImbuedTo(policy):
  *   print("The manager can provide the text encoding used")
  *
+ * resolved_asset = manager.resolve(
+ *         entity_reference,
+ *         {LocatableContentTrait.kId, TextEncodingTrait.kId},
+ *         ResolveAccess.kRead, context)
```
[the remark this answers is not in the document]
elliotcmorris: The reason we added this is to provide a motivation to check capability, so we can showcase both management policy + capability in the same example. Taking this out would imply adding a second example just for capability so we can showcase them seperately, you think that's the thing to do?

doc/doxygen/src/Examples.dox
```diff
@@ -175,19 +175,43 @@
...
+ * invariant, post-@fqref{hostApi.Manager.initialize} "initialize",
+ * therefore the check only needs to be made once per manager.
+ *
+ * @subsection Management Policy
+ *
  * The @fqref{hostApi.Manager.managementPolicy} "managementPolicy"
- * method allows a host to query a manager's capabilities and its
- * intentions towards different types of entity.
+ * method allows a host to query a manager's capabilities in relation
+ * to a given entity type.
```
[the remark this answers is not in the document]
feltech: The thinking was that it muddies the previous `hasCapability` doc if we say "capabilities and ...` as opposed to "capabilities in relation to ...".  I.e. `hasCapability` is the global capability, whereas `managementPolicy` is intended for querying capabilities around a particular type of entity (in a given `Context` etc).

doc/doxygen/src/Examples.dox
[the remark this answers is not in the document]
feltech: Yeah the problem we faced was choosing a capability to add to the example, then not actually using the functionality we tested for.  No big deal to remove it though.

doc/doxygen/src/Examples.dox
feltech: Yeah this echos my second point in https://github.com/OpenAssetIO/OpenAssetIO/pull/1141#discussion_r1348623413 - when do we eschew `hasCapability` because `managementPolicy` is enough?

The reasoning in this case is that the `managementPolicy` check can be avoided entirely if the capability is not even a possibility.

doc/doxygen/src/Examples.dox
feltech: The thumbnail sketch appears to follow on from the previous example, i.e. `managementPolicy` as assumed to already have been called, so we reasoned that `hasCapability` is similarly assumed to have already been called.

doc/doxygen/src/Examples.dox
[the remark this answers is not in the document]
foundrytom: Yeah - see the suggestion below - I think we should change capabilites here, as its definitely misleading.

>  whereas managementPolicy is intended for querying capabilities around a particular type of entity

I think we need to entirely avoid using the work 'capabilities' in relation to management policy, now we've made it a proper noun in relation to API methods. MP is also more than just "capability" too - it includes things like desired workflows, etc.. eg: "please make me a thumbnail", which definitely isn't the same as a capability of the manager... so if we can think of another phrase/word that we use with MP in place of capability, seems like that would sort it?

doc/doxygen/src/Examples.dox
foundrytom: Yeah, but if you do that, youre aborting publishing entirely - which isn't the intended approach, The example is meant to show you can still publish to the path you have. So we can't do the check for resolve here with an `and` with publishing. This  has to illustrate how to work with the variety of managers - some can give you the path up front, some cant.

It doesn't hurt to call `hasCapability before `ResolvesFutureEntitiesTrait.isImbuedTo(policy)` I guess, though it is redundant.  I guess it guards you against a manager messing up its implementation.

doc/doxygen/src/Examples.dox
foundrytom: I think its meant to be follow on from "some other code" rather than the one before, but as noted, the other one shouldn't be aborting publishing if resolve isn't supported.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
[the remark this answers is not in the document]
foundrytom: I think it'd be nice to make it really obvious for people, we've seen before that people don't always follow refs and links and join the dots.

doc/doxygen/src/Examples.dox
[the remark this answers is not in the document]
foundrytom: The sentence above could be tweaked a ilttle to say something like:

"This example demonstrates how to query a manager in relation to a specific entity type - in this case a simple text file, and inspect its capabilities and the data it may be able to resolve for it"

doc/doxygen/src/Examples.dox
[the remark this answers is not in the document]
foundrytom: Is "policy" itself enough?

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
```diff
@@ -134,16 +117,64 @@ def test_names_indices_match_constants(self):
...
+    def test_is_pure_virtual(
+        self, manager_interface, a_host_session, a_context, pure_virtual_error_msg
+    ):
+        with pytest.raises(RuntimeError, match=pure_virtual_error_msg.format("managementPolicy")):
+            manager_interface.managementPolicy(
+                [set()], access.PolicyAccess.kRead, a_context, a_host_session
+            )
+
+
+class Test_ManagerInterface_isEntityReferenceString:
```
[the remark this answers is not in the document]
elliotcmorris: Yep. : https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/6f3cca7b520b2043781b91166168a4d50f5ffd79

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -2433,17 +2433,40 @@ def test_method_defined_in_cpp(self, method_introspector):
...
-    def test_context_is_created_with_expected_properties(
+    def test_context_is_created_with_locale(self, manager, mock_manager_interface, a_host_session):
+        mock_manager_interface.mock.hasCapability.return_value = False
+
+        context_a = manager.createContext()
+
+        assert isinstance(context_a.locale, TraitsData)
+        assert context_a.locale.traitSet() == set()
+
+    def test_when_custom_state_not_supported_then_returned_context_contains_no_state(
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/010c3e5404b6a02102da56e96c4940222ab300b0

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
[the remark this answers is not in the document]
elliotcmorris: I've reintroduced them all so we can do them in one pass.
https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/4d10417c687c997b046ac808a94c262ea2873e58

src/openassetio-python/tests/package/managerApi/test_managerinterface.py
```diff
@@ -352,6 +377,53 @@ def test_is_entity_reference_string_is_not_called(self, mock_manager_interface):
...
+
+
+@pytest.fixture
+def pure_virtual_error_msg():
+    return 'Tried to call pure virtual function "ManagerInterface::{}"'
+
+
+@pytest.fixture
+def unimplemented_method_error_msg():
+    return "The '{}' method has not been implemented by the manager"
```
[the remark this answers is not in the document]
elliotcmorris: :+1: 

tests : https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/404376f6f6ce5370788ade967469632e6ce35685
impl : https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/9d81827087022f5e0dfada649bdcebcc6cd16bde

I've used the capability name for the message, so it reads like 

> The resolve method has not been implemented by the manager. Check manager capability for resolution by calling `manager.hasCapability`.

doc/doxygen/src/Examples.dox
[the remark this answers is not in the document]
elliotcmorris: I tend to think it's best to veer on the side of using a single word for a single concept as much as possible, and as mangementPolicy has a method called managementPolicy, giving it another alias seems like a pessimism to understanding, at least to me,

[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/a4cc7e98f781f7ba5961b9ffdf845e438f09b98b

As we're getting rid of the word "capability" for managementPolicy, reverted the phrasing to previous, intentions is a good word.
Also added a line hopefully serving to clarify the difference between ManagementPolicy and Capability, which I think serves to say why you would want to check it, but also dosen't imply that it's non-optional.

>  Failing to check the policy leaves a host vulnerable to recieving
>  empty, null or invalid data by making queries outside of the managers
>  realm of understanding.

doc/doxygen/src/Examples.dox
```diff
@@ -175,19 +175,43 @@
...
+ * @fqref{hostApi.Manager.hasCapability} "hasCapability"  is runtime
+ * invariant, post-@fqref{hostApi.Manager.initialize} "initialize",
+ * therefore the check only needs to be made once per manager.
+ *
+ * @subsection Management Policy
+ *
  * The @fqref{hostApi.Manager.managementPolicy} "managementPolicy"
- * method allows a host to query a manager's capabilities and its
- * intentions towards different types of entity.
+ * method allows a host to query a manager's capabilities in relation
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/a4cc7e98f781f7ba5961b9ffdf845e438f09b98b

doc/doxygen/src/Examples.dox
```diff
@@ -175,19 +175,43 @@
...
+ * to a given entity type.
  *
  * This method should be used wherever possible to adapt the host's
  * behaviour so that:
  *
  * - The manager is not invoked in relation to unsupported entity types.
- * - User workflows are appropriate to the supported capabilities.
+ * - User workflows are appropriate to the supported capabilities for a
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/a4cc7e98f781f7ba5961b9ffdf845e438f09b98b

doc/doxygen/src/Examples.dox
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/14487089e409808bac466d124c0cd13ecd3f3913

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -1707,6 +1753,11 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    *
    * @return Updated reference to use for future interactions with the
    * resulting new entity.
+   *
```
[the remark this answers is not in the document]
elliotcmorris: Added here, along with additional note for hint to call getCapability : https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/3596f290647e066f044c34c6294fe034b4b8a263

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -1414,6 +1455,11 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    *
    * @return Updated reference to use for future interactions as part of
    * the publishing operation
+   *
```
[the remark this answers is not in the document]
elliotcmorris: Added here, along with additional note for hint to call getCapability : https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/3596f290647e066f044c34c6294fe034b4b8a263

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -856,6 +877,11 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * BatchElementErrorPolicyTag::Exception.
    *
    * @return Populated data.
+   *
```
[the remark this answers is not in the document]
elliotcmorris: Added here, along with additional note for hint to call getCapability : https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/3596f290647e066f044c34c6294fe034b4b8a263

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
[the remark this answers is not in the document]
elliotcmorris: Added here, along with adding the note to all the completely missed methods. : https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/3596f290647e066f044c34c6294fe034b4b8a263

[Question] I wonder if there's additional doc we should add to the `Exception` tag overloads now, as they have a `@throws` from a seperate mechanism, it may be less obvious that they do exception conversion from BatchElementErrors(There is a `@see` link that tells you, but it's not in the exceptions section.) We could maybe do `@throw Converted BatchElementException See @ref BatchElementErrorPolicyTag::Exception.` so it's clearer when just looking at the exceptions section of the function documentation.

doc/doxygen/src/Examples.dox
elliotcmorris: What do you think of? https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/1dc973e36afe14f4900fda1c04252db9c8354273

We still show capability being a less fine-comb up front precondition for `publishing`, but leave a note saying "hey, see how this trait means you don't need to do a redundant check."

doc/doxygen/src/Examples.dox
elliotcmorris: What do you think of? https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/1dc973e36afe14f4900fda1c04252db9c8354273

We still show capability being a less fine-comb up front precondition for `publishing`, but leave a note saying "hey, see how this trait means you don't need to do a redundant check."

doc/doxygen/src/Examples.dox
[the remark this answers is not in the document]
elliotcmorris: Aye https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/db699306fe3528726c8da01f8a23385fbfeecfaf

doc/doxygen/src/Examples.dox
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/6b6976c97d5f7ddc45d17ddac628ffa696e5d3b3

Have just added the publishing check, see comment on the above example. If we keep the "Hey, note how we don't use capabilities" example there, I don't think we need to keep repeating ourselves and put it here as well.

doc/doxygen/src/ForHosts.dox
elliotcmorris: Yeah, this makes things a fair bit simpler, just check for publishing up front when doing publishing. https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/8798160fb07a0ee76a72d751d66ba177c9f6abec

doc/doxygen/src/ForManagers.dox
elliotcmorris: Yeah. https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/515e1d9ebebcd4826bcde2307b1c40c968b1accd

Changed a note below about what methods to implement. We've changed the categorisation since this has been written, and I think with the required methods we can just say to implement the publishing capability, which is ideal imo.

RELEASE_NOTES.md
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/205e205d8dff12b385c4a18c5c5082ad38facc68

It's in the improvement section, I think github may have collapsed the diff inbetween the headings.

src/openassetio-python/tests/package/hostApi/test_manager.py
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/010c3e5404b6a02102da56e96c4940222ab300b0

doc/doxygen/src/Examples.dox
foundrytom: Works for me 👍

doc/doxygen/src/Examples.dox
foundrytom: @feltech You happy with omitting here?

RELEASE_NOTES.md
foundrytom: Sorry my mistake!

doc/doxygen/src/Examples.dox
```diff
@@ -255,6 +291,9 @@
  *     # TextFileSpecification
  * )
  *
+ * # Ensure that the manager can do publishing.
```
foundrytom, line 294: [consider]
```suggestion
 * # Ensure that the manager supports publishing.
```

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -589,11 +595,20 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
    * @return a `TraitsData` for each element in @p traitSets.
+   *
+   * @throws errors.NotImplementedException by default when this method
+   * is not implemented by the manager. Implementations must therefore
+   * not invoke the base class implementation.
+   * This exception being thrown constitutes a manager implementation
+   * error, as all managers must implement this method in one of their
+   * component plugins.
+   *
+   * @see @ref Capability.kExistenceQueries
```
feltech, line 606: [fix] Incorrect capability

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -807,11 +843,19 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
    * by the manager.
    *
+   * @throws errors.NotImplementedException by default when this method
+   * is not implemented by the manager. Implementations must therefore
+   * not invoke the base class implementation.
+   * This exception being thrown constitutes a manager implementation
+   * error, as all managers must implement this method in one of their
+   * component plugins.
+   *
+   * @see @ref Capability.kExistenceQueries
```
feltech, line 853: [fix] Incorrect capability

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -260,6 +260,13 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * @param terms Map of terms to be substituted by the manager.
    *
    * @return Substituted map of terms.
+   *
+   * @throws errors.NotImplementedException Thrown when this method is not
```
feltech, line 264: [consider][minor] Here and other copy-pasted text - line wrapping.

src/openassetio-core/src/managerApi/ManagerInterface.cpp
```diff
@@ -14,11 +16,30 @@ namespace managerApi {
...
+        "by the manager."};
+  }
+}
+
+// NOLINTNEXTLINE(performance-unnecessary-value-param)
+StrMap ManagerInterface::updateTerminology([[maybe_unused]] StrMap terms,
                                            [[maybe_unused]] const HostSessionPtr& hostSession) {
-  return terms;
+  throw errors::NotImplementedException{
+      UNIMPLEMENTED_ERROR(ManagerInterface::kCapabilityNames[static_cast<int64_t>(
```
feltech, line 41: [consider] Put the `static_cast` and name lookup inside the macro.

feltech (commented):
Just minor niggles

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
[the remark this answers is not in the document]
foundrytom: Feel free to add anything that clarifies 👍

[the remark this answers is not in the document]
foundrytom: Resolving for the original omission.

foundrytom (approved):
Thanks @elliotcmorris for all those, LGTM aside from the points highlighted by @feltech

doc/doxygen/src/Examples.dox
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/4ff4959ddd51fbd485172b3b759415855417bda6

src/openassetio-core/src/managerApi/ManagerInterface.cpp
elliotcmorris: Aye good shout https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/59c07d2ae8ea8fe9dc8a526ef807407eb106e47c

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
elliotcmorris: Dealt with here : https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/b13bc4c4e04c619dd07d4e539652a99ee5b20a0d

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
elliotcmorris: Dealt with here : https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/b13bc4c4e04c619dd07d4e539652a99ee5b20a0d

doc/doxygen/src/Examples.dox
feltech: You mean omitting `managementPolicy`? Fine by me

src/openassetio-core/src/managerApi/ManagerInterface.cpp
```diff
@@ -14,11 +16,29 @@ namespace managerApi {
 
 ManagerInterface::ManagerInterface() = default;
 
+#define UNIMPLEMENTED_ERROR(capability)                                                           \
+  fmt::format(                                                                                    \
+      "The '{}' method has not been implemented by the manager. Check manager capability for {} " \
+      "by calling `manager.hasCapability`.",                                                      \
+      __func__, ManagerInterface::kCapabilityNames[static_cast<int64_t>(capability)])
```
feltech, line 23: [question] Don't wanna hold up the merge, but just curious why `int64_t` rather than `size_t` or `kCapabilityNames::size_type` or similar.  Semi-surprised there's no linter warning about signed conversions

elliotcmorris: Hmm, there actually was a warning because I used `UL` initially, that being the direct type of the `std::array`, but the warning message told me to use a sized-type like this. I took the type directly from the message without thinking about it.

But yeah, thinking about it for a second, this is a bit of a silly cast, I'll change it.

doc/doxygen/src/Examples.dox
feltech: Ah I think I get you now, as in no need to add `hasCapability(kResolution)` because `ResolvesFutureEntities` implies it?  Yeah that's fine.

src/openassetio-core/src/managerApi/ManagerInterface.cpp
elliotcmorris: [64afc94](https://github.com/OpenAssetIO/OpenAssetIO/pull/1141/commits/64afc9424aa28a8d9d9d604ed1d2d02264687346)

feltech (commented):
LGTM

Issue #807, opened 2023-01-31
[Lint] `pylint` does not check tests due to fatal error


feltech:
## What

`pylint` aborts with a fatal error when checking the `tests ` directory due to missing `__init__.py`.  This is reproducible locally.  

This does not fail the CI run. 

Linter errors in the main sources still get reported and will fail the CI run. 

Adding the missing `__init__.py` and re-running reveals many linter errors (mostly stuff we just forgot to explicitly ignore).

From [a recent CI run](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/4055787830/jobs/6979402597)
```
/opt/hostedtoolcache/Python/3.9.16/x64/bin/pylint --disable=fixme --ignore-paths=src/openassetio-python/tests/package/pluginSystem/resources/symlinkPath --rcfile=src/openassetio-python/pyproject.toml --output-format=json src/openassetio-python/package/openassetio src/openassetio-python/tests/package
[
    {
        "type": "fatal",
        "module": "package",
        "obj": "",
        "line": 1,
        "column": 0,
        "endLine": null,
        "endColumn": null,
        "path": "src/openassetio-python/tests/package/__init__.py",
        "symbol": "parse-error",
        "message": "error while code parsing: Unable to load file src/openassetio-python/tests/package/__init__.py:\n[Errno 2] No such file or directory: 'src/openassetio-python/tests/package/__init__.py'",
        "message-id": "F0010"
    }
]
```

feltech:
This may be something we need to fix in `fn-pylint-action`.

feltech:
Possible workaround: https://stackoverflow.com/a/54963445/535103

foundrytom:
Also, `fn-pylint-action` doesn't fail if pylint is not installed :roll_eyes:

PR #1146, opened 2023-10-06
[CI] Run pylint on all Python tests not just package

feltech:
## Description

We were limiting our `pylint` checks of test files to only those of the package, when we have several tests under `cmodule` and `bridge`. The new `--recursive` switch added since #807 means that these tests will be trivially discovered by starting to scan one level up.

## Reviewer Notes

Just adding this quick fix before I forget.  Newly surfaced violations currently remain.  This PR can be picked up again when we can schedule some time to it.

## Test Instructions

Equivalent local command line:

```
pylint --disable=I --recursive=y --disable=fixme --rcfile=src/openassetio-python/pyproject.toml src/openassetio-python/package/openassetio src/openassetio-python/tests
```


foundrytom (approved):
Thanks for sorting

elliotcmorris (approved):

PR #1151, opened 2023-10-10
[Docs] Remove misleading use of the term 'stateless'

### LINKED ISSUE #1143: Correct the use of 'stateless' in documentation
## What

Update documentation that gives the impression that managers are "stateless". This is somewhat incoherent due to the `settings`/`initialize` behaviour, which inherently defines some state in the manager, and the entire back-end/data of an asset management system is also inherently stageful.

## Why

This was originally penned to re-enforce that the functional methods of the API (eg `resolve`, `register`, etc.) should not depend on any existing state within the interface for their function or subsequent API use - in the way that say, OpenGL, has a `glGetError` which is explicitly stateful based on previous call.

API methods must inherently affect the state of the _asset management system_.

What the documentation is really trying to say, is that the API methods post-initialize should be:

- theadsafe
- re-entrant
- Only act based on the inputs to the method, and any other state established by the settings passed to `initialize`

Its fine for the manager instance to be statefull (eg: a server connectin based on settings).

## Acceptance Criteria

- Any relevant documentation updated to be unambiguous.

foundrytom:
This came from good intentions but is ultimlately misleading, given the settings mechanism, etc.

This also updates to the American form of 're-enterant' to align with our US English docs poicy.

Closes #1143

### DIFF
--- a/doc/doxygen/src/ForManagers.dox
+++ b/doc/doxygen/src/ForManagers.dox
@@ -20,9 +20,6 @@
  *
  * - All interaction between the @ref host and a @ref manager occurs
  *   through the manager's implementation of the @ref ManagerInterface.
- *   This is a publicly stateless interface, where all associated
- *   information about the session, and caller are provided to each method
- *   by the API middleware.
  *
  * - A manager's implementation of the @ref ManagerInterface supplied
  *   through its @ref PythonPluginSystemManagerPlugin is wrapped in the
@@ -31,11 +28,16 @@
  *   session state management and other auditing/logging functionality.
  *   It also provides a degree of isolation against future API changes.
  *
- * - The @ref ManagerInterface is designed to be stateless. The
- *   response to any method should only depend on the underlying asset
- *   data and any other objects passed into each call. The same logical
- *   operation may be spread out over time and space. It is critical that
- *   any particular implementation does not rely on local in-memory state.
+ * - Methods of the @ref ManagerInterface are required to be reentrant
+ *   and thread-safe. The response to any method should only depend on
+ *   the local state established during
+ *   @fqref{managerApi.ManagerInterface.initialize} "initialization",
+ *   the underlying asset data and any other objects passed into each
+ *   call. The same logical operation may be spread out over time and
+ *   across processes. It is critical that any particular implementation
+ *   does not rely on local in-memory state across different API
+ *   requests. The only exception here being read-through caches etc. as
+ *   long as they are suitably invalidated when upstream data changes.
  *
  * - The main currency in the interaction with a host is the @ref
  *   entity_reference. These are URIs that uniquely identify an @ref
--- a/doc/doxygen/src/Glossary.dox
+++ b/doc/doxygen/src/Glossary.dox
@@ -293,11 +293,13 @@
  *
  * @section manager_state Manager State
  *
- * The @ref ManagerInterface is, by design, a stateless, re-entrant API.
- * Consequently, the result of any call to an instance of this class
- * should solely depend on the methods arguments. In real-world use
- * however, it is often required to understand that a series of discreet
- * calls to the API may be part of the same logical user action.
+ * The @ref ManagerInterface is a reentrant API by design, and its
+ * implementation must be thread-safe. Consequently, the result of any
+ * call to an instance of this class should solely depend on the
+ * initialization settings, underlying asset data and the method
+ * arguments. In real-world use however, it is often required to
+ * understand that a series of discreet calls to the API may be part of
+ * the same logical user action.
  *
  * The most common (and extreme) example of this is a distributed
  * render. Where multiple processes, spread across multiple hosts are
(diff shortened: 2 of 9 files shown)


doc/doxygen/src/Glossary.dox
```diff
@@ -295,11 +295,12 @@
  *
  * @section manager_state Manager State
  *
- * The @ref ManagerInterface is, by design, a stateless, re-entrant API.
- * Consequently, the result of any call to an instance of this class
- * should solely depend on the methods arguments. In real-world use
- * however, it is often required to understand that a series of discreet
- * calls to the API may be part of the same logical user action.
+ * The @ref ManagerInterface is, by design, a reentrant and thread-safe
```
feltech, line 298: [consider] Just a little wary of saying "thread-safe by design". E.g. given our judicious use of `shared_ptr`, claiming a thread-safe design could raise eyebrows.  Perhaps "thread-safe by contract" or similar?

I wonder if at some point we should support extracting a copy of the underlying object in our `shared_ptr`s, so that manager plugins can be certain that no other thread will mess with the data.

feltech: Just saw the docs for `ManagerStateBase`, which put it a bit better
```
* The manager interface is by design reentrant, and must be
* thread-safe
```

feltech (approved):
LGTM

doc/doxygen/src/Glossary.dox
foundrytom: Good point. Tweaked both of those a little [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/1151/commits/aa9af3a604210d2d694a80cddc846e5782338f25)

feltech (approved):

doc/doxygen/src/ForManagers.dox
```diff
@@ -31,11 +28,14 @@
...
- *   operation may be spread out over time and space. It is critical that
- *   any particular implementation does not rely on local in-memory state.
+ * - Methods of the @ref ManagerInterface are required to be reentrant
+ *   and thread-safe. The response to any method should only depend on
+ *   the local state established during
+ *   @fqref{managerApi.ManagerInterface.initialize} "initialization",
+ *   the underlying asset data and any other objects passed into each
+ *   call. The same logical operation may be spread out over time and
+ *   space. It is critical that any particular implementation does not
+ *   rely on local in-memory state.
```
elliotcmorris, line 38: [Consider] I think this could be a cause for confusion. saying that the method mad depend on "local state established during initialization", but then going onto say that an implementation must not depend on "local in-memory state".

This reads to me as though initialize will store state outside of memory, which I don't think is the case. Not sure a better alternative, maybe "does not rely on state created from interactions with any given manager instance." ?

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -166,7 +166,8 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
    * Constructor.
    *
    * No work is done here - ManagerInterface instances should be cheap
-   * to construct and stateless.
+   * to construct. Any heavy-lifting should be done in @ref initialize,
+   * where relevant settings are also available.
```
elliotcmorris, line 170: [Question] Does it matter if a user creates cheap local-state here? I think it dosen't, just checking that there isn't a rule that _all_ immutable state must be created in `initialize`

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -467,9 +468,12 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
   /**
    * Prepares for interaction with a host.
    *
-   * This is a good opportunity to initialize any persistent connections
-   * to a back end implementation. It is fine for this call to block for
-   * a period of time.
+   * This method is passed a settings dictionary, that can be used to
+   * configure required local state to service requests. For example,
```
elliotcmorris, line 472: [consider][minor]?
```suggestion
   * configure required local state for service requests. For example,
```

foundrytom: Was thinking of service as a verb - does that still work?

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
foundrytom: Technically no reason they couldn't - the motivation here was to make things like [this](https://github.com/OpenAssetIO/OpenAssetIO/blob/d2fb418501d316a23ea84562590053ce50404397/src/openassetio-core/src/hostApi/ManagerFactory.cpp#L51) be suitably performant.

doc/doxygen/src/ForManagers.dox
elliotcmorris: Aye.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
elliotcmorris: Aye it does.

elliotcmorris (approved):

elliotcmorris (approved):

PR #1156, opened 2023-10-11
[Core] Verify that managers support required capabilities

### LINKED ISSUE #1148: Ensure base capabilities satisfied when initializing a manager
## What

Provide early checks to determine if the resulting manager is missing any required capabilities early on in the bootstrap process.

- `kEnityReferenceIdentification`
- `kManagementPolicy`

## Why

In order to facilitate multi-language plugins, some required methods are not enforced at the implementation level (as they may be provide by a corresponding plugin in another language). This means there are no build-time guards in place to protect against programming errors in an implementation.

## Acceptance Criteria

- `Manager.initialize` validates that the resulting `ManagerInterface` instance has the `kEntityReferenceIdentification` and `kManagementPolicy` capabilities _after_ initializing the underlying interface.
  - If either is missing, a `ConfigurationError` is raised with a message denoting that the requested implementation does not support the required capabilities, and which capabilities are missing.

foundrytom:
`Manager` will now verify that the wrapped interface implementation suports the required core capabilities. This is done here, as `hasCapability` can only be called post-initialize so we can support proxy interfaces etc...

Closes #1148

As ever, I have no idea any more what constitutes "best practice c++" these days, so attempted to go for 'readability'. C+C most welcome....

### DIFF
--- a/src/openassetio-core/src/hostApi/Manager.cpp
+++ b/src/openassetio-core/src/hostApi/Manager.cpp
@@ -1,7 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2013-2023 The Foundry Visionmongers Ltd
+#include <array>
 #include <stdexcept>
 #include <utility>
+#include <vector>
 
 #include <fmt/format.h>
 
@@ -23,6 +25,41 @@
 namespace openassetio {
 inline namespace OPENASSETIO_CORE_ABI_VERSION {
 namespace {
+
+/**
+ * Validate the supplied ManagerInterface supports all required
+ * capabilities, or throw a ConfigurationException.
+ *
+ * These capabilities are required at runtime, but not enforced at
+ * compile time, as multi-language implementations may only provide the
+ * implementation in one of the component languages.
+ */
+void verifyRequiredCapabilities(const managerApi::ManagerInterfacePtr &interface) {
+  using managerApi::ManagerInterface;
+
+  static constexpr std::array kRequiredCapabilities = {
+      ManagerInterface::Capability::kEntityReferenceIdentification,
+      ManagerInterface::Capability::kManagementPolicyQueries};
+
+  std::vector<std::string> missingCapabilities;
+  for (const ManagerInterface::Capability capability : kRequiredCapabilities) {
+    if (!interface->hasCapability(capability)) {
+      missingCapabilities.emplace_back(
+          ManagerInterface::kCapabilityNames[static_cast<size_t>(capability)]);
+    }
+  }
+
+  if (missingCapabilities.empty()) {
+    return;
+  }
+
+  std::string msg =
+      fmt::format("Manager implementation for '{}' does not support the required capabilities: {}",
+                  interface->identifier(), fmt::join(missingCapabilities, ", "));
+
+  throw errors::ConfigurationException(msg);
+}
+
 /**
  * Extract the entity reference prefix from a manager plugin's info
  * dictionary, if available.
@@ -81,6 +118,12 @@ InfoDictionary Manager::settings() { return managerInterface_->settings(hostSess
 
 void Manager::initialize(InfoDictionary managerSettings) {
   managerInterface_->initialize(std::move(managerSettings), hostSession_);
+
+  // Verify the manager has required capabilities. This must only be
+  // done after initialization, to ensure we can support proxy interface
+  // implementations that need initializing to configure the proxied
+  // implementation
+  verifyRequiredCapabilities(managerInterface_);
 
   entityReferencePrefix_ =
       entityReferencePrefixFromInfo(hostSession_->logger(), managerInterface_->info());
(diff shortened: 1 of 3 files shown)


src/openassetio-core/src/hostApi/Manager.cpp
```diff
@@ -23,6 +23,44 @@
...
+ *
+ * These capabilities are required at runtime, but not enforced at
+ * compile time, as multi-language implementations may only provide the
+ * implementation in one of the component languages.
+ */
+void verifyRequiredCapabilities(const managerApi::ManagerInterfacePtr &interface) {
+  using managerApi::ManagerInterface;
+  using CapabilityVector = std::vector<ManagerInterface::Capability>;
+
+  static const CapabilityVector kRequiredCapabilities = {
```
elliotcmorris, line 39: [consider][minor] Using `std::array` as the type here might let you get away with declaring this `constexpr` in c++17.

src/openassetio-core/src/hostApi/Manager.cpp
```diff
@@ -23,6 +23,44 @@
...
+  if (missingCapabilities.empty()) {
+    return;
+  }
+
+  std::string msg =
+      fmt::format("Manager implementation for '{}' does not support the required capabilities:",
+                  interface->identifier());
+  for (const ManagerInterface::Capability capability : missingCapabilities) {
+    msg += fmt::format(" {}", ManagerInterface::kCapabilityNames[static_cast<size_t>(capability)]);
+  }
```
elliotcmorris, line 59: [Consider][minor]
This implementation is perfectly fine. You might consider extracting the string conversion from the message construction though, it seems cleaner to me, makes it clear that the intent is to have a comma (or space) seperated list, but this is very much a personal style thing.
:
```suggestion
  // Convert missing capabilities to their string representations
  std::vector<std::string> missingCapabilityStrings;
  std::transform(missingCapabilities.cbegin(), missingCapabilities.cend(), 
  std::back_inserter(missingCapabilityStrings), [](const ManagerInterface::Capability cap) {
      return ManagerInterface::kCapabilityNames[static_cast<size_t>(capability)]; 
 });
  
  // Construct error message
  std::string msg =
      fmt::format("Manager implementation for '{}' does not support the required capabilities: {}",
                  interface->identifier(), fmt::join(missingCapabilityStrings, ", ")););
  }
```

elliotcmorris (approved):

src/openassetio-core/src/hostApi/Manager.cpp
foundrytom: Nice, thanks. Updated [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/1156/commits/90a0e40cdf4fd45e4a7143a26a2ae404b4fc7e67)

src/openassetio-core/src/hostApi/Manager.cpp
foundrytom: Ooooooh, I like the join, lets me do what I wanted to in the first place with a comma. Updated based on this [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/1156/commits/90a0e40cdf4fd45e4a7143a26a2ae404b4fc7e67)

elliotcmorris: Oh yeah just doing it with strings in the first pass, makes hella sense.

elliotcmorris (approved):
LGTMSAMWR

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -240,6 +241,54 @@ def test_when_entity_ref_prefix_type_invalid_then_debug_log_printed(
         )
 
 
+class Test_Manager_initizlize_capablility_check:
```
feltech, line 244: [consider][minor] Spelling
```suggestion
class Test_Manager_initialize_capablility_check:
```

src/openassetio-core/src/hostApi/Manager.cpp
foundrytom: Teamwork for the win!!

src/openassetio-python/tests/package/hostApi/test_manager.py
```diff
@@ -240,6 +241,54 @@ def test_when_entity_ref_prefix_type_invalid_then_debug_log_printed(
...
+    def test_has_capability_called_after_initialize(
+        self, manager, a_host_session, mock_manager_interface
+    ):
+        manager.initialize({})
```
feltech, line 248: [question] Interesting that this passes without having to set up `hasCapability`'s `return_value`, is the default return being coerced to a `bool`?

feltech (approved):
Neat!

src/openassetio-python/tests/package/hostApi/test_manager.py
foundrytom: ooh yeah - it must be!

feltech (approved):

PR #1158, opened 2023-10-12
[Core] Remove unused `EmptyEntityReferencePagerInterface`

foundrytom:
This is a relic from the old default implementation.


elliotcmorris (approved):
Oh well remembered. :+1:

PR #1159, opened 2023-10-12
[Doc] Add throws doc to exception policy overloads

elliotcmorris:
## Description

The exceptional conveniences did not document that they threw. This was somewhat implicit, however with the addition of the NotImplementedException, it is now quite suspicious, as it looks like they only throw NotImplementedExceptions, which isn't true from the host perspective.

Didn't log an issue for this, comes from https://github.com/OpenAssetIO/OpenAssetIO/pull/1141#discussion_r1349794207, I just thought doing this as a seperate PR would avoid holding up the prior one.

Looks like : 
image


foundrytom (approved):
Perfection! Thanks for taking the initiative here 🙏

Issue #1114, opened 2023-09-20
Platform agnostic CI build configuration


feltech:
## What

Some ideas I wanted to capture before I forgot 

* Use MSVC CMake generator on Windows.
* Use bash shell.
* Remove unnecessary bootstrapping.

## Why

We spend too much time waiting for CI runs when we could be doing more productive things. We should have an easy time of making changes and reusing bits of CI across workflows/repos, so should simplify it where possible.

* Using the MSVC generator means we can remove the `vcvarsall` dance. 
  - This is also a pre-requisite to using `bash` in Windows, since `vcvarsall.bat` cannot be called in bash.
  - To allow cross-platform `cmake` command lines, we should use the "Ninja Multi-Config" generator for non-Windows platforms, and get in the habit of passing `--config Release` to `cmake` commands. This includes CMake targets that call `cmake --build` as a subprocess (otherwise `--build` will build _all_ configurations, rather than the target configuration).
* It turns out `bash` is installed on Windows runners, so we can remove some `cmd`-specific conditional steps.
* ~~We removed the `cpython` Conan package dependency, but still have a bunch of steps in the `bootstrap` shell scripts specifically to accommodate it.~~

## Notes

See https://github.com/OpenAssetIO/OpenAssetIO/pull/1183/files for examples

feltech:
Did some experimentation (https://github.com/OpenAssetIO/OpenAssetIO/pull/1275). 

The boilerplate involved in littering the CI/CMake config with `--config` is quite extensive. It's more platform and CMake generator -agnostic, so is probably a good thing, but complicates things slightly (i.e. easy to forget).

Using `Ninja Multi-Config` means we can remove `-DBUILD_TYPE=` from the `cmake` command-line, in favour of the more uniform `--config`, but then the generator type must be in the build matrix, so the `preamble` entry in the build matrix is just replaced by an entry for the generator, meaning we cannot simplify the build matrix to a single array of platforms. 

We can eschew `Ninja Multi-Config` to use the default generator for the platform, which is perhaps better, since it reflects platform-specific defaults. However, we then need _both_ `-DBUILD_TYPE=` and `--config` on every `cmake` configuration line. This isn't so bad, but then we have `CMakePresets.json` presets that set `BUILD_TYPE` (e.g. for sanitizers), but don't set `--config`. To add `--config` to the presets would require a build preset (as opposed to configure), or adding `--config` to the command line (which would diverge in how config is set per-platform).

[...]

PR #1162, opened 2023-10-13
[CI] Remove `apt install` of Conan `cpython` deps

feltech:
## Description

Part of #1114. The additional system packages for Linux builds are no longer necessary since #1038 - we now use the system-provided Python rather than the `cpython` Conan package, which required all these extra system dependencies.

### DIFF
--- a/resources/build/bootstrap-ubuntu-20.04.sh
+++ b/resources/build/bootstrap-ubuntu-20.04.sh
@@ -7,19 +7,6 @@ sudo apt-get update
 # Install gcc, linters, build tools used by conan and Python 3.
 sudo apt-get install -y build-essential pkgconf clang-format-12 clang-tidy-12 python3-pip ccache
 
-# Install system packages required for the Python 3.9 conan package.
-# These would be installed as part of the conan package install, but
-# we're caching the conan directory via the `actions/cache` Github
-# action, so a fresh Github VM is left without these system packages.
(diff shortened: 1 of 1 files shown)


foundrytom (approved):
Always love a nice bit of deletion.

resources/build/bootstrap-ubuntu-20.04.sh
```diff
@@ -7,19 +7,6 @@ sudo apt-get update
 # Install gcc, linters, build tools used by conan and Python 3.
```
foundrytom, line 7: [Question] Is there anything in the "python 3" part of that comment that needs updating?

elliotcmorris (approved):

resources/build/bootstrap-ubuntu-20.04.sh
feltech: I guess its still relevant since it's installing `pip`, which is needed on the next line. 

We could probably get rid of `build-essential pkgconf` now, since none of our Conan dependencies need building... and could get rid of the `--build=missing` from the `conan` command line too.

feltech: Ah, though this file is still used as part of the Vagrant VM instructions, so `build-essential` is probably required for that.

foundrytom: Cool beans, lets leave as is.

foundrytom:
FFTSAMAYL

PR #1174, opened 2023-10-19
Add deprecations for errors/exceptions

foundrytom:
Should help allow code to work in `a14` and `b1.rev0`


elliotcmorris (changes requested):

elliotcmorris (approved):

feltech (approved):
LGTM. A bit concerned about our discovery that `errors::BatchElementError` is not exported, but that's a separate issue.

PR #1180, opened 2023-10-27
[CI] Add TraitGen C++ integration tests

### LINKED ISSUE #841: [CI] Enable TraitGen CMake CTest tests
## What

Enable TraitGen CTest integration tests.

## Why

Since https://github.com/OpenAssetIO/OpenAssetIO-TraitGen/issues/11 (also see follow-up https://github.com/OpenAssetIO/OpenAssetIO-TraitGen/issues/17) there are C++ specific tests for TraitGen.  

We currently run the TraitGen tests alongside other Python-only tests that use OpenAssetIO installed as a Python `pip` package, which the TraitGen CMake CTest project cannot use to build against.

Currently the tests are skipped, since we don't set the `OPENASSETIO_TRAITGENTEST_CMAKE_PRESET` environment variable.  Also see #840.

## Acceptance Criteria

- TraitGen C++ test runs in `integrations.yml` (the CMake CTest tests).

## Implementation notes

We may be able to move the TraitGen test from the `python-integrations` job to the `cmake-integrations` job.

feltech:
## Description

Closes #841. Since OpenAssetIO/OpenAssetIO-TraitGen#11 (also see follow-up OpenAssetIO/OpenAssetIO-TraitGen#17) there are C++ specific tests for TraitGen.

We were running the TraitGen tests alongside other Python-only tests that use OpenAssetIO installed as a Python pip package, which the TraitGen CMake CTest project cannot use to build against.

So add a separate test run of TraitGen under the "CMake package" integration tests. Only run the tests labelled `ctest`, since the tests for Python codegen are still executed under the "Python e2e" tests.

Since we use `awk` to assert that TraitGen test dependencies don't overwrite OpenAssetIO, we must use `bash` as the shell for all platforms. This means `vcvarsall.bat` can't be used on Windows. So switch to using the default generator for the platform (rather than Ninja), which on Windows will be MSVC and so not requiring `vcvarsall.bat`.

### DIFF
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -89,7 +89,7 @@ jobs:
           repository: OpenAssetIO/OpenAssetIO-TraitGen
           path: external/TraitGen
 
-      - name: Install TraitGen
+      - name: Install TraitGen (Python)
         run : |
           python -m pip install external/TraitGen
 
@@ -115,8 +115,7 @@ jobs:
       matrix:
         config:
         - os: windows-2019
-          preamble: call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat" x64
-          shell: cmd
+          shell: bash
         - os: ubuntu-20.04
           shell: bash
         - os: macos-11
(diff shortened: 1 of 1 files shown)


feltech:
:laughing: was a bit optimistic with this one. Multiple CI issues varying across platforms. Closing for further experimentation in a fork.

PR #1183, opened 2023-10-31
[CI] Add TraitGen C++ integration tests

### LINKED ISSUE #841: [CI] Enable TraitGen CMake CTest tests

feltech:
## Description

Closes #841. Since OpenAssetIO/OpenAssetIO-TraitGen#11 (also see follow-up OpenAssetIO/OpenAssetIO-TraitGen#17) there are C++ specific tests for TraitGen.

We were running the TraitGen tests alongside other Python-only tests that use OpenAssetIO installed as a Python pip package, which the TraitGen CMake CTest project cannot use to build against.

So add a separate test run of TraitGen under the "CMake package" integration tests. Only run the tests labelled `ctest`, since the tests for Python codegen are still executed under the "Python e2e" tests.

Since we use `awk` to assert that TraitGen test dependencies don't overwrite OpenAssetIO, we must use `bash` as the shell for all platforms. This means `vcvarsall.bat` can't be used on Windows. So switch to using the default generator for the platform (rather than Ninja), which on Windows will be MSVC and so not requiring `vcvarsall.bat`.

Using bash across all platforms is also part of a broader plan, see #1114.

Since we're using the default CMake generators, MSVC becomes multi-config, so we must specify `--config` when executing CMake command lines. Since we're not using Ninja, builds won't be parallelised automatically, so we must specify `--parallel` on the CMake build command line.

TraitGen hardcodes to `RelWithDebInfo` CMake builds, so use that across the board to ensure compatibility.

Note that TraitGen's `pytest -m ctest` triggers a `conan install`. So update the CONAN_USER_HOME to use the same directory as OpenAssetIO, to try to ensure packages are cached. In practice, this may not work since other jobs may lock the cache (found during dev testing). However, the TraitGen test structure is due to change in
OpenAssetIO/OpenAssetIO-TraitGen#32 so it doesn't seem worth spending too much time on finessing.

### DIFF
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -89,7 +89,7 @@ jobs:
           repository: OpenAssetIO/OpenAssetIO-TraitGen
           path: external/TraitGen
 
-      - name: Install TraitGen
+      - name: Install TraitGen (Python)
         run : |
           python -m pip install external/TraitGen
 
@@ -115,10 +115,9 @@ jobs:
       matrix:
         config:
         - os: windows-2019
-          preamble: call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat" x64
-          shell: cmd
+          site-packages: Lib/site-packages
         - os: ubuntu-20.04
-          shell: bash
+          site-packages: lib/python3.9/site-packages
         - os: macos-11
           # MacOS toolchain doesn't search /usr/local by default:
           # https://gitlab.kitware.com/cmake/cmake/-/issues/19120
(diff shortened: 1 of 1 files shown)


.github/workflows/integrations.yml
```diff
@@ -162,6 +159,24 @@ jobs:
         run : |
           python -m pip install external/TraitGen
 
+      - name: Install TraitGen test dependencies
+        run : >
+          python -m pip install -r external/TraitGen/tests/requirements.txt
+          | awk '{print} /Attempting uninstall: openassetio/{found=1} END{exit found}'
```
elliotcmorris, line 165: [Question] This seems like a very sensible thing to be testing. I wonder if there are other places we can keep this in mind.

elliotcmorris (approved):
👍

foundrytom (approved):

Issue #1181, opened 2023-10-31
Publish and use C++ distribution artifacts in CI


feltech:
# Implementation

* Upload `dist` using tried and tested `upload-artifact` action.
* File name as detailed in epic.
* Update dependent projects (`TraitGen`, `usdOpenAssetIOResolver`, `Test-CMake`)

# Out of scope

* Release artifacts (only concerned with CI for now)
  - Hence CPack is less useful.

PR #1187, opened 2023-11-03
[CI] Add Build Archive CI script.

elliotcmorris:
Part of https://github.com/OpenAssetIO/OpenAssetIO/issues/1181

Add archive behaviour to merges to main, so OpenAssetIO is built on all of our runner platforms and archived in github.

Intent is to unlock much better CI behaviour for us, as we are spending a lot of time and effort building OpenAssetIO across dependent repos, as well as just being generally convenient to have.

This will publish an artifact to every PR run, as well as one for each commit to main. The idea being that dependent repos are by default targetted to the `main` branch with the download-artifact action.

### DIFF
--- a/.github/workflows/build-archive.yml
+++ b/.github/workflows/build-archive.yml
@@ -1,0 +1,70 @@
+# Build on commit to main and archive  for easy retrieval.
+
+name: Build Archive
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-openassetio-atrifact:
+    name: ${{ matrix.config.os }}
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # We can't properly align to the VFX Reference Platform as this
+        # requires glibc 2.17, which is older than any of the available
+        # environments.
+        config:
+        - os: windows-2019
+          #Echo adds a newline character without it being interpreted as a command to vcvarsall
+          preamble: call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat" x64
+          shell: cmd
+        - os: ubuntu-20.04
+          shell: bash
+        - os: macos-11
+          # MacOS toolchain doesn't search /usr/local by default:
+          # https://gitlab.kitware.com/cmake/cmake/-/issues/19120
+          # The CMake FindPython module's Python::Python target (used
+          # via pybind11::embed in python-bridge-test) transitively adds
+          # linker flags to system libs, which fail due to this issue.
+          preamble: export LDFLAGS="-L/usr/local/lib"
+          shell: bash
+    defaults:
+      run:
+        # Annoyingly required here since `matrix` isn't available in
+        # the `shell` property of individual steps.
+        shell: ${{ matrix.config.shell }}
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build-archive.yml
```diff
@@ -0,0 +1,70 @@
...
+        cmake --build build --parallel --config RelWithDebInfo
+
+        cmake --install build --config RelWithDebInfo
+      env:
+        CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
+
+    - name: Upload archive
+      uses: actions/upload-artifact@v3
+      with:
+        name: openassetio-${{ runner.os }}
```
elliotcmorris, line 69: The requirements stated we wanted to get the OS version as well, ie `windows 2019`, whereas this is just `windows`. However, i couldn't figure out a way to get that info without integrating a custom action, which didn't seem worthwhile to me.

.github/workflows/build-archive.yml
```diff
@@ -0,0 +1,70 @@
...
+        cmake -S . -B build --install-prefix ${{ github.workspace }}/dist
+        -DOPENASSETIO_ENABLE_C=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo
+
+        cmake --build build --parallel --config RelWithDebInfo
+
+        cmake --install build --config RelWithDebInfo
+      env:
+        CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
+
+    - name: Upload archive
```
foundrytom, line 66: [question]
Im a bit concerned that we're not testing this build, and then planning to re-using it extensively downstream. This build stage could get out of sync with the one we are testing. It is also incredibly helpful when published artifacts correlate to other checks, especially for debugging. Would it be safer to publish the one from the main `ctest`  runs vs a custom untested build? This would need us to split build-test so we can make sure the artifact is uploaded even when the tests fail.

.github/workflows/build-archive.yml
foundrytom: `${{ matrix.config.os }}` should hopefully get you that value

foundrytom (changes requested):
This is going to be mega useful, thanks. Main concern is that we're not testing this build.

.github/workflows/build-archive.yml
elliotcmorris: :facepalm:

.github/workflows/build-archive.yml
elliotcmorris: Because the build commands/runner configurations are duplicated, so we could be building an artifact configured different to the artifact we're running tests on?

Yeah I can see that, although I also saw it as a positive, as we gain coverage by more properly being a consumable artifact, (ie, not have the tests built in at all.)

foundrytom: Yeah, exactly that. Def with you on the downstream repos using it - but if they also use the one we test here then I think you end up with the maximum amount of coverage?

foundrytom: > (ie, not have the tests built in at all.)

Yeah, that would be ideal, would `ctest` be clever enough to be able to reuse just the `dist` part?

foundrytom: Though if we publish the artifact before running `ctest` is there much difference? Aside from the chance the tests are using something outside of `dist`?

.github/workflows/build-archive.yml
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1187/commits/5b334247f552ce12ae566b9ac320ad003bf99316

PR #1190, opened 2023-11-08
Python->C++ exception translation

### LINKED ISSUE #947: Python exception translation
## What

Ensure information is not lost from Python exceptions when bridged to C++

Also see #866 (duplicate issue)

## Why

We currently lose information about Python exceptions when caught in C++.

When Python code calls a C++ function, and that C++ function raises an exception, there is a built-in hook via pybind11 to translate the C++ exception to a Python exception (see usages of `register_exception_translator`).

However, when C++ calls a Python function, and that Python function raises an exception, pybind11 will detect the Python exception and throw an (unhelpfully named) `error_already_set` exception. There is no pybind11 hook available to modify this. 

In this case, consumers (hosts or manager plugins) must catch `std::exception` and interrogate the `what()` string to disambiguate.

We really need to devise a way to translate Python exceptions to the appropriate C++ exceptions, so that consumers can disambiguate using the exception type (i.e. in `catch` blocks).

An unfortunate wrinkle is in the special status that pybind11's `error_already_set` exception has within pybind11 itself. If this exception propagates _back_ into Python, then the original Python exception is restored, complete with stack trace etc.

So ideally we would only conditionally translate an `error_already_set` exception into a more appropriate C++ exception, by somehow figuring out whether the exception will be caught in C++ or Python code.

## ACs 

- Define approach for the above

feltech:
## Description

Closes #947. Establish a strategy for single-sourcing Python<->C++ exception mapping, add a mechanism to translate OpenAssetIO-specific Python exceptions to corresponding C++ exceptions, and augment the pybind11 override macros to do the translation.

This only affects pybind11 "trampoline" functions - i.e. C++ virtual methods that can call out to a Python implementation.

## Reviewer Notes

The final commit is optional. A comment in the GitHub issue mentions `createPythonPluginSystemManagerImplementationFactory`, so the final commit was added to be able to use the new functionality there. However, on inspection it didn't actually seem necessary.

Testing every affected class isn't tackled. Instead, the macros are tested via dedicated contrived `cmodule` (`_openassetio_test`) tests. Testing for all classes can be added if it seems worthwhile, but will be a big chunk of code (similar to the GIL tests). Otherwise, we're OK as long as we trust contributors to always use the `OPENASSETIO_`-prefixed pybind11 macros (as detailed in the CODING_STANDARDS).

Only OpenAssetIO-specific exceptions are converted.  It wouldn't be much extra effort to also cover other  exceptions.

### DIFF
--- a/doc/contributing/CODING_STANDARDS.md
+++ b/doc/contributing/CODING_STANDARDS.md
@@ -232,6 +232,25 @@ a null pointer.
     .def("setHost", py::arg("host").none(false))
 ```
 
+### Polymorphic override macros
+
+In order to support C++ class instances transparently calling into a
+mirrored Python instance, pybind11 has support for so-called
+["trampoline" classes](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#overriding-virtual-functions-in-python),
+which dispatch either to the Python method override, or to the C++ base
+class implementation if no Python override is found. Method bodies
+use convenience macros to implement this logic.
+
+In pybind11, any exception that occurs in Python, and propagates to C++,
+is translated to an `error_already_set` C++ exception. In order to
+support Python->C++ exception type translation, we must augment the
+pybind11 macros.
+
+So instead of using the `PYBIND11_OVERRIDE_*` family of macros, we must
+use `OPENASSETIO_PYBIND11_OVERRIDE_*`, which decorates the pybind11
+implementation with additional functionality, such as exception
+translation.
+
 ### Python object lifetime
 
 Due to a [Pybind issue](https://github.com/pybind/pybind11/issues/1333),
@@ -268,7 +287,7 @@ member functions from, for example
 
 ```c++
 std::shared_ptr<MyReturnType> PyMyClass::myMethod(myArg) override {
-  PYBIND11_OVERRIDE(
+  OPENASSETIO_PYBIND11_OVERRIDE(
       std::shared_ptr<MyReturnType>, MyClass, myMethod, myArg);
 }
 ```
@@ -277,7 +296,7 @@ to
 
 ```c++
 std::shared_ptr<MyReturnType> PyMyClass::myMethod(myArg) override {
-  PYBIND11_OVERRIDE(
+  OPENASSETIO_PYBIND11_OVERRIDE(
       PyRetainingSharedPtr<MyReturnType>, MyClass, myMethod, myArg);
 }
 ```
@@ -422,14 +441,6 @@ provided externally (e.g. manager plugins).
 
 Any methods not released should be in O(1) time and guaranteed to not
 cause a deadlock by calling out to python off-thread.
-
-Until https://github.com/pybind/pybind11/issues/4878 is fixed, pure
-virtual bindings should use the `OPENASSETIO_PYBIND11_OVERRIDE_PURE`
-macro in the pybind11 [trampoline
-class](https://pybind11.readthedocs.io/en/stable/advanced/classes.html),
-rather than `PYBIND11_OVERRIDE_PURE`. See the issue description for more
-details.
-
 
 ## Environment variables
 
--- a/src/openassetio-core/include/openassetio/errors/exceptions.hpp
+++ b/src/openassetio-core/include/openassetio/errors/exceptions.hpp
@@ -4,6 +4,7 @@
 
 #include <stdexcept>
 #include <string>
+#include <tuple>
 #include <utility>
 
 #include <openassetio/export.h>
@@ -106,6 +107,12 @@ struct OPENASSETIO_CORE_EXPORT BatchElementException : OpenAssetIOException {
    */
   BatchElementError error;
 };
+
+/// List of all OpenAssetIO-specific exceptions. Useful for
+/// introspection in language bindings.
+using AllExceptions =
+    std::tuple<OpenAssetIOException, InputValidationException, ConfigurationException,
+               NotImplementedException, UnhandledException, BatchElementException>;
 
 /**
  * @}
--- a/src/openassetio-python/cmodule/src/_openassetio.hpp
+++ b/src/openassetio-python/cmodule/src/_openassetio.hpp
@@ -21,54 +21,6 @@ OPENASSETIO_FWD_DECLARE(hostApi, HostInterface)
 OPENASSETIO_FWD_DECLARE(hostApi, ManagerImplementationFactoryInterface)
 
 /**
- * Similar to PYBIND11_OVERRIDE, but allows different arguments to be
- * given to the Python implementation and the C++ fallback (base
- * class) implementation.
- *
- * For example, this is useful to decorate std::function callbacks
- * with `PyRetainingSharedPtr`s, so that Python objects are kept alive
- * as they pass through C++ callbacks.
- *
- * The CppArgs parameter must be a parentheses-enclosed,
- * comma-separated, list of arguments. These are given to the C++ base
- * class implementation, if no Python override is found.
- *
- * All remaining arguments following the CppArgs (not
- * parentheses-enclosed) are passed to the Python override
- * implementation, if one exists.
- */
-#define OPENASSETIO_PYBIND11_OVERRIDE_ARGS(Ret, Class, Fn, CppArgs, ... /* PyArgs */) \
-  PYBIND11_OVERRIDE_IMPL(PYBIND11_TYPE(Ret), PYBIND11_TYPE(Class), #Fn, __VA_ARGS__); \
-  return Class::Fn CppArgs
-
-/**
- * Work around https://github.com/pybind/pybind11/issues/4878
- *
- * In a Debug build, where `assert` is enabled, pybind11 will do an
- * `assert(!PyErr_Occurred())` before throwing "Tried to call pure
- * virtual function". Since Python 3.9, the PyErr_Occurred() function
- * requires the GIL to be acquired.
- *
- * The following macro duplicates PYBIND11_OVERRIDE_PURE_NAME and
- * inserts a `gil_scoped_acquire`, in order to work around the problem
- * until it is fixed upstream.
- *
- * @todo Revert to using PYBIND11_OVERRIDE_PURE once upstream fix is
- * available.
- */
-#define OPENASSETIO_PYBIND11_OVERRIDE_PURE_NAME(ret_type, cname, name, fn, ...)               \
-  do {                                                                                        \
-    PYBIND11_OVERRIDE_IMPL(PYBIND11_TYPE(ret_type), PYBIND11_TYPE(cname), name, __VA_ARGS__); \
-    const pybind11::gil_scoped_acquire gil{};                                                 \
-    pybind11::pybind11_fail(                                                                  \
-        "Tried to call pure virtual function \"" PYBIND11_STRINGIFY(cname) "::" name "\"");   \
-  } while (false)
-
-#define OPENASSETIO_PYBIND11_OVERRIDE_PURE(ret_type, cname, fn, ...)                              \
-  OPENASSETIO_PYBIND11_OVERRIDE_PURE_NAME(PYBIND11_TYPE(ret_type), PYBIND11_TYPE(cname), #fn, fn, \
-                                          __VA_ARGS__)
-
-/**
  * Declare a `RetainPyArgs` alias with common template arguments.
  *
  * The template arguments are those types that we expect to be derived
--- a/src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
+++ b/src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2013-2023 The Foundry Visionmongers Ltd
+#include <map>
 #include <string_view>
+#include <tuple>
+#include <type_traits>
 
 #include <pybind11/eval.h>
 #include <pybind11/operators.h>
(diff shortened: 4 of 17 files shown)


src/openassetio-core/include/openassetio/errors/exceptions.hpp
```diff
@@ -107,6 +108,12 @@ struct OPENASSETIO_CORE_EXPORT BatchElementException : OpenAssetIOException {
   BatchElementError error;
 };
 
+/// List of all OpenAssetIO-specific exceptions. Used in Python
+/// exception translation.
```
foundrytom, line 112: [Consider][minor] As this is core, maybe mention languages more generically if the intention is to enable more programatic binding generation for any target language?

src/openassetio-core/include/openassetio/errors/exceptions.hpp
```diff
@@ -107,6 +108,12 @@ struct OPENASSETIO_CORE_EXPORT BatchElementException : OpenAssetIOException {
...
+using AllExceptions =
+    std::tuple<OpenAssetIOException, InputValidationException, ConfigurationException,
+               NotImplementedException, UnhandledException>;
```
foundrytom, line 115: [Question] Why isn't `BatchElementException` here?

src/openassetio-python/private/include/openassetio/private/python/exceptions.hpp
```diff
@@ -0,0 +1,362 @@
...
+ *
+ * @tparam I Indices of exceptions in AllExceptions.
+ */
+template <std::size_t... I>
+constexpr bool assertAllExceptionsAreInList([[maybe_unused]] std::index_sequence<I...> unused) {
+  (assertExceptionIsInList<I>(CppExceptionsAndPyClassNames::kIndices), ...);
+  return true;
+}
+/**
+ * Assert that all exceptions in the AllExceptions list are also found
```
foundrytom, line 201: [Consider][minor] For a new reader, it might be nice to differentiate that the above are the mechanism by which the last actually does the assertion.

```suggestion
 * Execute compile-time assertions that all exceptions in the AllExceptions list are also found
```

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -11,87 +14,376 @@
...
+ * @param mod Python module to hold new Python exception class.
+ * @param registeredExceptions Previously registered exceptions, to use
+ * when looking up Python base class to inherit from, and to add this
+ * exception to once registered.
+ */
+template <>
+void registerPyExceptionClass<ConfigurationException>(const std::string_view pyClassName,
+                                                      const py::module_ &mod,
+                                                      PyObjectByName &registeredExceptions) {
+  const py::object pyBaseClass = registeredExceptions.at("InputValidationException");
```
foundrytom, line 166: [question][naive] Thinking out loud, from a very uninformed starting point, could you use `kCppExceptionsAndPyClassNames` to auto-derive `pyBaseClass` from the introspected base of the template so you dont need a specialisations? Might not be worth it.

src/openassetio-python/tests/cmodule/test_errors.py
```diff
@@ -66,3 +70,82 @@ def test_when_ConfigurationException_thrown_then_can_be_caught_as_InputValidatio
...
+            errors.BatchElementError.ErrorCode.kInvalidEntityReference, "Explosion!"
+        )
+        exception = exception_type(123, error, "Another explosion!")
+    else:
+        exception = exception_type("Explosion!")
+
+    return exception
+
+
+class PyExceptionThrower(_openassetio_test.ExceptionThrower):
```
foundrytom, line 131: [Consider] This is really well documented in c++ - might be worth a cursory note here on what this is doing and how.

doc/contributing/CODING_STANDARDS.md
```diff
@@ -232,6 +232,25 @@ a null pointer.
...
+class implementation if no Python override is found. Method bodies
+use convenience macros to implement this logic.
+
+In pybind11, any exception that occurs in Python, and propagates to C++,
+is translated to an `error_already_set` C++ exception. In order to
+support Python->C++ exception type translation, we must augment the
+pybind11 macros.
+
+So instead of using the `PYBIND11_OVERRIDE_*` family of macros, we must
+use `OPENASSETIO_PYBIND11_OVERRIDE_*`, which decorates the pybind11
```
foundrytom, line 250: [question] if wonder if its worth a grep test in CI to assert this? like we do for OAIO?

foundrytom (commented):
Only superficial comments from me as others are far better placed to comment on the implementation. But great job on the thoroughness of approach/testing. Minor Q - would you imagine squashing the last two commits into the previous commit-but-one?

foundrytom:
> . A comment in the GitHub issue mentions createPythonPluginSystemManagerImplementationFactory, so the final commit was added to be able to use the new functionality there. However, on inspection it didn't actually seem necessary.

I think the intent was that that code could raise OpenAssetIO exceptions from Python, that would end up in C++.

feltech:
> > . A comment in the GitHub issue mentions createPythonPluginSystemManagerImplementationFactory, so the final commit was added to be able to use the new functionality there. However, on inspection it didn't actually seem necessary.
> 
> I think the intent was that that code could raise OpenAssetIO exceptions from Python, that would end up in C++.

Yeah that was the thought, but looking at the code, no OpenAssetIO exceptions can be raised there. We could wrap the construction of `PythonPluginSystemManagerImplementationFactory` just in case we modify the Python constructor in future to raise an OpenAssetIO exception?

The `instantiate` method is already covered because it's pybind11 trampoline method. It was just potentially the constructor that's not covered.

feltech:
> Minor Q - would you imagine squashing the last two commits into the previous commit-but-one?

Yeah I guess that would make the history a bit easier to grok, assuming we're happy to keep the new shared private header, even though its not currently necessary.

foundrytom:
> Yeah that was the thought, but looking at the code, no OpenAssetIO exceptions can be raised there. We could wrap the construction of `PythonPluginSystemManagerImplementationFactory` just in case we modify the Python constructor in future to raise an OpenAssetIO exception?

Ahh yes - maybe we need to worry about `ManagerInterfaceFactoryInterface` then? As that calls into Python?

foundrytom:
> Ahh yes - maybe we need to worry about `ManagerInterfaceFactoryInterface` then? As that calls into Python?

Oh and `PagerInterface`? And anything else like that? Ie many of the places we had to touch the GIL?

src/openassetio-python/cmodule/src/_openassetio.hpp
```diff
@@ -20,10 +20,33 @@ OPENASSETIO_FWD_DECLARE(log, LoggerInterface)
 OPENASSETIO_FWD_DECLARE(hostApi, HostInterface)
 OPENASSETIO_FWD_DECLARE(hostApi, ManagerImplementationFactoryInterface)
 
+/// @note Update errorsTest.cpp if adding more override macros below.
+
+/**
+ * Decorate PYBIND11_OVERRIDE_NAME with exception type translation.
+ */
+#define OPENASSETIO_PYBIND11_OVERRIDE_NAME(ret_type, cname, name, fn, ...)                      \
```
elliotcmorris: [Consider] I might think about defining these in their own place within the cmodule dir, `exception_translation_macros.hpp` or something. They get in the way of the "point" of this file imo.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -462,3 +462,129 @@ void registerExceptions(const py::module &mod) {
...
+ * to @p Exception class.
+ * @param thrownPyExc pybind11-wrapped Python exception to potentially
+ * wrap in a HybridException.
+ * @param thrownPyExcName Python exception class name of @p thrownPyExc
+ * to test against @p expectedPyExcName.
+ */
+template <class Exception>
+void throwHybridExceptionIfMatches(const std::string_view expectedPyExcName,
+                                   const py::error_already_set &thrownPyExc,
+                                   const std::string_view thrownPyExcName) {
```
elliotcmorris: [Consider] could you extract the name from the PyExc here in the function, rather than require an external extraction and argument pass??

elliotcmorris (changes requested):
Really solid engineering imo. I couldn't have done this as well.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -11,87 +14,376 @@
...
+ */
 
-void registerBatchElementException(const py::module &mod) {
-  // Pybind has very limited support for custom exception types. This is
-  // a well-known tricky issue and is apparently not on the roadmap to
-  // fix. The only direct support is for an exception type that takes a
-  // single string parameter (message). However, we need `index` and
-  // `error` parameters to mirror C++.
+/**
+ * @section topy Conversion from C++ exception to Python exception.
```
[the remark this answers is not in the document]
elliotcmorris: Oh no it's not. its To-py, Doh!

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
elliotcmorris: Heh, i noted in my review how I thought resisting the urge to do this was a good call :P.
Either works, but don't let the genericism override the clarity.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -11,87 +14,376 @@
...
+ * identifiers.
+ *
+ * @tparam Type List of types.
+ */
+template <class... Type>
+struct TypesAndIds {
+  static constexpr std::size_t kSize = sizeof...(Type);
+  using Types = std::tuple<Type...>;
+  const std::array<std::string_view, kSize> ids;
+};
```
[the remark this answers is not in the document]
elliotcmorris: Not actually sure how that works, if std::array would reject an initializer list or a vector of innapropriate size, or if there'd be an implicit conversion leaving some empty string views on the end.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -11,87 +14,376 @@
...
+/// Mapping of Python class name to Python class object.
+using PyObjectByName = std::map<const std::string_view, const py::object>;
+
+/**
+ * Register a new Python exception.
+ *
+ * General case of a simple exception (only takes a single string
+ * constructor argument) with OpenAssetIOException as its base class.
+ *
+ * @tparam Exception Corresponding C++ exception type.
```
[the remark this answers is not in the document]
hartmutbehrens: `tparam` -> template parameter, so probably fine.

feltech:
> > Ahh yes - maybe we need to worry about `ManagerInterfaceFactoryInterface` then? As that calls into Python?
> 
> Oh and `PagerInterface`? And anything else like that? Ie many of the places we had to touch the GIL?

As long as calling out to Python from C++ is done transparently via a C++ base class virtual method (i.e. pybind11 trampoline), then exceptions raised in Python will be translated, due to the new/updated `OPENASSETIO_PYBIND11_OVERRIDE*` macros.

If there are places where we explicitly call a method on a `py::object`, then we'd need to decorate that call to translate any OpenAssetIO exceptions that propagate out of it.  But I don't think we do this anywhere, apart from `createPythonPluginSystemManagerImplementationFactory`?

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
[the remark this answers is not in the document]
feltech: Yeah it fails at compile time if you try to give too few or too many elements to the array.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -11,87 +14,376 @@
...
+ *
+ * In particular there are two lists, such that given a single index,
+ * the C++ exception class or the Python class name can be queried at
+ * compile time.
+ */
+struct CppExceptionsAndPyClassNames {
+  using Type = decltype(kCppExceptionsAndPyClassNames);
+  /// "Array" of C++ exception classes.
+  template <std::size_t I>
+  using Exceptions = std::tuple_element_t<I, typename Type::Types>;
```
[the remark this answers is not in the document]
feltech: Yeah the idea was to make it "look" like an array when used, especially when using `kClassNames` and `Execptions` together with the same index.

I think `std::get` only retrieves values, as opposed to types. Maybe it has specialisations, will take a look.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -11,87 +14,376 @@
...
+/**
+ * Execute compile-time assertions that no class in the
+ * CppExceptionsAndPyClassNames list has a base class at a lower index
+ * in the list.
+ *
+ * The resulting value is unused - it's just kludge to make the
+ * compiler make the attempt and so fail on any violated static_asserts.
+ */
+[[maybe_unused]] constexpr bool kIsClassListOrderValid =
+    assertClassesAreInHierarchyOrder(CppExceptionsAndPyClassNames::kIndices);
```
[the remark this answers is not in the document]
feltech: Oh interesting, odd kinda no-op assert, but getting rid of the pointless variable would be good.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -11,87 +14,376 @@
...
+   */
+  static constexpr auto kIndices = std::make_index_sequence<kSize>{};
+};
+
+/**
+ * @}
+ */
+
+/**
+ * @section asserts Compile-time assertions.
```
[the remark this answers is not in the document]
feltech: I'm not keen on that, since the `exceptionsValidation.hpp` header on its own would have lots of red squiggles in my editor (since it depends on stuff defined "above").

I am keen on finding a way to hide some of this assertion logic though, since I totally agree it causes issues with readability.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -11,87 +14,376 @@
...
+void tryCatch<0>(const py::module_ &pyModule, std::exception_ptr pexc) {
+  try {
+    std::rethrow_exception(std::move(pexc));
+  } catch (CppExceptionsAndPyClassNames::Exceptions<0> &cppExc) {
+    setPyException(cppExc, pyModule, CppExceptionsAndPyClassNames::kClassNames[0]);
+  }
+}
+
+/// Mapping of Python class name to Python class object.
+using PyObjectByName = std::map<const std::string_view, const py::object>;
```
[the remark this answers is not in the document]
feltech: I tried `unordered_map`, but `std::string_view` wouldn't work as a key.  I assumed there was no `std::hash` overload.  A quick google implies that's not the case and it should be possible, though, so I'll take another look.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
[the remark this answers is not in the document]
feltech: Similar to Tom's comment: https://github.com/OpenAssetIO/OpenAssetIO/pull/1190#discussion_r1389127543 - it's a nice idea, will take a look and see if it's straightforward enough.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
feltech: Could do, but it would be repeated multiple times (once for every call) rather than once in the caller and passed in.

foundrytom:
> As long as calling out to Python from C++ is done transparently via a C++ base class virtual method (i.e. pybind11 trampoline), then exceptions raised in Python will be translated, due to the new/updated `OPENASSETIO_PYBIND11_OVERRIDE*` macros.

Oh nice, I hadn't realised we were already using our custom macros there, so 👍 Does [this](https://github.com/OpenAssetIO/OpenAssetIO/pull/1190#issuecomment-1805633069) need one thought?

feltech:
> Does [this](https://github.com/OpenAssetIO/OpenAssetIO/pull/1190#issuecomment-1805633069) need one thought?

I'm not sure what you mean? Were you meant to link to a different comment?

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
elliotcmorris: Isn't that already the case? Looks like `throwHybridExceptionIfMatches` is only called from inside `convertPyExceptionAndThrow`, and the extraction is being done each time that function is being called anyway.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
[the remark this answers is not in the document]
feltech: `std::decay_t<decltype(std::get<I>(std::declval<Type::Types>()))>` :laughing: - I think the existing version is a bit neater.

foundrytom:
> > Does [this](https://github.com/OpenAssetIO/OpenAssetIO/pull/1190#issuecomment-1805633069) need one thought?
> 
> I'm not sure what you mean? Were you meant to link to a different comment?

Sorry no idea what went on there, the link was meant to be https://github.com/OpenAssetIO/OpenAssetIO/blob/2198d02b8415785effb779ed76eba78ed6362fb9/src/openassetio-python/cmodule/src/managerApi/EntityReferencePagerInterfaceBinding.cpp#L40

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
feltech: But `convertPyExceptionAndThrow` is only called once.

feltech:
> Sorry no idea what went on there, the link was meant to be
> 
> https://github.com/OpenAssetIO/OpenAssetIO/blob/2198d02b8415785effb779ed76eba78ed6362fb9/src/openassetio-python/cmodule/src/managerApi/EntityReferencePagerInterfaceBinding.cpp#L40

Still not sure I follow. `EntityReferencePagerInterfaceBinding.cpp` already used/has been updated to use `OPENASSETIO_*` macros.

src/openassetio-core/include/openassetio/errors/exceptions.hpp
feltech: Made more generic in b1373437

src/openassetio-python/private/include/openassetio/private/python/exceptions.hpp
feltech: Updated docstring as suggested in 960a54e4

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
feltech: Had some fun in f4993c24 - can drop if you think it's excessive (it only helps remove a single case). It does have the neat side-effect of ensuring that the Python class hierarchy mirrors C++.

src/openassetio-python/tests/cmodule/test_errors.py
feltech: Added a docstring in 9103b05c

doc/contributing/CODING_STANDARDS.md
feltech: Yeah good idea. I was thinking we could change the `business-language` check to `disallowed-phrases` and add it there. Could also be useful if/when we do inclusive terminology checks too.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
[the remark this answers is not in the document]
feltech: Simplified as suggested in 13b69491

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -11,87 +14,376 @@
...
+                    const std::string_view pyClassName) {
+  const py::object pyClass = pyModule.attr(pyClassName.data());
+  const py::object pyInstance = pyClass(exception.what());
+  PyErr_SetObject(pyClass.ptr(), pyInstance.ptr());
+}
+
+/**
+ * Set the current Python exception in this thread.
+ *
+ * Overload to handle more complex case of BatchElementException.
```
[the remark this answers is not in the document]
feltech: With the change in https://github.com/OpenAssetIO/OpenAssetIO/pull/1190#discussion_r1389127543 there are only two special cases remaining (`OpenAssetIOException` and `BatchElementException`), and handling them all in the same method could remove a little duplication.

Will have to sketch it out and see if how neat it looks.

src/openassetio-python/cmodule/src/_openassetio.hpp
feltech: Good idea, they are using up a lot of space now. The `_openassetio.hpp` has become a bit of a dumping ground.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
foundrytom: > It does have the neat side-effect of ensuring that the Python class hierarchy mirrors C++.

For me thats a useful benefit, as it makes future maintenance easier and less prone to human error. @elliotcmorris what are your thoughts?

foundrytom:
> > Sorry no idea what went on there, the link was meant to be
> > https://github.com/OpenAssetIO/OpenAssetIO/blob/2198d02b8415785effb779ed76eba78ed6362fb9/src/openassetio-python/cmodule/src/managerApi/EntityReferencePagerInterfaceBinding.cpp#L40
> 
> Still not sure I follow. `EntityReferencePagerInterfaceBinding.cpp` already used/has been updated to use `OPENASSETIO_*` macros.

Sorry I had somehow slipped back to browsing `main` not your tree.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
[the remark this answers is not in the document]
elliotcmorris: Yep, me too :P

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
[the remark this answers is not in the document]
elliotcmorris: Resolving in favour of the other thread.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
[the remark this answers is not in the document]
elliotcmorris: Yeah that's a fair point.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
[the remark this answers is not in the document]
feltech: Updated in a16f9571 - the trick was to get rid of the `const` from the key type.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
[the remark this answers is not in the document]
feltech: `if constexpr`ed a bunch in 8de8f03a.  I'm two minds about `registerPyExceptionClass` - it only saves one line of duplication, at the expense of a biig function (albeit mostly comments), but it does at least keep it all together. What do you think?

src/openassetio-python/cmodule/src/_openassetio.hpp
feltech: Good call. Put in their own header in a8e8db56 - much nicer and YAGNI.

doc/contributing/CODING_STANDARDS.md
feltech: CI job updated in 18647406

Showing that "OAIO" is still caught: https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/6876912503/job/18703496245?pr=1190

Showing that bad pybind11 macros is caught: https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/6876921789/job/18703536758?pr=1190

The ID of the job has changed and its a required check, so we can either change the ID back or update the project settings.

foundrytom: Awesome sauce. The new name is better so we can just update the branch protection requirement.

foundrytom (approved):

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
elliotcmorris: > It does have the neat side-effect of ensuring that the Python class hierarchy mirrors C++.

Suppose that's fair. It's not really that much more complicated either.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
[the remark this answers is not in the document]
elliotcmorris: I like the smaller ones above, I think that's an improvement. But I see what you mean about `registerPyExceptionClass`. I do think the if/else fallthrough logic is easier to understand than specialisation dispatch though, despite being a bit less elegant feeling.

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
elliotcmorris: Resolving this as it's not important, even though i still don't quite follow the argument... probably just text chat fog-of-war.

elliotcmorris (approved):

feltech:
I've squashed down the commits, in case anyone wants to check the commit messages/breakdown makes sense.

I've also added a fixup 7ec0b72d to move the static assertions to their own .cpp file, since I figured they were obscuring the public interface of the header, and pessimised build times by re-running the static assertions in every translation unit that included it.

foundrytom (approved):
Great idea @feltech - like it.

PR #1193, opened 2023-11-10
Ensure non-null `TraitsData` copy-construction

### LINKED ISSUE #1153: `TraitsData::make` copy-constructor factory segfaults if param is `nullptr`
## What

If `TraitsData::make` is given a `TraitsDataConstPtr` that holds a `nullptr`, then it will attempt to dereference it and segfault.

In particular, this affects `Manager.createChildContext` when the parent `Context` has no `locale`.

## ACs

- `TraitsData::make` throws `InputValidation` if the supplied ptr is `null`, binding for cc has `.none(false)` (predicates c++ tests)
- `Context` make initializes `locale` with an empty `TraitsData` is none is supplied

feltech:
## Description

Closes #1153. 

Ensure `TraitsData` copy-construction throws an error, rather than segfaults, if given a null `TraitsDataPtr` to copy from.

Ensure the default no-argument constructor of `Context` creates a blank (rather than null) `locale`, so that copy-construction of the `locale` (e.g. in `createChildContext`) can succeed. Note that direct construction of `Context` is non-standard (`createContext` should be used instead), but is still used e.g. in tests.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,6 +10,12 @@ v1.0.0-beta.x.x
   in Python.
   [#1186](https://github.com/OpenAssetIO/OpenAssetIO/issues/1186)
 
+### Bug fixes
+
+- Throw an exception when attempting to copy-construct a null
+  `TraitsData`.
+  [#1153](https://github.com/OpenAssetIO/OpenAssetIO/issues/1153)
+
 v1.0.0-beta.1.0
 ---------------
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,6 +16,10 @@ v1.0.0-beta.x.x
   `TraitsData`.
   [#1153](https://github.com/OpenAssetIO/OpenAssetIO/issues/1153)
 
+- Modified no-argument constructor of `Context` to create an empty (but
+  non-null) `locale`.
+  [#1153](https://github.com/OpenAssetIO/OpenAssetIO/issues/1153)
+
 v1.0.0-beta.1.0
 ---------------
 
--- a/src/openassetio-python/cmodule/src/ContextBinding.cpp
+++ b/src/openassetio-python/cmodule/src/ContextBinding.cpp
@@ -14,15 +14,26 @@ void registerContext(const py::module& mod) {
   using openassetio::Context;
   using openassetio::ContextPtr;
   using openassetio::managerApi::ManagerStateBasePtr;
-  using openassetio::trait::TraitsDataPtr;
+  using openassetio::trait::TraitsData;
+  using PyRetainingTraitsDataPtr =
+      openassetio::PyRetainingSharedPtr<openassetio::trait::TraitsData>;
   using PyRetainingManagerStateBasePtr =
       openassetio::PyRetainingSharedPtr<openassetio::managerApi::ManagerStateBase>;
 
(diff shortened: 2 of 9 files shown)


RELEASE_NOTES.md
```diff
@@ -1,6 +1,19 @@
 Release Notes
 =============
 
+v1.0.0-beta.1.x
```
foundrytom, line 4: [fix] As were adding breaking changes (probably good habit anyway so we think properly at release time).
```suggestion
v1.0.0-beta.x.x
```

RELEASE_NOTES.md
```diff
@@ -1,6 +1,16 @@
...
+---------------
+
+### Breaking changes
```
foundrytom, line 7: [question] Is this more an improvement as it would segfault before, so not sure if its really "breaking", if you're relying on the segfault then..........

RELEASE_NOTES.md
```diff
@@ -10,6 +10,9 @@ v1.0.0-beta.1.x
   `TraitsData`.
   [#1153](https://github.com/OpenAssetIO/OpenAssetIO/issues/1153)
 
+- Modified no-argument constructor of `Context` to create an empty (but
```
foundrytom, line 13: [consider] I'd also perhaps argue this is an improvement or probably even a bugfix too as it was never meant to null before.

src/openassetio-python/cmodule/src/ContextBinding.cpp
```diff
@@ -14,15 +14,21 @@ void registerContext(const py::module& mod) {
...
   using PyRetainingManagerStateBasePtr =
       openassetio::PyRetainingSharedPtr<openassetio::managerApi::ManagerStateBase>;
 
   py::class_<Context, ContextPtr> context{mod, "Context", py::is_final()};
 
   context
-      .def(py::init(RetainCommonPyArgs::forFn<&Context::make>()),
-           py::arg_v("locale", TraitsDataPtr{}), py::arg_v("managerState", ManagerStateBasePtr{}))
+      .def(py::init(
+               [](PyRetainingTraitsDataPtr locale, PyRetainingManagerStateBasePtr managerState) {
```
foundrytom, line 27: [fix] Can we note why were doing this (its in the commit, but would be good here too)

foundrytom (changes requested):

RELEASE_NOTES.md
feltech: Updated in 1989a8dc

RELEASE_NOTES.md
feltech: As suggested in https://github.com/OpenAssetIO/OpenAssetIO/pull/1193#discussion_r1389264049, changed to "Bug fixes" in 7789fe6b

RELEASE_NOTES.md
feltech: I like bug fix, that seems more appropriate to both changes.  Updated in 7789fe6b

src/openassetio-python/cmodule/src/ContextBinding.cpp
feltech: Added a comment in 5eaa8493

foundrytom (approved):

elliotcmorris (approved):

PR #1197, opened 2023-11-10
Add `__repr__` support to `EntityReference`

### LINKED ISSUE #1186: Add `__repr__` to `EntityReference`
## What

Add `__repr__` support to `EntityReference`

## Why

```
[<openassetio._openassetio.EntityReference object at 0x10e9b69f0>, <openassetio._openassetio.EntityReference object at 0x10e9b69f0>, <openassetio._openassetio.EntityReference
 object at 0x10e9b69f0>]
```

Makes debugging really hard.

## Acceptance Criteria

```
e = EntityReference("aRef")
repr(e)
<openassetio.EntityReference aRef>
```

foundrytom:
We opted not to go for `EntityReference("theStr")` as it gives the impression they may be directly contructed. Outside of tests, this is discouraged, as `manager.createEntityReference` shoud be used instead.

The present implementation avoids quoting the reference and skirts the issue of escaping angle brackes, as this should only be used for debug introspection.

## Note to reviewers

We were discussing the right way to identify the class. The first commit opts for brevity (`<EntityReference ...>`). The other option, which is perhaps better from a broader software ecosystem perspective in [this commit](https://github.com/OpenAssetIO/OpenAssetIO/pull/1197/commits/e76e5a87f5d2290abf1d779848776d5167659092) extends this to `<openassetio.EntityReference ...>`. This could be dropped if desired. Thoughts? It disambiguates when seen in the context of multiple libraries, as is common in a real-world integration.

Closes #1186

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,15 @@
 Release Notes
 =============
 
+v1.0.0-beta.x.x
+---------------
+
+### Improvements
+
+- Added the string representation of an `EntityReference` when `repr`'d
+  in Python.
+  [#1186](https://github.com/OpenAssetIO/OpenAssetIO/issues/1186)
+
 v1.0.0-beta.1.0
 ---------------
 
--- a/src/openassetio-python/cmodule/src/EntityReferenceBinding.cpp
+++ b/src/openassetio-python/cmodule/src/EntityReferenceBinding.cpp
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2013-2022 The Foundry Visionmongers Ltd
+#include <fmt/format.h>
 #include <pybind11/operators.h>
 #include <pybind11/pybind11.h>
 
@@ -14,5 +15,9 @@ void registerEntityReference(const py::module& mod) {
       .def(py::init<openassetio::Str>(), py::arg("entityReferenceString"))
       .def("toString", &EntityReference::toString)
       .def("__str__", &EntityReference::toString)
+      .def("__repr__",
+           [](const EntityReference& self) {
+             return fmt::format("<openassetio.EntityReference {}>", self.toString());
+           })
       .def(py::self == py::self);  // NOLINT(misc-redundant-expression)
 }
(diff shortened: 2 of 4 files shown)


RELEASE_NOTES.md
```diff
@@ -1,6 +1,15 @@
 Release Notes
 =============
 
+v1.0.0-beta.x.x
```
elliotcmorris, line 4: [Question] Did i see somewhere we were doing `beta.1.x` here?

elliotcmorris (approved):
> . This could be dropped if desired. Thoughts? It disambiguates when seen in the context of multiple libraries, as is common in a real-world integration.

I think `openassetio.EntityReference` is a good idea. Counterargument would be if openassetio entity-references were to become the main type of exchange even inside other external systems, but I suspect that wont be the case, and there will be some conversions between `assetManager.EntityReference` to `openassetio.EntityReference`.

RELEASE_NOTES.md
foundrytom: Yeah, but that needed fixing. We dont know whether its a major/minor bump in the beta till release depending on what the changeset is. just one `x` implies its only a minor.

feltech:
> > . This could be dropped if desired. Thoughts? It disambiguates when seen in the context of multiple libraries, as is common in a real-world integration.
> 
> I think `openassetio.EntityReference` is a good idea.

+1

src/openassetio-python/cmodule/src/EntityReferenceBinding.cpp
```diff
@@ -1,18 +1,23 @@
...
+void registerEntityReference(const py::module& mod) {
   using openassetio::EntityReference;
 
   py::class_<EntityReference>{mod, "EntityReference", py::is_final()}
       .def(py::init<openassetio::Str>(), py::arg("entityReferenceString"))
       .def("toString", &EntityReference::toString)
       .def("__str__", &EntityReference::toString)
+      .def("__repr__",
+           [](const EntityReference& self) {
+             return fmt::format("<openassetio.EntityReference {}>", self.toString());
```
feltech, line 20: [consider][minor] Put the string in quotes? Don't mind whether single or double

feltech (approved):

src/openassetio-python/cmodule/src/EntityReferenceBinding.cpp
foundrytom: Had pondered that - but then do we need to escape any of which ever character we pick in the presentation?  Given its only for debug - opted to dodge that entire issue for now (as technically we should maybe be escaping `>` too?). Seem reasonable?

Issue #1209, opened 2023-11-16
Add `kManagerDriven` and `kRequired` to `PolicyAccess`


foundrytom:
(Capturing some nascent rumination on issues that came up working on #31) 

## What

- Add `kManagerDriven` and `kRequired` to `PolicyAccess` for when a Host wishes to determine which traits a Manager requires for publishing or can provide for a future entity.
- Remove `ResolvesFutureEntitiesTrait` from MediaCreation

Such that, from the Hosts perspective, `PolicyAccess` cases are defined as follows for _entity_ trait sets:

- `kRead`: What is your policy for existing entities with the supplied trait set, and which traits are you capable of resolving data for.
- `kWrite`: What is your policy when I wish to publish entities with the supplied trait set, that I already hold the data for, and which traits are you capable of storing data for?
- `kManagerDriven`: What is your policy when I wish to publish entities with the supplied trait set, and which of the traits are you capable of providing data for (via resolve) for the yet-to-be-published entity. 
- `kRequired`: What is your policy when I wish to publish entities with the supplied trait set, and which of the traits are required.
- `kCreateRelated`: As per `kWrite` (included for parity with `PublishAccess`.

## Why

Managers may required the data for certain properties to be present in order to publish an entity.  The core API also defines the `prefilght` -> `resolve` -> `register` workflow for when a Host wishes or requires to derive some of an entities properties from the Manager during publish. For example - the location to store a render, or what frame range to export.

There is however, no way for a Host to query ahead of time which traits the Manager requires or may be able to resolve for future entities. This means that the Host has no way to present a coherent user experience without "trying first". The only indication that a Manager is even capable of resolving future data is via the tenuous [`ResolvesFutureEntities`](https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/blob/4f73fb6aa062d5807c4eb8010ff792da391b7d4b/traits.yml#L221C7-L221C29) trait from the optional and distinct MediaCreation library, which provides no information about which specific traits are derivable.

We need a solution within the core API, given these are are core workflows.

Using `managementPolicy` to provide this data has the benefit that as the method does not take an Entity Reference, it is a broad application-level query which allows Hosts to preemptively manage user experience if they are not able to provide data a manager needs or the manager isn't able to provide data the host needs. The resulting low call count hopefully offsets that several calls may be needed to define final behaviour.

[...]

PR #1211, opened 2023-11-21
[Docs] Remove `ResolvesFutureEntities` use in examples

### LINKED ISSUE #1210: Remove `ResolvesFutureEntities` from publishing examples
> 🤔 mystery option C - how about something like
> 
> ```python
>  # We then check if the manager can tell us where to save the file.
> if manager.hasCapability(kResolution):
> 
>     working_data = manager.resolve(
>            working_ref, TextFileSpecification.kTraitSet,
>            ResolveAccess.kWrite, context)
> ```

With a note that data will be blank if the manager doesn't support it?  Though the manager could give an `EntityAccessError`...

_Originally posted by @feltech in https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/issues/68#issuecomment-1820713733_

foundrytom:
This has been removed upstream in MediaCreation pending #1209, which will provide a first-class mechanism for managing resolving future entity properties.

This change hopefully keepps the example essentially "legal" in the mean time, albeit cumbersome.

Closes #1210

### DIFF
--- a/doc/doxygen/src/Examples.dox
+++ b/doc/doxygen/src/Examples.dox
@@ -284,9 +284,7 @@
  * from openassetio.access import (
  *     PolicyAccess, ResolveAccess, PublishingAccess)
  * # Commented imports are illustrative and may not exist yet
- * from openassetio_mediacreation.traits.managementPolicy import (
- *     ManagedTrait, ResolvesFutureEntitiesTrait
- * )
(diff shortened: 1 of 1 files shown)


elliotcmorris (approved):

feltech (approved):
If it wasn't for the `EntityAccessError`, this would be a pretty neat workflow - i.e. `resolve` and branch on whether the trait/property exists in the result.

PR #1212, opened 2023-11-28
[CI] Remove mediacreation workaround

elliotcmorris:
Part of https://github.com/OpenAssetIO/OpenAssetIO-Manager-BAL/issues/72

Can remove this dependency workaround now that BAL is sorted properly.
I think this is the only case of installing BAL dependencies in CI.

This should be merged after https://github.com/OpenAssetIO/OpenAssetIO-Manager-BAL/pull/91, which is why this is a draft.

### DIFF
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -70,10 +70,6 @@ jobs:
       - name: Test BAL
         run: |
           python -m pip install external/BAL
-          # We can't install BAL's requirements.txt as this includes
-          # openassetio... we really need to sort
-          # https://github.com/OpenAssetIO/OpenAssetIO-Manager-BAL/issues/72
-          python -m pip install openassetio-mediacreation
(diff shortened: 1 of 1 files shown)


.github/workflows/integrations.yml
```diff
@@ -70,10 +70,6 @@ jobs:
       - name: Test BAL
         run: |
           python -m pip install external/BAL
```
feltech, line 72: [consider] Combine `pip install`s into a single line (pretty sure that'll work) and do a `grep` test to ensure `openassetio`/`openassetio-mediacreation` is definitely not overwritten?

Actually, that might be complicated by Windows unless/until we use `bash` throughout, i.e. #1114

feltech (commented):
Reserving judgement until we can see the final CI logs.

.github/workflows/integrations.yml
foundrytom: Do we care if `media-creation` is?

PR #1214, opened 2023-11-29
Add `entityTraits`

### LINKED ISSUE #31: Add `Manager.entityTraits` method
# What

Add a method to `ManagerInterface` that allows introspection of the notional "type" of the entity referenced by any given entity reference. ie the trait set that describes - or effectively constraints - the qualities it possess or must posess, regardless of what property values the manager supports for `resolve` or `register`.

# Why

In order to support generic import style workflows (eg: retrieving an entity reference from a drop event), it is necessary for a host to determine the characteristics of an entity. 

This method allows the appropriate reader/loader/etc. to be created. At present, it is not possible to determine an entities traits from an entity ref.

When publishing it also allows introspection of the expected traits so behaviour can be adapted to ensure successful operation.

Unlike `managementPolicy` the resulting trait set should contain all known traits that represent qualities of the entity in question (or future entity), it should not omit any traits that have no properties, or the manager is unable to provide data for.

Passing the result of this through `managmentPolicy` would tell you which traits a manager may be able to populate/persist.

## ACs

- `ManagerInterface::entityTraits(const EntityReferences& entityReferences, TraitsAccess access, const ContextPtr& context, <callbacks/session>);`
  - `TraitsAccess` would habe `kRead` and `kWrite`
  - `using EntityTraitsSuccessCallback = function::<void(size_t, TraitSet)>;`
  - empty trait set when not known/applicable (not error as its a legitimate case, eg new asset with no type constraints)
  - error when :
     - kRead: entity doesn't exist
     - kWrite: entity is read-only or publishing is not allowed to that reference
     - all: malformed reference
  - `apiCompliance` checks for edge cases.
  - Simple example of drag-drop/paste of entity reference (ie via mime-drop or clipboard read) in `Examples.dox`
   
- Required as part of the `kEntityTraitIntrospection` interface-only capability. Validated along with the other required interface traits.

## Notes

First task is to write/review the docs to check were happy with the final definition before implementation.

feltech:
## Description

Closes #31. Add the `entityTraits` method.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,12 @@ v1.0.0-beta.x.x
 
 ### Breaking changes
 
+- Added new required core API method
+  `Manager`/`ManagerInterface.entityTraits`, which can be used to
+  determine the traits that an existing entity has, or that a new entity
+  must have, for a given entity reference.
+  [#31](https://github.com/OpenAssetIO/OpenAssetIO/issues/31)
+
 - Access policy keyword argument names in the Python bindings for
   `ManagerInterface` has changed  to `"resolveAccess"` for `resolve` and
   `"policyAccess"` for `managementPolicy`, in order to conform to the
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,13 @@ Release Notes
 v1.0.0-beta.x.x
 ---------------
 
+### Breaking changes
+
+- Access policy keyword argument names in the Python bindings for
+  `ManagerInterface` has changed  to `"resolveAccess"` for `resolve` and
+  `"policyAccess"` for `managementPolicy`, in order to conform to the
+  corresponding C++ argument names.
+
 ### New Features
 
 - Propagate `OpenAssetIOException`-derived Python exceptions as a
--- a/doc/doxygen/src/Examples.dox
+++ b/doc/doxygen/src/Examples.dox
@@ -174,6 +174,60 @@
  * url = LocatableContentTrait(resolved_asset).getLocation()  # May be None
  * @endcode
  *
+ * @section examples_inspecting_a_reference Inspecting a reference
+ *
+ * This example shows how an entity reference of unknown provenance can
+ * be inspected to determine the qualities of the entity it points to.
+ *
+ * Let's say we're an audio application written in Qt, and someone has
+ * just dragged and dropped a line of text into the application. We want
+ * to know if the text is an entity reference, and if so, does it point
+ * to an audio clip.
+ *
+ * @code{.py}
+ * from openassetio.access import EntityTraitsAccess
+ * from openassetio_mediacreation.traits.content import LocatableContentTrait
+ * from openassetio_mediacreation.specifications.audio import SampledAudioResource
+ *
+ * class MyAudioApp(QMainWindow):
+ *
+ *   def __init__(self, manager):
+ *     super().__init__()
+ *     self.__manager = manager
+ *     self.__context = manager.createContext()
+ *     self.setAcceptDrops(True)
+ *     # [... Other Qt setup]
+ *
+ *   def dragEnterEvent(self, event):
+ *     if event.mimeData().hasUrls():
+ *       event.acceptProposedAction()
+ *
+ *   def dropEvent(self, event):
+ *     if not event.mimeData().hasUrls():
+ *       return
+ *     uri = event.mimeData().urls().first().toString()
+ *     ref = self.__manager.createEntityReferenceIfValid(uri)
+ *     if ref is None:
+ *       return
+ *
+ *     entity_trait_set = self.__manager.entityTraits(
+ *       ref, EntityTraitsAccess.kRead, self.__context)
+ *
+ *     # Ignore this drop if the entity is not of the correct type for
+ *     # this application.
+ *     if not SampledAudioResource.kTraitSet <= entity_trait_set
+ *       return
+ *
+ *     resolved_asset = manager.resolve(
+ *        entity_reference, {LocatableContentTrait.kId},
+ *        ResolveAccess.kRead, context)
+ *     url = LocatableContentTrait(resolved_asset).getLocation()
+ *
+ *     self.load_audio_from_url(url)
+ *
+ *  # [... Other application setup]
+ * @endcode
+ *
  * @section example_discovering_capability Discovering capability
  *
  * In the previous example, we saw how to resolve a specific trait for
--- a/src/openassetio-core/include/openassetio/access.hpp
+++ b/src/openassetio-core/include/openassetio/access.hpp
@@ -74,6 +74,27 @@ enum class ResolveAccess : std::underlying_type_t<internal::access::Access> {
 };
 
 /**
+ * Access pattern for entity trait set queries.
+ */
+enum class EntityTraitsAccess : std::underlying_type_t<internal::access::Access> {
+  /**
+   * Used to query the full trait set of an existing entity.
+   *
+   * For example, when an entity is known to exist, but is of unknown
+   * classification.
+   */
+  kRead = internal::access::Access::kRead,
+  /**
+   * Used to query the minimal trait set that must be specified when
+   * publishing to a particular entity reference.
+   *
+   * For example, when validating that a user-supplied entity reference
+   * is appropriate for a publishing operation.
+   */
+  kWrite = internal::access::Access::kWrite,
+};
+
+/**
  * Access pattern for @ref publish "publishing".
  */
 enum class PublishingAccess : std::underlying_type_t<internal::access::Access> {
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -1741,10 +1741,6 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * `ShotSpecification` entity to without error. The resulting entity
    * reference should then reference the newly created Shot.
    *
-   * @note All supplied trait::TraitsDatas should have the same trait
-   * sets. If you wish to register different "types" of entity, they
-   * need to be registered in separate calls.
-   *
    * @warning When registering traits that contain URLs or file paths
    * (for example the MediaCreation LocatableContent trait), it should
    * never be assumed that the resulting @ref entity_reference will
--- a/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/Manager.hpp
@@ -390,6 +390,10 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    * know in advance if you can expect the configured manager to be able
    * to provide data you may require.
    *
+   * This method gives the global policy for how the manager wishes to
+   * interact with certain categories of entity. See @ref entityTraits
+   * for entity-specific introspection.
+   *
    * @note Because traits are specific to any given application of the
    * API, please refer to the documentation for any relevant companion
    * project(s) that provide traits and specifications for your specific
@@ -609,11 +613,10 @@ class OPENASSETIO_CORE_EXPORT Manager final {
    */
 
   /**
-   * @name Entity Reference inspection
+   * @name Entity Reference Inspection
    *
-   * Because of the nature of an @ref entity_reference, it is often
-   * necessary to determine if some working string is actually an @ref
-   * entity_reference or not, to ensure it is handled correctly.
+   * Functionality for validating and creating entity references, and
+   * the existence or kind of entity that they point to.
    *
    * @{
    */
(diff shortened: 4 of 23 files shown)


src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -938,6 +937,77 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+   * Callback signature used for a successful entity trait set query.
+   */
+  using EntityTraitsSuccessCallback = std::function<void(std::size_t, trait::TraitSet)>;
+
+  /**
+   * Provides the host with the @ref trait_set of one or more @ref
+   * entity "entities".
+   *
+   * For example, a host may use this to validate that a user-provided
+   * entity reference is appropriate for an operation.
```
elliotcmorris, line 950: [Consider][minor] "For example" this early on soft-implies to my mind that this is _the_ use case for this.

```suggestion
   * An example use case of this could be a host validating that a user-provided
   * entity reference is appropriate for an operation.
```

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -938,6 +937,77 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+   * If @ref access.EntityTraitsAccess.kWrite "kWrite" is given, respond
+   * with the minimal trait set required to publish to the entity
+   * reference. If an entity is read-only, then call the error callback
+   * using the @ref
+   * errors.BatchElementError.ErrorCode.kEntityAccessError
+   * "kEntityAccessError" code.
+   *
+   * The trait set should include all traits that an entity possesses,
+   * not just those with `resolve`able properties, so that the kind of
+   * entity can be determined by the host.
```
elliotcmorris, line 971: [Consider] Reads a bit strange coming out of the write section where we say return a minimal traitset. Assuming you're trying to make this point about both read and write, perhaps?

```suggestion
   * When populating the `entityTraits` return, you should include all
   * relevant traits for the access mode, not just those with `resolve`able properties.
```
The "so that the kind of entity can be determined by the host" seems only correct when applied to read, i'd just omit it.

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
```diff
@@ -119,6 +119,16 @@ struct PyManagerInterface : ManagerInterface {
                                   errorCallback);
   }
 
+  void entityTraits(const EntityReferences& entityReferences,
+                    const access::EntityTraitsAccess resolveAccess, const ContextConstPtr& context,
```
elliotcmorris, line 123: [fix] Presumed copy/paste typo

```suggestion
                    const access::EntityTraitsAccess entityTraitsAccess, const ContextConstPtr& context,
```

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
```diff
@@ -247,6 +257,10 @@ void registerManagerInterface(const py::module& mod) {
            py::arg("context").none(false), py::arg("hostSession").none(false),
            py::arg("successCallback"), py::arg("errorCallback"),
            py::call_guard<py::gil_scoped_release>{})
+      .def("entityTraits", &ManagerInterface::entityTraits, py::arg("entityReferences"),
+           py::arg("access"), py::arg("context").none(false), py::arg("hostSession").none(false),
```
elliotcmorris, line 261: [Question] I see that it's consistent, but why are the accesses not named their specific names rather than just `access` here?

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -754,6 +753,66 @@ class OPENASSETIO_CORE_EXPORT Manager final {
...
+   *
+   * If @ref access.EntityTraitsAccess.kWrite "kWrite" is given, the
+   * response will be the minimal trait set required to publish to the
+   * entity reference. If an entity is read-only, the error callback
+   * will be invoked using the @ref errors.BatchElementError.ErrorCode.kEntityAccessError
+   * "kEntityAccessError" code.
+   *
+   * The trait set will include all traits that an entity possesses, not
+   * just those with `resolve`able properties, so that they can be used
+   * to determine the kind of entity.
```
elliotcmorris, line 786: [Consider] This seems irrelevent to the host, I don't know why they would presume otherwise.

src/openassetio-python/tests/cmodule/gil/test_manager_gil.py
```diff
@@ -56,8 +56,8 @@ def test_all_methods_covered(self, find_unimplemented_test_cases):
             for method in unimplemented:
                 print(
                     f"""
-    def test_{method}(self, a_threaded_manager, a_context, a_host_session):
-        a_threaded_manager.{method}(a_context, a_host_session)
```
elliotcmorris, line 60: [Question] Why the removal of `host_session`. This is searching for unimplemented manager methods isn't it, shouldn't `host_session` remain?

doc/doxygen/src/Examples.dox
```diff
@@ -174,6 +174,57 @@
...
+ *      ref = self.__manager.createEntityReferenceIfValid(event.mimeData().text())
+ *      if ref is None:
+ *        return
+ *
+ *      entity_trait_set = self.__manager.entityTraits(
+ *        ref, EntityTraitsAccess.kRead, self.__context)
+ *
+ *      # Ignore this drop if the entity is not of the correct type for
+ *      # this application.
+ *      if not SampledAudioResource.kTraitSet <= entity_trait_set
```
elliotcmorris, line 215: [Comment] Makes me want a `SampledAudioResource.FulfilledBy(entity_trait_set)`

RELEASE_NOTES.md
```diff
@@ -4,6 +4,11 @@ Release Notes
 v1.0.0-beta.x.x
 ---------------
 
+### Breaking changes
+
+- Added new required core API method `entityTraits`.
```
foundrytom, line 9: [fix] Can we elaborate a little here - eg that its a manager method, and whats it's for.

doc/doxygen/src/Examples.dox
```diff
@@ -174,6 +174,57 @@
...
+ *
+ *   def __init__(self, manager):
+ *     super().__init__()
+ *     self.__manager = manager
+ *     self.__context = manager.createContext()
+ *     self.setAcceptDrops(True)
+ *     # [... Other Qt setup]
+ *
+ *   def dragEnterEvent(self, event):
+ *    if event.mimeData().hasFormat("text/plain"):
```
foundrytom, line 202: [consider] `text/uri-list` seems to [be a thing](https://www.iana.org/assignments/media-types/text/uri-list).

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
foundrytom: Good spot - seems like this slipped through the net with resolve. We generally are supposed to match the c++ sig arg names I think.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
foundrytom: I think this is quite important, as a host must not assume that a trait here means the data is resolvable or publishable (I probably would). Be good to add a note about using `managementPolicy` to determine which of these traits the manager may be able to resolve/persist data for.

Maybe invert the phrasing? (The trait set will include all traits that an entity possesses, note that not all traits will be resolvable or persistable by the manager....)

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
```diff
@@ -754,6 +753,66 @@ class OPENASSETIO_CORE_EXPORT Manager final {
...
+   *
+   * An empty trait set is a valid response, for example if the entity
+   * is a new asset with no type constraints.
```
foundrytom, line 789: [consider] Here, and on the other side, moving to be with the `kWrite` text as I think this only applies there, where as the previous para is for both read/write.

doc/doxygen/src/Examples.dox
```diff
@@ -174,6 +174,57 @@
  * url = LocatableContentTrait(resolved_asset).getLocation()  # May be None
  * @endcode
  *
+ * @section examples_inspecting_a_reference Inspecting a reference
```
foundrytom, line 177: [plan] I know we didn't put this in the ACs, but would be nice to extend the publishing/resolve methods docs/docs to indicate that `entityTraits` exists and can be of assistance. Perhaps the glossary entry for trait sets too.

foundrytom (changes requested):
Nice one!

src/openassetio-core/include/openassetio/access.hpp
```diff
@@ -73,6 +73,29 @@ enum class ResolveAccess : std::underlying_type_t<internal::access::Access> {
...
+/**
+ * Access pattern for entity trait set queries.
+ */
+enum class EntityTraitsAccess : std::underlying_type_t<internal::access::Access> {
+  /**
+   * Used to query the full trait set of an existing entity.
+   *
+   * For example, when an entity reference to read from is blindly
+   * dropped into an application, this can be used to determine the type
+   * of widget(s) that could/should be created to resolve it.
```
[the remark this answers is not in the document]
feltech: Hmm, how about a bit more generic "this can be used to trigger the appropriate code path"?

src/openassetio-python/cmodule/src/managerApi/ManagerInterfaceBinding.cpp
feltech: Yeah good spot, will ninja a prior commit to address this for the other methods, and then fixup `entityTraits`

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
foundrytom: Whatever phrasing you think works is cool - just feel its important to state that as well as some of those traits not being resolvable by a manager, just because a trait is in `entityTraits` for `kWrite` it doesn't mean the manager is capable of storing that traits properties during publish. I dont think that is obvious to a new-comer without already knowing about management policy.  The docs should try to bi-directionally link so which ever you come across first exposes the whole concept.

src/openassetio-python/tests/cmodule/gil/test_manager_gil.py
feltech: It was a copy-paste error when it was introduced - no `Manager` method takes a `HostSession` instance.  I must have copy-pasted it from `ManagerInterface`.

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
foundrytom: > there's just no properties to fill in so it's pointless.

Btw, Im explicitly talking about the fact that a trait _with_ properties may not be resolvable by any given manager, or that manager storing those properties during publish, _regardless_ of whether that trait is required to characterize the entity.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
foundrytom: fwiw, I liked it.

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
```diff
@@ -347,7 +347,23 @@ class OPENASSETIO_CORE_EXPORT ManagerInterface {
...
+    kDefaultEntityReferences = internal::capability::manager::Capability::kDefaultEntityReferences,
+
+    /**
+     * Manager can be queried for the @ref trait "traits" of a given
+     * entity.
+     *
+     * @warning Support for this capability is required by all managers.
+     * In situations where plugins are implemented as multiple component
+     * plugins (e.g. Python and C++) at least one of the component
+     * plugins must implement this capability.
```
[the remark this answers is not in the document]
feltech: Moved in bf8e8f29

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: Did as suggested in d4ec12ca, but then paired with Tom on a rewrite in 188db754 and 1dfd5f05

src/openassetio-core/include/openassetio/hostApi/Manager.hpp
feltech: Initial rewrite in b095cd05, then paired with Tom on further updates in 188db754 and 1dfd5f05

doc/doxygen/src/Examples.dox
feltech: Added some cross-referencing in 5c975239 and 8837daae and 1632609c

feltech:
Minor unrelated docs fixes in 59e2f7a6 and 1cdd32f4

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
feltech: I'm struggling to improve on what's there, and since this is only a minor consideration, and Tom likes it, I'm going to leave as-is.

src/openassetio-core/include/openassetio/access.hpp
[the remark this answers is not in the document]
feltech: Made the examples a bit more generic in cb597487

doc/doxygen/src/Examples.dox
foundrytom: Nice, thanks.

foundrytom (approved):
Looking great

src/openassetio-core/include/openassetio/managerApi/ManagerInterface.hpp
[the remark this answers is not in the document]
elliotcmorris: > Yeah that is better, will do, but defining enums out of order just feels wrong.

I know what you mean. Ty for doing it.

src/openassetio-python/tests/cmodule/gil/test_manager_gil.py
elliotcmorris: Right, i see. Thing about `Manager` and `ManagerInterface` differently from the external perspective still bites me.

elliotcmorris (approved):

Issue #1043, opened 2023-08-14
Review publishing workflow docs


foundrytom:
## What

There have been numerous changes to the publishing workflow of late. Review the documentation at a holistic level to check we haven't missed any key cross-references/updates.
 
## Why

We need a joined-up story!

feltech:
Note that the glossary entry for `preflight` could be misinterpreted to mean that `preflight` doesn't need to be called if the `ResolvesFutureEntities` trait is not imbued in the `managementPolicy`:

> If a manager has said it is capable of resolving future entities ahead of time, the host must call the [preflight](https://openassetio.github.io/OpenAssetIO/classopenassetio_1_1v1_1_1host_api_1_1_manager.html#a5310cd7307fbdf5cb15998129a6f3336) method and resolve the returned [Entity Reference](https://openassetio.github.io/OpenAssetIO/glossary.html#entity_reference) to determine the manager's chosen location for the data and use this for any write operations.

i.e. if you assume the converse is also true, then this could imply that if the manager has said it is _not_ capable of resolving future entities ahead of time, the host _doesn't need_ to call the preflight method.  This is not the case.

PR #1224, opened 2023-12-11
WIP [Docs] Revise publishing docs

foundrytom:
Attempting to capture the main sentiment behind the OpenAssetIO approach to publishing as part of #1043.

Ran out of time so this is in need of a good edit and links fixing etc. The idea would be then to expand upon this in each of the respective main doc/method sections with more details from the host/manager perspectives.

### DIFF
--- a/doc/doxygen/src/Glossary.dox
+++ b/doc/doxygen/src/Glossary.dox
@@ -453,33 +453,110 @@
  *
  * @section publish publish
  *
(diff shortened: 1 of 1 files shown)


elliotcmorris (approved):
I've made some changes to this, proof stuff and fixing some links.

To be really fastidious, i'd request another review, especially given that Tom and I paired on this, but given the circumstances with the christmas break and all, I'm just gonna merge this.

PR #1228, opened 2023-12-14
[CI] Limit cmake version in wheel builds to < 3.28

elliotcmorris:
Causing CI to fail when building wheels, because wheel builds request the latest cmake version via pip.
Looks to be due to a genuine cmake bug,
https://discourse.cmake.org/t/3-28-segmentation-fault-on-macos-11-runner/9588

This is temporary, we should revert this once cmake has shipped a patch.

### DIFF
--- a/src/openassetio-python/pyproject.toml
+++ b/src/openassetio-python/pyproject.toml
@@ -1,10 +1,12 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2021-2022 The Foundry Visionmongers Ltd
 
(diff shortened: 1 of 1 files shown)


SamCrooksFoundry (approved):
I'm sure the patch won't take long to ship out, we should check back in Jan

elliotcmorris:
Logged an issue as an odd-job and put it into the next iteration, should remind us when we look at the board : https://github.com/OpenAssetIO/OpenAssetIO/issues/1229

PR #1230, opened 2023-12-18
Revert Cmake segfault workaround.

elliotcmorris:
Cmake has shipped a patch, fixing the segfault on macos devices, revert the workaround.


SamCrooksFoundry (approved):
Thanks for keeping an eye on this Elliot

PR #1234, opened 2024-01-02
[Python] Fix HybridException base class RTTI

### LINKED ISSUE #1225: Hybrid exception cannot be caught as `std::exception`
## What
This was noticed in the usdResolver, but I think it's an OpenAssetIO problem so logging here.
We have code in the usd resolver tests that triggers a python exception, and attempts to verify the error message, catching by `std::exception&`. https://github.com/OpenAssetIO/usdOpenAssetIOResolver/blob/de7b465d0af08712bc34002d45746484c40505d1/src/resolver.cpp#L168

However, what happens is that we end up catching in the Wildcard exception handler, and logging "unknown non-exception type caught"
(CI logs here, although it dosen't show the full output : https://github.com/elliotcmorris/usdOpenAssetIOResolver/actions/runs/7062456693/job/19226520598)

## Why
Unknown, very likely an RTTI issue across exception boundrys., although I wouldn't rule out a diamond inheritance problem either.  (See here : https://godbolt.org/z/e6PE5b9n4, it isn't exactly what we're seeing in the usd example, but it does show our exception heirarchy demonstrating catch behaviour strangeness.)

Another thought is a standard library incompatibility between clang 14.3 (what our CI builds+runs the usd resolver on) and clang 10.5, (what our CI builds openassetio on,) however, I did test, albeit hastily, running on a 10.5 docker image and it still didn't seem to function.

Information from my experiments:
- The error, which is a MalformedEntityReference error from bal, thrown as a BatchElementException in the openassetio middleware, is catchable as an `OpenAssetIOException` or a `BatchElementException` (or even a HybridException<BatchElementException>, i went to the effort of making that type public to check)
- When caught as one of these exceptions, typeid reports the type correctly as a HybridException<BatchElementException>
- The standard exception types show up in exported symbols, (checked with `nm`) but the `HybridException` template instantiations dont.
- Attempting to explicitly instantiate the HybridException Types in both OpenAssetIO and the calling code itself (using the same hacked version making it public,) had no effect. The idea was to force the template type information to be exported, although I could never manage it.

I'm at a bit of a loss. This isn't a super emergency, as it will only occur from a python manager implementation, using the exceptional conveniences into a C++ host, attempting to catch as `std::exception` or `std::runtime_error`. That being said it's a bit of a footgun. My instinct says this is to do with being a template exception across a module boundry, and we may need to rethink.

What I havn't done but really _should_ have done is removed USD from this whole equation and attempted to catch a hybrid exception in-process in openassetio itself, that would tell us for sure if there's RTTI export nonsense involved or not, and that's probably where i'd pick this investigation up.

feltech:
## Description

Closes #1225. 

The `HybridException` mechanism used multiple inheritance to allow an exception to be caught as either an OpenAssetIO exception or a pybind11 exception, allowing it to propagate from Python through C++ and back out to Python without losing information.

However, this entailed diamond inheritance, where the two branches of the inheritance hierarchy had a root base class of `std::exception`.

This meant that a `HybridException` could not be caught as a `std::exception` due to the ambiguity. Frustratingly, this manifests as if `std::exception` wasn't part of the hierarchy at all, at least on GCC, and the exception would simply fall through the `catch`.

So rework the `HybridException` mechanism to use composition, rather than inheritance. I.e. we compose the Python exception proxy object `pybind11::error_already_set` rather than add it to the base class list.

This means we need additional explicit detection of `HybridException`s when translating exceptions from C++ to Python, to ensure we re-raise the original Python exception, rather than creating a new Python exception and losing e.g. stack trace data.

Luckily, when we `throw` a `pybind11::error_already_set` from within a pybind11 `register_exception_translator` callback, it's handled by pybind11 as we expect and the original Python exception is restored.

- [ ] ~~I have updated the release notes.~~ - since bugfix to unreleased feature

## Test Instructions

This was detected via the unit tests for the `usdOpenAssetIOResolver` project. A fix was applied to that repo, but can be reverted to test this PR in a "real" scenario. However, the unit tests added by this PR are sufficient to illustrate the problem.

### DIFF
--- a/src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
+++ b/src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
@@ -20,6 +20,7 @@ using openassetio::errors::NotImplementedException;
 using openassetio::errors::OpenAssetIOException;
 using openassetio::errors::UnhandledException;
 using openassetio::python::exceptions::CppExceptionsAndPyClassNames;
+using openassetio::python::exceptions::HybridException;
 
 /**
  * @section topy Conversion from C++ exception to Python exception.
@@ -63,6 +64,12 @@ void setPyException(const Exception &exception, const py::module_ &pyModule,
  * given `exception_ptr` is rethrown and propagates up the nested
  * try-catch blocks until it is caught.
  *
+ * If the exception is a `HybridException` type, i.e. was originally a
+ * Python exception that has propagated through C++ and is now
+ * propagating back out to Python, then rethrow the original Python
+ * error. Otherwise, construct a new Python exception from the C++
+ * exception.
+ *
  * @tparam I Index of exception type to catch in
  * CppExceptionsAndPyClassNames list.
  * @param pyModule Python module containing Python exception classes.
@@ -76,6 +83,8 @@ void tryCatch(const py::module_ &pyModule, std::exception_ptr pexc) {
     } else {
       tryCatch<I - 1>(pyModule, std::move(pexc));
     }
+  } catch (const HybridException<CppExceptionsAndPyClassNames::Exceptions<I>> &cppExc) {
+    throw cppExc.originalPyExc;
   } catch (const CppExceptionsAndPyClassNames::Exceptions<I> &cppExc) {
     setPyException(cppExc, pyModule, CppExceptionsAndPyClassNames::kClassNames[I]);
   }
--- a/src/openassetio-python/private/include/openassetio/private/python/exceptions.hpp
+++ b/src/openassetio-python/private/include/openassetio/private/python/exceptions.hpp
@@ -93,19 +93,13 @@ struct CppExceptionsAndPyClassNames {
 /**
  * Hybrid of OpenAssetIO and pybind11 exception class.
  *
- * Multiple inheritance means that in C++ we can `catch` this exception
- * as either `error_already_set` or the given OpenAssetIO exception. If
- * (re)thrown, it can be caught further up the stack (again) as either
- * of the two exceptions. In this way we can satisfy the two use-cases:
- * - Catching the exception in C++ as an OpenAssetIO C++ exception.
- * - Allowing the exception to propagate back to Python via pybind11,
- *   which will translate it back into a Python exception (see
- *   registerExceptions).
+ * This template class inherits from its template argument, allowing it
+ * to be caught transparently as the wrapped C++ exception type.
  *
- * Note that calling `.what()` on this exception is ambiguous and will
- * cause a compiler error. However, this is a good thing, since this
- * exception should never appear in a `catch` (should only be caught as
- * one of the base classes).
+ * The corresponding Python exception (as reflected by a pybind11
+ * `error_already_set` object), is stored for potential recall if this
+ * C++ exception propagates back out to Python - see
+ * `registerExceptions`.
  *
  * Generalisation assuming that the given exception type is a simple
  * exception that takes a string message as its only constructor
@@ -113,11 +107,20 @@ struct CppExceptionsAndPyClassNames {
  *
  * @tparam CppException C++ exception type corresponding to given
  * Python exception.
+ *
+ * @todo Remove PYBIND11_EXPORT once tests that require it are
+ *   redesigned. I.e. _openassetio_test throws HybridException which is
+ *   then caught in _openassetio, requiring global RTTI info for libc++.
+ *   This is not representative of real-world use, since HybridException
+ *   should be private to _openassetio. See
+ *   https://github.com/OpenAssetIO/OpenAssetIO/issues/1237
  */
 template <class CppException>
-struct HybridException : py::error_already_set, CppException {
+struct PYBIND11_EXPORT HybridException : CppException {
   explicit HybridException(const py::error_already_set &pyExc)
-      : py::error_already_set{pyExc}, CppException{pyExc.what()} {}
+      : CppException{pyExc.what()}, originalPyExc{pyExc} {}
+
+  py::error_already_set originalPyExc;
 };
 
 /**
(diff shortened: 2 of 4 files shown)


src/openassetio-python/private/include/openassetio/private/python/exceptions.hpp
```diff
@@ -115,9 +109,11 @@ struct CppExceptionsAndPyClassNames {
  * Python exception.
  */
 template <class CppException>
-struct HybridException : py::error_already_set, CppException {
+struct PYBIND11_EXPORT HybridException : CppException {
```
feltech, line 112: [question] This is just beyond the edge of my knowledge. The `HybridException` is only ever thrown and caught internally within the `_openassetio` dll, so does the template class _really_ need exporting?  

The wrinkle is that an intermediate dll may intercept the `HybridException`, i.e. catch (as the base type) and re-`throw`, to be caught again in `_openassetio` (as a `HybridException`). I have assumed, without much evidence, that this case requires the explicit symbol export.

The practical reason for adding the export here is that the tests fail (on MacOS / libc++).  This is due to the separate `_openassetio_test` Python extension module reusing `HybridException`, which is a special case that will not occur in production, but has some similarities to the case above, i.e. interception by an intermediate dll.

I think a better test would be to properly simulate the intermediate dll `catch`-and-re`throw` case, but would require some time to implement.  Worth adding?

feltech: I'm actually struggling to think of how to "slot in" an arbitrary 3rd-party dll that catches and re-throws the exception.  However, it is possible (even likely) that we'll do this in `libopenassetio-core` as part of logging/auditing.  Below are a couple of sequence diagrams to hopefully illustrate:

Current Python<->Python e2e exception propagation
```mermaid
sequenceDiagram
    host.py->>_openassetio (host): Manager.resolve(...)
    _openassetio (host)->>libopenassetio-core: Manager::resolve(...)
    libopenassetio-core->>_openassetio (manager): ManagerInterface::resolve(...)
    _openassetio (manager)->>manager.py: MangerInterface.resolve(...)
    manager.py-->>_openassetio (manager): throw OpenAssetIOException(...)
    _openassetio (manager)-->>libopenassetio-core: throw HybridException(...)
    libopenassetio-core-->>_openassetio (host): propagate exception
    _openassetio (host)-->>host.py: unpack OpenAssetIOException
```

Potential future addition of exception intercept for intermediate logging and re-throw
```mermaid
sequenceDiagram
    host.py->>_openassetio (host): Manager.resolve(...)
    _openassetio (host)->>libopenassetio-core: Manager::resolve(...)
    libopenassetio-core->>_openassetio (manager): ManagerInterface::resolve(...)
    _openassetio (manager)->>manager.py: MangerInterface.resolve(...)
    manager.py-->>_openassetio (manager): throw OpenAssetIOException(...)
    _openassetio (manager)-->>libopenassetio-core: throw HybridException(...)
    libopenassetio-core->>libopenassetio-core: catch OpenAssetIOException + log + re-throw
    libopenassetio-core-->>_openassetio (host): propagate exception
    _openassetio (host)-->>host.py: unpack OpenAssetIOException
```

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -63,6 +64,12 @@ void setPyException(const Exception &exception, const py::module_ &pyModule,
  * given `exception_ptr` is rethrown and propagates up the nested
  * try-catch blocks until it is caught.
  *
+ * If the exception is a `HybridException` type, i.e. was originally a
+ * Python exception that has propagated through C++ and is now
+ * propagating back out to Python, then rethrow the original Python
+ * error. Otherwise, construct a new Python exception from the C++
+ * exception.
```
elliotcmorris, line 71: [Question] Just so i'm clear. A hybrid exception is _always_ a python exception, in that, it's the c++ representation of an exception that was originally thrown in python?

src/openassetio-python/private/include/openassetio/private/python/exceptions.hpp
elliotcmorris: As testing the below flow indicates things work as expected, ie, the only reason we need to export is because of our unrepresentative test module, I think the best thing to do from this is to leave the export, (with a big comment explaining the situation,) and [plan] to address our python test infrastructure so we can remove it.

src/openassetio-python/tests/cmodule/resources/_openassetio_test/errorsTest.cpp
```diff
@@ -401,6 +408,20 @@ void registerExceptionThrower(py::module_& mod) {
...
+      "throwPythonExceptionAndCatchAsStdException",
+      [](ExceptionThrower& exceptionThrower) {
+        try {
+          exceptionThrower.throwFromOverride();
+        } catch (const std::exception&) {  // NOLINT(*-empty-catch)
+        }
+        try {
+          exceptionThrower.throwFromOverride();
+        } catch (const std::runtime_error&) {  // NOLINT(*-empty-catch)
+        }
```
elliotcmorris, line 421: [Question] Again just to make sure I understand. The idea is that an uncaught exception here fails the test implicitly because it falls through the catch and escapes as an uncaught exception, causing a general error?

Might be an idea to just note that in the catch statements `//uncaught exception causes error` or something to jog the mind.

elliotcmorris (approved):

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
feltech: Yes exactly

src/openassetio-python/private/include/openassetio/private/python/exceptions.hpp
feltech: Added issue:

- https://github.com/OpenAssetIO/OpenAssetIO/issues/1237

feltech: Added an `@todo` with a link to the above issue in b9cbaa9e

src/openassetio-python/tests/cmodule/resources/_openassetio_test/errorsTest.cpp
feltech: Added a comment in 743e0b0b

src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
elliotcmorris: Hmm, i wonder if it might be better named, like, `PythonExceptionCppWrapper` or something. I was under the impression until now that it could be either/or.

One for later anyhow.

src/openassetio-python/private/include/openassetio/private/python/exceptions.hpp
elliotcmorris: Perfect

elliotcmorris (approved):

Issue #812, opened 2023-02-10
Upgrade to Ubuntu 22.04 in CI and use newer clang linter tools


feltech:
## What

Upgrade to Ubuntu 22.04 in CI and use the more recent `clang-tidy` and `clang-format` (v14), available in the default `apt` repo for 22.04.

## Why

Currently our CI uses Ubuntu 20.04 for C++ linter checks. This means we're limited to `clang-tidy`/`clang-format` v12 in the default `apt` repo.

`clang-tidy` 14 has some useful extra checks, e.g. `misc-const-correctness`, which will flag when a variable can be `const`.  `clang-format` 14 has additional formatting fixits, e.g. `QualifierAlignment`, which will place `const` on the appropriate side of a variable.

Useful links: 
* `clang-tidy` 12: https://releases.llvm.org/12.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/list.html 
* `clang-format` 12: https://releases.llvm.org/12.0.0/tools/clang/docs/ClangFormatStyleOptions.html
* (can change the `12` in the URLs to a different version)

## Acceptance Criteria

* CI updated to use 22.04 runners.
* CI linter bootstrap to install latest available Clang tooling (14 according to [the repo website](https://packages.ubuntu.com/jammy/clang-tidy)).
* OpenAssetIO (and our other related projects) CMake config updated to prefer `clang-tidy-14` and  fixed to conform to the new checks, with appropriate config customisation, as discovered.

feltech:
I also found that our `.clang-tidy`'s use of `-*` disables some useful checks from `clang-analyzer`/`-diagnostic`, e.g. unused variables. We may want to explicitly re-enable `clang-*` after the `-*` (and otherwise investigate what other checks we're losing)

feltech:
I compiled the current codebase with clang-tidy v17 and get 2140 warnings(!) - almost all of them are the new IWYU-like check `misc-include-cleaner`.

The `misc-include-cleaner` is rather strict - it doesn't allow transitive headers from the associated header of a cpp (i.e. headers included by `file.hpp` aren't considered for `file.cpp`).  Apparently [that's deliberate](https://github.com/llvm/llvm-project/issues/67140), since it makes automated refactoring harder.

Without `misc-include-cleaner` I get a lot of warnings about the CMake-generated `export.h`, so we'll need to exclude that.

[...]

PR #1253, opened 2024-01-25
Redesign Python test-specific bindings

### LINKED ISSUE #1237: Redesign `_openassetio_test` tests
## What

Change `_openassetio_test`-based tests to be more representative of production use cases, in particular with respect to cross-dll interactions, where `_openassetio` is an isolated dll in production that doesn't share its internals with other dlls.

## Why

Following on from https://github.com/OpenAssetIO/OpenAssetIO/pull/1234#discussion_r1440435303

The `_openassetio_test` Python extension module is used to exercise low-level Python binding code that is generic across all our bindings.  It defines and binds C++ functions and types that are only useful for test purposes and shouldn't be distributed.

Code in the main `_openassetio` Python extension module is designed to be entirely private, with no public headers and no symbols exported. 

However, in order to re-use the same machinery in `_openassetio_test` that is used by `_openassetio`, we have to share code between them. This has led to compromises in project structure, and, most recently, exporting symbols (`HybridException`, see comment thread linked above) that should be hidden. This odd splitting of test functions/types yet sharing of internal machinery also means that tests are less representative of production scenarios.

## Acceptance Criteria

- Test functionality is conditionally appended to `_openassetio` python module if testing is enabled
- `_openassetio_test` is removed
- Tests that depend on `_openassetio_test` are refactored to continue working
- "Test pollution" is kept to a minimum in non-test code.
- Test symbols moved into `_openassetio` are placed in a "_testutils" submodule (We can't just call it "test" because that's what we called api compliance suite.)
- Remove now-unnecessary `PYBIND11_EXPORT` from `HybridException`

feltech:
## Description

Closes #1237.

Removes the separate `_openassetio_test` Python extension module, in favour of a static library providing the test-specific bindings, which is conditionally linked to and called from within the `_openassetio` Python extension module initialisation.

This allowed a little tidying of code that was previously internally-public (public to other OpenAssetIO components, but not to users).

## Test Instructions

Running the usdOpenAssetIOResolver tests against this PR (on Ubuntu) seems to work just fine, but I may be forgetting something?


src/openassetio-python/cmodule/src/_openassetio.cpp
```diff
@@ -3,6 +3,14 @@
 
 #include "_openassetio.hpp"
 
+#ifdef OPENASSETIO_ENABLE_TESTS
+// In order to have representative tests of internal functionality, we
+// must unfortunately pollute this extension module with test-specific
+// bindings. The static library providing the implementation of this
```
elliotcmorris, line 9: [consider][minor] Seems harsh. It's only conditionally compiled in and it serves a just purpose, and you dub it pollution? You are a cruel ruler indeed.

elliotcmorris (approved):
Looks great. Can't believe you got this done so quick, I reckon you should take friday off now.

Q on the clang tidy bits? Are you using a more modern clang-tidy than we specify, giving you extra checks? I'm wondering how code that didn't conform got in otherwise.

feltech:
> Q on the clang tidy bits? Are you using a more modern clang-tidy than we specify, giving you extra checks? I'm wondering how code that didn't conform got in otherwise.

Yeah, the clang-tidy bundled with CLion is newer than version used in our CI. 

There is precedent for making such changes in #874. Looking through our chat history, I think that was prompted by the fact that MacOS doesn't have a clang version old enough to match CI. 

Perhaps we should prioritise #812 (note the comment that we don't actually have to upgrade Ubuntu, just `pip install` newer tools)

src/openassetio-python/cmodule/src/_openassetio.cpp
feltech: I made it a bit less harsh. I took the liberty of squashing the commits.  The updated text can be found in 7e30300e.

elliotcmorris:
Added https://github.com/OpenAssetIO/OpenAssetIO/issues/812 to the next iteration so we can consider it during planning.

elliotcmorris (approved):

Issue #1203, opened 2023-11-15
Persistent cache mechanism


foundrytom:

foundrytom:
Capturing discussion: 

- Add a `CacheInterfacePtr` (`cache`?) to `ManagerStateBase` that is set by the interface implementation in `createState` and/or `createChildState`
- Managers can then use the interface to the cache to pre-populate results, or serialize etc. to a path, and encode in `persistenceTokenForState`, so the cache could be re-populated and set back in `stateFromPersistenceToken`.
- `Manager.cpp`, will attempt to lookup an API call in the cache if there is one in the context.

foundrytom:
Ask people for input - broad assumption is that the distributed serialized on disk cache is a read-only cache to prevent the explosion that is networked concurrent writes to the same cache.

foundrytom:
The manager is general in charge of which cache is used as it knows about the pipeline (or exposes that configuration)

foundrytom:
Consider making this in the same ilk as the GRPC prototype, so its just another wrapper ManagerInterface + factory so the core API doesn't know about it.

elliotcmorris:
It's painful, but categorising this as non essential as we could release without it, and it's a big thing.
This dosen't mean it's not still on the 1.0.0 roadmap.

PR #1254, opened 2024-01-30
[Doc] Add DR23, Traits and Specs versioning

elliotcmorris:
## Description
Add the decision record for traits and specifications versioning. This document is an attempt to provide a tour through our thoughts, which are currently swirling and incomplete, as is this document.

At a high level, this DR attempts to wrestle with the choice between versioning the Openassetio-mediacreation traits + spec schema as a whole, vs versioning traits and specs individually, whilst exploring the technical/ecosystem implications those choices would have.

Due to the nature of the decision, this is a more speculative DR than most, with the possibility matrix multiplying very fast. We have attempted to make some assumptions that we feel are decisions that can be made separately to this choice, which are described at the beginning of the document.

We acknowledge that we are by no means experienced in this space, and are cognizant of how wrong most of this probably is. We're putting it out for necessary feedback, hopefully to drive discussion at a TSC meeting or otherwise.

### DIFF
--- a/doc/decisions/DR023-Versioning-traits-and-specifications.md
+++ b/doc/decisions/DR023-Versioning-traits-and-specifications.md
@@ -1,0 +1,702 @@
+
+
+# DR023 Versioning traits and specifications
+
+- **Status:** To be decided
+- **Impact:** High
+- **Driver:** @elliotcmorris
+- **Approver:** @feltech
+- **Contributors:** @elliotcmorris @feltech @reinecke @antirotor
+- **Outcome:**  TBC
+
+### Preface
+This document refers to the versioning of trait and specification
+libraries in general. However, as `openassetio-mediacreation` is the
+only one in play at the moment, and since reading "trait and
+specifications library" over and over would be tiresome, we will just
+use the term `mediacreation` throughout this document.
+
+# Problem Statement
+In order to achieve OpenAssetIOs interoperability goal, hosts and
+managers must agree on the structure of data exchanged between them.
+
+The `openassetio-mediacreation` defines the agreed standard for interop
+objects (called `traits` and `specifications`) across the VFX industry.
+Hosts and Managers must agree on the structure of these objects. [These
+objects](https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/blob/main/traits.yml)
+are currently in a first-pass state, and it's understood that the only
+way to come to a solid set of useful shared concepts is via use and
+iteration.
+
+OpenAssetIO wants to be thin, and does not want to be "the problem". To
+this end, we need a change and versioning process for the traits and
+specifications schema which allows maximum compatibility, especially in
+the face of the unpredictable version adoption in the VFX industry.
+
+[`openassetio-mediacreation`](https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation)
+is a generative library, in that it produces trait and specification
+view classes representing the objects specified in the source `.yml` via
+the `openassetio-traitgen` tool, currently in both the c++ and python
+languages. As such, the challenge of versioning is not only descriptive,
+but also technical.
+
+# Aims :
+- Allow new mediacreation versions to be adopted by managers without
+  breaking compatibility with hosts running previous versions.
+- Deliver a solid but minimal versioning foundation, that leaves open
+  opportunities for implementation conveniences at a later date.
+- Allow infinite backwards compatibility for manager implementations.
+- Low cognitive load, the OpenAssetIO development ecosystem cannot
+  support a complicated/technically intricate solution.
+- Minimize burden on manager implementors, whilst keeping in mind the
+  prior aims.
+
+# Simplifying assumptions
+Versioning is a vast topic, and OpenAssetIO has a lot of moving parts.
+For the purposes of not going mad, we will make some tentative
+assumptions to simplify discussions.
+
+## Version Format
+For illustrative purposes, we will presume flat versioning (v1, v2) etc
+for schema versions by default throughout this document. However, we are
+considering semver, and may note some semver specific considerations.
+
+## Backwards Compatability Mechanisms
+
+A necessary goal in versioning a schema is to provide mechanisms to be
+backwards compatible.
+
+There are a few mechanisms to do this, such as.
+- Have the manager bundle all the old versions of `mediacreation` that
+  they wish to support in their package
+  - A convenient alternative to this would be having
+    `openassetio-traitgen` generate all the old versions of the trait
+    and specification view types into the library, under a separate
+    namespace.
+- Provide mechanisms to upgrade and downgrade `mediacreation` types.
+
+Given the effort we have to spend on this, we are going to assume for
+the moment that `openassetio-traitgen` generates older versions of trait
+types into the library, for use by the manager to perform conversions.
+This does not preclude the future addition of automatic
+upgrade/downgrade functionality.
+
+For example:
+```python
+openassetio_mediacreation.legacy.traits.content.LocatableContentTrait_v1
+```
+
+## Who adapts
+We are going to assume that in most cases, the manager will be
+performing trait conversions between versions, with the host defining
+the "working version".
+
+Reasons:
+
+- Hosts update less often, and such are more likely to remain on older
+  versions of `mediacreation` for longer.
+- Managers are likely to be more capable of implementing capable trait
+  conversions between versions.
+- Hosts "start the conversation", and for some methods that means
+  sending a trait or specification along, implying an opinion on
+  version. (Although, hosts _could_ query managers for what version of a
+  trait they should send via a version aware `managementPolicy` or some
+  other mechanism, so this is perhaps a weak argument. That arrangement
+  seem intuitively more complex though)
+- In many cases, manager plugins are privately owned by pipeline
+  authors, who are also in control of when to update the hosts in their
+  pipeline.
+
+## Manager/Host Namespace Collisions
+With the current state of affairs, a large problem with versioning is
+presented when you consider the possibility of namespace collision
+between hosts and managers.
+
+For the purpose of this decision, we will assume that two versions of
+`mediacreation` can exist in the same process without problematic
+namespace collisions, although this remains a problem to be solved.
+
+(Note: Practically, this may mean that we end up strongly recommending
+ vendoring mediacreation into manager plugins, especially for python
+ libraries which can't do C++ tricks with private symbols.)
+
+## Schema version, package version
+
+The schema version (the version number in the `.yml`) and the package
+version (the `openassetio-mediacreation` library version) are separate.
+
+It is possible that updates to `openassetio` or `openassetio-traitgen`
+could cause breakages that necessitate a major version update to the
+`openassetio-mediacreation` package.
+
+We are assuming that this does **not** imply a schema version update.
+
+
+# What to Version
+In considering how to version the openassetio traits and specifications,
+two options immediately present themselves. Version every trait and
+specification individually, or version the entire schema as a whole.
+
+## Version every trait
+Every trait gets its own disconnected version. This version would have
+to travel across the api-boundary along with the trait data. Whilst the obvious
+thing would be to enhance the low level `TraitsData` type to have a
+first class understanding of versioning, this cannot be so because not
+every method takes a `TraitsData` as an argument. The lowest-friction
+course of action might be to pack the version into `TraitId`, essentially
+doing string-munging to communicate the version across the api-boundary.
+
+Hosts and managers would need to be inspect the version data, which we
+could provide conveniences for, and construct the correct version
+of the trait view.
+
+```yml
+# traits.yml
+Camera:
+  description: >
+    The entity represents a device that captures a view through
+    re-projection or encoding into some other form.
+  usage:
+    - entity
+  version: 3
+```
+
+```python
+# Manager
+# In this example, CameraTrait is on version 3
+# SpatialTrait is on version 2
+#
+# The manager is responding to a resolve, needing to populate a
+# TraitsData with a camera trait and a spatial trait of the expected
+# version
+cameraVersion = mediacreation.versionFromTraitId(mediacreation.CameraTrait.kId)
+spatialVersion = mediacreation.versionFromTraitId(mediacreation.SpatialTrait.kId)
+
+# Support all known major versions.
+# Camera
+success_result = TraitsData()
+if (cameraVersion >= 3 and cameraVersion < 4):
+     trait = mediacreation.CameraTrait(success_result)
+     ... # Up to date camera population
+elif (cameraVersion >= 2 and cameraVersion < 1):
+     trait = mediacreation.legacy.CameraTrait_v2(success_result)
+     ... # V2 camera population
+elif (cameraVersion >= 1):
+     trait = mediacreation.legacy.CameraTrait_v1(success_result)
+     ... # V1 camera population
+else:
+  raise RuntimeError()
+
+# Spatial
+if (spatialVersion >= 2 and spatialVersion < 3):
+    trait = mediacreation.SpatialTrait(success_result)
+    ... # Up to date spatial population
+elif (spatialVersion >= 1):
+    trait = mediacreation.legacy.SpatialTrait_v1(success_result)
+     ... # V1 to date spatial population
+else:
+  raise RuntimeError()
+```
+
+### Benefits
+- Maximally flexible, hosts and managers can disregard versioning for
+  traits they don't care about.
+- Obvious from the YAML which version a trait is.
+- Individual traits will only update version when they change, thus they
+  will not have "gaps", they will always go v1->v2->v3.
+- A single trait can be updated in isolation without any effect on other
+  traits in the package.
+- Semver could be used to provide additional runtime compatibility
+  information.
+
+### Drawbacks
+- Version has to be encoded in `traitID` because several methods
+  (including `resolve`) take `traitSet`, which is only a collection of
+  `traitIDs`. There is no `TraitsData` to inspect. This would be a major
+  breaking change as currently conceived.
+- Version support may become fragmented if hosts and managers provide
+  versioned support for only certain traits
+- When packing multiple traits, manager has to branch on each trait as
+  they could be different versions.
+- When extended to specification this _really_ causes problems, as we
+  need to invent a new place to pack specifications version across the
+  wire.
+
+## Version the whole schema.
+This option places a version number is placed at the top of the
+`traits.yml` file or equivalent. This defines the version for all the
+traits within the package.
+
+The manager will still need a mechanism to know which trait-library
+version the host will be expecting. This could be achieved by a
+`HostInterface` method.
+
+```cpp
+//HostInterface implementation
+SchemaVersion HostInterface::schemaVersion(std::string_view library_identifier) {
+  if (library_identifier == "openassetio-mediacreation"){
+    return openassetio_mediacreation::version();
+  }
+}
+```
+### Benefits
+- Version dosen't need to be packed into any `traitId`
+- A traits and specifications library gets a version as a unit, making
+  talking about supported versions easier, and hopefully reducing
+  support fragmentation
+- Manager can branch at a higher level
+
+### Drawbacks
+- Unknown at runtime which traits or specifications have been updated in
+  a version bump.
+- Encourages redundant branching when in many/most cases traits will
+  remain compatible between versions.
+- Doesn't provide a way to enforce strict compatibility on a
+  trait-by-trait basis programmatically, it's all or nothing.
+
+```python
+# The manager is responding to a resolve, needing to populate a
+# TraitsData with a camera trait and a spatial trait of the expected 
+# version
+hostMediaCreationV = hostInterface.schemaVersion("openassetio-mediacreation")
+
+...
+
+success_result = TraitsData()
+match hostMediaCreationV:
+  case 1:
+    camTrait = mediacreation.legacy.CameraTrait_v1(success_result)
+    spatialTrait = mediacreation.legacy.SpatialTrait_v1(success_result)
+    ... # V1 Camera + Spatial population
+  case 2:
+    camTrait = mediacreation.legacy.CameraTrait_v2_(success_result)
+    spatialTrait = mediacreation.legacy.SpatialTrait_v2(success_result)
+    ... # V2 Camera + Spatial population
+  case 3:
+    camTrait = mediacreation.CameraTrait(success_result)
+    spatialTrait = mediacreation.SpatialTrait(success_result)
+    ... # Up to date Camera + Spatial population
+  case _
+    raise RuntimeError()
+```
+
+### Last-Changed
+A concept that kept coming up was the idea of a "last-changed" property
+on the traits, when related to whole schema versioning, which we'll
+just briefly mention.
+
+```yml
+# traits.yml
+schemaVersion: 5
+
+Camera:
+  description: >
+    The entity represents a device that captures a view through
+    re-projection or encoding into some other form.
+  usage:
+    - entity
+  last_changed: 2
+```
+- This would allow managers to support swathes of versions in a single
+  branch, as the traits would remain compatible with eachother across a
+  known set
+- Opens the door up to more easily supporting automatic upgrade flows in the future
+  given an injection of prewritten upgrade/downgrade methods.
+
+# Worked Example
+Let's work an example. Take the ever so popular `LocatableContent`
+trait.
+
+
+```yml
+# Some descriptions omitted for brevity.
+schemaVersion: 1 <------------ NEW
+
+LocatableContent:
+        description: >
+          This trait characterizes an entity whose data is persisted
+          externally to the API through data accessible via a valid
+          URL.
+          The `location` property holds the most applicable location
+          of the entity's content for the current process environment
+          - considering platform, host, etc. Location is in the form
+          of a URL.
+        usage:
+          - entity
+        properties:
+          location:
+            type: string
+            description: ...
+          mimeType:
+            type: string
+            description: ...
+          isTemplated:
+            type: boolean
+            description: ...
+```
+
+Let's work through making a change to alter the name `location` to
+`url`, to better convey that this string should be an url.
+
+We will assume the schema is versioned as a unit for this example, but
+it isn't so hard to imagine how this would look if the versions were
+taken directly from the traits instead.
+
+## Host
+For context, the host code looks like this. The host is using the
+version of `mediacreation` shown above, where `mediacreation` is on
+version `1`. This won't change during this example.
+
+```py
+from openassetio.access import ResolveAccess
+from openassetio_mediacreation.traits.content import LocatableContentTrait
+
+...
+
+resolved_asset = manager.resolve(
+    entity_reference, {LocatableContentTrait.kId},
+    ResolveAccess.kRead, context)
+ url = LocatableContentTrait(resolved_asset).getLocation()  # May be None
+```
+
+## Manager
+The manager needs to resolve this reference, and call the success
+callback with the location.
+
+```python
+from openassetio_mediacreation.traits.content import LocatableContentTrait
+
+def resolve(
+        self,
+        entityReferences,
+        traitSet,
+        resolveAccess,
+        context,
+        hostSession,
+        successCallback,
+        errorCallback,
+    ):
+
+  success_result = TraitsData()
+  trait = LocatableContentTrait(success_result)
+  trait.setLocation(managed_filesystem_locations[ref.toString()])
+  successCallback(idx, success_result)
+```
+
+At this point, `LocatableContent` has never had a version update, so
+there is no need for the manager to do any version handling.
+
+Now, let's make our change.
+```yml
+schemaVersion: 2 <------------ UPDATED
+
+# Some descriptions omitted for brevity.
+LocatableContent:
+        description: >
+          This trait characterizes an entity whose data is persisted
+          externally to the API through data accessible via a valid
+          URL.
+          The `url` property holds the most applicable location
+          of the entity's content for the current process environment
+          - considering platform, host, etc. Location is in the form
+          of a URL.
+        usage:
+          - entity
+        properties:
+          url: <---------------------- CHANGED
+            type: string
+            description: ...
+          mimeType:
+            type: string
+            description: ...
+          isTemplated:
+            type: boolean
+            description: ...
+```
+The manager sees this, and updates its mediacreation version. To support
+hosts that use this new schema version, as well as hosts that havn't
+upgraded yet, the manager now has to branch.
+
+```py
+
+def resolve(
+        self,
+        entityReferences,
+        traitSet,
+        resolveAccess,
+        context,
+        hostSession,
+        successCallback,
+        errorCallback,
+    ):
+
+  hostMediaCreationV = hostInterface.schemaVersion("openassetio-mediacreation")
+
+  success_result = TraitsData()
+  if (hostMediaCreationV == 2):
+    trait = mediacreation.traits.content.LocatableContentTrait(success_result)
+    trait.setUrl(managed_filesystem_locations[ref.toString()])
+    successCallback(idx, success_result)
+  elif (hostMediaCreationV == 1):
+    trait = mediacreation.legacy.traits.content.LocatableContentTrait_v1(success_result)
+    trait.setLocation(managed_filesystem_locations[ref.toString()])
+    successCallback(idx, success_result)
+  else:
+    raise RuntimeError()
+```
+
+
+## What about a non breaking change?
+
+Good question. Up until now we've been demonstrating flat versioning for
+brevity, but let's switch to semver for this example.
+
+(Flat versioning may also support this, either via the OTIO approach of
+just not updating the version for a minor breaking change, or having
+`lastMajorChanged` and `lastMinorChanged` properties on traits.)
+
+Lets take this example. Adding a new property `isLocal` to fulfil the
+imaginary need to know if your content is on your local machine or not.
+Going from where we left off
+```yml
+# Some descriptions omitted for brevity.
+LocatableContent:
+        description: ...
+        usage:
+          - entity
+        properties:
+          url:
+            type: string
+            description: ...
+          mimeType:
+            type: string
+            description: ...
+          isTemplated:
+            type: boolean
+            description: ...
+          isLocal: <------------------ NEW
+            type: boolean
+            description: ...
+        version: "2.1.0" <------------ UPDATED
+```
+
+With this being a non breaking change, the manager shouldn't have to
+update. Imagine another host who is ahead of the manager, and has picked
+up this new version.
+
+```py
+from openassetio.access import ResolveAccess
+from openassetio_mediacreation.traits.content import LocatableContentTrait
+
+...
+
+resolved_asset = manager.resolve(
+    entity_reference, {LocatableContentTrait.kId},
+    ResolveAccess.kRead, context)
+
+isLocal = LocatableContentTrait(resolved_asset).getIsLocal()  # May be None
+```
+Thankfully, we have already accounted for this. Our traitViews
+understand missing properties, and return `None`/`Optional`, so the host
+will know that there is no information on this new property from the
+current version of the manager.
+
+# Versioning Specifications
+Traits are not all we have to consider, lets take a moment to think
+about specifications.
+
+```yml
+# For example, Workfile, A specification with the Entity, Work, and LocatableContent traits.
+Workfile:
+  description: >
+    The entity is the product of some manual task or process,
+    that often defines how to produce or derive other data.
+
+
+    Common examples are the documents opened, worked on and saved
+    by artists using DCC tools to create models/images/etc.
+  usage:
+    - entity
+  traitSet:
+    - namespace: usage
+      name: Entity
+    - namespace: application
+      name: Work
+    - namespace: content
+      name: LocatableContent
+```
+
+Conceptually, specifications are "things" that exhibit a set of traits.
+Current, specifications are purely unordered sets of `traitId`'s, raising similar problems as how to encode the version that individual traits have.
+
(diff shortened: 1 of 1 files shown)


doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+## Version every trait
+Every trait gets its own disconnected version. This version would have
+to travel across the api-boundary along with the trait data. Whilst the obvious
+thing would be to enhance the low level `TraitsData` type to have a
+first class understanding of versioning, this cannot be so because not
+every method takes a `TraitsData` as an argument. The lowest-friction
+course of action might be to pack the version into `TraitId`, essentially
+doing string-munging to communicate the version across the api-boundary.
+
+Hosts and managers would need to be inspect the version data, which we
```
DuncanHopkinsFoundry, line 149: This sentence is hard to read.

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+success_result = TraitsData()
+if (cameraVersion >= 3 and cameraVersion < 4):
+     trait = mediacreation.CameraTrait(success_result)
+     ... # Up to date camera population
+elif (cameraVersion >= 2 and cameraVersion < 1):
+     trait = mediacreation.legacy.CameraTrait_v2(success_result)
+     ... # V2 camera population
+elif (cameraVersion >= 1):
+     trait = mediacreation.legacy.CameraTrait_v1(success_result)
+     ... # V1 camera population
```
DuncanHopkinsFoundry, line 186: What about an alternate/secondary constructor that accepts a version container. The host prefills this with the versions is understands and is handed one of those versions back?

```
# Ask for versions 0 to 3 and 5. But not 4!
trait = mediacreation.legacy.CameraTrait(success_result, {[0,3},{5,5}])
```

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+- Version support may become fragmented if hosts and managers provide
+  versioned support for only certain traits
+- When packing multiple traits, manager has to branch on each trait as
+  they could be different versions.
+- When extended to specification this _really_ causes problems, as we
+  need to invent a new place to pack specifications version across the
+  wire.
+
+## Version the whole schema.
+This option places a version number is placed at the top of the
```
DuncanHopkinsFoundry, line 226: Hard to read sentence.

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+```cpp
+//HostInterface implementation
+SchemaVersion HostInterface::schemaVersion(std::string_view library_identifier) {
+  if (library_identifier == "openassetio-mediacreation"){
+    return openassetio_mediacreation::version();
+  }
+}
+```
+### Benefits
+- Version dosen't need to be packed into any `traitId`
```
DuncanHopkinsFoundry, line 243: Would this only be true for the host?
Does the manager still need to track the versions? I am assuming multiple apps/hosts can be active at a time, they might all have different version combinations?

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+
+### Versioning specs individually
+- Raises the question of whether each trait in a specification needs to
+  be pinned to a specific version of a trait.
+- Have to pack specification version into the type, or ask the manager
+  to somehow deduce specification version
+
+The complexity of thinking about disconnected versioning when you take
+specifications into account does seem to bias towards versioning the
+entire schema as a unit.
```
DuncanHopkinsFoundry, line 560: As long as the spec and trait versioning is handled separately, I would advocate the opposite.  The per spec version does not have to be effected by trait versions.

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+  pipeline.
+
+## Manager/Host Namespace Collisions
+With the current state of affairs, a large problem with versioning is
+presented when you consider the possibility of namespace collision
+between hosts and managers.
+
+For the purpose of this decision, we will assume that two versions of
+`mediacreation` can exist in the same process without problematic
+namespace collisions, although this remains a problem to be solved.
```
SamCrooksFoundry, line 117: Is it worth including the version within a namespace to eliminate this problem?

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+  traits in the package.
+- Semver could be used to provide additional runtime compatibility
+  information.
+
+### Drawbacks
+- Version has to be encoded in `traitID` because several methods
+  (including `resolve`) take `traitSet`, which is only a collection of
+  `traitIDs`. There is no `TraitsData` to inspect. This would be a major
+  breaking change as currently conceived.
+- Version support may become fragmented if hosts and managers provide
```
SamCrooksFoundry, line 217: I'm wondering how big a risk fragmentation is.  Do we predict that there will be a lot of different versions of similar traits?  My uninformed guess would be that traits would likely settle down after a few years and few would change.

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+  versioned support for only certain traits
+- When packing multiple traits, manager has to branch on each trait as
```
SamCrooksFoundry, line 219: If we used SemVer or similar, and had the concept of none breaking changes, it feels to me like branching could be minimal for most traits.

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+
+
+This would strongly put multi-package specifications into the camp of
+being an extension method used by organizations wishing to use
+openassetio inside their private pipeline, as it's doubtful productized
+management systems or hosts would want to account for all the specific
+permutations here. A major downside.
+
+## Simplify, ignore it.
+Also an option, maybe we can think about this later. How and if external
```
SamCrooksFoundry, line 686: This feels like the best option to me.  For OAIO v1 release, are non-medicreation traits a required feature?  I still think we should be versioning mediacreation traits, but perhaps we can defer third-party supplied traits until the library has seen some real world use.

doc/decisions/DR023-Versioning-traits-and-specifications.md
feltech: :thinking: there is still the problem of the manager knowing what version was used by the host.

Perhaps missing from this example is that the `trait` object has methods on it, which may change depending on version, necessitating the branching.

doc/decisions/DR023-Versioning-traits-and-specifications.md
feltech: In any given communication across the API boundary there is only one host communicating with one manager.

As an advanced use case, it is possible for a host to make use of multiple managers at runtime (e.g. for different types of content).  But in any given API call there is only one manager and one host.

A manager only needs to consider multiple hosts in that, ideally, they develop their plugin once and it can be loaded into any host application and "just work".  But at runtime, it only has to deal with one host.

A key (educated) assumption is that hosts will use a single version of a trait package, and it is up to manager plugins to adapt based on the version the host uses.  Therefore a single manager plugin developed for use across multiple different host applications (and different versions of the same host application) may well need to branch in its code, based on the trait package version used by the host it is loaded into.

doc/decisions/DR023-Versioning-traits-and-specifications.md
feltech: The ideal solution would indeed have Specifications continue to work regardless of changes in the traits making up their trait set.

There are technical reasons why that's difficult
* Specifications currently have no identity of their own in the API - i.e. when you make an API query you never provide the name of the Specification you're using, just its constituent traits. So there's currently no mechanism for the manager to know what Specification was used by the host at all, let alone the version of it. Adding such information would be a fairly big API rethink, but it is something we've discussed.
* The Specification classes generated from the YAML schema (using `traitgen`) currently have a hard dependency on the classes generated for their Traits, meaning mismatched Trait and Specification versions could lead to false-negative queries, or worse. This dependency is something we're considering removing as a consequence.

doc/decisions/DR023-Versioning-traits-and-specifications.md
feltech: That's the dream, but especially whilst we're in the early stages of teasing out the core set of traits from the community, we expect there to be a fair bit of churn.

Though admittedly _most_ changes are likely to be non-breaking, i.e. new traits, rather than changes to existing ones.

doc/decisions/DR023-Versioning-traits-and-specifications.md
feltech: In OTIO they don't even change the version for non-breaking changes.

I think the problem here is that you can't have one big outer branch on trait package version, and instead have to check the version individually for every trait.

doc/decisions/DR023-Versioning-traits-and-specifications.md
feltech: The intention has always been that host applications will supply their own application-specific Traits and Specifications.  E.g. a `BlenderWorkfileSpecification` that is bundled with Blender and extends the traits of `WorkfileSpecification` with a `Blender` trait, allowing managers to filter queries to only show Blender project files.

The idea is that such application-specific traits/specs will be made public by ISVs so that manager plugins can, if they wish, adapt their code to support them.

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+        name: Camera
+      - namespace: companycorp-traits.identifier
+        name: CompanyCorp
+```
+
+This is troublesome to consider for a few reasons. It's tempting to just
+handwave this sort of thing away for later due to how complex it can
+get, but no, let's bite the bullet.
+
+## Explicit versioning
```
feltech, line 588: [consider] Would like to see a worked example, similar to the "Whole-schema versioning" alternative below, since I suspect they'll result in similarly excessive branching.

feltech (commented):
Gave it a read through and its looking good. I've refrained from nitpicks, for now :wink: .

I'd like to see more worked examples and a little bit of expansion of the existing worked examples to illustrate the problem a bit clearer (e.g. we must branch, but _why_ as per https://github.com/OpenAssetIO/OpenAssetIO/pull/1254#discussion_r1472936249)

doc/decisions/DR023-Versioning-traits-and-specifications.md
feltech: That's a really good point, there may well be 3rd party libraries (perhaps even studio-specific plugins) using the host-side API :thinking:

doc/decisions/DR023-Versioning-traits-and-specifications.md
elliotcmorris: Even given that, that dosen't neccesarily mean non-mediacreation _specifications_ are a required feature. We've always expected mixin traits to be appended to formal specifications, if there's only a few examples of mixin traits to expect (like the host identifying on, for example,) we might be able to move forward asking managers to just check for those if they need em.

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+
+```python
+# Manager
+# In this example, CameraTrait is on version 3
+# SpatialTrait is on version 2
+#
+# The manager is responding to a resolve, needing to populate a
+# TraitsData with a camera trait and a spatial trait of the expected
+# version
+cameraVersion = mediacreation.versionFromTraitId(mediacreation.CameraTrait.kId)
```
tomc-cinesite, line 172: [question] Would the version be determined from the constant from the runtime package, or the requested trait ID in the `resolve` request.  The latter may be beneficial in case this is running out-of-process (eg: via gRPC) where the imported `openassetio-mediacreation` package may not be the version the host has imported.

doc/decisions/DR023-Versioning-traits-and-specifications.md
tomc-cinesite: If the version is in the id, then the version the host wants is in the `resolve` request?

tomc-cinesite: Is the other challenge that in reality the population of the data will vary from version to version, so you'd still need the conditionals so you can route to the correct code path (eg `trait.setOldName(value)` -> `trait.setNewName(value)`)

doc/decisions/DR023-Versioning-traits-and-specifications.md
elliotcmorris: In this example it was coming from the traitID itself yes, (essentially we were thinking of appending/prepending it to the string.)

I don't think there's any future in determining the version from the runtime package itself, for exactly the reasons you say. The host will have to inform the manager somehow of the version directly.

tomc-cinesite: > ...essentially we were thinking of appending/prepending it to the string....

> ...The host will have to inform the manager somehow of the version directly, ...

If its in the id string then do you need another mechanism?

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+- A single trait can be updated in isolation without any effect on other
+  traits in the package.
+- Semver could be used to provide additional runtime compatibility
+  information.
+
+### Drawbacks
+- Version has to be encoded in `traitID` because several methods
+  (including `resolve`) take `traitSet`, which is only a collection of
+  `traitIDs`. There is no `TraitsData` to inspect. This would be a major
+  breaking change as currently conceived.
```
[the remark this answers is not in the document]
themissingcow: Something else to consider - which I think might actually be quite important - is that we've often talked about "simple" managers that don't need first class understanding of every field in the data.  This probably extends to resolver caches too.

If the versioning information is _not_ part of the data, and is always driven by the host, then you end up effectively saying it its perhaps not even possible to write this kind manager any more. And that has generally been quite an important part of the API - that managers can store/recall host provided data without having to understand/manipulate this for each request. There may very well be only certain trait properties that are actively managed.

If the versioning is in the id, then you always know what version persisted data is/was, and in the a host is also capable of supporting older data as it _may be the only thing that is capable to mapping it to a newer schema version_.

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+
+[`openassetio-mediacreation`](https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation)
+is a generative library, in that it produces trait and specification
+view classes representing the objects specified in the source `.yml` via
+the `openassetio-traitgen` tool, currently in both the c++ and python
+languages. As such, the challenge of versioning is not only descriptive,
+but also technical.
+
+# Aims :
+- Allow new mediacreation versions to be adopted by managers without
```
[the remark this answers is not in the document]
feltech: This raises the spectre of some kind of negotiation between host and manager, and requiring _both_ hosts and managers to bloat their codebase with lots of conditionals, rather than isolating that bloat to just the (typically more nimble) manager implementers.

I'm sure pipeline people in the past have told us that they're used to new versions of DCC tools breaking their plugins, and have processes in place to adapt when upgrading. Not to mention the cadence of adopting new DCC tool releases tends to be slow enough that they'd have plenty of warning of breaking changes (and there would likely be  other breaking changes besides OpenAssetIO).

Commercial plugins are an interesting case, though I would hope they'd be self-motivated to keep up to date.

Having said all that, I'm not totally averse to a negotiation mechanism, I'm just keen to limit the scope of a versioning mechanism to the minimum that would be generally helpful. If trait version negotiation is really required for that, then so be it.

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+  breaking compatibility with hosts running previous versions.
+- Deliver a solid but minimal versioning foundation, that leaves open
+  opportunities for implementation conveniences at a later date.
+- Allow infinite backwards compatibility for manager implementations.
+- Low cognitive load, the OpenAssetIO development ecosystem cannot
+  support a complicated/technically intricate solution.
+- Minimize burden on manager implementors, whilst keeping in mind the
```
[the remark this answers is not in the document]
feltech: I suspect this was about our assumption that the manager would be responsible for most of the work to adapt to whatever the host advertises as its supported schema version; but we want to make that adaptation as painless as possible. For hosts it would be trivial. However, this assumption may be upended by https://github.com/OpenAssetIO/OpenAssetIO/pull/1254/files#r1482525401

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+## Who adapts
+We are going to assume that in most cases, the manager will be
+performing trait conversions between versions, with the host defining
+the "working version".
+
+Reasons:
+
+- Hosts update less often, and such are more likely to remain on older
+  versions of `mediacreation` for longer.
+- Managers are likely to be more capable of implementing capable trait
```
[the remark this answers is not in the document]
feltech: It was/is indeed a guiding principle, but that seems to have come into question now, and if we change our stance it quite significantly affects the decision and design - see https://github.com/OpenAssetIO/OpenAssetIO/pull/1254#discussion_r1482525401

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+  conversions between versions.
+- Hosts "start the conversation", and for some methods that means
```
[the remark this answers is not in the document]
feltech: Yep, more to discuss on this - see https://github.com/OpenAssetIO/OpenAssetIO/pull/1254#discussion_r1482525401

doc/decisions/DR023-Versioning-traits-and-specifications.md
feltech: Yeah you'd likely reuse code for traits that haven't changed, but how you do that is more constrained with per-trait versioning.  

E.g. with per-trait versions you must branch around every trait when you receive it, leading to fragmented handler code, whereas with a global schema version you could maintain a single handler (using OO techniques to reuse code) dedicated to a specific schema version.

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+  wire.
+
+## Version the whole schema.
+This option places a version number is placed at the top of the
+`traits.yml` file or equivalent. This defines the version for all the
+traits within the package.
+
+The manager will still need a mechanism to know which trait-library
+version the host will be expecting. This could be achieved by a
+`HostInterface` method.
```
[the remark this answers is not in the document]
feltech: We did discover that the `HostInterface` is available at `initialize()` time, so the manager _could_ configure its adaptation as part of its initialization, avoiding the need to make the call with each request (and sidestep the out-of-process callback issue).

Another solution would be to add a new argument to `initialize` that advertises the trait schema versions in use by the host.

If we require trait schema version negotiation between host and manager (see ), then I suspect we'd need a different set of functions altogether (or perhaps the `initialize` process could incorporate this). Also see

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
feltech: Some really good arguments for encoding the version in the trait ID there, I'm coming around to it, but still wary :thinking:

doc/decisions/DR023-Versioning-traits-and-specifications.md
```diff
@@ -0,0 +1,702 @@
...
+```
+### Benefits
+- Version dosen't need to be packed into any `traitId`
+- A traits and specifications library gets a version as a unit, making
+  talking about supported versions easier, and hopefully reducing
+  support fragmentation
+- Manager can branch at a higher level
+
+### Drawbacks
+- Unknown at runtime which traits or specifications have been updated in
```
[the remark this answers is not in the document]
feltech: I'm not sure how that can be avoided though, other than automated upgrade/downgrade scripts (which we're trying to avoid due to the development cost).

Regardless how granular the versioning is, the developer still has to adapt their code on a version bump. We need to ensure release notes document the traits that change, of course.

What is the use-case for runtime introspection of individual trait versions?

doc/decisions/DR023-Versioning-traits-and-specifications.md
feltech: That's a really good argument to encoding version in the trait ID - also good arguments in https://github.com/OpenAssetIO/OpenAssetIO/pull/1254#discussion_r1482555651

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
tomc-cinesite: "I'm developing a manger that only deals with locatable content trait and I don't want to have to change my code unless there is a breaking change with that trait ". With package versioning, release notes are the only indication that anything may have changed. There could be hundreds of traits in a package over time.

Runtime introspection comes in the case of "Im a host author publishing my own trait data to a manger that stores and recalls this without concrete understanding" , which I would imagine is pretty common.

I resolve a reference, I can immediately know which trait was stored, and adapt my handling of that data. If you only have the package version, there could be a large number of versions where that specific trait is identical. In many of these situations, and I have to rely on magic knowledge that v2-25 of the package dont need special handling. The host is the only one that can deal with data migration from old to new.

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
tomc-cinesite: The concern is more about making requests more stateful. Which in many scenarios then forces a custom manager state object.

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
tomc-cinesite: What is the reticence?

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
tomc-cinesite: Yeah, I think a "negotiation mechanism" (which includes one way demands) are my main concern, as it implies that one or more parties have to understand _all_ data. The original API design was that "managers faithfully store and recall data, unless in specific circumstances they _do_ have first-class understanding (eg paths)" and that case was usually in the minority. Any kind of negotiation mechanism significantly affects the viability of that. The data being self describing however still facilitates this and places the burden on whoever wishes to specifically _interpret_ the data.

doc/decisions/DR023-Versioning-traits-and-specifications.md
tomc-cinesite: I guess I was thinking that you it probably be unlikely to have a single handler per-package.

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
tomc-cinesite: ☝️ that thing - sorry not been very good at explaining things with typing 🙃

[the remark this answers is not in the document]
tomc-cinesite: Runtime introspection at least allows you to look at the `TraitsData` per-response to learn what data came back. This could easily happen with older assets published a long time ago being mixed with newer assets, where the manager doesn't actively manage that data.

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
tomc-cinesite: Maybe this comes down to a question of what the fundamental interaction model is. Historically, its always been that Hosts can store/recall data in/from a Manager, Managers may opt-in to layer on first-class understanding of the data. 

The rational here was that in a future (largely unrealistic) world, all data would have originated from one host or another. So the manager acts more like a database, where the ideal manager allows the host to store/recall assets that it doesn't even understand. Eg DCC -> DCC could exchange specific data without the manager needing to understand it per-se. This was a critical part of the abstraction. Not all managers may support this, but it is possible, and should be aimed for unless there is a compelling reason not to.

In practice this is probably less likely to be the case unless a manager embraces the API fully (though that has happened with USD...), the fact that the API was designed to facilitate that was a conscious choice.

  https://openassetio.github.io/OpenAssetIO/index.html#manager_implementation_concept

Management policy and others are a concession that many managers may have compromised support, but at least you can determine this accurately.

If we switch to the "host determines a version and the managers have to obey", then it pretty much turns this upside down, to a model where the manager is required to have first-class understanding of all data in any request. This would preclude host A publishing one version of a proprietary asset and host B resolving it having requested another version. If this is a vendor specific asset type, the manager may have no way to perform this migration, but the hosts - as they consume/produce that data with first class understanding - may well be able to. 

I think its the same reason that we chose not to enforce 'required' fields in the middleware, as the host is the only one who knows if a missing trait property is actually an error or not. The manager or middleware never can.

A world (probably not entirely dissimilar to OTIO?) were hosts in the ecosystem can exchange assets not-fully understood by the manger between themselves seems incredibly valuable, and seems to be readily facilitated where the data model is coherent and self-describing allows the responsibility of dealing with differing versions to lie with whoever wishes to layer on concrete understanding of that data not the broker in the middle. 

Auto-magic support could be layerd on if desired, but isn't a pre-requisite as long as it is possible to pass this through the manager as a 'hub' in the so-called hub/spoke model.

Keeping data interpretation de-coupled from its storage and retrieval feels really important if we want to make a long-term flexible interchange mechanism.

The existing API methods all traffic trait IDs, which allows any cleverness in any part host or manager to then be implemented if desired without any structural changes. eg. `resolve`  is requesting a specific version, so a host could resolve n trait versions and see which come back from a 'simple' manager. A complex manager could opt-in to doing version conversion from the requested id to the stored one it if wanted /could. I guess here per-trait versioning is beneficial as it means you dont need as much out-of-band knowledge as discussed else where.  This sounds familiart with my understanding of OTIO too (though I may well be wrong) so may also mean its not wheel re-invention?

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
feltech: > managers faithfully store and recall data, unless in specific circumstances they do have first-class understanding

To be clear, that's only for managers that advertise such a capability (via `managementPolicy`), not all managers.

To make sure I understand the wider point - a naive "pass-through" store-and-recall manager would never be able to respond positively to a negotiation, since it has no opinion on what trait/schema version is used?

I guess similarly, a generic host (e.g. `simpleResolver`) would have the same problem, if it was expected that the host had to provide a schema version to the manager, when it has no opinion.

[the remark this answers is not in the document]
tomc-cinesite: > To make sure I understand the wider point - a naive "pass-through" store-and-recall manager would never be able to respond positively to a negotiation, since it has no opinion on what trait/schema version is used?

sorry missed that bit - yeah, thats my main concern with any active negotiation mechanism.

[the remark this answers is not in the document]
tomc-cinesite: > Could it still work, though, as long as there is a way for the host/manager to signal in its negotiation "I don't care" - perhaps that's the default implementation.

Visits a previous question - is versioning optional? This negotiation mechanism sounds complex in practice and requires a lot of extra code, if its not part of how you handle data inherently, I bet most people would skip it- as its hard enough as is to understand an integration as a newcomer?

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
themissingcow: As an example, https://github.com/OpenAssetIO/OpenAssetIO/issues/1203 feels it would get a lot more complex/challenging with out-of-band version coordination.

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
feltech: Downsides I can think of to encoding the version in the trait ID:

* Adding a property is a breaking change. I.e. version would have to be bumped, so then older trait views would think the trait is not imbued (at least as currently implemented).  If the version wasn't part of the trait ID, then views could detect if a trait is imbued regardless of version, and make a best effort to read/write properties they understand.  Of course, such a tolerant approach has its own downsides.
* Any trait with multiple versions would require branching around individually. If the schema version was agreed at initialisation, the host/manager has more opportunities to adapt their control flow ahead-of-time (e.g. effectively an entirely different plugin implementation could be constructed). Though both solutions _could_ live side-by-side.
* Specifications must be trait version aware, and these versions must be kept in sync somehow, or we require a rethink in how they're implemented and what they represent.

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
feltech: An alternate view of OTIO is that it has per-schema versioning, but their schemas are much smaller.

I do agree the idea of a pass-through manager with no first-class understanding of the traits is something we need to keep in mind with any solution. Also related: https://github.com/OpenAssetIO/OpenAssetIO/pull/1254#discussion_r1496051054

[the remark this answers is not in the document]
themissingcow: > An alternate view of OTIO is that it has per-schema versioning, but their schemas are much smaller.

Yeah - though I guess I was meaning more the part of "the version is in the data" rather than "the version is ephemeral runtime metadata".

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
feltech: > Putting it in the traits data instead of an ephemeral runtime quantity could be a middle ground, but then it precludes the use of any traitID centric introspection methods

Yeah we excluded `TraitsData` changes as a possibility for that reason.

> Would the expectation be that a manager has to then independently store and track the version in use at the time of storage - which kind of gets you back to what you get with it being in the id in the first place?

Well yes, when we chatted with Ynput about this exact question they said it's not ideal, but they could store the trait version alongside the trait ID.

I also assumed that it's the manager's responsibility to massage data from its backend, whether structured as traits or their own bespoke structure, into a form suitable for the host. But this assumption does have negative implications for the ideal of a "pass-through" manager scenario, as you brilliantly described elsewhere (i.e. https://github.com/OpenAssetIO/OpenAssetIO/pull/1254#discussion_r1495946276)

> You never see specs at the transport layer

This is a major wrinkle we wrestled with this option: how to ensure Specifications can be recognised across the API boundary, when there's no communication of Specification version, just the versions of the constituent traits.

If per-schema versioning was used, and that was encoded in the trait IDs, and all traits in a set use the same schema version, then you _could_ figure out the appropriate collection of possible Specification views.

If per-trait versioning was used, figuring out the appropriate Specification view is much harder. You would presumably have to iterate over all available MediaCreation packages, trying all their Specifications until one fit.

With the schema version communicated/agreed at initialisation time, you know immediately which Specifications are valid candidates.

Specifications need a rethink if trait IDs encode their version. If we remove the trait accessor methods (e.g. `.locatableContentTrait()`), then things become easier... E.g. Specifications could be redesigned to be trait version agnostic.

themissingcow:
Conscious there are several threads here with overlapping conversations, please feel free to close any we me on. Attempting to summarise some considerations here, then I'll go away 🤣 .

It feels like there are perhaps two things being discussed?

### Versioning is part of the data vs versioning is a side-channel

Over the years, we have gone to great lengths to uphold the following design patterns in the API:

1.  Manager implementations should not require internal state to service API requests  - other than the inherent "state" of the back-end data store. Requests should be atomic and contain all required inputs to formulate a response. In the documentation we take trouble to state that a managers response should solely depend on the input to a request plus the back-end state. A custom ManagerInterfaceState object should never be a requirement, and only exists to allow caching or other opt-in optimisations.
2. Data transport/the core API is separated from domain specific concepts. If you do not intend to interpret or process data in a meaningful way, you do not need to have any first-class understanding of it. i.e why the Trait/Specification views are a layer on top of the naive `TraitsData` container that is used at the API boundaries and not part of the core API. You should _always_ be able to work at the simple traitId/`TraitsData` if required and implement (albeit limited) functioning Hosts and/or Managers.
3. Required mechanisms should be as simple and explicit as possible. e.g we moved `access` out of `Context` as it was too easy to overlook as it was.

(1) and (2) have been critical in allowing flexibility in adding assorted middleware, gRPC, shared gRPC, caching, 'shim' managers, etc., and allowing round-trip of data not understood by all parties, and future adaptation to OpenAPI/HTTP etc. Two hosts should be able to exchange asset data via a compliant manager as a 'hub' without the manager needing specific understanding of that data - short of storing/recalling traits datas. This also facilitates generic presentation of data such as in an asset browser, info panel etc... where neither the host or manager has any inherent opinion on versions or a way to select a specific one, and just wants to show the data "as-is".

If versioning is a sidecar mechanism outside of a request then it feels to preclude (1) and (2) unless it the mechanism is optional. If it is optional, then how should implementations deal with the other party not upholding their end of bargain?  It also relies on host and managers both implementing additional logic to ensure consistency between the versioning mechanism responses and the underlying data which may be prone to programming error or even difficult to satisfy in the case of transient dependencies between packages. Different parts of the same actor may even have different package versions (eg: a plugin to a DCC using its own version of a traits package but making requests through a previously initialised manager/context). In other DCCs the API requests may be _entirely_ parameterised by user input post initialisation, there may not even be a versioned traits package to 'import'. This would make (3) hard to achieve.

If versioning is included in the data itself, all the above are seemingly satisfied at the expense of making versioning a required consideration. Some mitigations to the burden this may bring this could be provided in entirely within the view classes to an allow specific implementation (Manager _or_ Host) to choose to risk ignoring versioning (which is different to making it optional), or `>=`, et.al flexibility. It also allows 'simple' hosts/managers, and has the benefit that no additional coding is required for self-consistent data, as versioning is implicitly baked into every request that deals with traits with the correct values. Existing introspection mechanisms (`entityTraits`, `managementPolicy` et.al.) would largely work as-is in a version-aware way. A notable tradeoff is that `entityTraits` for `kWrite` may need further consideration.

### Trait .vs. Specification versioning

This one feels simpler, and seems to focus more about the tradeoff of the potential for high-level code path changes at the expense of requiring "outside knowledge" encoded in documentation in business logic.

## Thoughts

Stepping back, I'd flag the following considerations when looking at a solution:

- Is versioning data optional?
- Published data may exist for extended time periods and be combined with new data.
- Is breaking the current separation between the interpretation of data, and its transport/storage something we really want to do?
- Is requiring statefulness at the API level something we should even consider? I believe it is somewhat an anti-pattern in scalable/distributed architectures.
- How would these mechanisms affect the transport-layer mechanisms we have/need (caching, gRPC, HTTP etc) or their future development.
- Can we definitely say that either the Host or the Manager is specifically responsible for version coersion? Rather than saying it is the responsibility of who ever wishes to apply first-class understanding to the data.
- Keeping versioning with data seems to be common in most file formats, the USD proposal and in OTIO, and version transformations are applied at the point of interpretation (I imagine largely due to the third point above). Is our situation really that different?

_Personally_, it feels like decoupling the identity of the data from the data is at odds with the majority of prior-art in systems design I've encountered, and with OTIO/USD and other industry peers (please correct me if I'm off-mark here). Moving away from a stateless interface design could preclude a significant number of opportunities in the future, including ones we already have modelled (gRPC variants) and that many third-parties have requested (HTTP). If the mechanism is optional, then it creates even more runtime permutations and further complicates implementation. I'm also not sure if a host would genuinely be able to determine all versions up front at initialisation time (in Gaffer for example, wed build an `OpenAssetIOResolve` node that is parameterised by string trait IDs, that the user could employ as they see fit), I'd hate to loose the ability to build these kinds of low-level hosts. Managers being able to blindly store/recall data (albeit with limited facility) feels hugely valuable, and would allow hosts who control their own traits to manage versions transitions themselves.

fwiw

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
elliotcmorris: > I bet most people would skip it- as its hard enough as is to understand an integration as a newcomer?
I tend to agree. Openassetio has already broken the seal on requiring a certain level of manual diligence on the implementors part, "good behaviour", etc, but it does seem like negotiated would be a humongous addition to that. If an openassetio integration isn't already impossible without a reference currently, it definitely would be under that paradigm.

doc/decisions/DR023-Versioning-traits-and-specifications.md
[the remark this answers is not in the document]
themissingcow: > I'm not sure how that can be avoided though, other than automated upgrade/downgrade scripts (which we're trying to avoid due to the development cost).

Random thought - the OTIO metadata (iirc) looks similar to a TraitsData. Could you literally use the same code OTIO uses to define their schemas with up/downgrade to manage traits? You'd just copy the trait ID to an `OTIO_SCHEMA` key?

elliotcmorris:
Closing this PR. We learned a lot from the discussion, but the decision obviously has many more facets that we initially thought, so we won't be making a decision based on this DR.

The conversation continues here : https://github.com/OpenAssetIO/OpenAssetIO/pull/1267, where we attempt to make a smaller decision based on one of the points that emerged here, namely in-data vs out-of-data versioning.

PR #1267, opened 2024-02-22
[Docs] Add DR23 (again), trait versioning method

elliotcmorris:
Closes https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/issues/83

## Description
So the previous attempt at DR23 was, as we soon discovered, both underthought and attempting to tackle too much.

This DR attempts to take the learnings from that and make a more measured decision on a very specific point, which seems to be the root of the issue, that being in-data or out-of-data versioning.

Add initial draft of DR exploring mechanism of communicating trait and specifications library version in an OpenAssetIO session.

### DIFF
--- a/doc/decisions/DR023-Versioning-traits-and-specifications-method.md
+++ b/doc/decisions/DR023-Versioning-traits-and-specifications-method.md
@@ -1,0 +1,468 @@
+# DR023 Trait & Spec versioning paradigm, in-data or out-of-data.
+
+- **Status:** Decided
+- **Impact:** High
+- **Driver:** @elliotcmorris
+- **Approver:** @felltech @elliotcmorris @themissingcow
+- **Outcome:**
+
+## Background
+This DR follows on from a [previous attempt at a
+DR](https://github.com/OpenAssetIO/OpenAssetIO/pull/1254), which sought
+(somewhat haphazardly), to nail down format and type of versioning that
+traits and specifications libraries would use. That draft DR, in
+hindsight, attempted to tackle too broad a range of decisions, and
+consequently ended up deciding on not much. Nonetheless, we are now more
+informed and are coming back for a second attempt.
+
+This DR attempts to narrow the focus onto a single point of contention
+that became a recurring theme, whether OpenAssetIO trait and
+specification versioning information should be communicated as part of
+the data, or agreed upon as a secondary "handshake" interaction between
+host and manager. For the purposes of this DR, we will refer to these
+two categories as "in-data" and "out-of-data".
+
+In attempting to keep its scope limited, this DR does not propose a
+specific solution, but merely provides representative examples of
+what a solution in each proposed paradigm might look like. It is generally
+assumed further facilitative work (for example, convenience utilities
+in `traitgen` and `mediacreation` will be decided upon and implemented
+at a later date.)
+
+## Relevant data
+
+Over the course of our investigations, it became clear that we had
+either failed to consider entirely, or handwaved away as less important
+than they actually were, several key criteria in the name of simplicity.
+The main themes are listed below, as we currently understand them.
+
+### Two way compatibility
+In the [previous
+DR](https://github.com/OpenAssetIO/OpenAssetIO/pull/1254), we attempted
+to make a simplifying assumption based around the idea that managers
+were more capable of managing version updates. Whilst we still believe
+that, it has come to our attention that there are valid reasons that a
+host would also want to maintain backwards compatibility when updating
+versions. Due to this, we now consider the ability for both the host and
+manager to choose to maintain compatibility essential.
+
+### Distributed workflows/RPC
+During investigations, the concept of systems that work remotely, (eg
+[gRPC](https://github.com/foundrytom/OpenAssetIO-gRPC)) came up as a
+potential difficulty with some implementations, given that any callback
+based communication, ie querying the `hostInterface` for a version,
+would be especially difficult to implement in this paradigm.
+We now consider it essential that we take an approach where versioning
+does not impose undue extra effort on distributed workflows.
+
+### No internal state
+OpenAssetIO has long attempted to stick to the philosophy that each
+method call represents a stateless interface, and that the manager
+should not persist state between calls. This is true and essential to
+model potential workflows/integrations in the future (GRPC/HTTP), etc.
+
+However, we do not believe this necessarily excludes any handshake style
+interactions. OpenAssetIO already provides a mechanism to configure
+managers at the beginning of the session with
+[initialize](https://openassetio.github.io/OpenAssetIO/classopenassetio_1_1v1_1_1host_api_1_1_manager.html#aa52c7436ff63ae96e33d7db8d6fd38df).
+Whilst we agree that state must not be utilized during a session,
+injecting version state before or at initialize time remains a valid
+option.
+
+### Naive managers
+A pitfall that many of our earlier sketches fell into was assuming
+managers would have a first class understanding of the traits they were
+being asked to persist, which led to us assuming they would be able to
+perform upgrade and downgrade behaviour on them.
+
+This isn't, nor has it ever been this case, and the idea that a manager
+can and should persist data it doesn't understand has been a long held
+principle of the library. This supports the idea stated above that both
+hosts and managers must each be able to adapt to version divergence, as
+it cannot be known which of them understands the data, and thus
+understands how to perform the migration.
+
+### Is versioning optional?
+Some concerns have been raised about the optionality of versioning,
+specifically around the idea that a complicated versioning handshake may
+lead to implementors simply ignoring it, a valid concern. Whilst we can,
+and likely will declare at least some support for versions a necessary
+part of any implementations, we have seen before that implementors can
+ignore necessary implementational steps due to ignorance of their
+necessity. This point is well taken, and will weigh into any decisions.
+
+### Reconsidering schema version
+The achilles heel of decision records ... interconnected concerns. the
+prior DR attempt _tried_ to isolate itself to higher level concerns,
+version format, schema vs trait version, etc. One of the decisions we
+_thought_ we made was deciding on schema versions (as opposed to trait
+by trait versions).  However, choosing to place the version in the data
+makes us reconsider the value of schema-wide versions. They are simpler
+conceptually, but if each trait could come in with a different schema
+version anyhow, are we really gaining much by avoiding every trait
+simply having its own version?
+
+## Options considered
+
+### Option 1 - Out of data
+
+This option imagines that an agreement is made via the OpenAssetIO
+interface on the version of traits and specifications that are in
+play. Hosts and Managers will inspect each other, and make a "handshake"
+deal on agreed versions. From that point on, the versions are known
+by both sides, and both sides must communicate using data formatted
+to the schema of the agreed upon version.
+
+Our initial, "what is the simplest thing we could do" was the idea that
+the host knows the versions in play right? Let's just add a way for
+the host to let the manager know what versions it wants to use, and let
+the manager adapt. See example below.
+### Host side
+```python
+# When setting up a manager
+settings = manager.settings()
+settings["server"] = "my.server.com"
+
+mediacreation_version = openassetio_mediacreation.version()
+
+manager.initialize(settings, [mediacreation_version])
+```
+
+### Manager side
+```python
+ def initialize(self, managerSettings, schemaVersions, hostSession):
+   # Store versions for later use
+   self.__schemaVersions = schemaVersions
+```
+
+If you've read the principles above, you can likely see that this
+violates a few of them. Notably, the host has no method of supporting
+multiple versions, and all the onus of migrating between versions is on
+the manager, despite the fact that it may not understand it at all.
+
+Therefore, an out-of-data implementation will likely be a fair bit more
+complicated, and look more like a **negotiation**.
+
+### Negotiated
+
+In this conception, one side of the exchange (manager in this case),
+advertises a set of versions it supports. The other side then queries
+this list, selects the specific versions it would prefer to communicate
+in, then informs the other side which versions should be the medium
+of exchange going forward. Keep in mind, this is merely a passable
+representation of the concept of a negotiated, out of data approach,
+and not a final design.
+
+### Host side
+
+```python
+# During initial setup
+
+# In this example, supportedSchemaVersions returns a dictionary of the
+# provided schemas, mapped to a list of versions the manager supports
+# for those schemas
+# {
+#   "openassetio.mediacreation", [1,2,3]
+#   "companycorp.custom_traits", [1]
+# }
+supported_versions = manager.supportedSchemaVersions()
+
+# Verify that the trait packages we require are supported at all
+# by the manager
+if not openassetio_mediacreation.packageId() in supported_versions:
+  incompatible_with_manager_error()
+  return;
+
+if not companycorp_traits.packageId() in supported_versions:
+  incompatible_with_manager_error()
+  return;
+
+mediacreation_version = None
+companycorp_version = None
+
+# Verify the versions of the trait packages we are capable of supporting
+# are supported by the manager
+# This host is capable of working with mediacreation version 3 and 4,
+# and companycorp custom traits version 1. Given the choice, it would
+# prefer to work with mediacreation version 4.
+if 4 in supported_versions[openassetio_mediacreation.packageId()]:
+  mediacreation_version = 4
+else if 3 in supported_versions[openassetio_mediacreation.packageId()]:
+  mediacreation_version = 3
+else:
+  incompatible_with_manager_error()
+  return
+
+if 1 in supported_versions[companycorp_traits.packageId()]:
+  companycorp_version = 1
+else:
+  incompatible_with_manager_error()
+  return
+
+# Tell the manager up front what versions you are going to use.
+# These must be versions reported from `supportedSchemaVersions`.
+negotiated_schema_versions =
+{
+  openassetio_mediacreation.packageId():mediacreation_version,
+  companycorp_traits.packageId():companycorp_version
+}
+manager.informSchemaVersions(negotiated_schema_versions)
+
+... # Go on to use openassetio methods with trait views of the now
+    # agreed upon version for each trait package.
+```
+### Manager side
+```python
+def supportedSchemaVersions(this, hostSession):
+  # As a manager, we can likely directly declare which schema versions
+  # we support
+  return {
+   openassetio_mediacreation.packageId(), [1,2,3]
+   companycorp_traits.packageId(), [1]
+  }
+
+def informSchemaVersions(this, schemaVersions, hostSession):
+  # Record the schema versions that the host is going to use.
+  # We can use these to branch our internal functionality
+  this.__mediacreation_schema_version = schemaVersions[openassetio_mediacreation.packageId()]
+  this.__companycorp_schema_version = schemaVersions[companycorp_traits.packageId()]
+```
+
+### Pros
+- Decides on version up front, reducing low level trait by trait branching.
+- Versioning is technically optional, which may allow "best-effort" data
+  access. E.g. if the trait ID remains unchanged, and only new
+  properties are added, then a trait version bump is a non-breaking
+  change - `isImbued()` still returns `true` and older trait properties
+  are still accessible.
+- Specifications continue to work as they are currently conceived. If
+  both host and manager agree on the schema version, then the valid
+  selection of Specification view classes is trivially known.
+
+### Cons
+- Complex dance needs to be done up front, people may end up just
+  skipping it.
+- The potential to deliberately skip version negotiation to allow "best
+  effort" access patterns is very bug-prone, may be difficult to debug,
+  and in general introduces confusion about whether versions should be
+  accounted for.
+- Managers and hosts now have additional data related to specific
+  entities that they must communicate, increasing the complexity of
+  "pass-through" hosts and managers, and complexity of data transfer
+  across multiple hops and/or storage in multiple systems. This could be
+  especially problematic for naive managers that simply wish to be data
+  recorders, as they are almost forced to be version aware.
+- Implies probable negotiation, which implies introducing state.
+  See [no internal state](#no-internal-state).
+- Precludes the ability to communicate with multiple versions of the
+  same trait in the same session.
+- Upgrading traits requires out-of-code knowledge about what traits
+  are compatible between schema versions.
+### Option 2 - In data
+
+In this conception, version information is sent "over the wire" so to
+speak. Each trait has a version packed into it, which is unpacked on
+the other side, and then used for any version branching.
+
+> **Note**
+>
+> Practically, this would likely be packed into the `traitId` of each
+> trait, something like
+> "`openassetio-mediacreation-v-4:LocatableContent`". This would
+> effectively make each version of a trait a new trait as far as the
+> `openassetio` core library is concerned.
+
+
+### Host side
+```python
+# The host side implementation is almost entirely unchanged from
+# from standard, excepting a new BatchElementError type
+trait_set = {LocatableContentTrait.kId}
+context = self.createTestContext()
+
+result = [None]
+
+def success_cb(idx, traits_data):
+    result[0] = traits_data
+
+def error_cb(idx, batchElementError):
+    if instanceOf(batchElementError, VersionNotSupportedError):
+      incompatible_with_manager_error()
+    ...
+
+self._manager.resolve(
+    [entity_reference], trait_set, ResolveAccess.kRead, context, success_cb, error_cb
+)
+```
+### Manager side
+```python
+# On the manager side, each trait gets unpacked (probably via a utility
+# method provided by traitgen or openassetio), and branching is done
+# on a trait by trait basis
+
+def resolve(
+        self,
+        entityReferences,
+        traitSet,
+        resolveAccess,
+        context,
+        hostSession,
+        successCallback,
+        errorCallback,
+    ):
+
+  # This function is imagined a pre-provided utility, possibly generated
+  # via traitgen, that would error in some way if the trait set does not
+  # contain the locatableContent trait.
+  locatableContent_version = locatableContent_version_from_trait_set(traitSet)
+
+  success_result = TraitsData()
+  if (locatableContent_version == 2):
+    trait = mediacreation.traits.content.LocatableContentTrait(success_result)
+    trait.setUrl(managed_filesystem_locations[ref.toString()])
+    successCallback(idx, success_result)
+  elif (locatableContent_version == 1):
+    trait = mediacreation.legacy.traits.content.LocatableContentTrait_v1(success_result)
+    trait.setLocation(managed_filesystem_locations[ref.toString()])
+    successCallback(idx, success_result)
+  else:
+    raise RuntimeError()
+
+```
+
+#### Pros
+- More flexible, traits of different versions can be mixed in the same
+  session, which could allow hosts/managers to perform partial upgrades
+  in the case they are blocked on small specifics, and allow components
+  of a system to each work with a different schema version and merge
+  their results.
+- Hosts and managers would be capable of working with mismatched
+  schema versions, so long as the versions of the individual required
+  traits are compatible.
+- Hosts don't have to interrogate and declare which library versions
+  are in use, which helps especially in the case of second-order
+  dependencies.
+- Seems to fit with existing OpenAssetIO design, especially with
+  `BatchElementError`, as an incompatible version wouldn't necessarily
+  need to error the whole batch.
+- No new API, core `openassetio` library remains version unaware.
+- Cannot be ignored, unlike other "layered on top" mechanisms.
+- Managers not required to consider storage/communication of additional
+  version data to enable interpretation by future/downstream systems.
+- Stateless by default.
+- Allows per-trait versioning, which avoids the need for "insider
+  knowledge" as to which traits actually change in which package
+  versions.
+- Prior art (e.g. OTIO) encodes versions in their data structure,
+  presumably for good reason.
+- Allows 'simple' hosts to be written that don't understand specific
+  data, and thus wouldn't be able to meaningfully negotiate.
+
+#### Cons
+- No "primary" version across the session, increasing implementation
+  complexity.
+- Would likely require some utilities in `traitgen`, otherwise, due to
+  each new version effectively being a brand new trait, constructing the
+  correct trait view would be become a tedious dance.
+- Introspection methods become more complex (see [What's
+  versioned](#whats-versioned)).
+- Unclear how specification versioning is to work without an out-of-data
+  version.
+- The most obvious implementation, (using `traitID` to store the
+  version), would effectively make any change to a trait version a
+  breaking change. Allowing non breaking changes (such as adding a new
+  property) would require significant rework to the `traitgen` output.
+
+### In data - Extended
+The implementation of in-data versioning is very basic, which is part of
+its strength, but it necessitates a "try and see" approach on the part
+of the host. If the host wants to be adaptive to the manager, it may
+benefit from a more up front introspection of the supported manager
+versions.
+
+To this end, the `supportedSchemaVersions` method (or similar concept,)
+from Option 1 could be orthogonally added without any conceptual
+conflict. This would have the added benefit of allowing hosts to also
+support a range of version/have backwards compatibility, which is a huge
+benefit.
+
+This stops short of negotiation, being merely introspection similar to
+our other manager introspection methods.
+
+### What's versioned?
+In this example, we have been working under the assumption that we'll be
+inserting version info into `traitId`, which in the current environment
+essentially makes each new version of a trait a unique trait, which has
+both pros and cons.
+
+A decision intentionally not made in this DR is whether or not this sort
+of presumption is the way forward with an in-data approach. it is
+possible to pack version data into `traitID` and yet not presume that
+`mediacreation-v1-locatableContent` is a conceptually different trait
+from `mediacreation-v2-locatableContent`.
+
+There are reasons we might want to do this, for example `entityTraits`
+and perhaps even `managementPolicy` could potentially benefit from being
+unversioned. Does a user want to know that an entity supports
+`locatableContent-v2` specifically, or are we simply asking "What is
+this thing?", in which case, an unversioned, or "all-version" method of
+talking about traits may be helpful.
+
+`managementPolicy` presents a unique decision point with an in-data
+solution, as it is supposed to be entity agnostic, which could present
+problems to a simple manager if we choose to follow the implications
+of versioning it.
+
+We intentionally leave this choice to a subsequent decision, as there
+are numerous technical options available.
+## Decision
+
+OpenAssetIO will move forward with [Option 2 -- in
+data.](#option-2---in-data)
+
+### Rationale.
+This has been a very complex decision, and is only a subset of a larger
+strategy, many decisions remain to be made before versioning can be said
+to be complete, and in many ways versioning will not be complete until
+after real world usage has hardened our approach.
+
+Once the non-negotiable preconditions had been considered, the
+controversy in the decision came down to four general areas
+
+- The addition of state to the interface.
+- The possibility and ease of naive host/manager implementation.
+- The possibility and ease of exchanging different versions in the same
+  session.
+- What to do about specifications.
+
+There were of course other points, but on all but the latter of these
+points of contention, an in-data solution came out ahead.
+
+The question of specifications is a large one, and can be said to be
+out-of-data versioning greatest appeal at this point. We will move
+forward under the assumption that specifications will remain
+un-versioned (in the sense that specifications themselves will not have
+an in-data version number, the traits within them will implicitly be
+versioned). This may introduce complexity to implementors using
+specification views in their applications, and they may need to resort
+to using out-of-code knowledge to resolve compatibility issues. The
+working view of specifications however is that they are merely optional
+implementor conveniences, and seeing as they have no real presence in
+the API itself, we feel comfortable enough putting them to the side for
+this decision, although there is certainly more thinking to be done
+here.
+
+Nonetheless, a large reason for the adoption of an in-data solution is
+how little of the API it needs to disrupt in order to become useful, and
+yet how open it is to extension.
+
+We acknowledge there may be complexity challenges by moving forward with
+direct in-data versioning with current OpenAssetIO methods, as the idea
+that each version of a trait effectively becomes a new trait introduces
+[questions](#whats-versioned) with introspection methods. However, the
+fact that the solution will function under present assumptions without
+the need for _any_ API changes is very appealing, and allows a layered
+iterative approach to reduce this complexity, rather than having to do a
+big bang drop full of speculation all at once. We look forward with
+excited expectation as to how `openassetio-traitgen`, with its generated
+views, can be leveraged to make this experience simpler and easier.


doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+- Introduces "Try and see" negotiation on the host side, although
+  introspection behaviour to negate this could be introduced at a later
+  date, similar to other introspection mechanisms. (See Option 3 -
+  Extended)
+- Would likely require some utilities in `traitgen`, otherwise, due to
+  each new version effectively being a brand new trait, constructing the
+  correct trait view would become a tedious dance.
+- EntityTraits becomes more complex, as we either have to declare it
+  version agnostic somehow, or managers would need to return the correct
+  versions/entire swathes of versions.
```
feltech, line 332: [consider][minor] Is `entityTraits` really a special case? In the simplest formulation, its the mirror on the host side of what the manager must do, i.e. branch on every trait/specification its interested in.

I.e. re. "managers would need to return the correct versions/entire swathes of versions" - you could apply the same argument to `resolve` - the host could resolve swathes of versions.

It's back to that question of "who's in charge here".

```suggestion
- Hosts and managers must detect and branch on every supported trait version and/or communicate all their supported versions in an API request (e.g. host->manager via `resolve`, manager->host via `entityTraits`).
```

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+version format, schema vs trait version, etc. One of the decisions we
+_thought_ we made was deciding on schema versions (as opposed to trait
+by trait versions).  However, choosing to place the version in the data
+makes us reconsider the value of schema-wide versions. They are simpler
+conceptually, but if each trait could come in with a different schema
+version anyhow, are we really gaining much by avoiding every trait
+simply having its own version?
+
+## Decision
+TBD
```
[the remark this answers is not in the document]
feltech: Ah, this is at least partially addressed by the "In data - Extended" section

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+To this end, the `supportedSchemaVersions` method (or similar concept,)
+from Option 2 could be orthogonally added without any conceptual
+conflict. This would have the added benefit of allowing hosts to also
+support a range of version/have backwards compatibility, which is a huge
+benefit.
+
+This stops short of negotiation, being merely introspection similar to
+our other manager introspection methods.
+
+### Reconsidering Schema Version
```
feltech, line 352: [consider] I think this section could be part of the "Relevant data"

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+  else:
+    raise RuntimeError()
+
+```
+
+#### Pros
+- Both managers and hosts have the option to be backwards compatible.
+- More flexible, traits of different versions can be mixed in the same
+  session, which could allow hosts/managers to perform partial upgrades
+  in the case they are blocked on small specifics.
```
feltech, line 310: [consider] Also
```suggestion
- More flexible, traits of different versions can be mixed in the same
  session, which could allow hosts/managers to perform partial upgrades
  in the case they are blocked on small specifics, and allow components 
  of a system to each work with a different schema version and merge
  their results.
```

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+    trait = mediacreation.legacy.traits.content.LocatableContentTrait_v1(success_result)
+    trait.setLocation(managed_filesystem_locations[ref.toString()])
+    successCallback(idx, success_result)
+  else:
+    raise RuntimeError()
+
+```
+
+#### Pros
+- Both managers and hosts have the option to be backwards compatible.
```
feltech, line 307: [consider] I think I'd remove this, since negotiation also has this Pro, it's just that backward compatibility is ahead-of-time vs. just-in-time?

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
feltech: OK, so the next Pro "helps especially in the case of second-order dependencies" kinda covers this addition, though I think its worth expanding on it.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+- More flexible, traits of different versions can be mixed in the same
+  session, which could allow hosts/managers to perform partial upgrades
+  in the case they are blocked on small specifics.
+- Hosts don't have to interrogate and declare which library versions
+  are in use, which helps especially in the case of second-order
+  dependencies.
+- Seems to fit with existing OpenAssetIO design, especially with
+  `BatchElementError`, as an incompatible version wouldn't necessarily
+  need to error the whole batch.
```
feltech, line 316: [consider] Related point
```suggestion
- Seems to fit with existing OpenAssetIO design, especially with
  `BatchElementError`, as an incompatible version wouldn't necessarily
  need to error the whole batch.
- Mismatched schema versions in use by the host and manager don't 
  entirely preclude them working together from the outset, as long as
  the versions of individual required traits are compatible.
```

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+#### Cons
+- No "primary" version across the session, increasing implementation
+  complexity.
+- Introduces "Try and see" negotiation on the host side, although
+  introspection behaviour to negate this could be introduced at a later
+  date, similar to other introspection mechanisms. (See Option 3 -
+  Extended)
+- Would likely require some utilities in `traitgen`, otherwise, due to
+  each new version effectively being a brand new trait, constructing the
+  correct trait view would become a tedious dance.
```
feltech, line 329: [consider] I would couch this more as a performance consideration, since you _could_ apply all the trait views for all supported versions and check `isImbued`
```suggestion
- Applying trait views for all supported versions to find the correct one introduces a performance pessimisation, and would likely require some utilities in `traitgen` to optimise this process.
```

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+### Cons
+- Complex dance needs to be done up front, people may end up just
+  skipping it.
+- Still muddies the water for naive call store and recall managers. Does the
+  declaration of a supported version set implicitly exclude all other
+  traits in other traits & specs libraries. Reintroduces the spectre of optionality.
+- Manager reporting a supported schema version could imply every trait
+in that schema is supported, despite the need to still call
+managementPolicy.
+
```
feltech, line 232: [consider]
```suggestion
- Requires manager plugins to be stateful (at least at `initialize()` time), which goes against our philosophy (even if in most cases its practically unavoidable).
```

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+  in the case they are blocked on small specifics.
+- Hosts don't have to interrogate and declare which library versions
+  are in use, which helps especially in the case of second-order
+  dependencies.
+- Seems to fit with existing OpenAssetIO design, especially with
+  `BatchElementError`, as an incompatible version wouldn't necessarily
+  need to error the whole batch.
+- No new API, core `openassetio` library remains version unaware.
+- Cannot be ignored, unlike other "layered on top" mechanisms.
+
```
feltech, line 319: [consider]
```suggestion
- Prior art (e.g. OTIO) encodes versions in their data structure, presumably for good reason.

```

feltech (commented):
Really like this DR, very well introduced, well written in general, and much easier to reason about than the previous one.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+- Both host and manager can choose to be backwards compatible
+- Decides on version up front, reducing low level trait by trait branching.
+
+### Cons
+- Complex dance needs to be done up front, people may end up just
+  skipping it.
+- Still muddies the water for naive call store and recall managers. Does the
+  declaration of a supported version set implicitly exclude all other
+  traits in other traits & specs libraries. Reintroduces the spectre of optionality.
+- Manager reporting a supported schema version could imply every trait
```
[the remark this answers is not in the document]
elliotcmorris: Yeah, I think I agree, I think I was reaching for cons, but you've provided more here that illustrate this point better. Will remove.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+
+However, we do not believe this necessarily excludes any handshake style
+interactions. OpenAssetIO already provides a mechanism to configure
+managers at the beginning of the session with
+[initialize](https://openassetio.github.io/OpenAssetIO/classopenassetio_1_1v1_1_1host_api_1_1_manager.html#aa52c7436ff63ae96e33d7db8d6fd38df).
+Whilst we agree that state must not be utilized during a session,
+injecting version state before or at initialize time remains a valid
+option.
+
+### Naive managers
```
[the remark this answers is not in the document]
elliotcmorris: This critique only holds if versioning is optional right? In that, in a world where versioning was always required, the distinction between whether a trait does or does not have a version isn't there, you always have to assume there's a version and store it, the only difference being that it's easier if it's in the data because you don't have to do anything extra.

If so, I feel an additional con on out-of-data would convey that better, rather than mentioning that specific behaviour here.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+A pitfall that many of our earlier sketches fell into was assuming
+managers would have a first class understanding of the traits they were
+being asked to persist, which led to us assuming they would be able to
+perform upgrade and downgrade behaviour on them.
+
+This isn't, nor has it ever been this case, and the idea that a manager
+can and should persist data it doesn't understand has been a long help
```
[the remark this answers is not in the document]
elliotcmorris: addressed in : https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/ad29c0f4d97814f8ee0380a281ee6c6d36bc63d4

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+
+- **Status:** To be decided
+- **Impact:** High
+- **Driver:** @elliotcmorris
+- **Approver:** @felltech @elliotcmorris
+- **Outcome:**
+
+## Background
+This DR follows on from a [previous
+attempt at a DR](https://github.com/OpenAssetIO/OpenAssetIO/pull/1254), which sought
```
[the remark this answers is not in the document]
elliotcmorris: addressed in : https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/ad29c0f4d97814f8ee0380a281ee6c6d36bc63d4

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+  need to error the whole batch.
+- No new API, core `openassetio` library remains version unaware.
+- Cannot be ignored, unlike other "layered on top" mechanisms.
+
+#### Cons
+- No "primary" version across the session, increasing implementation
+  complexity.
+- Introduces "Try and see" negotiation on the host side, although
+  introspection behaviour to negate this could be introduced at a later
+  date, similar to other introspection mechanisms. (See Option 3 -
```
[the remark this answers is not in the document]
elliotcmorris: addressed in : https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/ad29c0f4d97814f8ee0380a281ee6c6d36bc63d4

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+  Extended)
+- Would likely require some utilities in `traitgen`, otherwise, due to
+  each new version effectively being a brand new trait, constructing the
+  correct trait view would become a tedious dance.
+- EntityTraits becomes more complex, as we either have to declare it
```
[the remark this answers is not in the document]
elliotcmorris: addressed in : https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/ad29c0f4d97814f8ee0380a281ee6c6d36bc63d4

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+  # Record the schema versions that the host is going to use.
+  # We can use these to branch our internal functionality
+  this.__mediacreation_schema_version = schemaVersions[openassetio_mediacreation.packageId()]
+  this.__companycorp_schema_version = schemaVersions[companycorp_traits.packageId()]
+```
+
+### Pros
+- Both host and manager can choose to be backwards compatible
+- Decides on version up front, reducing low level trait by trait branching.
+
```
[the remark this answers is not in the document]
elliotcmorris: Taken verabim : https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/016807eb1d426bb95c2770333864ccf189d7d6c5

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
elliotcmorris: Removed : https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/cea60bd734c0e023d658bec2a4df1e6f963bfc07

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
elliotcmorris: I don't think that's entirely true, although I know what you're getting at.
The decision we make now will define the pattern and methodology of version communication, and even if we do add out-of data / in-data utilities after the fact, they would remain utilities, the paradigm would stay the same.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+        context,
+        hostSession,
+        successCallback,
+        errorCallback,
+    ):
+
+  # This function is imagined a pre-provided utility, possibly generated
+  # via traitgen, that would error in some way if the trait set does not
+  # contain the locatableContent trait.
+  locatableContent_version = locatableContent_version_from_trait_set(traitSet)
```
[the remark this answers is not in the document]
elliotcmorris: In any case it seems to me either way we're imagining new mediacreation/traitgen functionality to make the example work.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
elliotcmorris: Yeah, very good : b90709aaeba5d04546a80d65323bbde781c6a235

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+  date, similar to other introspection mechanisms. (See Option 3 -
+  Extended)
+- Would likely require some utilities in `traitgen`, otherwise, due to
+  each new version effectively being a brand new trait, constructing the
+  correct trait view would become a tedious dance.
+- EntityTraits becomes more complex, as we either have to declare it
+  version agnostic somehow, or managers would need to return the correct
+  versions/entire swathes of versions.
+- Unclear how specification versioning is to work without an out-of-data
+  version.
```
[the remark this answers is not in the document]
elliotcmorris: Isn't this a bit too strong? Specifications will still "work", they'd just be unversioned, and unable to be versioned with a direct in-data representation. I grant the significant rework think is probably worth saying.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+
```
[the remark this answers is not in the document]
elliotcmorris: This is true enough, although I don't want to give the impression that we wouldn't take on this work, it's significant yes, but not out of reach to have a version aware traitgen able to present combatible views across non-breaking changes.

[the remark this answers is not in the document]
elliotcmorris: Reworded this to try and be more specific about what redesign exactly we're talking about, as I think it's worth drawing a distinction between core library redesign (as we advertise it as stableish) and traitgen redesign, which is much more open to change.
d37bd4d855182c73e85bdd7eb3d55ca5d1c5f518

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Resolve was a given in my mind, the reason to call out `entityTraits` to my mind is this is a specific drawback of an in-data solution.`entityTraits` is asking like "what is this entity, what does it do?", and when asking that question, the version information seems _almost_ redundant. The in-data approach forces versioning in a first class concern of every method, as you say, but that uniquely influences `entityTraits` imo.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Aye, fair enough, moved it : 21688f11635f47e872d6976902f14a65af81d2da

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Seems reasonable, taken verbatim : 5ac410a68f871434dcd51bdec7855e6f828cc7b5

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Yeah you're right, holdover from when there was more than 2 options. deae53fb12c77c2f16365f60114d7a2d92697101

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Isn't this just restating the "more flexible" point at the top of this list?

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Surely it's both then, applying all the trait views for all the supported versions and checking `IsImbued` is definately a tedious dance. I wouldn't say the performance aspect is the main problem here, but I'll fold it in.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: I like it, taken verbatim : b04b02a0b16012fa7269b4b1958f107c6ad15e83

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+> trait, something like
+> "`openassetio-mediacreation-v-4:LocatableContent`". This would
+> effectively make each version of a trait a new trait as far as the
+> `openassetio` core library is concerned.
+
+
+### Host side
+```python
+# The host side implementation is almost entirely unchanged from
+# from standard, excepting a new BatchElementError type
```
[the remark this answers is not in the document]
elliotcmorris: Added quite a long section into the in-data option about this. Worth a read.
68d887e855b6c8cdb15e517abd61b9af8573c3f2

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,365 @@
...
+  return
+
+# Tell the manager up front what versions you are going to use.
+# These must be versions reported from `supportedSchemaVersions`.
+negotiated_schema_versions =
+{
+  openassetio_mediacreation.packageId():mediacreation_version,
+  companycorp_traits.packageId():companycorp_version
+}
+manager.informSchemaVersions(negotiated_schema_versions)
```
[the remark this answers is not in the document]
elliotcmorris: I see what you mean here, but I feel putting it in has made the example a bit less clear, bringing questions of python style imports vs how this would work in c++, as well as just making the example a fair bit longer. I'm gonna leave this out.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,412 @@
...
+  in that schema is supported, despite the need to still call
+  managementPolicy.
+- Managers and hosts now have additional data related to specific
+  entities that they must communicate, increasing the complexity of
+  "pass-through" hosts and managers, and complexity of data transfer
+  across multiple hops and/or storage in multiple systems. This could be
+  especially problematic for naive managers that simply wish to be data
+  recorders, as they are almost forced to be version aware.
+- Requires manager plugins to be stateful, at least at `initialze` time.
+  See [no internal state](#no-internal-state).
```
tomc-cinesite, line 259: [fix] Add:
- Hosts or Managers may not know the versions of packages used in the case of code that works at the TraitsData level (eg BAL, and/or no packages at all), or proxying input from another source. Eg user input.
- Hosts may need/wish to query multiple versions of a particular trait (eg a plugin running inside a DCC with de-coupled code bases).
- Precludes host-host data exchange where the hosts are the only ones that understand the schemas.
- Precludes per-trait versioning which requires prior-knowledge of what traits are compatible between package versions.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
tomc-cinesite: Feels reasonable given the alternative proposal imagines significant core API changes?

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,412 @@
...
+- Seems to fit with existing OpenAssetIO design, especially with
+  `BatchElementError`, as an incompatible version wouldn't necessarily
+  need to error the whole batch.
+- No new API, core `openassetio` library remains version unaware.
+- Cannot be ignored, unlike other "layered on top" mechanisms.
+- Managers not required to consider storage/communication of additional
+  version data to enable interpretation by future downstream systems.
+- Prior art (e.g. OTIO) encodes versions in their data structure,
+  presumably for good reason.
+
```
tomc-cinesite, line 351: [fix] Add:
- Allows per-trait versioning, which avoids the need or "insider knowledge" as to which traits actually change in which package versions.
- Stateless
- Allows 'simple' managers and hosts to be written that don't understand specific data.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
tomc-cinesite: It might be good to clarify what you mean by "un versioned". 

The existing specifications approach would inherently version the traits - without change. What you wouldn't get with the current design was the ability to version the specifications themselves in a way that was visible in the resulting data, which is certainly a con against pre-negotiation of a package specific version.

At present however, Specifications deliberately don't exist at the data/core API level. They're just shorthand for a particular trait set.

Adding the ability to store alternate trait sets as different versions of a specification in the generated view classes doesn't really seem any more onerous that generating multiple classes for the trait views themselves. 

If you want to go from a trait set to a spec view class, it doesn't seem any different from as it stands without versioning- in that there is no concrete 1:1 mapping anyway. There just may be n number of well-known specifications that overlap a given trait set. There are just more known view classes if you have versioned specs. That problem exists beyond the scope of versioning I think as it will always run up against one traitset being a subset of another (which is deliberate, by design).

The main downside seems to be about having hosts/managers agree on _specifications_ as an atomic unit instead of a trait set, which could be possible with package versioning and the out-of-data mechanism.

I think perhaps a fairer version of this is that with in-data versioning, specifications work exactly as they do now, but out-of-data has a pro, that it would be possible to have more consensus about which specifications to use directly.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,412 @@
...
+- Manager reporting a supported schema version could imply every trait
+  in that schema is supported, despite the need to still call
+  managementPolicy.
+- Managers and hosts now have additional data related to specific
+  entities that they must communicate, increasing the complexity of
+  "pass-through" hosts and managers, and complexity of data transfer
+  across multiple hops and/or storage in multiple systems. This could be
+  especially problematic for naive managers that simply wish to be data
+  recorders, as they are almost forced to be version aware.
+- Requires manager plugins to be stateful, at least at `initialze` time.
```
tomc-cinesite, line 258: [fix] Only if this is done as a pre-negotiated mechanism, nothing to stop you putting the versions list in the Context instead of `intialize`.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,412 @@
...
+- Cannot be ignored, unlike other "layered on top" mechanisms.
+- Managers not required to consider storage/communication of additional
+  version data to enable interpretation by future downstream systems.
+- Prior art (e.g. OTIO) encodes versions in their data structure,
+  presumably for good reason.
+
+#### Cons
+- No "primary" version across the session, increasing implementation
+  complexity.
+- Introduces "Try and see" negotiation on the host side, although
```
tomc-cinesite, line 355: [question] Is this strictly true as the existing `managementPolicy`methods would , without change, allow introspection of support?

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,412 @@
...
+its strength, but it necessitates a "try and see" approach on the part
+of the host. If the host wants to be adaptive to the manager, it may
+benefit from a more up front introspection of the supported manager
+versions.
+
+To this end, the `supportedSchemaVersions` method (or similar concept,)
+from Option 2 could be orthogonally added without any conceptual
+conflict. This would have the added benefit of allowing hosts to also
+support a range of version/have backwards compatibility, which is a huge
+benefit.
```
tomc-cinesite, line 384: [question] Any reason that `managementPolicy` can't be used as-is for this?

tomc-cinesite (commented):
Good re-work! Sorry bit short on time, so I've probably missed some things on one side or the other.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
feltech: :laughing: That's not the con that I was highlighting, I was referring to

> Manager reporting a supported schema version could imply every trait
> in that schema is supported, despite the need to still call
> managementPolicy.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
feltech: I think we basically assume a negotiation mechanism, because simply communicating the supported versions from host to manager violates some requirements (as mentioned in the sentence just above the `### Negotiated` section)

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
feltech: > Feels reasonable given the alternative proposal imagines significant core API changes?

I'm not sure which you're saying feels more reasonable, inventing utility methods vs. using existing methods for the example?

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
feltech: I was trying to contrast this to ahead-of-time negotiation, where it's all-or-nothing. I agree it's more-or-less covered by the first point... I think perhaps the first point is covering too much under the "more flexible" umbrella, and could be split into multiple Pros.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
feltech: I would keep the `/` in `future/downstream`, or use `future and/or downstream`.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
tomc-cinesite: I was just meaning that assuming some future development in order to make the example seems representative of a possible future seemed reasonable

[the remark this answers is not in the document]
tomc-cinesite: .. eg using `LocatableContentTrait.imbuedVersion` or similar

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
tomc-cinesite: I guess it feels like all the options violate something somewhere so is it definitely worth ruling out that as an alternative tradeoff? ie if you dont like 'in-data' but also dont want to become stateful?

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
themissingcow: I think this is a huge con, capturing state at initialize only doesn’t make it any less stateful, and makes it almost impossible to write a host that parameterises its input.  Eg the user specifies which trait to resolve - as everytime the uses changes that you’d need to reinit the whole manager.

Not sure I follow “practically unavoidable”? There feels to be a huge difference between settings as to what back end to talk to, and something that is effectively saying that you’re changing the meaning of an every requests inputs.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
themissingcow: I think the “who’s in charge here” thing is key. 
If it’s this becomes a democratic negotiatiation to a happy place, then the code is in charge somehow (and that feels confusing as it’s not clear from an API perspective who ultimately makes a decision). An alternative way to look at it, which is what you get in file world, (which is more similar than perhaps thought, if you abstract the transport layer as we do and desire caching middleware) where the **data** is in charge. The host will provide vX of data, the manager may provide vY of data, its at the point that specific interpretation is needed where conversion lies, not the transport layer. So either end can be “simple” as it it’s just becoming part of the transport. Rather than reading both parties to be active, all the time, which creates the “swathes” problem. `entityTraits` then is simply either the actual version of the data - in the simple BAL case, or “the preferred/only version of the code used by the manager” in dynamic translation cases.

A key thought experiment for me on this all would be “what does BAL, simpleResolver and a the generic relationship/entity browser look like with these two approaches. And critically how does their use/functionality change.

Re versioning being redundant in entity traits - if different versions of traits are incompatible the is versioning actually very important in the entity traits method? As an entity having the qualities of one version or another may be very different! Eg adding a property to a trait would mean entities with the older version are inherently not going to have that data - which may be required (I think we suggested leaning towards small traits with only required properties right?)

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
themissingcow: > At present however, Specifications deliberately don't exist at the data/core API level. They're just shorthand for a particular trait set.

Ruminating on this a little - wondered if it might be worth just highlighting why this is currently the case. As it might suggest whether knowing specification versions is actually important or not beyond programmer convenience.

The reason they’re not expressed at the API level is that the interoperability concept of the API relies specifically on specific sets of _traits_ not the specification that happened to make some data, to make programmers lives easier. In the majority of cases one side of the equation only probably cares about a subset of these traits in particular - and this is what allows interop.

For example, the read node in nuke only cares that entity has locatable content. In the future, it may care about image related traits. As long as target entity has those traits as a _subset_ of its traits then all is well. Nuke may use an ImageSpecification of sorts for consistency, but it doesnt matter if the original entity was authored by some DelightRenderAOVSpecificstion view class or by hand in a test case json file, or made up on the fly, as long as it has compatible versions of the specific traits - locatable content etc… 

There is the mirror in publishing where a manager may only want to know if it has locatable content to store and just stash the rest for filtering later.

There are certainly cases where specific combinations may be required, and versioning the classes used can help with alignment, but ultimately it’s just the traits themselves that actually make any meaningful difference - which is why specifications don’t exist at the API level today. All a spec is doing is helping check which traits are there.

Something @elliotcmorris often mentioned is - why do we even need them? I imagine we probably don’t, compatibility of data is implicit in the trait set so doesn’t need another layer, they’re just there to help communicate which traits are mutually agreed on by some subset of a user base when alignment is helpful.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,412 @@
...
+of presumption is the way forward with an in-data approach. it is
+possible to pack version data into `traitID` and yet not presume that
+`mediacreation-v1-locatableContent` is a conceptually different trait
+from `mediacreation-v2-locatableContent`.
+
+There are reasons we might want to do this, for example `entityTraits`
+and perhaps even `managementPolicy` could potentially benefit from being
+unversioned. Does a user want to know that an entity supports
+`locatableContent-v2` specifically, or are we simply asking "What is
+this thing?", in which case, an unversioned, or "all-version" method of
```
themissingcow, line 405: I think there are some really interesting ideas there - you could even say that `resolve` with no version means “whatever you have”, I think `entityTraits` for read would be useful with versioning info as then you know what data you might be able to get (which helps significantly with migrations or user-facing comparability feedback, and writing generic browsers/views)

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
elliotcmorris: > Adding the ability to store alternate trait sets as different versions of a specification in the generated view classes doesn't really seem any more onerous that generating multiple classes for the trait views themselves.

I agree on this, this would be the "significant rework", but I don't think it's enormous. it's more of a consistency problem, and entirely within the domain of programmer convenience. Given an in-data approach, how does a method know which specific version of a specification it should be expecting to work with, you're left with a very nuts and bolts try-and-see approach, is a host expected to scan through managementPolicy with every possible specification version until it gets the one the manager happens to support? Seem onerous, and is not something that would be expected currently.

>  think perhaps a fairer version of this is that with in-data versioning, specifications work exactly as they do now, 

Whilst true, surely specifications must eventually be versioned somehow if they are to remain useful. If anything, I would expect the structure of specifications to be the most volatile over time, if specifications end up being used as the first class units of exchange they are intended to be. The downside of in-data versioning is that as a strategy, it will not work on specifications precisely because they are not concepts that exist in the api. This strikes me as a downside, as we were expecting an hoping for a solution that worked for both traits & specs (which you can understand when you view this exercise from the perspective of adding versions to the items defined in the .yml file)

> ultimately it’s just the traits themselves that actually make any meaningful difference - which is why specifications don’t exist at the API level today. All a spec is doing is helping check which traits are there.

Something @elliotcmorris often mentioned is - why do we even need them?

I continue to mention this, to the disdain of my colleagues. It's something i've never quite been able to take on faith, the usefulness of these trait groupings, as from what I've seen, and what I expect, managers and hosts will _mostly_ treat traits as atomic units, and perhaps have some exceptions for certain groupings (traitsets,) but i'm not sure it'll be the standard practice, so I continue to question the usefullness of specifications in general, and wonder if their inability to be versioned should stand in the way here.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: > I think this is a huge con, capturing state at initialize only doesn’t make it any less stateful, 
Even capturing _this_ state? It seems somewhat contrived to me that the trait schema version a host  

> Not sure I follow “practically unavoidable”? There feels to be a huge difference between settings as to what back end to talk to, and something that is effectively saying that you’re changing the meaning of every requests inputs.

The feeling of the state isn't relevant to that phrase, settings _is_ state, and thus managers are already stateful.
I don't personally see it as a major step up in state myself, settings will majorly alter the output data returned from the manager. The level of idempotence is preserved as post-initialize idempotence. What am I missing, why is this such a difference?

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: I believe you would still need to do a try-and-see implementation pattern though, in that you'd need to provide a set of all possible versions to `managementPolicy` and have it tell you which ones it likes.

However, i'll grant I wasn't thinking of that when I wrote this con, and that approach is not really any less maintainable than a "ask for what versions are good" approach, at least from a hosts perspective. Making `managementPolicy` even _more_ complicated for the manager though ... eesh.

What do you think @feltech? I am tempted to remove this as a con based on this observation.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Probably not,  see https://github.com/OpenAssetIO/OpenAssetIO/pull/1267#discussion_r1506772966

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: We benefit from having generated trait views here, as it gives us options. I can imagine an "all-versions" view which presents the interface of every property that has ever existed in the history of the trait, but errors/excepts with version errors if the data dosen't line up. We'd have to think what that means re: `isImbued` etc, but the compatibility is appealing, and more than that, the idea that we can segregate versions away from identity is appealing. 

I can't help but thinkg that it's not gonna be a fun time if you're trying to find out if an entity is an image, but you get `false` results because _actually_ your entity is `imagev4` and you asked about `imagev3`, (because your build step hadn't ran traitgen recently enough, or whatever). I also worry about the difficulty of implementation on the managers side for version aware introspection methods, could get hairy.

I fear making this decision is going to take months and months...

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: I wouldn't rule it out, seems like a decent possibility to me. I swear we discussed it at some point ... but things have moved so much from then it was probably in a completely different ... no pun intended ... context.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
themissingcow: > Given an in-data approach, how does a method know which specific version of a specification it should be expecting to work with,

That's what I was trying to get at in the second reply - it doesn't matter what specification is used, as long as the traits match. There never will be a 1:1 mapping of a trait set to a specification, due to the whole "add more traits to be more specific" thing. The traits are the only important thing.

>  perhaps have some exceptions for certain groupings (traitsets,)

To me thats all they're for - they're a "light weight convenience" not a core-currency, so I totally agree with you, that if it makes this versioning thing easier to grok, just ignore them! I really don't believe they belong at the API level, just as you said - to help communicate expectations for certain groupings.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
themissingcow: Yeah true, for me, there are two differences that help try to ensure we can take advantage of a myriad of more modern/scalable design patterns:

 - Settings are entirely optional. It is possible to write a totally functional manager without them - if all the data you need is in the entity reference. This could be true for some file-path based manager, or where all back-end info is held in the ref itself.

- Settings are intended to change how/what data is retrieved, not the shape of it is for a specific request. It's like the DNS server setting on your computer, yes you may get a different IP address back when looking up a host name, but you still get an IP address back. The nature of the returned data is solely dependent on what you feed into the request, its values may change, but not the shape of it.  The proposed changes mean that resolving a trait may get different properties (by name) back depending on the state at initialisation - so the shape _and_ the values for any given 'trait' are changing.  This is the very reason the HostSession is passed as a parameter _not_ just at initialise time.

As an example, let's pretend I'm a host wishing to migrate some data from v1 and v2 of an entity trait (I have to do this as no-one else knows how to do that). With the out-of-data approach, because (if I understand the proposal correctly) the manager is now stateful with respect to the input parameters of the request (trait set in the case of resolve), and that state can only be set  at initialise time, I'd have to have n separately initialised manager _instances_ for each trait version that I need access to. If the manager wasn't stateful in terms of versioning, I'd be able to make, at worst n _requests_, and at best, a single request to retrieve data for multiple versions. 

I think this is the clarification we tried to make [here](https://github.com/OpenAssetIO/OpenAssetIO/pull/1151), but probably failed. There is inherently going to need to be some state in the "instance" in many real world situations - for bootstrapping communications with another service, but that is the intended limit to that scope, it wasn't meant to be state that affects the Host <> ManagerPlugin interaction. Just  ManagerPlugin <> private implementation.

Once very concrete reason for this was to allow the existence of additional/alternate middleware that allow shming/spoofing etc... just as you see with scalable web services, their ability to be scaled (which is usually via middleware that short-circuits or distributes evaluation) is often down to whether a requests encapsulates all of the required inputs to formulate a response.

OpenAPI is a great example of this, if you try to write an endpoint definition for this so you could write a plugin that talks to any server implementation, as stands you can pretty much entirely have it working aside from two problems:

- Logging
- HostInterface

They're both ways in which we break/blur the lines of "everything you need is in the request". Logging isn't so bad, as you can just capture the logging and send it back in the response as it's not a two-way thing. HostInterface however is a problem.  It is, at least, part of the request itself, but it technically now requires a two-way channel. So it's better than being stateful, but is still a significant problem for simple, distributed systems. The current implementation (well gPRC, but it's the same thing) captures the identifier/display name and basically makes the hostinterface just another dict.  

If we had additional state that wasn't included in the request (ie captured at initialise), the implementation would just end up putting it back in the request so the server could build the response, or end up having creating a session on the server and capture a handle to that in the request - as the server is servicing multiple independent clients. I think active sessions with servers have been long proven as not being a scalable approach.

To me both of these illustrate how this data belongs with the request one way or another at least, and that a drive to make it initialisation-time state seems to perhaps be driven more out of convenience for a specific programming paradigm, rather than a consideration of the underlying data model.

Putting it in the context would help, (or even host interface, but that forces two-way coms which have other issues).

fwiw

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
themissingcow: And I suppose it surfaces a potential con that management policy is meant to be entity agnostic (as its about what the  manager can support), and that different entities in a simple manager may not have the same version. But I guess it has always been the case that not every entity would have any specific trait though even though "it should have", so maybe its no different to the situation were already in... as 'simple' managers usually respond more generally ie "yes to everything".

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
themissingcow: > get false results because actually your entity is imagev4 and you asked about imagev3

If we're saying Image v4 is a breaking change from Image v3 then can you definitely say they are equivalent and the false is probably correct? Or if they are suitably equivalent - did the traits need a new version?   Did I hear OTIO don't bump versions for additions? As we already have too manage missing properties where data is consumed.

themissingcow: > I fear making this decision is going to take months and months...

Being cheeky: pick something that already exists instead of making a new thing, OTIO/USD/files all seem to be going in the same way, and that way also means its not a core API concern.

themissingcow: Another potential benefit of aligning with in-data and OTIO or USD as closely as possible is that it’s not another thing for people to learn. 

If I understand correctly, it looks like it would also be a lot more straight forward to store/recall data from one in the other. Eg OTIO meta data or USD prim schemas could all become traits and visa versa.

elliotcmorris: > Being cheeky: pick something that already exists instead of making a new thing

We're not trying to invent something new, really, it's just that OpenAssetIO has uniquenesses about it. We tried to just do what seemed simplest and most normal to us in the last DR to poor results...
The in-data approach here, which is what i'm leaning towards currently, _is_ a thing that already exists right, we've seen it around. It's mainly just its interaction with specifications that gives me pause.

> At the risk of being glib, I bet you could probably borrow the entire OTIO upgrade/downgrade mechanism and point it at at TraitsData with minor changes

As far as I recall, OTIO has a central registry of all possible types that can exist, and has strict requirements about the existence of upgrade/downgrade conversion methods. Ie, if you change the schema, they enforce the requirement for upgrade/downgrade methods to be committed along with the schema change, so it's not 1:1. The main thing is that we don't have visibility or control of external trail libraries that may appear, so we can't make strict guarantee about version upgrades occurring.

Nonetheless, this dosen't preclude in-data upgrade abilities in the trait libraries (ie mediacreation,) and we definately want to add those, but we're trying to limit the scope at first to a mechanism that can work without complex automated utilities, (this is based on TSC feedback, that for the moment, managers at least would be happy to do the conversions themselves. ... Hosts are perhaps another story, we wern't thinking about hosts doing conversions at that point.)

A thing to note is that writing those methods is more burdensome on contributors/us that it is on OTIO, because it must happen at the trait library level, we'd have to support at least 2 languages, ie 2 versions of each upgrade/downgrade function.

Also, the question of speed of iteration came up in regards to this, do we really want to enforce the need for upgrade/downgrade functions to be written when we know mediacreation is early days and requires iteration?

themissingcow: > > Being cheeky: pick something that already exists instead of making a new thing
> 
> We're not trying to invent something new, really, it's just that OpenAssetIO has uniquenesses about it. 

I guess that was the cheeky part - is it _really_ that different if you step back far enough? I just has extras that can stray into think that it is different? An active, out-of-data version negotiation mechanism just felt relatively "new" in comparison to the prior art in the industry doing on the whole, similar things?

> The in-data approach here, which is what i'm leaning towards currently, _is_ a thing that already exists right, we've seen it around. It's mainly just its interaction with specifications that gives me pause.

Sure, personally I think the reality of specifications is maybe just that they're closer to what you've always talked about than perhaps they are being portrayed. Keeping them a 'programmer convenience managed by traitgen' feels to work for me, and then keeps them fairly compatible. Forcing the conversation about trait sets between independent parties. And you can then use specs as a bookmark for formally defining which one you need. Having a runtime registry of classes to trait sets is pretty straight forward (we had that before). Hosts/Managers shouldn't get too hung up on 1:1 specification matches, as that defeats the composition principal.

> As far as I recall, OTIO has a central registry of all possible types that can exist, and has strict requirements about the existence of upgrade/downgrade conversion methods. Ie, if you change the schema, they enforce the requirement for upgrade/downgrade methods to be committed along with the schema change, so it's not 1:1. The main thing is that we don't have visibility or control of external trail libraries that may appear, so we can't make strict guarantee about version upgrades occurring.

When I meant glibly "take OTIOs mechanism" - I meant take _all_ of it 🙃.
 
> Nonetheless, this dosen't preclude in-data upgrade abilities in the trait libraries (ie mediacreation,) and we definately want to add those, but we're trying to limit the scope at first to a mechanism that can work without complex automated utilities, (this is based on TSC feedback, that for the moment, managers at least would be happy to do the conversions themselves. ... Hosts are perhaps another story, we wern't thinking about hosts doing conversions at that point.)

I guess it felt like it all works without them, just you don't get stuff for free without code that makes it free? It's purely a choice that we make as end users as to how compatible/rigourous you want to be with inter for versions. Some hosts/managers may refuse to support anything other than the latest etc... regardless of what mechanisms you put in place - simply to avoid being in a world of infinite permutations to maintain.

> A thing to note is that writing those methods is more burdensome on contributors/us that it is on OTIO, because it must happen at the trait library level, we'd have to support at least 2 languages, ie 2 versions of each upgrade/downgrade function.

For sure - I wonder would it be any more effort than the whole mechanics and ongoing maintenance in every implementation of new core api methods and/or the negotiation handshake?

> Also, the question of speed of iteration came up in regards to this, do we really want to enforce the need for upgrade/downgrade functions to be written when we know mediacreation is early days and requires iteration?

I guess thats a choice of when you decide to change the version number, rather than the mechanism? Like how we treat alphas and betas differently to post v1 server?

themissingcow: > but we're trying to limit the scope at first to a mechanism that can work without complex automated utilities, (this is based on TSC feedback, that for the moment, managers at least would be happy to do the conversions themselves. 

Thinking further on this - In a way - is that a benefit of the in-data approach? It can be explored with, I believe, no changes to the core API (other than potentially documentation), and manageable iterations to traitgen and select implementations that wish to participate.

With the in-data approach, If managers wish to do the work in the first instance, a host could leave its code essentially unchanged (bar picking up a new package version and perhaps minor class name changes), the manager can use version introspection mechanisms in traitgen code to see which versions a host is using and adapt accordingly if required - the manual work suggested. 

This would potentially only need some changes to trait-get and how the yaml is defined. Managers could even work with existing un-modified hosts treating them if traitgen treated them as "v0" or some such.

With the out-of-band approach, it looked like both parties would have to do some work to add in negotiation, and you'd need core API changes to introduce the mechanism.

It seems like either way, in the long run if host or manager code needs to "consider" versions, and there doesn't seem to be substantially different orders of magnitude of work between either solution, but the impact of change is significantly different.

There is one other perspective that, what ever the initial outcome of this decision process is, it will be considered lacking from _some_ world view or another and need refinement. The in-data solution seems to have a much smaller blast-radius than an actively negotiated mechanism that touches the core API. This would allow it to be more easily iterated through practical application in a much smaller part of the ecosystem, and also be much more amenable to future changes.

To me facilitating iteration in a more decoupled, less intrusive way is quite a significant benefit, given the uncertainly about the decision to date.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: So is the issue here more that like, it is no longer possible to be stateless, as some managers may at the moment, not require settings, so can be stateless. However versioning _is_ a proper hard requirement to be passed through initialize, so even simple/explicitly engineered managers can't be stateless?

themissingcow: Yeah,  if you wish to communicate to a stateless back end to fulfil the request, you have to capture the state and put it in the request anyway... something like a first-party OpenAPI definition would have to do the same.

The point about a host also needing multiple manager instances highlights the impact on the other side of the fence too.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
feltech: There's then perhaps a separate Con that "unversioniong" of some methods may be required.  

You could conceivably apply the argument to `resolve` - if you assume that the host can branch on the `TraitsData` traits returned from `resolve`, then the input to `resolve` could be unversioned trait IDs.

I think my argument that `entityTraits` isn't necessarily a special case holds. It's perhaps more susceptible to a rethink of whether it should be unversioned than other methods, but not exclusively.

tomc-cinesite: Yeah. I guess one nice thing with the in-data option is it isn't an API signature change just a calling convention change, so the decisions around that are relatively easy to iterate. It might even be that both make sense. Eg a host resolves a specific version if that's all it supports, or leaves it unversioned in the utopian 'auto up/down grade traitgen world'

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
feltech: Yeah I think there is another good Con to tease out here

> Hosts that want to work with multiple schema versions (e.g. for a read-upgrade-save workflow) must initialize multiple manager instances.

Re. APIs, _some_ session-level state is inevitable (e.g. user ID), unless the API is completely unrestricted. That state _could_ be constructed anew with every request, or bundled in the `Context`, I guess. 

If `HostInterface` exposed the negotiated schema version, then statelessness is recovered, since the schema version(s) could be sent with every request (rolling back on previous assumption that `HostInterface` was off limits).

themissingcow: > Re. APIs, some session-level state is inevitable (e.g. user ID), unless the API is completely unrestricted. 

This reminded me of something we didn't quite get into when we were chatting last week. I think it's really, really, important to make sure this is not the case. It is worth noting that "state" and "negotiation" are really two separate things in this scenario. It perhaps illustrates how we've been inadvertently combining them at times in our discussions.

Looking at the design of modern secure and zero-trust APIs (i.e. signed API requests/OAuth, etc), they share our goal of being stateless - partially for scalability reasons, but I think mostly for security reasons. Making an _endpoint_ stateful in a way that affects the handling of subsequent requests allows for someone to tailgate.

Im not saying there is no state in the entire system (otherwise you'd never need to talk to the manager in the first place), they key is that there is no state in the mechanism that facilitates interaction - ie the API mechanism _itself_. This is exactly what we're tying to achieve with the OpenAssetIO too. There is always some state - the asset data - but there shouldn't be any need for state in the API itself - as a requirement to introspect the real "state" i.e. the data. Which is why settings need to be optional, and solely exist to work around certain practicalities that arise from the design of some existing systems.

User ID is a great example of this, and one thats readily solvable without state in the API layer. When identity needs to be a first-class thing, we can just use what existing HTTP APIs do. Most zero-trust/secure APIs have initial negation (not dissimilar to what you've been talking about with the out-of-data approach), the difference is however, the result is not held as state in the endpoint, it is returned to the requestor, and provided with each subsequent request.  e.g you get an auth token, but you have to include that with every request - the endpoint doesn't remember which token a given client is using so it only has to send it _once_ - as that would invalidate it as a security mechanism.  

What we modelled in the MovieLabs workflow was exactly this - there was some negotiation (OpenLoginIO or whatever), that determines a barer token (often just a normal oauth flow) its not OpenAssetIOs job to get involved in this most likely. This token is then passed through the context so it is available to the implementation - which in turn passes that token along with the requests it has to make to the underlying back-end systems on a per-request basis. These back-ends aren't stateful with respect to the request source itself, which is why it has to be provided every time - they only validate the request based only on the data in the specific request itself (the token, the source ip etc, private keys etc). The validation may correlate this with data stored in the back-end (which is also "state"), but that state isn't in the API layer itself, so it is still atomic to each request. Ie, you could switch servers satisfying the request - as load ballancers may do.

Using this model - version negotiation could then always be layered on to in-data versioning, to determine which versions _should_ be used, if this becomes necessary or is desirable. It is just good to do this in a way that doesn't make the _endpoint_ stateful.  So you would take the result of the negotiation and use it to form the request for the agreed versions. This keeps the API itself stateless, and all the benefits that brings for distributed and scalable systems, whilst allowing any higher-level alignment to be agreed if specifically required in any given scenario.

The design of the Context and manager state object in the API already employs this mechanism. There is a negotiation (Manager::createContext) and the fact that the resulting context is held by the host, and given back in associated requests to the manager in order to correlate them, such that the manager knows "related" API requests without needing any API related state held within itself. This is what enabled remote gRPC to a shared server to be implemented with minimal effort compared to a per-instance server.

If you really didn't like the idea of the Host having to track that data - then thats where the Manager wrapper could come in, as it "handles tracking state" already - in the way that it auto-populates the host session etc... so that the ManagerInterface implementation conceptually doesn't know any different, and just sees it as part of the request.

I think the only thing that this approach really precludes is you can't "switch in your code once at startup", but given that security and the underlying data will most likely already require an implementation to deal with per-entity or per-request variation, it seems a worthy concession, if it means we follow the well-established patterns more broadly use across client/server APIs.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,412 @@
...
+
+However, we do not believe this necessarily excludes any handshake style
+interactions. OpenAssetIO already provides a mechanism to configure
+managers at the beginning of the session with
+[initialize](https://openassetio.github.io/OpenAssetIO/classopenassetio_1_1v1_1_1host_api_1_1_manager.html#aa52c7436ff63ae96e33d7db8d6fd38df).
+Whilst we agree that state must not be utilized during a session,
+injecting version state before or at initialize time remains a valid
+option.
+
+### Naive managers
```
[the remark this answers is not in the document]
elliotcmorris: This section is specifically about how managers should be able to persist data naively, and how we made a bad assumption regarding that last DR.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
elliotcmorris: Uh ... oops.
I'm actually going to keep both removals, they're both sort of saying the same thing, ie, "we'd have to doc something", which is your same criticism. 3fda5557971290eeecc91521f35d57a2de430b1a

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: [1] I don't believe this is a seperate con to saying that there's now additional data that must be communicated and stored, which is a con, but already listed. In the case of non-versioned data from another source, this would hit the same problem even with an in-data approach, a version would have to be invented.
[2] Seems valid
[3] This has never really come up before and i'm not sure it's a legit con. It dosen't preclude this concept, just means you have to write a somewhat generic negotiation mechanism in your manager, which I concede would be harder, but it seems awful niche for a con.
[4] I'm reading this as "Upgrading traits requires out-of-code knowledge of what's compatible", but I find the snippet hard to parse. Seems more like a critique of schema version vs trait by trait versioning, although I guess valid as trait by trait versioning would be harder with an out-of-data formulation. We'd try to solve this with mediacreation utils though.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Ok, fair enough it is covering quite a lot. I added the point.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/ac7d5b5a95c20e509edbfb2d9330b75b611bd2a7

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/964b40ca700c948a2ee025aa48ad828b19c67654

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: [1] Again this seems like a criticism of a schema version rather than in-data versioning, but also again, there is some overlap so will add.
[2] Fine, but rephrasing because we both know it's not that simple.
[3] This restates the above point "Managers not required to consider storage/communication of additional
  version data to enable interpretation by future downstream systems."

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
elliotcmorris: I'm going to leave this as a con, as leaving things as they are, I believe even just to a new user looking at mediacreation, specification versioning will become a question we'll have to be able to explain.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Tbh i think that's captured in the "seems to fit with OpenAssetIO design" pro. Cause it does seem to fit.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/61ba40856e4bda0344f7fcaeaac5c74a7602ac65

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Changed this up a bit, simplified just to call out the complexity, and link to the relevant section. Removed the try-and-see thing as you're correct it's not strictly true. Think that works.

https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/066101d9e574e43e101aea7e2590ddf0527b2c3d

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
elliotcmorris: Aye sorry, just a transcription error. https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/f290ae35ffd21c25a875e47f75d4fbc5ea3a9690

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: This got simplified in general, there's a whole section about this so the con can just link now : https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/066101d9e574e43e101aea7e2590ddf0527b2c3d

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: If we take this out, then statelessness can't be a pro of the in-data solution.

I've softened the language, as I think negotiation is pretty strongly implied by the other conditions, but admit that it's not a definite requirement to out-of-data. https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/6c01b0a09d88724bade5c0ebdba54eeeeb08b1b5

elliotcmorris:
@feltech @tomc-cinesite @themissingcow 

Made an attempt to make a decision with a rationale : https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/a24bb70e0580364557ed32cbc0f0b00661a575a4

Hoping to move this DR to a close, as I think we're in alignment on this point.

(I still need to actually rename the file since it's DR024 at the moment, but i'll do that after as I don't want to break all the links)/

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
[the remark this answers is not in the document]
themissingcow: Cool cool, was just wondering it highlighting that naive hosts are perhaps equally as important?

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
themissingcow: > [1] I don't believe this is a seperate con to saying that there's now additional data that must be communicated and stored, which is a con, but already listed. 

Sorry yeah - this would be a con specific to an "initialization time only" approach as it was attempting to highlight that that information is not always available at initialization time.

> [3] This has never really come up before and i'm not sure it's a legit con. It dosen't preclude this concept, just means you have to write a somewhat generic negotiation mechanism in your manager, which I concede would be harder, but it seems awful niche for a con.

I don't think its that niche, Host A communicating to Host B via a manager as a hub, with proprietary traits is a highly valuable and explicitly supported workflow in the original API design.

Eg. Katana publishes renders with assorted additional "Foundry specific" traits. Nuke then loads those, and uses the Foundry traits to do comp stuff automatically, all with the manager not understanding any of it, other than storing/recalling the traits.  There may well be different versions of those traits published over time for different entities, or the different hosts may bump versions at different times.

How would you see that working with the manager having to negotiate a version in a non-entity specific way?

> [4] I'm reading this as "Upgrading traits requires out-of-code knowledge of what's compatible", but I find the snippet hard to parse. Seems more like a critique of schema version vs trait by trait versioning, although I guess valid as trait by trait versioning would be harder with an out-of-data formulation. We'd try to solve this with mediacreation utils though.

Yeah - it is really scheme vs trait versioning - were you imaging a view of the out-of-data approach where each specific trait is negotiated? I may have missed it but I recall it reading like the out-of-band approach was leaning towards schema versioning - I may well be wrong though.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
themissingcow: > [3] This restates the above point "Managers not required to consider storage/communication of additional version data to enable interpretation by future downstream systems."

It was the **hosts** part in particular that didn't seem to be getting much air time in other points.

themissingcow:
> Made an attempt to make a decision with a rationale : [a24bb70](https://github.com/OpenAssetIO/OpenAssetIO/commit/a24bb70e0580364557ed32cbc0f0b00661a575a4)
> 
> Hoping to move this DR to a close, as I think we're in alignment on this point.

❤️ ❤️ ❤️ ❤️ ❤️ ❤️ 

I know how much effort everyone has put in here, this feels like a really good call/rationale to me. Obviously I'm biased. 🙃  Thanks for indulging all the discussion!

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: No you wern't misreading, there is a bias in that direction for the out-of-band approach.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: That seems natural, hosts don't have the requirement to persist information on entities they don't understand, so don't run up against this problem.

tomc-cinesite: (for the record, as were in a good place on the DR anyway, but this is perhaps a useful long-term point). I was thinking more about the importance of being able to write a host that doesn't understand the data. Eg simpleResolver, a generic browser, systems that migrate assets from one system to another, how Gaffer would implement it etc... They're not storing the data, but they have no way to engage in a meaningful negotiation... if you see what I mean.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,465 @@
...
+  are in use, which helps especially in the case of second-order
+  dependencies.
+- Seems to fit with existing OpenAssetIO design, especially with
+  `BatchElementError`, as an incompatible version wouldn't necessarily
+  need to error the whole batch.
+- No new API, core `openassetio` library remains version unaware.
+- Cannot be ignored, unlike other "layered on top" mechanisms.
+- Managers not required to consider storage/communication of additional
+  version data to enable interpretation by future/downstream systems.
+- Stateless managers remain possible
```
feltech, line 352: [fix] I think we can remove this, since out-of-data can also be conceived this way, as per our discussions on bundling negotiated schema versions in `Context` etc.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,465 @@
...
+- Allows per-trait versioning, which avoids the need or "insider
```
feltech, line 353: [fix] Typo
```suggestion
- Allows per-trait versioning, which avoids the need for "insider
```

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,465 @@
...
+- Prior art (e.g. OTIO) encodes versions in their data structure,
+  presumably for good reason.
+
+#### Cons
+- No "primary" version across the session, increasing implementation
+  complexity.
+- Would likely require some utilities in `traitgen`, otherwise, due to
+  each new version effectively being a brand new trait, constructing the
+  correct trait view would be become a tedious dance.
+- Introspection methods become more complex. (See [What's
```
feltech, line 365: [fix] Grammar
```suggestion
- Introspection methods become more complex (see [What's
```

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,465 @@
...
+
+### In data - Extended
+The implementation of in-data versioning is very basic, which is part of
+its strength, but it necessitates a "try and see" approach on the part
+of the host. If the host wants to be adaptive to the manager, it may
+benefit from a more up front introspection of the supported manager
+versions.
+
+To this end, the `supportedSchemaVersions` method (or similar concept,)
+from Option 2 could be orthogonally added without any conceptual
```
feltech, line 382: [fix] Typo
```suggestion
from Option 1 could be orthogonally added without any conceptual
```

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,465 @@
...
+conflict. This would have the added benefit of allowing hosts to also
+support a range of version/have backwards compatibility, which is a huge
+benefit.
+
+This stops short of negotiation, being merely introspection similar to
+our other manager introspection methods.
+
+### What's versioned?
```
feltech, line 390: [consider][minor] I think this section would be better placed before the Pros and Cons, since its introducing context for the decision.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
```diff
@@ -0,0 +1,465 @@
...
+- Introspection methods become more complex. (See [What's
+  versioned](#whats-versioned)).
+- Unclear how specification versioning is to work without an out-of-data
+  version.
+- The most obvious implementation, (using `traitID` to store the
+  version), would effectively make any change to a trait version a
+  breaking change. Allowing non breaking changes (such as adding a new
+  property) would require significant rework to the `traitgen` output.
+
+### In data - Extended
```
feltech, line 374: [consider] I wonder if we really need this section, since its essentially saying that introspect/negotiation can be layered on, which is a Pro, but the same applies to the out-of-data approach (i.e. in-data could be layered on top of negotiation at a later date), so doesn't really differentiate the options in any way.

I think it's worth adding a little note in the "Rationale" section that choosing this option doesn't preclude layering on introspection or negotiation at a later date, i.e. the approach is extensible.

feltech (approved):
Looking really good. Just a few last minute niggles.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Heh, adding this at all was a response to a [fix] comment from Tom. I would appreciate if folks could align to absolute universal truth in their fix comments ... that'd be great :P

It's another overlap of decision thing, with the example given in the DR, this is true when compared to the alternate, and statelessness is for sure more like ... favoured?, in the in-data conception at the moment.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: I don't agree, I think desciptiveness at this level of depth is better served after folks have digested the example.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: I know it's a _tad_ awkward, but I think it's a neccesary callout as a large part of the rationale is due to how in-data is more naturally extensible and fits into the openassetio system better.

That is already called out in the rationale. 

>  a large reason for the adoption of an in-data solution is
how little of the API it needs to disrupt in order to become useful, and
yet how open it is to extension.

and 

> ...allows a layered
iterative approach to reduce this complexity, rather than having to do a
big bang drop full of speculation all at once. We look forward with
excited expectation as to how `openassetio-traitgen`, with its generated
views, can be leveraged to make this experience simpler and easier.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/f0a64e5a87c14d0cc35c6a95f907a7fe2a371bd8

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/f0a64e5a87c14d0cc35c6a95f907a7fe2a371bd8

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Changed wording slightly to try and map what we're getting at here. I think it would be disingenuous not to mention state at all, it was a large differentiator in the discussion, and the in-data approach in our head does come out ahead on that regard.
https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/ea0fdcfe027d20622ddf240cbcca770427c86706

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: Oh right, i guess that makes sense. Added it, although only about hosts and I think managers get called out on this area already in a slightly different way. https://github.com/OpenAssetIO/OpenAssetIO/pull/1267/commits/beed7f51c92087712fbb9dd42cf2e0690db65298

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
elliotcmorris: I am resolving this, we're getting into implementational decisions that should come under a different record. This decision is not and cannot be perfect given that there is more to this decision yet to be written. So long as we know we havn't blocked any routes off for ourselves.

doc/decisions/DR024-Versioning-traits-and-specifications-method.md
feltech: The rationale section doesn't explicitly mention layering on introspection/negotiation as a possibility. But if we're keeping this "Extended" section, then fine.

feltech (approved):
Make it so.

PR #1274, opened 2024-02-28
[CI] Remove MacOS LDFLAGS preamble

### LINKED ISSUE #1245: Remove CI MacOS LDFLAGS /usr/local preamble
## What

See if we can remove the `preamble:` that adds `/usr/local` path to `LDFLAGS` in MacOS CI build jobs.

## Why

This was added back when we had a Conan `cpython` package dependency, and might no longer be needed since we removed it.

Though a small and seemingly inconsequential change, doing this along with #1114 would then allow us to collapse our list of build configurations down to a "normal" GitHub build matrix and completely remove the preamble from build steps.  I.e. we'd no longer have need for platform-specific tweaks. This would give us more readable CI config that's easier to maintain.

feltech:
## Description

Closes #1245. We removed the Conan `cpython` package dependency in #1038, meaning this preamble is no longer needed.

Removing this should enable yet further CI simplification, i.e. #1114.

### DIFF
--- a/.github/workflows/upload-release-builds.yml
+++ b/.github/workflows/upload-release-builds.yml
@@ -41,21 +41,13 @@ jobs:
         - os: ubuntu-20.04
           shell: bash
           build-type: Release
-          # MacOS toolchain doesn't search /usr/local by default:
-          # https://gitlab.kitware.com/cmake/cmake/-/issues/19120
-          # The CMake FindPython module's Python::Python target (used
-          # via pybind11::embed in python-bridge-test) transitively adds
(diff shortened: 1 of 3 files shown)


.github/workflows/upload-release-builds.yml
```diff
@@ -47,15 +47,12 @@ jobs:
           # via pybind11::embed in python-bridge-test) transitively adds
```
SamCrooksFoundry, line 47: This comment block doesn't seem relevant once the link flags are removed

SamCrooksFoundry (approved):

PR #1275, opened 2024-02-28
[CI] Platform-agnostic build/test

### LINKED ISSUE #1114: Platform agnostic CI build configuration

feltech:
## Description

Closes #1114. We spend too much time waiting for CI runs when we could
be doing more productive things. We should have an easy time of making
changes and reusing bits of CI across workflows/repos, so should
simplify it where possible.

So use the default generator for the platform and platform-agnostic
CMake command lines.

This also makes downstream adoption more natural, removing the
dependency on Ninja and the need for the `vcvarsall.bat` preamble, and
ensures our CI tests reflect the standard build configuration across
platforms.

Modernise to use parallel builds where possible, and use all available
cores on runners.

### DIFF
--- a/.github/bootstrap_platform/action.yml
+++ b/.github/bootstrap_platform/action.yml
@@ -22,15 +22,15 @@ runs:
     uses: actions/cache@v3
     with:
       path: ~/conan
-      key: ${{ matrix.config.os }}-conan-${{ github.run_id }}
-      restore-keys: ${{ matrix.config.os }}-conan-
+      key: ${{ matrix.os }}-conan-${{ github.run_id }}
+      restore-keys: ${{ matrix.os }}-conan-
 
   - name: Install dependencies (Unix)
     if: runner.os != 'Windows'
     # Configure the system and install library dependencies via conan
     # packages.
     run: |
-      source resources/build/bootstrap-${{ matrix.config.os }}.sh
+      source resources/build/bootstrap-${{ matrix.os }}.sh
     env:
       WORKSPACE: ${{ github.workspace }}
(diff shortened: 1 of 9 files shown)


feltech:
Abandoning due to boilerplate involved, and, critically, Windows MSVC generator parallel `ctest` failures. See https://github.com/OpenAssetIO/OpenAssetIO/issues/1114#issuecomment-1968683513

Issue #521, opened 2022-07-12
Document/use exceptions when manager implementation not found


feltech:
# What

Document the appropriate exception that should be thrown when a manager implementation with a given identifier is not found by a `ManagerInterfaceFactoryInterface` (/ `ManagerImplementationFactoryInterface`) when calling `instantiate()`.

# Why

Currently the correct behaviour is not documented. The Python plugin system implementation will raise an `exceptions.PluginError`, which is Python-only (currently) and perhaps too specific to plugin factories, rather than factories in general.

Documenting the expected behaviour (and obeying it in our implementations) ensures that hosts can guard against the appropriate exception, to give proper user/developer feedback.

# Notes

The Python `exceptions` module has not been migrated to C++. If we require a dedicated exception class, then this should be available in C++ (and bound to Python).

elliotcmorris:
Might fall out of the wash in : #1115 

Note: There's an incorrect doc in `PythonPluginsystemManagerImplementationFactory.py`, saying it still returns `PluginError`

PR #1291, opened 2024-04-11
[Doc] Fix incorrect PluginError doc

elliotcmorris:
All that's left of #521


feltech:
Ah, you hit the same CI failure as in my PR, where I posted a fix: https://github.com/OpenAssetIO/OpenAssetIO/pull/1282#issuecomment-2050276909

Probably should have pulled that commit out as a separate PR, can do if you think its worthwhile.

feltech (approved):

PR #1296, opened 2024-04-17
Add LoggerInterface.isSeverityLogged

### LINKED ISSUE #1014: Add severity check to LoggerInterface
## What

Provide a mechanism for the middleware to ask the `LoggerInterface` if the message they are about to construct will be logged, so they can avoid constructing the string if not neccesary.

## Why

Constructing log strings can be expensive, especially in hot code paths.  Doing so only for the log severity level to be such that the string is never printed is suboptimal, to say the least.

See review conversation https://github.com/OpenAssetIO/OpenAssetIO/pull/1011#discussion_r1256109927

## Acceptance Criteria
- New method `isSeverityLogged(Severity)` added to `LoggerInterface`
- Default implementation for this method that returns true.
- Implement and obey this additional `LoggerInteface` flag in addition to current `maxSeverity` in `SeverityFilter`

feltech:
## Description

Closes #1014. 

Add a new method `isSeverityLogged`, which returns `true` by default, but can be overridden by subclasses in order to hint whether a log string provided to the logger is likely to be output, and so allow users to short-circuit complex string construction logic.

Update `SeverityFilter` to implement the new method, and make use of the method of the wrapped logger in determining whether a message should be filtered.

## Reviewer Notes

Note that the C++ bindings for `LoggerInterface::isSeverityLogged` are tested implicitly via the `SeverityFilter`, since the `SeverityFilter` is injected with and wraps the Python `mock_logger`.  So calling the `SeverityFilter.log`/`.isSeverityLogged` method from Python means a Python->C++->Python call stack.


elliotcmorris (approved):
LGTM.

As usual am perturbed at not having benchmarking around whilst making performance changes like this, but I admit i'm 99% sure performing this check is faster in practically every case.

feltech:
> As usual am perturbed at not having benchmarking around whilst making performance changes like this, but I admit i'm 99% sure performing this check is faster in practically every case.

It would be impossible to benchmark in the general case, since we don't know what the logger call patterns will be by library consumers (frequency, size/complexity of string).

We could do some benchmarking in the more specific case of the upcoming #1206 work (but even that requires making  assumptions about host call volume and manager latency).

Perhaps the PR description/commits were couched too much on the performance side. An API method to discover if a severity is enabled seems like useful introspection, regardless. But practically, the only use I can think of is performance.

Issue #1103, opened 2023-09-19
Ensure #pragma once is at top of files.


elliotcmorris:
## What
Ensure `#pragma once` is always at the top of the file.
## Why
In some of our files `#pragma once` comes after some `#include` declarations. For example the files in the `log` directory. [Example.](https://github.com/OpenAssetIO/OpenAssetIO/blob/bef5b5cd165601a94faecbcf00dc3d840a90fd4b/src/openassetio-core/include/openassetio/log/ConsoleLogger.hpp#L7)

`#pragma once` is subject to macro expansions, and it's conceivable, albeit unlikely, that this could cause a hard to find issue.

If we ever get a moment we should just move these up. Definately a minor odd-job.

## Acceptance Criteria
`#pragma once` is always at the very top of all cpp headers files in the repo.

Gmaj9:
Hello, I would like to take this on for ASWF Dev Days.

Gmaj9:
I only found one .h/.hpp file that didn't have `#pragma once` at the top. However, I noticed whitespace inconsistencies throughout. In some places `#pragma once` is directly beneath the comment at the top of the file; other times there is a newline in between the two. The same goes for whether or not there is a newline between `#pragma once` and the `#include` statements following it. Should I address this, and if so, what would be preferred?

feltech:
> I noticed whitespace inconsistencies throughout.

I like the way you think! Personally, I'd love to be militant about consistency. But realistically, we have to draw the line somewhere, for pragmatic reasons 😉 .  So I'd say, if you want to make them consistent, go for it, I'll happily review it. But it's really not necessary.

In terms of preference, after looking at a random selection of headers, I think the most common case is
```
// Copyright stuff
/**
 * Optional (rare) file docstring
 */
#pragma once

#include <blah> 
```

So converging on that most-common style would be best.

Gmaj9:
OK. And — if I read the docs correctly — the whitespace changes can go in this PR, but each in their own commit?

feltech:
> OK. And — if I read the docs correctly — the whitespace changes can go in this PR, but each in their own commit?

Yeah separate commits would be good. But since the two sets of changes are so closely related it could also make sense as a single commit.

Gmaj9:
[...]

PR #1301, opened 2024-04-18
Move Pragma Once to Top of File

elliotcmorris:
Part of https://github.com/OpenAssetIO/OpenAssetIO/issues/1103, fixes the one we know about.


feltech (approved):
LGTM. Would be good to put the issue/PR number in the commit message.

PR #1304, opened 2024-04-25
[Test] Load C++ plugins in test harness

### LINKED ISSUE #1294: API Compliance Suite support for C++ plugins
# What

Load C++ plugins in the API compliance suite if no Python plugin is found.

# Why

The C++ plugin system has now been added, but the API Compliance Suite needs updating to be able to validate these plugins.

# ACs

- API Compliance Suite test harness attempts to load a Python plugin, and if one is not found it then tries to load a C++ plugin.
- Testing this involves unit tests only (mock the factories).

elliotcmorris:
## Description
When a python plugin of the provided identifier cannot be found when loading a test plugin, fallback to attempting to load a c++ plugin.

As specified during refinement, testing only that the initialisation of the factory is provided the correct type of implementation factory, via mocks, as the functionality for the `ManagerFactory` to load plugins once provided a correct implementation factory is tested elsewhere

Closes #1294

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -15,6 +15,11 @@ v1.0.0-beta.x.y
   a `CppPluginSystemManagerPlugin` object.
   [#1115](https://github.com/OpenAssetIO/OpenAssetIO/issues/1115)
 
+- `openassetio.test.manager`, notably used for Api Compliance tests,
+  updated to automatically load C++ plugins if a python plugin of the
+  requested identifier cannot be found.
+  [#1294](https://github.com/OpenAssetIO/OpenAssetIO/issues/1294)
+
 - Added `LoggerInterface.isSeverityLogged` to allow users to
   short-circuit log message construction if the logger indicates the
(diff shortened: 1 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -15,6 +15,10 @@ v1.0.0-beta.x.y
   a `CppPluginSystemManagerPlugin` object.
   [#1115](https://github.com/OpenAssetIO/OpenAssetIO/issues/1115)
 
+- Test harness updated to automatically load C++ plugins if a python
```
feltech, line 18: [consider] Perhaps clarify this is the API Compliance Suite test harness, `openassetio.test.manager`, not our internal unit tests.

feltech (approved):
LGTM!

RELEASE_NOTES.md
elliotcmorris: :+1:  Did this, but am just rebasing and merging, don't think it needs a re-review.

PR #1306, opened 2024-05-08
Add .pyi stub generation

### LINKED ISSUE #1252: Add .pyi stub files to Python package
## What

Following on from #747, add `.pyi` stub files to our Python package covering the `_openassetio` Python extension module.

## Why

IDEs can use `.pyi` stub files located in the package to greatly enhance code-completion and documentation for Python C extension modules.

## Acceptance Criteria

- Installing via `pip install` includes an `_openassetio` directory containing `.pyi` stub files (satisfying [PEP 561](https://peps.python.org/pep-0561/)), which PyCharm (and other IDEs) can detect and use to enhance code completion.
- `.pyi` stubs are included in the binary wheels on PyPI.
-  `.pyi` stubs are installed when doing a from-source install i.e. `pip install .`
- ~~`.pyi` stubs are installed (to the source tree, just like the extension module itself) when doing a `--editable` install.~~
- Installing via `cmake --install`  bundles the `.pyi` stubs in the install tree.

# Notes

See #747 for investigation with proposed implementation.

feltech:
## Description

Closes #1252. IDEs can use .pyi stub files located in the package to greatly enhance code-completion and documentation for Python C extension modules.

Add as part of `setup.py` installs, as well as CMake via new `OPENASSETIO_ENABLE_PYTHON_STUBGEN` variable (`ON` by default).

Note that stubs wont install with `pip install --editable`.  We can install them, but in local testing neither PyCharm nor Visual Studio Code could use them.

## Test Instructions

Worth checking alternative IDEs.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -31,7 +31,10 @@
         "enable-sanitizers",
         "enable-targets-all",
         "enable-tests"
-      ]
+      ],
+      "cacheVariables": {
+        "OPENASSETIO_ENABLE_PYTHON_STUBGEN": "OFF"
+      }
     },
     {
       "name": "test",
--- a/src/openassetio-python/cmodule/CMakeLists.txt
+++ b/src/openassetio-python/cmodule/CMakeLists.txt
@@ -124,3 +124,102 @@ if (OPENASSETIO_ENABLE_TESTS)
     target_compile_definitions(openassetio-python-module PRIVATE OPENASSETIO_ENABLE_TESTS)
     target_link_libraries(openassetio-python-module PRIVATE openassetio-python-module-test)
 endif ()
+
+
+#-----------------------------------------------------------------------
+# Generate .pyi stubs
+#
+# pybind11-stubgen will create openassetio/_openassetio subdirectories
+# in the build output directory. We use this knowledge to assemble a
+# Python pseudo-package in the build output directory, such that stubgen
+# can import the module to do its thing then dump its output alongside.
+
+if (OPENASSETIO_ENABLE_PYTHON_STUBGEN)
+    # Check pybind11-stubgen is available.
+    execute_process(
+        COMMAND "${Python_EXECUTABLE}" -m pybind11_stubgen --help
+        OUTPUT_VARIABLE _pybind11_stubgen_output
+        ERROR_VARIABLE _pybind11_stubgen_output
+        RESULT_VARIABLE _pybind11_stubgen_exit_code
+    )
+
+    # Fatal error if pybind11-stubgen isn't available.
+    if (NOT _pybind11_stubgen_exit_code EQUAL "0")
+        message(
+            FATAL_ERROR
+            "OPENASSETIO_ENABLE_PYTHON_STUBGEN=${OPENASSETIO_ENABLE_PYTHON_STUBGEN} but"
+            " pybind11-stubgen not found: status=${_pybind11_stubgen_exit_code}:"
+            " ${_pybind11_stubgen_output}"
+        )
+    endif()
+
+    # Create temporary pseudo-package directory.
+    add_custom_command(
+        TARGET openassetio-python-module POST_BUILD
+        COMMAND
+        "${CMAKE_COMMAND}" -E make_directory "${CMAKE_CURRENT_BINARY_DIR}/openassetio"
+    )
+
+    if (WIN32)
+        # On Windows, copy openassetio.dll dependency into the
+        # pseudo-package directory, so that stubgen can import
+        # _openassetio.
+        # Note that this must be done _after_ the build, otherwise
+        # Windows (Visual Studio CMake generator) complains that the
+        # build directory already exists.
+        add_custom_command(
+            TARGET openassetio-python-module POST_BUILD
+            COMMAND
+            "${CMAKE_COMMAND}" -E copy_if_different
+            "$<TARGET_FILE:openassetio-core>"
+            "${CMAKE_CURRENT_BINARY_DIR}/openassetio/"
+        )
+    endif()
+
+    # Execute commands to generate .pyi stubs.
+    add_custom_command(
+        TARGET openassetio-python-module POST_BUILD
+        COMMAND
+        "${CMAKE_COMMAND}" -E echo "Generating .pyi stubs with pybind11-stubgen..."
+        # Copy the Python extension module under the pseudo-package
+        # directory.
+        COMMAND
+        "${CMAKE_COMMAND}" -E copy_if_different
+        "$<TARGET_FILE:openassetio-python-module>"
+        "${CMAKE_CURRENT_BINARY_DIR}/openassetio/"
+        # Ensure we have a py.typed file at the root of our package, to
+        # signal to IDEs that stubs exist.
+        COMMAND
+        "${CMAKE_COMMAND}" -E touch "${CMAKE_CURRENT_BINARY_DIR}/openassetio/py.typed"
+        # Execute pybind11-stubgen, modifying PYTHONPATH so it can
+        # locate the openassetio._openassetio module.
+        COMMAND
+        "${CMAKE_COMMAND}" -E env
+        --modify "PYTHONPATH=path_list_prepend:${CMAKE_CURRENT_BINARY_DIR}"
+        --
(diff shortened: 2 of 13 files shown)


.github/workflows/build-instructions.yml
```diff
@@ -26,15 +26,15 @@ jobs:
 
     - name: Build and install (Docker)
       run: |
-        docker run -v `pwd`:/src ghcr.io/openassetio/openassetio-build bash -c '
+        docker run -v `pwd`:/src ghcr.io/openassetio/openassetio-build:2024.3.rc1 bash -c '
```
feltech, line 29: [fix] Push new latest and drop commit.

feltech:
I installed Visual Studio Code fresh (didn't have it installed before now) in order to test this works with IDEs other than PyCharm.

* `pip install .` works
* `pip install ./openassetio-1.0.0b2.post1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl` (i.e. pre-built wheel) works. So presumably the PyPI package will work too.
* `pip install --editable` _doesn't_ work - Visual Studio Code's default Python plugin doesn't seem to be able to detect the editable install _at all_, regardless of `.pyi` stubs. So there's a deeper issue there to do with either my set-up or a limitation of VSCode.
* `cmake --install`, plus `.pth` file in the venv pointing into the install directory (closest I could come up with to PyCharm's "Mark Directory As"), _doesn't_ work - similar problem to `--editable` - no trace of openassetio.

src/openassetio-python/cmodule/CMakeLists.txt
```diff
@@ -124,3 +124,102 @@ if (OPENASSETIO_ENABLE_TESTS)
...
+
+    # Fatal error if pybind11-stubgen isn't available.
+    if (NOT _pybind11_stubgen_exit_code EQUAL "0")
+        message(
+            FATAL_ERROR
+            "OPENASSETIO_ENABLE_PYTHON_STUBGEN=${OPENASSETIO_ENABLE_PYTHON_STUBGEN} but"
+            " pybind11-stubgen not found: status=${_pybind11_stubgen_exit_code}:"
+            " ${_pybind11_stubgen_output}"
+        )
+    endif()
```
elliotcmorris, line 154: [Consider] Should this error be fatal? Not being able to generate stubs feels like a warning sort of thing for me.

src/openassetio-python/cmodule/CMakeLists.txt
```diff
@@ -124,3 +124,102 @@ if (OPENASSETIO_ENABLE_TESTS)
...
+        COMMAND
+        "${CMAKE_COMMAND}" -E touch "${CMAKE_CURRENT_BINARY_DIR}/openassetio/py.typed"
+        # Execute pybind11-stubgen, modifying PYTHONPATH so it can
+        # locate the openassetio._openassetio module.
+        COMMAND
+        "${CMAKE_COMMAND}" -E env
+        --modify "PYTHONPATH=path_list_prepend:${CMAKE_CURRENT_BINARY_DIR}"
+        --
+        "${Python_EXECUTABLE}" -m pybind11_stubgen
+        # Fail the build and abort if any errors generating stubs.
```
elliotcmorris, line 201: [Consider] Similarly here, could this become a pain? I guess you can always disable stub generation overall...

src/openassetio-python/tests/package/_openassetio/test_pyi.py
```diff
@@ -0,0 +1,76 @@
+#
+#   Copyright 2023 The Foundry Visionmongers Ltd
```
elliotcmorris, line 2: [fix] 2024

src/openassetio-python/tests/package/_openassetio/test_pyi.py
```diff
@@ -0,0 +1,76 @@
...
+    "managerApi.pyi",
+    "pluginSystem.pyi",
+    "trait.pyi",
+    "utils.pyi",
+)
+
+
+def test_expected_pyi_files_generated(pyi_dir: pathlib.Path):
+    actual_pyi_files = set(p.name for p in pyi_dir.glob("*") if p.name != "_testutils")
+    assert actual_pyi_files == set(expected_pyi_files)
```
elliotcmorris, line 46: [Question] Maybe a stupid question, but does this interact at all with turning off stubs in the sanitizer build, or really any build with stubgen turned off? Are these tests not run in that case?

CMakePresets.json
```diff
@@ -31,7 +31,10 @@
         "enable-sanitizers",
         "enable-targets-all",
         "enable-tests"
-      ]
+      ],
+      "cacheVariables": {
+        "OPENASSETIO_ENABLE_PYTHON_STUBGEN": "OFF"
```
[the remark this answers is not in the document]
feltech: This is for CI.  It's annoying that there's no way to add code comments to JSON. From the commit message:

> Disable generation of stubs for the ASan CI job, since that would
require wrangling the `LD_PRELOAD` env var to allow stubgen to import
`_openassetio`, and going to that effort wouldn't give us much more
coverage, if any

src/openassetio-python/cmodule/CMakeLists.txt
feltech: It's tricky.  I wrestled with the best error/warning behaviour.  

In particular I had in mind that stubgen is enabled by default, so people cloning the repo and building from source may encounter this.  

So the first question is, should it be enabled by default?  I reasoned yes, because our default config is deliberately the same as the config used for a release artefact.

Given it's enabled by default because our default config is for a releasable build, should it be a fatal error if stubgen isn't available? I reasoned yes, because otherwise we (or someone else) could easily accidentally build and release without stubs, e.g. if the logs are hidden behind CI.

There is a hole in our tests for Python wheel builds, since the stubgen tests are skipped if stubs are not found and there is no env var saying they should exist.  For Python wheel builds the env var isn't set (since it's set inside of CMake and only applies to `ctest`).

I guess we could set that env var separately in CI during wheel builds, to force the stubgen tests to run.  That would mitigate somewhat the hole in our tests :thinking: ....?

src/openassetio-python/cmodule/CMakeLists.txt
feltech: I think in this case, if you definitely want stubs and stubgen is available, then failing to generate proper stubs should be a fatal error.

src/openassetio-python/tests/package/_openassetio/test_pyi.py
feltech: The magic for that is at the bottom of the test script.  There is a fixture that will skip all tests in the module if the expected stub directory is not found _and_ there is no env var signalling that stubs should definitely exist (which is set by CMake for `ctest` runs).

src/openassetio-python/cmodule/CMakeLists.txt
elliotcmorris: > because otherwise we (or someone else) could easily accidentally build and release without stubs, e.g. if the logs are hidden behind CI.

Damn fine point!

src/openassetio-python/cmodule/CMakeLists.txt
feltech: Forced stubgen tests to run in 6fc8db96

Added a drop! commit to illustrate - resulting workflow run: https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/9127570146/job/25098125757?pr=1306 - now removed drop! commit.

src/openassetio-python/cmodule/CMakeLists.txt
feltech: Forced stubgen tests to run in 6fc8db96

Added a drop! commit to illustrate - resulting workflow run: https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/9127570146/job/25098125757?pr=1306 - now removed drop! commit.

feltech: Renamed the environment variable in 60e6779a since now that it can be provided "from outside" it made it look like it would set the CMake variable, when it doesn't, it's only used in tests.

elliotcmorris (approved):

PR #1316, opened 2024-05-23
Docs typo fix

BryceGattis:
## Description

Small typo fix in docs.

### DIFF
--- a/doc/doxygen/src/Glossary.dox
+++ b/doc/doxygen/src/Glossary.dox
@@ -27,7 +27,7 @@
  *
  * @see @fqref{Context} "Context"
  *
- * @section DCC Digital Contentent Creation tool (DCC)
+ * @section DCC Digital Content Creation tool (DCC)
  *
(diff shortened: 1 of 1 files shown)


feltech (approved):
Good spot and thanks for the fix!

Only minor comment really is (as per [the docs](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/contributing/COMMITS.md)) it would be good to add a `[Docs]` tag preceding the commit message.

BryceGattis:
@feltech Should I try to amend the commit to fix that?

feltech:
> @feltech Should I try to amend the commit to fix that?

We would have merged this regardless.  However, since this PR needs rebasing against `main` anyway, it would be great if you could update the commit message at the same time, but not a big deal.

feltech:
@BryceGattis I just noticed you gave us permission to update your branch, so I've taken the liberty of rebasing, and finessing the commit message a bit.

PR #1319, opened 2024-05-30
Add `abidiff` check to CI

### LINKED ISSUE #1085: CI job checking for ABI breaks
## What

Add a CI job that ensures no unintentional ABI breaks and symbol leakage.

## Why

As we approach beta and v1 we need to ensure breaking changes are managed.  Part of that is ensuring no unintentional changes to the ABI. This was highlighted in recent work (https://github.com/OpenAssetIO/OpenAssetIO/issues/1070) where a new dependency was found to leak symbols. This particular case was spotted manually, but ideally our CI would flag such changes to the ABI.

Tools exist for this, e.g. https://github.com/lvc/abi-compliance-checker - as well as command-line tools like `nm` (must treat platforms independently though.

Could use a Docker container than can run locally, that's triggered by CI.

## Acceptance Criteria

- - Linux only
- Need to store either the dwarf symbol builds or abi symbol dumps for the most recent build

- If an MR breaks the ABI, the CI should raise the alarm

feltech:
## Description

Closes #1085. Add new CTest target and CI job that runs it. ABI snapshots are committed to the repository and checked against with every PR.

## Reviewer Notes

Deliberately breaking `drop!` commit is provided, to show the output.

If you want to run the steps in the new README, use the RC of the Docker image, i.e. `ghcr.io/openassetio/openassetio-build:2024.4.rc1`


.github/workflows/code-quality.yml
```diff
@@ -153,3 +153,27 @@ jobs:
...
           CCACHE_DIR: /tmp/ccache
+
+  abi-check:
+    name: ABI diff check
+    runs-on: ubuntu-latest
+    # We must test ABI on a consistent platform. So choose our VFX
+    # reference platform compatible Docker image, which we have
+    # pre-installed libabigail into for performing the ABI diff checks.
+    container:
+      image: ghcr.io/openassetio/openassetio-build:2024.4.rc1
```
feltech, line 164: Drop commit and push new 2024.4 and latest

feltech:
Passing CI job: https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/9315770655/job/25642641211?pr=1319

Deliberate failures incoming.

feltech:
`drop!` commit added to illustrate the output of breakages (see commit message for what breakages were introduced).

I tried to investigate the rather odd entry in the output
```
'class openassetio::v1::pluginSystem::CppPluginSystemManagerImplementationFactory at CppPluginSystemManagerImplementationFactory.hpp:48:1' changed:
  type size hasn't changed
  there are member function changes:
```
when `CppPluginSystemManagerImplementationFactory` hasn't changed.

Using the `--dump-diff-tree` can help.  Ultimately its because of the change to `ManagerInterface`, and that warning is saying the `CppPluginSystemManagerImplementationFactory` is affected by it.  However, this shouldn't show up as an independent ABI change. It's also odd that the list of "member function changes" is empty. So I think its a bug in libabigail.

libabigail 2.5 is out, but not available on CentOS 7 (i.e. our Docker image) so we're stuck on 2.4 for now..

resources/abi/README.md
```diff
@@ -0,0 +1,74 @@
+# ABI change monitoring
+
+When the OpenAssetIO C++ ABI changes, we need to know about it to inform
+our release notes and semver versioning.
+
+This is aided (though not solved!) by automated testing against a
```
elliotcmorris, line 6: [Question] How not solved? Like, there's an outside but not realistic shot at libabigail missing things, or it misses a lot of common things?

The exclamation mark scares me here, it's saying to me that I can't trust this tool and I need to do additional work myself, and I think, what do I as an implementor need to be aware of? There is a point where if we still have to be vigilant, the overhead of the tooling isn't worth it at all.

cmake/Testing.cmake
```diff
@@ -66,6 +66,51 @@ function(openassetio_add_generic_test_target target_name)
...
+    # the same name as the library, under `resources/abi` in the
+    # project root.
+    #
+    # Uses the build artifact location for convenience (i.e. handy
+    # generator expressions to get the path), rather than the install
+    # artifacts, but they are equivalent for `abidiff` purposes.
+    function(openassetio_add_abi_test_target target_name)
+        add_custom_target(
+            openassetio.test.abi.${target_name}
+            COMMAND ${CMAKE_COMMAND} -E echo
```
elliotcmorris, line 87: [Consider][Minor] I think most people wouldn't realise this is a newline, maybe a comment.

resources/abi/README.md
```diff
@@ -0,0 +1,74 @@
...
+unavoidable, then a new XML snapshot dump of the ABI must be created
+(and release notes updated as appropriate).
+
+To do this we use `abidw`. For convenience, the `openassetio-build`
+Docker image has libabigail pre-installed.
+
+The generated snapshot is then checked against new builds using
+`abidiff`.
+
+## Example of updating the ABI snapshot
```
elliotcmorris, line 26: [Consider] Great walkthrough, but since it's a neccesary workflow, I wonder if a link from PROCESS.md might be desireable

elliotcmorris, line 26: [Consider][Minor] Maybe this could more clearly be a set of steps you need to follow with an instructive tone. Ie, rather than 
> This will generate a new XML formatted ABI dump to overwrite the
existing XML dump under `resources/abi`.

do

> Generate a new XML formatted ABI dump and overwrite the existing XML dump under `resources/abi`

A couple of other things on this note. Perhaps the title could be "Updating the ABI snapshot" rather than "example of", as well as potentially presenting this as a numbered list of instructions, rather than an example. Not much changes to content, just presentation.

Just cause the goal for any process should be being able to do it on autopilot.

resources/abi/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+```sh
+docker exec openassetio-build \
+    bash -c 'cd /src && \
+        cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug \
+          -DOPENASSETIO_ENABLE_TESTS=ON \
+          -DOPENASSETIO_ENABLE_TEST_ABI=ON && \
+        ctest --test-dir build --output-on-failure \
+          -R openassetio.test.abi'
+```
```
elliotcmorris, line 74: [Consider] Missing a step to say the dump should be commited back to the git repo if an ABI break is expected and desired?

elliotcmorris (approved):

resources/abi/README.md
feltech: Oh yes we still have to be vigilant.  libabigail, from what I've seen, won't miss any changes in the binary, but they won't necessarily be breaking changes, so we need manual effort to decode the output and decide if and how the change is breaking.  

Most troubling is that changes in template classes or inline functions (i.e. things defined just in headers) won't be picked up.  This is eased as long as it's used in the binary.  E.g. I tried messing with `BatchElementError`, which is only defined in a header, but since it's used in the binary, the ABI change was picked up by libabigail.  

Note that using the command-line switch to point `abidiff` at the headers (`--headers-dir-1`) doesn't cause it to check their ABI, it's only used to filter the output from scanning the binary.

Also, although the docs you highlight is focused on ABI, I did have in mind things that affect versioning more generally, which are not necessarily ABI breaks - e.g. trait changes, changes to the documented contract, new non-breaking features.

resources/abi/README.md
feltech: Added a few links and mentions in 7450e2ae

feltech: Good idea, restructured in 99e93a2c

resources/abi/README.md
elliotcmorris: Do you think these specific points should be noted in this document? How else is someone meant to know what to look out for?

resources/abi/README.md
elliotcmorris: :+1:

elliotcmorris (commented):
changes all look good, just a Q about whether the high level "What you need to look out for" notes should be included.

resources/abi/README.md
feltech: I like the idea of a Limitations section :+1:

feltech: Added a section in bad49c8d.

Reviewed over-the-shoulder by @elliotcmorris.

Issue #1322, opened 2024-06-05
Support partial substitution in `substitute`


feltech:
## What

* Allow partial substitution in `utils.substitute` without throwing an exception, e.g. such that we can do `substitute("image.{frame}.{view}.exr", {"view": "right"}) == "image.{frame}.right.exr"`.
* Support for providing a default value, e.g. `{variable:format-default}` - TSC feedback
* Enhance syntax to support optional substrings e.g. `/mnt</lhs{show}/rhs>/file`, where `/lhs` and `/rhs` won't print if `{show}` is not provided - TSC  feedback.

## Why

We suspect that partial, or gradual, substitution will be useful in OpenAssetIO, especially in the context of serialisation and multi-component systems. However, the current implementation relies on libfmt, which [does not support this](https://github.com/fmtlib/fmt/issues/1048) in its public interface.

If a substitution variable is not provided, the source of the string (host/manager) may wish to provide a good default.

To make strings more re-usable e.g. across different but related file system layouts, it's useful to have a mechanism to make substrings around a token only printed when the token is substituted.

SamCrooksFoundry:
We want this, but it's hard to implement because libfmt doesn't support it.

feltech:
Updated with additional nice-to-haves suggested during the TSC meeting.

PR #1323, opened 2024-06-05
Add `substitute` utility function

### LINKED ISSUE #1213: OpenAssetIO string substitution utility
https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/pull/71#discussion_r1406479090

## What
Expose a utility API method to perform string substitution on any string in the "OpenAssetIO format".
Document the "OpenAssetIO format" as mirroring the libfmt format

## Why
Hosts receive templated URL from managers, such as `file:///some/shot-{frame%3A04d}`, (percent encoded colon), and it's important that managers dont have to branch and produce different formats depending on the specific host they're trying to communicate with.

LocatableContent has support for template substitutions, like `fmt`, with the curly brackets.
The normal ways you format this is via std::format or python fstrings, however these are not known to be fully compatible, so we've ended up saying the meaningless "OpenAssetIO Format". **We need to define what this means, or produce a method that allows both languages to do the substitutions in the same way (by just forwarding to libfmt, which we use internally)**

## Acceptance criteria
- Follow design of notebook in https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/issues/70

feltech:
## Description

Closes #1213. Many strings communicated between a host and a manager may contain placeholder tokens that are intended to be substituted by the host before use, e.g. an animated image sequence on disk as several files similarly named, except for a frame number, could be referred to collectively using `image.{frame}.exr`.

However, how is the host/manager to know what syntax to use for placeholder tokens, when they are, in general, separately developed codebases?

A related problem is how the host/manager knows whether and what placeholder tokens are allowed in a given string? This question has been moved out of the core library to be an industry-specific concern. Check the docs in the MediaCreation repository; but in short, the documentation of individual traits and properties dictate the allowed tokens.

So we define an "OpenAssetIO format" for substitution tokens, which, conveniently, is exactly the same as the libfmt format which, conveniently again, almost exactly matches Python format string format.

However, we only support named (as opposed to positional) arguments, and only officially support format specifiers for integer zero padding (e.g. `{frame:04d}`).  Other format specifiers will work, but we're reluctant to document support for them, since future substitution implementations (e.g. in other languages) may not be able use libfmt.

So add a `utils::substitute` function that takes a dictionary mapping token names to values, and returns a substituted string, or an error.

In particular, an error is thrown when insufficient variables are given in the dictionary to fill all placeholders. This is undesirable, but partial substitution is not supported in libfmt (or indeed, Python format strings). Future work may add such support, but it is non-trivial to add.

## Reviewer Notes

Note that partial substitution is not supported due to the difficulty in doing so via libfmt - so this has been punted to a follow-up issue #1322.

Also see https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/pull/97 for higher level documentation.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -27,6 +27,10 @@ v1.0.0-beta.x.y
   logger.
   [#1014](https://github.com/OpenAssetIO/OpenAssetIO/issues/1014)
 
+- Added `utils.substitute` providing a common string substitution format
+  using named placeholders in libfmt/Python format string style syntax.
+  [#1213](https://github.com/OpenAssetIO/OpenAssetIO/issues/1213)
+
 ### Improvements
 
 - Added `getRelationship(s)` overloads for convenience, providing
--- a/src/openassetio-core/CMakeLists.txt
+++ b/src/openassetio-core/CMakeLists.txt
@@ -93,6 +93,7 @@ target_sources(
     src/utils/path/windows/pathTypes.cpp
     src/utils/path/posix.cpp
     src/utils/path/posix/detail.cpp
+    src/utils/substitute.cpp
 )
 
 # Public header dependency.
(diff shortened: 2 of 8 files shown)


elliotcmorris (approved):

feltech:
^ minor commit message fix, mostly to trigger the stalled DCO CI job

PR #1329, opened 2024-06-18
Printable TraitsData + other types.

### LINKED ISSUE #1307: Ability to log `TraitsData` (incl. properties), access modes, entity references
## Acceptance Criteria

- All manager API can be printed as a string in C++ and Python - and you can read what the traits data contains
- Create a test that prints each of the data types
- Should work with fmt, to_string and stream operators
- Python should work with `str` and `repr`

For our purposes, `repr` returns the same as `str`

elliotcmorris:
## Description

Add fmt formatters for most useful OpenAssetIO types.
Wrap this internal formatting behaviour to implement public <<
operators for logging and printing.

Not totally consistent, and not totally all-encompasing either. Lots of decisions were made in a pairing context.

Closes #1307

Kept a bit split between python and C++ just for review convenience.

### DIFF
--- a/src/openassetio-core/include/openassetio/utils/ostream.hpp
+++ b/src/openassetio-core/include/openassetio/utils/ostream.hpp
@@ -1,0 +1,155 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2024 The Foundry Visionmongers Ltd
+#pragma once
+
+#include <memory>
+#include <ostream>
+#include <unordered_map>
+#include <vector>
+
+#include <openassetio/InfoDictionary.hpp>
+#include <openassetio/errors/BatchElementError.hpp>
+#include <openassetio/hostApi/Manager.hpp>
+#include <openassetio/managerApi/ManagerInterface.hpp>
+#include <openassetio/trait/collection.hpp>
+#include <openassetio/trait/property.hpp>
+#include <openassetio/typedefs.hpp>
+
+OPENASSETIO_FWD_DECLARE(Context)
+OPENASSETIO_FWD_DECLARE(managerApi, HostSession)
+OPENASSETIO_FWD_DECLARE(trait, TraitsData)
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
+class EntityReference;
+using EntityReferences = std::vector<EntityReference>;
+
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "an_entity_reference".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const EntityReference& formattable);
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "['an_entity_reference_1', 'an_entity_reference_2']".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const EntityReferences& formattable);
+
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "{'locale': {'aTrait': {'aProperty': propertyVal}}, 'managerState': memoryAddress}".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out, const ContextPtr& formattable);
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "{'locale': {'aTrait': {'aProperty': propertyVal}}, 'managerState': memoryAddress}".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const ContextConstPtr& formattable);
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "{'locale': {'aTrait': {'aProperty': propertyVal}}, 'managerState': memoryAddress}".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out, const Context& formattable);
+
+namespace managerApi {
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "humanReadableCapabilityName".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const ManagerInterface::Capability& formattable);
+}  // namespace managerApi
+
+namespace hostApi {
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "humanReadableCapabilityName".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const Manager::Capability& formattable);
+}  // namespace hostApi
+
+namespace errors {
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "humanReadableErrorCodeName".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const BatchElementError::ErrorCode& formattable);
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "humanReadableErrorCodeName: Error message.".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const BatchElementError& formattable);
+}  // namespace errors
+
+namespace trait {
+
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "{'aTrait': {'aTraitProperty': traitValue, 'anotherTraitProperty':
+ * anotherTraitValue}, 'anotherTrait': {aTraitProperty: traitValue}}".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const TraitsDataPtr& formattable);
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "{'aTrait': {'aTraitProperty': traitValue, 'anotherTraitProperty':
+ * anotherTraitValue}, 'anotherTrait': {aTraitProperty: traitValue}}".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const TraitsDataConstPtr& formattable);
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "{'aTrait': {'aTraitProperty': traitValue, 'anotherTraitProperty':
+ * anotherTraitValue}, 'anotherTrait': {aTraitProperty: traitValue}}".
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out, const TraitsData& formattable);
+
+namespace property {
+
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "Value", (or "'Value'" if formattable is a string)
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out, const Value& formattable);
+}  // namespace property
+}  // namespace trait
+}  // namespace OPENASSETIO_CORE_ABI_VERSION
+}  // namespace openassetio
+
+/* The types here are just typedefs to standard types, and thus must be defined
+ * in the standard namespace. */
+namespace std {
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "{'key1': 'value1', 'key2': 'value2'}"
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(
+    std::ostream& out, const openassetio::OPENASSETIO_CORE_ABI_VERSION::StrMap& formattable);
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "{'key1': value1, 'key2': value2}"
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(
+    std::ostream& out,
+    const openassetio::OPENASSETIO_CORE_ABI_VERSION::InfoDictionary& formattable);
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "{'trait1', 'trait2'}"
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(
+    std::ostream& out,
+    const openassetio::OPENASSETIO_CORE_ABI_VERSION::trait::TraitSet& formattable);
+/**
+ * Insertion operator for use with ostreams.
+ * Formats as "[{'trait1', 'trait2'}, {'trait3', 'trait4'}]"
+ */
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(
+    std::ostream& out,
+    const openassetio::OPENASSETIO_CORE_ABI_VERSION::trait::TraitSets& formattable);
+}  // namespace std
--- a/src/openassetio-core/src/utils/formatter.cpp
+++ b/src/openassetio-core/src/utils/formatter.cpp
@@ -1,0 +1,212 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2024 The Foundry Visionmongers Ltd
+
+#include "formatter.hpp"
+
+#include <algorithm>
+#include <sstream>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <openassetio/Context.hpp>
+#include <openassetio/EntityReference.hpp>
+#include <openassetio/errors/BatchElementError.hpp>
+#include <openassetio/hostApi/Manager.hpp>
+#include <openassetio/managerApi/HostSession.hpp>
+#include <openassetio/managerApi/ManagerInterface.hpp>
+#include "../errors/exceptionMessages.hpp"
+
+auto fmt::formatter<openassetio::EntityReference>::format(
+    const openassetio::EntityReference& entityRef, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  return fmt::formatter<string_view>::format(fmt::format("{}", entityRef.toString()), ctx);
+}
+
+auto fmt::formatter<openassetio::EntityReferences>::format(
+    const openassetio::EntityReferences& entityRefs, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  std::string result;
+
+  if (entityRefs.empty()) {
+    result = "[]";
+  } else {
+    result += "['";
+    result += fmt::format("{}", fmt::join(entityRefs, "', '"));
+    result += "']";
+  }
+
+  return fmt::formatter<string_view>::format(result, ctx);
+}
+
+auto fmt::formatter<openassetio::StrMap>::format(const openassetio::StrMap& strMap,
+                                                 format_context& ctx) const
+    -> decltype(ctx.out()) {
+  std::vector<std::string> flattenedStrMap;
+  flattenedStrMap.reserve(
+      strMap.size());  // Reserve space for the elements to avoid multiple allocations
+
+  for (const auto& kv : strMap) {
+    // Format each key-value pair into a single string
+    flattenedStrMap.push_back(fmt::format("'{}': '{}'", kv.first, kv.second));
+  }
+
+  return formatter<string_view>::format(fmt::format("{{{}}}", fmt::join(flattenedStrMap, ", ")),
+                                        ctx);
+}
+
+auto fmt::formatter<openassetio::managerApi::ManagerInterface::Capability>::format(
+    openassetio::managerApi::ManagerInterface::Capability capability, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  return formatter<string_view>::format(
+      openassetio::managerApi::ManagerInterface::kCapabilityNames[static_cast<std::size_t>(
+          capability)],
+      ctx);
+}
+
+auto fmt::formatter<openassetio::hostApi::Manager::Capability>::format(
+    openassetio::hostApi::Manager::Capability capability, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  return formatter<string_view>::format(
+      openassetio::managerApi::ManagerInterface::kCapabilityNames[static_cast<std::size_t>(
+          capability)],
+      ctx);
+}
+
+auto fmt::formatter<openassetio::errors::BatchElementError::ErrorCode>::format(
+    openassetio::errors::BatchElementError::ErrorCode errorCode, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  return formatter<string_view>::format(openassetio::errors::errorCodeName(errorCode), ctx);
+}
+
+auto fmt::formatter<openassetio::errors::BatchElementError>::format(
+    const openassetio::errors::BatchElementError& ber, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  openassetio::Str out = fmt::format("{}: {}", ber.code, ber.message);
+  return formatter<string_view>::format(out, ctx);
+}
+
+auto fmt::formatter<openassetio::trait::TraitSet>::format(
+    const openassetio::trait::TraitSet& traitSet, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  std::vector<std::string> quotedTraitSet;
+  quotedTraitSet.reserve(traitSet.size());
+  std::transform(traitSet.cbegin(), traitSet.cend(), std::back_inserter(quotedTraitSet),
+                 [](const std::string& trait) { return fmt::format("'{}'", trait); });
+
+  return fmt::formatter<string_view>::format(
+      fmt::format("{{{}}}", fmt::join(quotedTraitSet, ", ")), ctx);
+}
+
+auto fmt::formatter<openassetio::trait::TraitSets>::format(
+    const openassetio::trait::TraitSets& traitSets, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  return fmt::formatter<string_view>::format(fmt::format("[{}]", fmt::join(traitSets, ", ")), ctx);
+}
+
+namespace {
+struct ToStringVisitor {
+  std::string operator()(openassetio::Str arg) const { return fmt::format("'{}'", arg); }
+  std::string operator()(openassetio::Float arg) const { return std::to_string(arg); }
+  std::string operator()(openassetio::Int arg) const { return std::to_string(arg); }
+  std::string operator()(openassetio::Bool arg) const { return arg ? "True" : "False"; }
+};
+}  // namespace
+auto fmt::formatter<openassetio::trait::property::Value>::format(
+    const openassetio::trait::property::Value& val, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  return fmt::formatter<string_view>::format(std::visit(ToStringVisitor(), val), ctx);
+}
+
+auto fmt::formatter<openassetio::InfoDictionary>::format(
+    const openassetio::InfoDictionary& infoDict, format_context& ctx) const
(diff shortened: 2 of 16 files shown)


src/openassetio-core/tests/utils/PrintableTest.cpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    WHEN("A Capability is printed") {
+      THEN("It can be printed via fmt and <<") { checkBasicPrintable(capability, "publishing"); }
+    }
+
+    WHEN("A Context is printed") {
+      THEN("It can be printed via fmt and <<") {
+        // No closing brace on purpose, to account for variant managerState memory address
+        checkBasicPrintableContains(*context,
```
feltech, line 168: [fix] Test for the nullptr case

src/openassetio-core/tests/utils/PrintableTest.cpp
```diff
@@ -0,0 +1,191 @@
...
+                                    R"({'locale': {'aTrait': {'aIntTraitProperty': 2}}, )"
+                                    R"('managerState': 0x)");
+        checkBasicPrintableContains(context, R"('locale': {'aTrait': {'aIntTraitProperty': 2}}, )"
+                                             R"('managerState': 0x)");
+      }
+    }
+
+    WHEN("A TraitsData is printed") {
+      THEN("It can be printed via fmt and <<") {
```
feltech, line 177: [fix] Test for the nullptr case

src/openassetio-core/src/utils/formatter.cpp
```diff
@@ -0,0 +1,198 @@
...
+auto fmt::formatter<openassetio::Context>::format(const openassetio::Context& context,
+                                                  format_context& ctx) const
+    -> decltype(ctx.out()) {
+  openassetio::Str out = fmt::format("{{'locale': {}, 'managerState': {}}}", context.locale,
+                                     static_cast<void*>(context.managerState.get()));
+
+  return fmt::formatter<string_view>::format(out, ctx);
+}
+
+auto fmt::formatter<std::shared_ptr<openassetio::trait::TraitsData>>::format(
```
feltech, line 153: [consider][minor] `TraitsDataPtr`

src/openassetio-core/src/utils/formatter.cpp
```diff
@@ -0,0 +1,198 @@
...
+                     traitsData.getTraitProperty(&out, traitId, propertyKey);
+                     return fmt::format(R"('{}': {})", propertyKey, out);
+                   });
+
+    // Add the string that makes up the trait, which is the json dict section for the traitId,
+    // followed by all the property key value pairs.
+    traitStrings.push_back(fmt::format("'{}': {{{}}}", traitId, fmt::join(propertyStrings, ", ")));
+  }
+
+  // The idea here being that this is a valid json dict, hence all the extra brace formatting
```
feltech, line 195: [fix] Not JSON dict, Python dict.  The single quotes make it invalid JSON, sadly.

src/openassetio-core/src/utils/ostream.cpp
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2023 The Foundry Visionmongers Ltd
+#include <openassetio/Context.hpp>
+#include <openassetio/EntityReference.hpp>
+#include <openassetio/hostApi/Manager.hpp>
+#include <openassetio/managerApi/HostSession.hpp>
+#include <openassetio/utils/ostream.hpp>
+
+#include "../src/utils/formatter.hpp"
```
feltech, line 9: [consider] This confused me.  I think just `#include "formatter.hpp"` is better.

src/openassetio-core/tests/utils/PrintableTest.cpp
```diff
@@ -0,0 +1,191 @@
...
+#include <openassetio/Context.hpp>
+#include <openassetio/errors/exceptions.hpp>
+#include <openassetio/hostApi/HostInterface.hpp>
+#include <openassetio/hostApi/Manager.hpp>
+#include <openassetio/log/LoggerInterface.hpp>
+#include <openassetio/managerApi/ManagerInterface.hpp>
+#include <openassetio/managerApi/ManagerStateBase.hpp>
+#include <openassetio/trait/TraitsData.hpp>
+#include <openassetio/utils/ostream.hpp>
+#include "../src/utils/formatter.hpp"
```
feltech, line 17: [consider] Just "utils/formatter.hpp"

src/openassetio-core/tests/utils/PrintableTest.cpp
```diff
@@ -0,0 +1,191 @@
...
+                                    R"('managerState': 0x)");
+        checkBasicPrintableContains(context, R"('locale': {'aTrait': {'aIntTraitProperty': 2}}, )"
+                                             R"('managerState': 0x)");
+      }
+    }
+
+    WHEN("A TraitsData is printed") {
+      THEN("It can be printed via fmt and <<") {
+        checkBasicPrintableByCharacters(
+            *traitsData,
```
feltech, line 179: [fix] Remove `*` prefix, otherwise we're just asserting the same thing twice.

src/openassetio-core/tests/utils/PrintableTest.cpp
feltech, line 168: [fix] Remove `*` prefix

src/openassetio-python/cmodule/src/errors/BatchElementErrorBinding.cpp
```diff
@@ -32,7 +34,8 @@ void registerBatchElementError(const py::module& mod) {
       .def_readonly("code", &BatchElementError::code)
       .def_readonly("message", &BatchElementError::message)
       .def("__repr__", [](const BatchElementError& self) {
```
feltech, line 36: [fix] I think this should be `__str__` and `__repr__` should be more-or-less what it was before, to be consistent with other types, and the idiom that `__repr__` is roughly copy-pasteable into a Python terminal.

src/openassetio-python/tests/package/test_entityreference.py
```diff
@@ -42,9 +42,9 @@ def test_returns_constructor_supplied_string(self):
 
 
 class Test_EntityReferenve_repr:
-    def test_returns_angle_bracket_string_with_class_and_value(self):
+    def test_returns_bracket_string_with_class_and_value(self):
```
feltech, line 45: [question] Minor, just curious why keep this test when others were removed?

feltech (changes requested):
Looking good, just minor niggles.

src/openassetio-core/src/utils/formatter.cpp
```diff
@@ -0,0 +1,198 @@
...
+    const openassetio::InfoDictionary& infoDict, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  std::vector<std::string> flattenedStrMap;
+  flattenedStrMap.reserve(
+      infoDict.size());  // Reserve space for the elements to avoid multiple allocations
+
+  for (const auto& kv : infoDict) {
+    // Format each key-value pair into a single string
+    flattenedStrMap.push_back(
+        fmt::format("`{}`: {}", kv.first, std::visit(ToStringVisitor(), kv.second)));
```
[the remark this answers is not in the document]
feltech: Actually, 

[fix] Backticks!

src/openassetio-core/include/openassetio/utils/ostream.hpp
```diff
@@ -0,0 +1,80 @@
...
+                                                 const EntityReferences& formattable);
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out, const ContextPtr& formattable);
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const ContextConstPtr& formattable);
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out, const Context& formattable);
+
+namespace managerApi {
+
+OPENASSETIO_CORE_EXPORT std::ostream& operator<<(std::ostream& out,
+                                                 const ManagerInterface::Capability& formattable);
```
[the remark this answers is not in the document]
elliotcmorris: By that argument, it seems we should also provide `kCapabilityNames` or an equivilent on `Manager` too then?

src/openassetio-core/src/utils/formatter.hpp
```diff
@@ -0,0 +1,122 @@
...
+};
+
+template <>
+struct fmt::formatter<openassetio::errors::BatchElementError> : fmt::formatter<string_view> {
+  auto format(const openassetio::errors::BatchElementError& ber, format_context& ctx) const
+      -> decltype(ctx.out());
+};
+
+template <>
+struct fmt::formatter<openassetio::trait::TraitSet> : fmt::formatter<string_view> {
```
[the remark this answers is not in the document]
elliotcmorris: fmt has its own implementation of `string_view. `basic_string_view`, which this aliases to.

The docs : 
```
`fmt::basic_string_view`` is used for format strings even if ``std::string_view`` is available to prevent issues when a library is compiled with a different ``-std`` option than the client code (which is not recommended).
```

As for your second question, no I didn't. I figured we're trying to make strings, we should just use strings. There's more than enough trying to be clever going on as it is.

src/openassetio-core/src/utils/formatter.cpp
```diff
@@ -0,0 +1,198 @@
...
+auto fmt::formatter<openassetio::EntityReferences>::format(
+    const openassetio::EntityReferences& entityRefs, format_context& ctx) const
+    -> decltype(ctx.out()) {
+  std::vector<std::string> quotedRefs;
+  quotedRefs.reserve(entityRefs.size());
+  std::transform(
+      entityRefs.cbegin(), entityRefs.cend(), std::back_inserter(quotedRefs),
+      [](const openassetio::EntityReference& ref) { return fmt::format("'{}'", ref.toString()); });
+
+  return fmt::formatter<string_view>::format(fmt::format("[{}]", fmt::join(quotedRefs, ", ")),
```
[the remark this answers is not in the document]
elliotcmorris: I'm looking at that `join` call and don't understand it at all. Does it even take 3 arguments, I thought you could only put in a seperator?

src/openassetio-core/src/utils/formatter.cpp
```diff
@@ -0,0 +1,198 @@
...
+  std::vector<std::string> flattenedStrMap;
+  flattenedStrMap.reserve(
+      strMap.size());  // Reserve space for the elements to avoid multiple allocations
+
+  for (const auto& kv : strMap) {
+    // Format each key-value pair into a single string
+    flattenedStrMap.push_back(fmt::format("'{}': '{}'", kv.first, kv.second));
+  }
+
+  return formatter<string_view>::format(fmt::format("{{{}}}", fmt::join(flattenedStrMap, ", ")),
```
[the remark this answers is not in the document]
elliotcmorris: >  if you really want to get a fmtlib techie solution

I do not ... at all. Requiring this ticket be done as fmtlib formatters at all was a mistake in my opinion, leaning into it harder isn't something I want to contemplate.

src/openassetio-core/tests/utils/PrintableTest.cpp
elliotcmorris: You mean because the pointer overloads call through the the base implementation so there's no need to test both?
My thought was it was worth testing that a pointer and value overload both could be called publicly, as we're providing them as essentially public interface.

src/openassetio-core/src/utils/ostream.cpp
elliotcmorris: Huh, i'd gotten it into my head that these _had_ to be relative like this. :shrug:

src/openassetio-python/cmodule/src/errors/BatchElementErrorBinding.cpp
elliotcmorris: We discussed this during pairing. If we're going to do this, can you relax your insistance that the `str` formulation uses a colon rather than a comma. Then we can wrap the `str` one like we do in `EntityReference`, rather than having 2 different sources of truth for formatting.

src/openassetio-core/include/openassetio/utils/ostream.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include <openassetio/InfoDictionary.hpp>
+#include <openassetio/errors/BatchElementError.hpp>
+#include <openassetio/managerApi/ManagerInterface.hpp>
+#include <openassetio/trait/collection.hpp>
+#include <openassetio/trait/property.hpp>
+#include <openassetio/typedefs.hpp>
+
+OPENASSETIO_FWD_DECLARE(Context)
+OPENASSETIO_FWD_DECLARE_IN_NS(managerApi, HostSession)
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/64e5a271b2799da5a25e2b92927cb1916c1e23fc

src/openassetio-core/include/openassetio/utils/ostream.hpp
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/9d8521f7a75b7710003e9f3be7eed34e83cc6b33

src/openassetio-core/src/utils/formatter.cpp
```diff
@@ -0,0 +1,198 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2023 The Foundry Visionmongers Ltd
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/06f38d25ae3a791caa8462977552cb46f84139b8

src/openassetio-core/src/utils/formatter.hpp
```diff
@@ -0,0 +1,122 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2023 The Foundry Visionmongers Ltd
+#pragma once
+
+#include <memory>
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/042ad80becf7c6481d0ead358344313ce61ecbda

src/openassetio-core/src/utils/formatter.cpp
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/f8a93c2c24f514d83998e570e543464944a73b53

src/openassetio-core/tests/utils/PrintableTest.cpp
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/9b17e8a9244375e32257dc979fa75bd46c0e89f7

src/openassetio-core/tests/utils/PrintableTest.cpp
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/9b17e8a9244375e32257dc979fa75bd46c0e89f7

src/openassetio-core/src/utils/formatter.cpp
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/f390f1731165deab160844f9f19242d8525676fd

src/openassetio-core/src/utils/formatter.cpp
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/6eb7b04db9a17edd720dd1edf8f6f91bd2a7988c

src/openassetio-core/src/utils/ostream.cpp
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/667ffcd50fb62084709323736d8fbd9b236be8d8

src/openassetio-core/tests/utils/PrintableTest.cpp
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/667ffcd50fb62084709323736d8fbd9b236be8d8

src/openassetio-python/tests/package/test_entityreference.py
elliotcmorris: No need to be curious, just not thinking as usual.
https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/3c871896e1e0176d12448b798505fde2c2e3d321

src/openassetio-core/src/utils/formatter.hpp
```diff
@@ -0,0 +1,122 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <fmt/format.h>
+#include <openassetio/managerApi/ManagerInterface.hpp>
+#include <openassetio/trait/collection.hpp>
+#include <openassetio/typedefs.hpp>
+
+OPENASSETIO_FWD_DECLARE_IN_NS(errors, BatchElementError)
```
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/commit/64e5a271b2799da5a25e2b92927cb1916c1e23fc

src/openassetio-python/tests/package/utils/test_printable.py
```diff
@@ -0,0 +1,206 @@
...
+    def test_print_context(self, a_context):
+        # No closing brace on purpose, to account for variant managerState memory address
+        checkBasicPrintableContains(
+            a_context,
+            r"{'locale': {'aTrait': {'aIntTraitProperty': 2}}, 'managerState': 0x",
+            f"Context({str(a_context)}",
+        )
+
+    def test_print_traitsData(self, a_traitsData):
+        checkBasicPrintableByCharacters(
```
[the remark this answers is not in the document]
elliotcmorris: I don't see any reason to choose such a complicated solution vs one that anyone can understand.

src/openassetio-core/include/openassetio/utils/ostream.hpp
```diff
@@ -0,0 +1,80 @@
...
+OPENASSETIO_FWD_DECLARE_IN_NS(hostApi, Manager)
+OPENASSETIO_FWD_DECLARE_IN_NS(trait, TraitsData)
+
+namespace openassetio {
+inline namespace OPENASSETIO_CORE_ABI_VERSION {
```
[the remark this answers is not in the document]
elliotcmorris: :thinking: Fair enough, wouldn't have thought to doc insertion operators, but I guess they're no different to any other function.

There's not much to say on most of them, followed a pattern that just gave an example.
https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/1e2d09ed7ca5700830d57acc4449f092014b8a63

resources/build/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN git clone --branch v42 https://github.com/rollbear/trompeloeil
 RUN cd trompeloeil && cmake -S . -B build && cmake --build build && cmake --install build
 
 # Pull and install fmt
-RUN git clone --branch 9.1.0 https://github.com/fmtlib/fmt
+RUN git clone --branch 10.2.1 https://github.com/fmtlib/fmt
```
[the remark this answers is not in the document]
elliotcmorris: I don't know how to do ... any of that.

The ABI dump is using the `v9` inline namespace, which indicates that it's still using version 9, which is surprising.
Do you have a reference to where you originally discovered the forced export symbol, the original comment you left dosen't provide any guidance. When we were pairing you said that now version 10 dosen't export it so we should update, but I have no idea where or how you learned that.

Given that we changed this during development as a hail mary solve, can we just drop this commit? It's not neccesary and not part of any AC's related to the work done here. Could just plan to do it independently if it ever becomes neccesary ... which it may not for a long while.

src/openassetio-python/tests/package/utils/test_printable.py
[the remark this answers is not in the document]
feltech: I should have explained my reasoning more. The reason I suggest it is to ensure that the string we construct is indeed copy-paste-able to a Python terminal as a dict.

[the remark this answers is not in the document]
feltech: Perhaps an alternative would be to construct a dict and `str` that, then compare the strings as currently?  Would be interesting to see the difference if that doesn't work, at least.

resources/build/Dockerfile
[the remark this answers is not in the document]
feltech: I'm happy to drop the commit.

I think it was fixed in version 10 of the library, which might not correspond to the inline namespace(?)

You're right that we don't have any process instructions for updating the Docker image.  Usually it's obvious its needed because tests fail until you update, but obviously not the case here.

I had in mind to reset back to the commit that bumped fmtlib and run the ABI check, see if it fails, and if so why. The README details running the ABI check locally as one of the final steps.  But perhaps that should be moved to before the snapshot is updated, to reproduce the ABI check failure locally first before moving on to update the snapshots.

src/openassetio-core/include/openassetio/utils/ostream.hpp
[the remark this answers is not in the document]
feltech: Nice, thanks.

src/openassetio-core/include/openassetio/utils/ostream.hpp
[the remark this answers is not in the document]
feltech: Yeah does seem odd that `kCapabilityNames` only exists on `ManagerInterface`. But out of scope here, I guess.

src/openassetio-core/src/utils/formatter.cpp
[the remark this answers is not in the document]
feltech: There are just two arguments, the final "argument", i.e. comma,  is in quotes. I'll try to break it down

```c++
if (entityRefs.empty()) {
  return "[]";
}

std::string result;
result += "['";
result += fmt::join(entityRefs, "', '");
result += "']"
```

I think that works?

src/openassetio-core/tests/utils/PrintableTest.cpp
feltech: Yeah, but you're dereferencing the pointer before passing it to `checkBasicPrintableByCharacters`, so its not testing stringification of the pointer.

src/openassetio-python/cmodule/src/errors/BatchElementErrorBinding.cpp
feltech: So `str` would be `errorCode, 'errorMessage'` and `repr` would be `BatchElementError(errorCode, 'errorMessage')`

I think that still doesn't work really for `repr`, since `errorCode` would be the stringified enum value, rather than the Python type, so is not a valid argument to the `BatchElementError` constructor.

We do have sort-of precedent in breaking that rule for `TraitsData`, since we can't currently construct a `TraitsData` from a Python dict. But I think we should be able to... we've discussed it before, since we've had to implement ad-hoc utility functions to convert between dict and TraitsData more than once (e.g. for BAL and simpleResolver).

Another thought is that ideally we'd use the `str` as the message in a `BatchElementException`, which would suite a `: ` better than `, ` separating the code from the description.  We don't currently include the code in the exception message, though, so not a strong argument.

feltech: Just noticed `Context` is another example where the `repr` doesn't work as a copy-pasteable code. Argument is still valid, but much weaker, that a `Context` _should_ be constructible with a dict argument.

But that's probably a step too far. I think ideally `repr` for `Context` should be updated to match its constructor signature.

feltech: Perhaps the quickest easiest solution is to just remove `__repr__` definitions from `BatchElementError` and `Context` (and perhaps elsewhere)?  So `str(...)` will still work and `repr(...)` will do whatever Python was already doing before.

feltech (approved):
Looking good, other than previous unresolved comments.

src/openassetio-core/src/utils/formatter.cpp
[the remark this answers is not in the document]
elliotcmorris: Oh _I see_, yeah not a bad idea.

This may be quicker, i'd have to profile, although I suspect it's going to be equivalent at best. Temporary vector and Temporary string are similar in performance cost imo, but the real killer is the need to convert the output of `join` to a string.

:shrug: If  `fmt::format` itself is quite slow, you could make this argument, but I think without any metrics either implementation has a shot at being faster, so just leaving this as it is.

src/openassetio-core/tests/utils/PrintableTest.cpp
elliotcmorris: Right yeah, the actual function is called inside the helper duh.

elliotcmorris: ... you're gonna have to explain this one to me face to face. The idea is that you have a pointer, and you test both the stringificaton of the actual data and the pointer, by deferencing the pointer to test the value-date, and then immediately below, not referencing it to test it as a pointer.

feltech: Nope you're right, I completely misread :facepalm:

src/openassetio-core/tests/utils/PrintableTest.cpp
feltech: Ignore me, misread.

src/openassetio-python/tests/package/utils/test_printable.py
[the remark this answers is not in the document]
elliotcmorris: Ah I see. I didn't understand why you'd want to get complicated functions involved, but this lets you check that it's a actual python dict as well, which _is_ great.

https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/42cf2897175a0d89b9808cce3740cc8a00d57232

src/openassetio-core/src/utils/formatter.cpp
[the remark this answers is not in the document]
elliotcmorris: https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/8a6fff22ba018148b36f05e2a06f78b4307f715d

resources/build/Dockerfile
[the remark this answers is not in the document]
elliotcmorris: Aye. I just dropped the commit on this.

src/openassetio-python/cmodule/src/errors/BatchElementErrorBinding.cpp
elliotcmorris: Removed the repr's.

https://github.com/OpenAssetIO/OpenAssetIO/pull/1329/commits/397559c0a76b9ff00cb7e18a4285d8e92bcf0a52

src/openassetio-python/tests/package/utils/test_printable.py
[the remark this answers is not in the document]
feltech: Nice!

feltech (approved):
LGTM!

PR #1332, opened 2024-06-19
Revert tomlplusplus recipe revision pin

feltech:
## Description

An update to the Conan recipe for tomlplusplus meant that exceptions were disabled and there was no way to re-enable them, so we had to pin the recipe to the previous revision.

An upstream fix has now been pushed (conan-io/conan-center-index#24385), so we can revert the recipe revision pin.


feltech:
> Intentionally not reverting the allow-revisions variable? I suppose it dosen't hurt to leave it.

No that wasn't intentional. :thinking: In general I would say its useful global config to have, but since its explicitly specified as an env var it raises the question of "why?" to anyone reading it.  I'm of a mind to remove it.

elliotcmorris (approved):

PR #1334, opened 2024-06-21
[Release] Bump version to v1.0.0-beta.2.2

elliotcmorris:
## Description

This appears to be a minor change without any ABI or API breaks.
As such, I don't believe we need to do any additional annotation of the release notes? We were intending to split changes into manager and hosts breaks, but there arn't any, so nothing to do?

Sort of weird that this isn't a major change considering it has the C++ plugin system in it, but it's non-breaking as far as I know.


RELEASE_NOTES.md
```diff
@@ -1,9 +1,16 @@
...
-v1.0.0-beta.x.y
+v1.0.0-beta.2.2
 ---------------
 
+_This release introduces new features, and remains binary and source
+compatible for hosts, and source compatible for managers._
+
+_The addition of a new virtual method,
+ `LoggerInterface.isSeverityLogged`, introduces a binary incompatibility
+ for upgrading managers._
```
feltech, line 12: [fix] I think we have to be more strict than this, because on Windows apparently any new virtual method can cause ABI breaks.

So I think we remain source compatible for both hosts and managers, but that's it.

feltech (changes requested):

RELEASE_NOTES.md
elliotcmorris: Is this new information since we discussed on Friday, when we thought only managers were a problem?
In any case, i'm fine with being more careful than we need to be, updated the message to be broader.
https://github.com/OpenAssetIO/OpenAssetIO/pull/1334/commits/1308cf32066ca64fbd797cde3294845762211faa

PR #1335, opened 2024-06-24
[Build] Update all macos runners to macos13

elliotcmorris:
Macos-11 is being retired. We gotta update.


feltech (approved):
Would be nice to have a bit more detail in the commit message about why 12 vs 13/14/latest. Otherwise, LGTM.

PR #1343, opened 2024-07-02
Improve speed of CMake configure

feltech:
We were needlessly repeating the same work to compute platform warnings and IPO capability in the `openassetio_set_default_target_properties` function every time it was called.

On Linux the slowdown is barely noticeable. However, on Windows the slowdown is frustrating. A re-configure of a hot build directory goes down from 1 min 43 secs before this commit to a mere 8 secs, on Windows.

### DIFF
--- a/cmake/DefaultTargetProperties.cmake
+++ b/cmake/DefaultTargetProperties.cmake
@@ -1,5 +1,16 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2013-2022 The Foundry Visionmongers Ltd
+include_guard(GLOBAL)
+include(CompilerWarnings)
+include(CheckIPOSupported)
+
+if (OPENASSETIO_ENABLE_IPO)
+    check_ipo_supported(RESULT _is_ipo_supported)
+    if (NOT _is_ipo_supported)
+        message(WARNING "OPENASSETIO_ENABLE_IPO is not supported, option ignored")
+        set(OPENASSETIO_ENABLE_IPO OFF)
+    endif ()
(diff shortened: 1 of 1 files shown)


cmake/DefaultTargetProperties.cmake
```diff
@@ -1,5 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2013-2022 The Foundry Visionmongers Ltd
+include_guard(GLOBAL)
+include(CompilerWarnings)
+include(CheckIPOSupported)
+
+check_ipo_supported(RESULT _is_ipo_supported)
```
SamCrooksFoundry, line 7: Should these lines also be guarded with:
```
    if (OPENASSETIO_ENABLE_IPO)
```

cmake/DefaultTargetProperties.cmake
```diff
@@ -1,5 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2013-2022 The Foundry Visionmongers Ltd
+include_guard(GLOBAL)
```
SamCrooksFoundry, line 3: [comment/no action] I didn't realise that CMake had include_guards.  I can see why it needs them, but I think it's ugly that such a high level language requires them.

cmake/DefaultTargetProperties.cmake
feltech: Good spot, refactored a little in 758931b8

SamCrooksFoundry (approved):

Issue #1309, opened 2024-05-15
Investigate how to make .pyi stubs work in `--editable` installs


feltech:
## What

Investigate whether and how IDEs can make use of `.pyi` stub files for our `_openassetio` module in a pip `--editable` install.

## Why

We do not install `.pyi` stub files as part of a pip `--editable` install, since during development of https://github.com/OpenAssetIO/OpenAssetIO/pull/1306 neither PyCharm nor VS Code could made to make use of them. The effort to make this work was abandoned due to running out the clock, and this follow-up issue made instead.

During development of OpenAssetIO itself it would be very useful to work in `--editable` mode, such that IDEs provide code completions for both a recently compiled extension module as well as the live in-progress sources.  Currently we have to choose one or the other.

feltech:
Updated description to reflect the fact https://github.com/OpenAssetIO/OpenAssetIO/issues/1340 broke editable installs (and it wasn't worth fixing given the lack of usefulness). Kept this issue around since it's a nice-to-have. But at time of writing the maintenance burden isn't worth the small benefit.

Issue #1207, opened 2023-11-15
Consistent CMake style


feltech:

feltech:
Inconsistent styling does come up in reviews too e.g. https://github.com/OpenAssetIO/OpenAssetIO/pull/1348#discussion_r1678140343

Perhaps cmake-lint can enforce more than we've currently got it configured for?  

Otherwise we can at least document as part of our style guide, so we can point to it when it comes up in a review.

PR #1348, opened 2024-07-10
Fix C++ plugin loading from Python

### LINKED ISSUE #1340: Fix loading C++ plugins from Python
## What

Allow C++ plugins to be loaded from a Python environment for both CMake shared library installs as well as PyPI pip installs.

## Why

Currently, C++ plugins cannot be loaded by our PyPI package, because the Python extension module `openassetio._openassetio` statically links in the core C++ library, and hides all its symbols, so the plugin cannot link when loaded at runtime.

There are three issues at play here
- The `_openassetio` library hides all symbols from the (static) core C++ library, so the plugin cannot link when loaded.
- Python (on POSIX) loads with `RTLD_LOCAL`, meaning the symbols are unavailable to subsequently loaded libraries, even if they are public.
- Ideally the plugin would have no dependency on the core OpenAssetIO C++ library, and would instead rely on symbols being already available in the process environment when it is loaded. 
    - However, even if the `_openassetio` library and core C++ library are split into two shared libraries, `RTLD_LOCAL` is transitive, so would apply to the core C++ library too. So the plugin cannot rely on the symbols being present in the environment.
    - Windows is another question, and probably enforces `RTLD_LOCAL`-like behaviour regardless (needs more research).
    - A workaround, therefore, and possibly necessary (depending on Windows), is that the plugin must explicitly depend on the OpenAssetIO core C++ library.
 
Original investigation: https://github.com/OpenAssetIO/OpenAssetIO/issues/1324#issuecomment-2197460862

feltech:
## Description

Closes #1340. 

Building OpenAssetIO as a static library almost certainly precludes loading C++ manager plugins.  This is due to plugins depending explicitly on the OpenAssetIO core C++ shared library and/or depending on the symbols of that library being exported and public, in order to link properly at runtime.

The most straightforward and universal fix to this is to ensure OpenAssetIO is only ever built as a bunch of shared libraries.

So update the Python distribution to bundle OpenAssetIO as several shared libraries, rather than a single extension module.

When building/installing as a CMake package on Windows, we must unfortunately duplicate the dlls, with one copy in the `lib` subdirectory, for use by C++ consumers, and one copy in the `site-packages/openassetio` directory, for use by Python consumers. This was already being done, though as a slightly hacky fix early in the project to get Windows builds working (#244 - related follow-up to give Windows more love #538). So consolidate that logic with code comments explaining why.

With these changes, `pip install --editable` no longer works. Rather than fix it, disable `--editable` installs, for now. They're not very useful since the vast majority of code is in C++, so not worth the effort to support at the moment. Support will be re-visited in #1309.

Update BUILDING instructions to reflect the new structure of Python distributions. Plus fix some historic inaccuracies in that doc.

Also, ninja commit some fixes to running `ctest` with multi-config generators that were discovered whilst investigating this issue.

### DIFF
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -166,6 +166,8 @@ jobs:
           # warnings, sanitizers. Will need additional CMake preset(s).
           OPENASSETIO_TRAITGENTEST_CMAKE_PRESET: test
           CONAN_USER_HOME: ~/conan
+          # For Windows:
+          OPENASSETIO_DLL_PATH: ${{ github.workspace }}/dist/bin
 
       - name: Checkout OpenAssetIO-Test-CMake
         uses: actions/checkout@v4
@@ -193,6 +195,8 @@ jobs:
         env:
           CMAKE_PREFIX_PATH: ${{ github.workspace }}/dist
           CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
+          # For Windows:
+          OPENASSETIO_DLL_PATH: ${{ github.workspace }}/dist/bin
 
       # TODO(DF): SimpleCppManager should (also) be tested as part of
       # Python e2e tests, once we fix loading C++ plugins from Python.
@@ -226,3 +230,5 @@ jobs:
           OPENASSETIO_LOGGING_SEVERITY: 1
           OPENASSETIO_PLUGIN_PATH: ${{ github.workspace }}/dist/plugins
           OPENASSETIO_DEFAULT_CONFIG: ${{ github.workspace }}/examples/manager/SimpleCppManager/tests/resources/openassetio_config.toml
+          # For Windows:
+          OPENASSETIO_DLL_PATH: ${{ github.workspace }}/dist/bin
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -198,8 +198,49 @@ jobs:
           # For Windows:
           OPENASSETIO_DLL_PATH: ${{ github.workspace }}/dist/bin
 
-      # TODO(DF): SimpleCppManager should (also) be tested as part of
-      # Python e2e tests, once we fix loading C++ plugins from Python.
+  cpp-manager-plugin:
+    name: C++ manager plugin on ${{ matrix.config.os }}
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - os: windows-2019
+            preamble: call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat" x64
+            site-packages: Lib/site-packages
+            shell: cmd
+          - os: ubuntu-20.04
+            site-packages: lib/python3.9/site-packages
+            shell: bash
+          - os: macos-12
+            site-packages: lib/python3.9/site-packages
+            shell: bash
+    defaults:
+      run:
+        shell: ${{ matrix.config.shell }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Bootstrap
+        uses: ./.github/bootstrap_platform
+
+      - name: Install test dependencies
+        run: |
+          python -m pip install pytest
+
+      - name: Build/install OpenAssetIO (CMake)
+        run: >
+          ${{ matrix.config.preamble }}
+
+          cmake -S . -B build --install-prefix ${{ github.workspace }}/dist
+          -DCMAKE_BUILD_TYPE=RelWithDebInfo
+
+          cmake --build build --parallel --config RelWithDebInfo
+
+          cmake --install build --config RelWithDebInfo
+        env:
+          CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
 
       - name: Build/install SimpleCppManager
         run: >
@@ -219,9 +260,9 @@ jobs:
         env:
           CMAKE_PREFIX_PATH: ${{ github.workspace }}/dist
 
-      - name: Test SimpleCppManager with simpleResolver
+      - name: Test SimpleCppManager with simpleResolver (via CMake install tree)
         run: >
-          python '${{ github.workspace }}/examples/host/simpleResolver/simpleResolver.py'
+          python ${{ github.workspace }}/examples/host/simpleResolver/simpleResolver.py
           openassetio-mediacreation:identity.DisplayName simplecpp://test/entity/1
           | grep 'Test Entity 1'
 
@@ -232,3 +273,21 @@ jobs:
           OPENASSETIO_DEFAULT_CONFIG: ${{ github.workspace }}/examples/manager/SimpleCppManager/tests/resources/openassetio_config.toml
           # For Windows:
           OPENASSETIO_DLL_PATH: ${{ github.workspace }}/dist/bin
+
+      - name: Build/install OpenAssetIO (setup.py)
+        run: |
+          ${{ matrix.config.preamble }}
+          python -m pip install src/openassetio-python
+        env:
+          CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
+
+      - name: Test SimpleCppManager with simpleResolver (via Python distribution)
+        run: >
+          python ${{ github.workspace }}/examples/host/simpleResolver/simpleResolver.py
+          openassetio-mediacreation:identity.DisplayName simplecpp://test/entity/1
+          | grep 'Test Entity 1'
+
+        env:
+          OPENASSETIO_LOGGING_SEVERITY: 1
+          OPENASSETIO_PLUGIN_PATH: ${{ github.workspace }}/dist/plugins
+          OPENASSETIO_DEFAULT_CONFIG: ${{ github.workspace }}/examples/manager/SimpleCppManager/tests/resources/openassetio_config.toml
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-# Copyright 2013-2022 The Foundry Visionmongers Ltd
+# Copyright 2013-2024 The Foundry Visionmongers Ltd
 
 #-----------------------------------------------------------------------
 # CMake preamble.
@@ -110,7 +110,8 @@ include(CMakeDependentOption)
 # TODO(DF): We may end up needing to support building of both shared
 #   and static libs simultaneously. Also, using this means we inherit
 #   the option if openassetio is used as a CMake subproject.
-option(BUILD_SHARED_LIBS "Set to OFF to build static libraries" ON)
+option(BUILD_SHARED_LIBS
+    "Set to OFF to build static libraries. Static libaries are not recommended." ON)
 
 # Enable Python bindings build by default.
 option(OPENASSETIO_ENABLE_PYTHON "Build Python bindings" ON)
@@ -218,6 +219,20 @@ option(OPENASSETIO_ENABLE_CPPLINT "Enable cpplint linter during build" OFF)
 
 # Enable cmake-lint linter.
 option(OPENASSETIO_ENABLE_CMAKE_LINT "Enable cmake-lint linter during build" OFF)
+
+
+#-----------------------------------------------------------------------
+# Warn against static library builds
+
+if (NOT BUILD_SHARED_LIBS)
+    message(WARNING [[
+Building OpenAssetIO as a static library is not recommended. For C++
+plugins to load, the necessary symbols must be exported and be available
+to link against. In addition, C++ plugins will generally have a
+dependency on the OpenAssetIO core C++ shared library file, and if that
+file is not available (because it has been statically linked into a
+larger binary), then the plugin will fail to load.]])
+endif()
 
 
 #-----------------------------------------------------------------------
--- a/doc/BUILDING.md
+++ b/doc/BUILDING.md
@@ -108,8 +108,14 @@ cmake --install build
 The artifacts are installed to `/build/dist`. This default location can
 be overridden by setting `CMAKE_INSTALL_PREFIX` prior to build.
 
-By default, OpenAssetIO builds as a shared library. This can be
-overridden by setting `BUILD_SHARED_LIBS` prior to build.
+By default, OpenAssetIO builds as a shared library.
+
+> **Warning**
+>
+> It is strongly recommended that OpenAssetIO is built as a shared
+> library, to ensure that OpenAssetIO C++ manager plugins can load
+> correctly.
+
 If shared library builds are disabled, then the core library will
 be built and installed as a static library, and statically linked into
 the Python module.
@@ -168,7 +174,6 @@ behaviour of the build.
 
 | Option                                            | Description                                                           | Default |
 |---------------------------------------------------|-----------------------------------------------------------------------|---------|
-| `BUILD_SHARED_LIBS`                               | `ON` to build shared libraries. `OFF` to build static libraries.      | `ON`    |
 | `OPENASSETIO_ENABLE_PYTHON`                       | Additionally build python bindings                                    | `ON`    |
 | `OPENASSETIO_ENABLE_PYTHON_INSTALL_DIST_INFO`     | Create a dist-info metadata directory alongside Python installation   | `ON`    |
 | `OPENASSETIO_ENABLE_C`                            | Additionally build C bindings                                         | `OFF`   |
--- a/doc/BUILDING.md
+++ b/doc/BUILDING.md
@@ -124,13 +124,6 @@ In installing the Python module, OpenAssetIO creates a valid Python
 package, placed at
 `build/dist/lib/{python-version}/site-packages/openassetio`.
 
-> **Note**
->
-> When built this way, the `openassetio` Python package will not
-> function unless the binary artifacts are also made available to your
-> Python environment.
->
-
 ### Building via pip
 
 OpenAssetIO can alternately be built and installed via `pip`, which may
@@ -144,22 +137,6 @@ python -m pip install src/openassetio-python
 This will automatically build and install the Python extension module
 into your python environment, along with the Python sources. After this,
 you should be able to simply `import openassetio` from your python shell.
-
-> **Note**
->
-> `pip install src/openassetio-python` is not a substantially different
-> operation to calling CMake directly as described above (and as such,
-> all the same dependency requirements apply).
->
-> However, be aware that once the C++
-> library is built, it is statically, rather than dynamically, linked
-> into the python libraries, and is not usable as a standalone
-> OpenAssetIO C++ library.
->
-> This means that this method of building is only appropriate for the
-> exclusively python focused developer. Anyone wishing to write
-> non-python hosts or managers should build via the CMake method
-> described above.
 
 ### Sandboxed builds
 
--- a/doc/BUILDING.md
+++ b/doc/BUILDING.md
@@ -155,14 +155,6 @@ you should be able to simply `import openassetio` from your python shell.
 > non-python hosts or managers should build via the CMake method
 > described above.
 
-For users that wish to modify the Python sources and test their changes
-without reinstalling, [editable installs](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs)
-are also supported
-
-```shell
-python -m pip install --editable src/openassetio-python
-```
-
 ### Sandboxed builds
 
 For a more convenient "out of the box" build environments, OpenAssetIO
--- a/src/openassetio-python/cmodule/CMakeLists.txt
+++ b/src/openassetio-python/cmodule/CMakeLists.txt
@@ -30,8 +30,21 @@ install(
     TARGETS openassetio-python-module
     EXPORT ${PROJECT_NAME}_EXPORTED_TARGETS
     DESTINATION ${_install_subdir}
-    COMPONENT openassetio-python-module
+    COMPONENT openassetio-python-distribution
 )
+if (BUILD_SHARED_LIBS)
+    # For a shared library build, we must ensure the core C++ library
+    # dependency is installed when the Python extension module component
+    # is installed. The openassetio-python-distribution component is
+    # targeted for install when building Python wheels - see setup.py.
+    # In addition, ensure the Python bridge library is also available,
+    # to enable (the albeit rare case of) C++ plugin direct interaction
+    # with Python objects.
+    install(
+        TARGETS openassetio-core openassetio-python-bridge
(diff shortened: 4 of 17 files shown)


foundry-markf:
Regarding my offline comment earlier on

> When building/installing as a CMake package on Windows, we must unfortunately duplicate the dlls, with one copy in the lib subdirectory, for use by C++ consumers, and one copy in the site-packages/openassetio directory, for use by Python consumers.

It feels like you're trying to solve the DLL discovery problem in the _library_, whereas it's a _consumer_ problem.
By this I mean that the consumer of OpenAssetIO is expected to 1) distribute library files however it wishes, 2) set up DLL search paths in both C++ and Python to discover them. Not all applications are the same in their distribution structure, so making any assumptions it the library will not satisfy everyone.


You've mentioned PySide in the linked issue. On Windows, you'll find a shiboken DLL in its bin folder, while the Python extension modules are in lib/site-packages/shibokenX folder - no duplication.

Also, OpenUSD as a popular library with Python bindings, the C++ DLLs are in lib, and the Python C Extension modules are in a lib/python/pxr package. No duplication again.


Having pondered this a little longer, I imagine you might be saying there's a need to duplicate because there are two types of deployment for OpenAssetIO? One via a C++ package manager mechanism (for example) and one via PyPi? And they must be standalone?

I'm admittedly mostly thinking about consumers that have both C++ and Python binding support, so the duplication seems wasteful (and could be confusing).

With the current setup, if a consumer only takes a C++ package, or the pip package, but not both, then there is no duplication.

But, in the case where a consumer takes both, then there is a chance of duplication unless consumers can 'do the right thing' with their scripts to create a distributable?

foundry-markf:
> With these changes, pip install --editable no longer works

In what way did it break?

CMakeLists.txt
```diff
@@ -110,7 +110,7 @@ include(CMakeDependentOption)
 # TODO(DF): We may end up needing to support building of both shared
 #   and static libs simultaneously. Also, using this means we inherit
 #   the option if openassetio is used as a CMake subproject.
-option(BUILD_SHARED_LIBS "Set to OFF to build static libraries" ON)
+option(BUILD_SHARED_LIBS "Set to OFF to build static libraries. Not recommended." ON)
```
foundry-markf, line 113: I know what you mean here, but would another reader think 'Not recommended' is referring to 'don't use this option'?

Should it be clarified to 'Static libraries are not recommended'?

This would at least be consistent with the warning message that follows below.

.github/workflows/integrations.yml
```diff
@@ -194,8 +194,49 @@ jobs:
...
+
+      - name: Install test dependencies
+        run: |
+          python -m pip install pytest
+
+      - name: Build/install OpenAssetIO (CMake)
+        run: >
+          ${{ matrix.config.preamble }}
+
+          cmake -S . -B build --install-prefix ${{ github.workspace }}/dist
```
foundry-markf, line 232: Nit: As you've got a CMakePresets.json, I'm surprised it's not being used here. The advice for presets in the Professional CMake book is that they should describe your CI 'ready to release' setup.

doc/BUILDING.md
```diff
@@ -181,23 +146,24 @@ behaviour of the build.
...
 | `OPENASSETIO_ENABLE_TESTS`                        | Additionally build tests                                              | `OFF`   |
 | `OPENASSETIO_ENABLE_PYTHON_TEST_VENV`             | Automatically create environment when running tests                   | `ON`    |
 | `OPENASSETIO_WARNINGS_AS_ERRORS`                  | Treat compiler warnings as errors                                     | `OFF`   |
 | `OPENASSETIO_ENABLE_IPO`                          | Enable Interprocedural Optimization, aka Link Time Optimization (LTO) | `ON`    |
 | `OPENASSETIO_ENABLE_POSITION_INDEPENDENT_CODE`    | Enable position independent code for static library builds            | `ON`    |
 | `OPENASSETIO_GLIBCXX_USE_CXX11_ABI`               | When building under gcc, use the new C++11 library ABI                | `OFF`   |
 | `OPENASSETIO_ENABLE_NEW_DTAGS`                    | Set RUNPATH, overriding RPATH, on linux platforms                     | `OFF`   |
-| `OPENASSETIO_ENABLE_COVERAGE`                     | Enable coverage reporting for gcc/clang                               | `OFF`   |
```
foundry-markf, line 195: Just a comment about unrelated changes; I keep having to go back to the commit list to figure out _why_ a change occurred that seems unrelated to improving C++ plugin support, e.g. why has OPENASSETIO_ENABLE_COVERAGE been removed?

Would it not be better to have this as part of a separate PR?

src/openassetio-python/cmodule/CMakeLists.txt
```diff
@@ -151,7 +181,7 @@ if (OPENASSETIO_ENABLE_PYTHON_STUBGEN)
             " pybind11-stubgen not found: status=${_pybind11_stubgen_exit_code}:"
             " ${_pybind11_stubgen_output}"
         )
-    endif()
+    endif ()
```
foundry-markf, line 184: Nit: an unexpected change, as in CMake, commands are _functions_ so having the parentheses bind to the command name without a space is what I'd expect to see.

Why did this change?

src/openassetio-python/cmodule/CMakeLists.txt
```diff
@@ -30,8 +30,38 @@ install(
...
+    if (UNIX)
+        install(
+            TARGETS openassetio-core openassetio-python-bridge
+            COMPONENT openassetio-python-distribution
+        )
+    elseif (WIN32)
+        install(
+            TARGETS openassetio-core openassetio-python-bridge
+            COMPONENT openassetio-python-distribution
+            # On Windows we must, unfortunately, duplicate the
```
foundry-markf, line 52: Perhaps I'm not seeing the bigger picture here (unchanged files perhaps). But I'm not immediately seeing a duplication of DLLs.

I'm going to check build logs.

foundry-markf (commented):
Looking at this with very little hands prior experience of building OpenAssetIO in its current form, so 'feeling the diff' is harder.

feltech:
> > With these changes, pip install --editable no longer works
> 
> In what way did it break?
> 
> I did read [3ebab87](https://github.com/OpenAssetIO/OpenAssetIO/commit/3ebab87787375194afc31583cb57e92b9221af21) and it's to do with needing to distribute multiple shared libs, but I still wasn't clear on the actual bit that was broken.

I explained it a bit more in the follow-up issue https://github.com/OpenAssetIO/OpenAssetIO/issues/1309.

> Since https://github.com/OpenAssetIO/OpenAssetIO/issues/1340 we build OpenAssetIO as shared libraries by default, but a naive editable install will only install the Python extension module, and skip over the shared libraries it depends on.

So to fix it we would have to ensure the other shared libraries beyond just the Python extension module are "installed" - where "installed" in this case means "copied into the source tree after building".

CMakeLists.txt
feltech: Good point, rather ambiguous as it stands.

.github/workflows/integrations.yml
feltech: I think in this particular case a preset wouldn't give us much?  Other than perhaps setting the build type.

More generally we probably could make more use of presets in various bits of CI - its a little bit mixed at the moment.

There is a counter-argument, though - by not using presets we're dogfooding the defaults, which for OpenAssetIO is intended to produce a release-ready build, as would typically be desired by consumers.

doc/BUILDING.md
feltech: Yeah I wrestled with this a bit. 

I've started deliberately adding ninja commits to fix things "whilst I'm in the neighbourhood", rather than create whole new issues/PRs, to reduce the process burden, given the reduced human resource working on the project at the moment.

I did assume most reviewers would review commit-by-commit, so it wouldn't be too annoying.

But if it's going to be a frustration for reviewers, then I can revisit this policy.

src/openassetio-python/cmodule/CMakeLists.txt
feltech: :grin: This is CLion's auto-format of CMake files. I think most of our CMake conforms to CLion's preferences, but it's not been enforced.

Interesting point about commands being functions, I'd never thought about it that way, just assumed the brackets were an odd notational choice.

src/openassetio-python/cmodule/CMakeLists.txt
feltech: The duplication comes because of the non-standard `DESTINATION` here.  There are also install rules for those targets independently, which will install to the default destination (`lib`/`bin`).

feltech:
> Regarding my offline comment earlier on
> 
> > When building/installing as a CMake package on Windows, we must unfortunately duplicate the dlls, with one copy in the lib subdirectory, for use by C++ consumers, and one copy in the site-packages/openassetio directory, for use by Python consumers.
> 
> It feels like you're trying to solve the DLL discovery problem in the _library_, whereas it's a _consumer_ problem. By this I mean that the consumer of OpenAssetIO is expected to 1) distribute library files however it wishes, 2) set up DLL search paths in both C++ and Python to discover them. Not all applications are the same in their distribution structure, so making any assumptions it the library will not satisfy everyone.
> 
> You've mentioned PySide in the linked issue. On Windows, you'll find a shiboken DLL in its bin folder, while the Python extension modules are in lib/site-packages/shibokenX folder - no duplication.
> 
> Also, OpenUSD as a popular library with Python bindings, the C++ DLLs are in lib, and the Python C Extension modules are in a lib/python/pxr package. No duplication again.
> 
> Having pondered this a little longer, I imagine you might be saying there's a need to duplicate because there are two types of deployment for OpenAssetIO? One via a C++ package manager mechanism (for example) and one via PyPi? And they must be standalone?
> 
> I'm admittedly mostly thinking about consumers that have both C++ and Python binding support, so the duplication seems wasteful (and could be confusing).
> 
> With the current setup, if a consumer only takes a C++ package, or the pip package, but not both, then there is no duplication.
> 
> But, in the case where a consumer takes both, then there is a chance of duplication unless consumers can 'do the right thing' with their scripts to create a distributable?

As discussed offline, we do have to think about 3 use-cases
- Pure Python consumers, which could take a couple of forms
   - `pip install openassetio` 
   -  Point their `PYTHONPATH` to an OpenAssetIO build - e.g. Rez.
- Pure C++ applications that only care about the core C++ library and don't support Python.
- C++ applications with an embedded Python interpreter.

I think the root of the Windows packaging confusion problem boils down to how Windows has always been a second-class citizen in OpenAssetIO.

So lets talk POSIX first.

OpenAssetIO's CMake is configured for two cases
* By default, such that `cmake --install` puts things in the expected place assuming a POSIX install prefix (e.g. `/usr/local`). This includes shared libraries in `lib` and a Python extension module (and pure Python sources) in `lib/pythonX.Y/site-packages/openassetio`, where the Python module has an `RPATH` /`@loader_path` to `$ORIGIN/../../..` so it can locate its dependency.
* For crafting a Python distribution package (i.e. `pip install .`), such that `cmake --preset setuptools` combined with `cmake --install --component openassetio-python-distribution` will bundle all shared libraries alongside one-another under an `openassetio` subdirectory of the install prefix, ready to be consumed by Python's `setuptools` for creating a wheel.

Now for Windows. The second case above is actually the more straightforward, since bundling all shared libraries alongside one-another is supported out-of-the-box by the Python interpreter on Windows, without any additional need to `os.add_dll_directory`.

For the first case, a POSIX directory structure doesn't make as much (any?) sense for Windows, but for consistency and, again, because Windows hasn't received much love compared to POSIX, we do our best to square-peg-round-hole it. This means, naively, `.dll`s go in `bin`, `.lib`s go in `lib` and Python extension modules go in `Lib/site-packages`.

Using this scheme, for the pure C++ application use-case, all is good, just treat the dlls/libs as any other build/deploy dependency and ignore Python. For C++ applications with an embedded Python interpreter we're also _probably_ OK - it depends on the structure of the application (wrt dll load order vs. Python dll search paths) - but that's the application's problem, since they'll be building against and bundling, rather than using directly.

Its the `PYTHONPATH`-extending pure Python use-case where we get into a muddle. This is admittedly probably the rarest case, but we rely on this in various unit tests, and we know it's a common pattern for tools like Rez.

You can't simply point `PYTHONPATH` to the `site-packages` directory, because the Python extension module depends on one of the dlls in `bin`. You can't even fix this by setting `PATH` to point to `bin` because Python 3.8+ doesn't allow this.

We _could_ use  `os.add_dll_directory` to point to `bin`. But this has a huge disadvantage in that it hardcodes a search path that might not exist in the other use cases (C++ applications, Python wheels), so we'd have to make `add_dll_directory` either conditional or configurable. For example, I had wondered about having a separate `rpath.txt` file, whose contents would be passed at runtime to `add_dll_directory`, thus providing a customisation point for packagers (e.g. when assembling a conan package, or creating a Python wheel).

The simpler solution that satisfies all use-cases is to duplicate the dlls. This leaves a nasty taste in my mouth. And as you say, could be confusing to consumers.

An even simpler solution would be to say this use-case is unsupported out-of-the box and its up to the consumer to either copy dlls around or `add_dll_directory` themselves :thinking:  - would require some tweaking of our tests, but doable... perhaps that is the better way?

feltech:
> An even simpler solution would be to say this use-case is unsupported out-of-the box and its up to the consumer to either copy dlls around or add_dll_directory themselves

Did a quick investigation. Turned off the duplication of dlls, then ran `ctest` in Windows.  Most of the errors can be fixed with a snippet like the following at the top of the root test module(s).

```python
if os.name == "nt":
    spec = importlib.util.find_spec("openassetio")
    dll_dir = pathlib.Path(spec.origin).parent.parent.parent.parent / "bin"
    if (dll_dir / "openassetio.dll").exists():
        os.add_dll_directory(str(dll_dir))
```

However, the remaining error is a bit of a blocker. 

The `openassetio.test.manager` subpackage is an executable command-line utility. It is the API Compliance test suite to help managers validate their plugin.  It has its own test cases, and is used in downstream test cases. Executing this involves implicitly importing  `openassetio`, which imports the `_openassetio` extension module, which can't find its dll dependency.

The only quick fix would be to add the snippet above to the main openassetio top-level `__init__.py`, but as I said above

> this has a huge disadvantage in that it hardcodes a search path

I would rather not have that snippet above appear anywhere in the distributed code 🤔

feltech:
A flexible solution could be to support an `OPENASSETIO_DLL_PATH` environment variable, which can be set by Windows consumers, then read in the top-level `__init__.py` of `openassetio` where it will pass the value to `add_dll_directory`.

.github/workflows/integrations.yml
foundry-markf: > Oh I guess you can set the build directory with presets too.

Yes

> There is a counter-argument, though - by not using presets we're dogfooding the defaults
There isn't a need to set CMake cache variables in a preset - you can still use the defaults. It's just that

```
cmake --workflow --preset ci_Release
```

which does as much as you tie into a workflow (configure + build (+ test) (+packaging)) does it all in one command, which is very appealing for CI.


The point the documentation makes is that the presets are documenting (if used for CI) what you _expect_ to always work, so is a starting point for everyone, including new developers. If you don't go down the presets route, and there are a whole bunch of options, then you do wonder 'which ones should I set?' I find that a lot in the thirdparty libraries from upstream.

doc/BUILDING.md
foundry-markf: I suspect with the size of the team, that we'll need to be more flexible.

A mention in the PR description of 'oh I've also fixed this, but it's independent' sort of comment would help that it's intentionally here amongst the real meat of the PR.

src/openassetio-python/cmodule/CMakeLists.txt
foundry-markf: So this comes down to different people with different editors applying different code preference rules?

Maybe something to stick on the backlog of 'these are the rules, make sure your editor conforms'.

foundry-markf:
> > An even simpler solution would be to say this use-case is unsupported out-of-the box and its up to the consumer to either copy dlls around or add_dll_directory themselves
> 
> Did a quick investigation. Turned off the duplication of dlls, then ran `ctest` in Windows. Most of the errors can be fixed with a snippet like the following at the top of the root test module(s).
> 
> ```python
> if os.name == "nt":
>     spec = importlib.util.find_spec("openassetio")
>     dll_dir = pathlib.Path(spec.origin).parent.parent.parent.parent / "bin"
>     if (dll_dir / "openassetio.dll").exists():
>         os.add_dll_directory(str(dll_dir))
> ```

Something similar to that is common in our thirdparty package tests of libraries that use Python, so that their Python bindings can find their libraries. So it's not uncommon, and is the kind of thing I had in mind that 'the consumer of the library sets the environment up'.

> 
> However, the remaining error is a bit of a blocker.
> 
> The `openassetio.test.manager` subpackage is an executable command-line utility. It is the API Compliance test suite to help managers validate their plugin. It has its own test cases, and is used in downstream test cases. Executing this involves implicitly importing `openassetio`, which imports the `_openassetio` extension module, which can't find its dll dependency.
> 
> The only quick fix would be to add the snippet above to the main openassetio top-level `__init__.py`, but as I said above
> 
> > this has a huge disadvantage in that it hardcodes a search path
> 
> I would rather not have that snippet above appear anywhere in the distributed code 🤔

The bit that bothers me the most, is whether duplicating the DLLs is just techdebt that will come back to bite in the future.

> A flexible solution could be to support an OPENASSETIO_DLL_PATH environment variable, which can be set by Windows consumers, then read in the top-level __init__.py of openassetio where it will pass the value to add_dll_directory.

This appeals more than duplicating files.

As Python has this security workaround for search path locations only on Windows, it's not unreasonable to expect to do additional setup on Windows.

Perhaps with a nice error message if it's unset?

CMakeLists.txt
feltech: Updated in 9f4a1037

doc/BUILDING.md
feltech: I think I need to be as kind to reviewers as possible.  

I think a good middle ground is to continue (pragmatically) skipping creating issues, triage, estimation, scheduling, etc, and just do the work when I'm in the neighbourhood - but to at least split the work into a separate PR.  That's not _much_ more process burden (it does create possible merge conflicts and temporary CI failures), and is significantly easier to review when self contained and small.

With that in mind, I have removed the first two commits and put them up as separate PRs (#1361 and #1362).

src/openassetio-python/cmodule/CMakeLists.txt
feltech: We have a related issue https://github.com/OpenAssetIO/OpenAssetIO/issues/1207 - I've added a comment there to ensure we look into styling some more.

.github/workflows/integrations.yml
feltech: Definitely agree - added https://github.com/OpenAssetIO/OpenAssetIO/issues/1363

feltech:
> > A flexible solution could be to support an OPENASSETIO_DLL_PATH environment variable, which can be set by Windows consumers, then read in the top-level **init**.py of openassetio where it will pass the value to add_dll_directory.
> 
> This appeals more than duplicating files.
> 
> As Python has this security workaround for search path locations only on Windows, it's not unreasonable to expect to do additional setup on Windows.
> 
> Perhaps with a nice error message if it's unset?

Done in 4157393e. As expected, this causes Python 3.7 builds to fail. So a pre-requisite is to remove support for Python 3.7. This has been tacitly OK'd be the community (via ASWF Slack channel)

src/openassetio-python/package/openassetio/_windows.py
```diff
@@ -0,0 +1,81 @@
...
+
+    An ImportWarning might seem more appropriate than the default
+    (UserWarning), but Python will not print ImportWarnings unless the
+    user explicitly enables them.
+    """
+
+    if os.environ.get("OPENASSETIO_DLL_PATH") is not None:
+        # Join path to directory of package if it is relative, else use
+        # verbatim.
+        fullSearchPath = os.environ["OPENASSETIO_DLL_PATH"]
```
foundry-markf, line 53: Noting that the environment variable is fetched twice in two statements.

I'm imagining as public facing code, we ought to show best practices, and not have duplicate effort?
What's been the stance before?

src/openassetio-python/package/openassetio/_windows.py
foundry-markf, line 1: I note use of the older os.path APIs.

Is there a move torwards pathlib planned?

feltech:
> > > A flexible solution could be to support an OPENASSETIO_DLL_PATH environment variable, which can be set by Windows consumers, then read in the top-level **init**.py of openassetio where it will pass the value to add_dll_directory.
> > 
> > 
> > This appeals more than duplicating files.
> > As Python has this security workaround for search path locations only on Windows, it's not unreasonable to expect to do additional setup on Windows.
> > Perhaps with a nice error message if it's unset?
> 
> Done in [4157393](https://github.com/OpenAssetIO/OpenAssetIO/commit/4157393eafe67e46a927f6b47960053cf05a4685). As expected, this causes Python 3.7 builds to fail. So a pre-requisite is to remove support for Python 3.7. This has been tacitly OK'd be the community (via ASWF Slack channel)

Also added c22fa8bb to add a note in the API docs about this new environment variable, so it's a bit more discoverable.

foundry-markf (commented):
It's a question for the future, and a question on performance/duplicate effort/best practices in Open Source software.

So not requesting changes, but maybe.

src/openassetio-python/package/openassetio/_windows.py
feltech: Good point, will update. Indeed being public-facing, we have tried to aim for high quality and best practices (though pragmatism all too often wins out).

src/openassetio-python/package/openassetio/_windows.py
feltech: I started with `pathlib`, but it doesn't support `normpath` (there's `resolve`, but that's a bit more heavy-handed), so it was either use a mixture of `os.path` and `pathlib`, or just use `os.path`. I opted for the consistency of just using one.

feltech: Though I guess we can use `resolve` in the production code. Will switch to `pathlib` in `_windows.py`

foundry-markf: Doesn't that counter your consistency argument?

foundry-markf: (I'm good with consistency)

feltech: Updated to make more use of `pathlib` in 70f28a00

src/openassetio-python/package/openassetio/_windows.py
feltech: Updated as part of 70f28a00

foundry-markf (approved):
Happy with this. The duplicated DLLs on Windows had bothered me, and I do think it's reasonable to have Windows specific workflows to set up search paths, as it is Python specifically on Windows, that has the caveats.

Issue #1202, opened 2023-11-15
Hybrid Plugin System


foundrytom:
## What

Add a plugin system that combines multiple plugin systemss and dispatches based on advertised capability, or priority order if multiple factories advertise a capability.

## ACs

- Uses a shared identifier to locate the conceptual same manager across each plugin system.
- Make sure this is generic so it composes the output of N plugin systems.
- Uses `hasCapability` to route.
- Picks the first if present in both.

## Implementation notes

- Ideally have a internally `RoutingManagerInterface(vector<ManagerInterfacePtr> interfaces)` as this allows eg: Cpp, GPRC Python, in-process Python.

foundrytom:
https://lucascheller.github.io/VFX-UsdAssetResolver/resolvers/production.html
https://forum.aousd.org/t/batching-ar2-queries/818/2?u=foundrytom

feltech:
One potentially controversial change of behaviour worth calling out, required to reduce the amount of work to implement the hybrid plugin system, is to no longer require managers to reject (with an exception) unrecognized settings passed to `initialialize` - see

PR #1356, opened 2024-07-16
Fix terminology `Mapper` initialisation

feltech:
## Description

Discovered whilst working on  #1202. Since the `updateTerminology` API methods were moved to C++, the input terminology dictionary is no longer mutated. Instead, the dictionary is copied (by the Python bindings), and the manager plugin is expected to return a mutated copy.

The `Mapper` class was still assuming the input terminology dict would be mutated. This was never detected in the unit tests, because we were erroneously subclassing `Manager` rather than `ManagerInterface`, and so short-circuiting the Python->C++->Python code path.

So refactor the tests to instead create a mock `ManagerInterface` and catch this problem. Fix the problem by overwriting the `Mapper`'s internal terminology dictionary, rather than expecting it to be mutated.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -42,6 +42,10 @@ v1.0.0-beta.x.x
   multiple shared libraries, allowing external libraries to link.
   [#1340](https://github.com/OpenAssetIO/OpenAssetIO/issues/1340)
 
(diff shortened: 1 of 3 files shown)


SamCrooksFoundry (approved):
Good spot on this one, nice little tidy PR

PR #1357, opened 2024-07-16
Make non-subclassable C++ Python bindings classes `final`

feltech:
Discovered whilst working on #1202.

When a Python script defines a class that inherits from a C++ class, the natural expectation would be that calls to an overridden method in C++ via a C++ base class reference would call out to the Python method implementation.

However, this is not the case unless we have explicitly enabled it in the bindings by providing a so-called "trampoline" class.

So to prevent confusion, we add a tag to the Python bindings to declare the class as "final", so that it is an error at Python import time when a class inherits from that C++ class.

However, there were several classes that we had never added the `is_final` tag to.

So add this tag to appropriate Python bindings.

Also add to `CppPluginSystemManagerImplementationFactory`, despite it defining a trampoline class. This class was never intended to be derived from by Python classes. Hence remove its unnecessary trampoline class.

## Description

Closes # (issue)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,11 +4,19 @@ Release Notes
 v1.0.0-beta.x.x
 ---------------
 
+_This release might not be source compatible for Python hosts. See
+breaking changes, below._
+
 ### Breaking changes
 
 - Removed official support for Python 3.7. The minimum supported Python
   version is now Python 3.9.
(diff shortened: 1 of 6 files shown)


feltech:
Converted to draft since CI will fail until #1356 is merged.

foundry-markf (approved):
I can't really comment on whether these classes should be final, as I'm not that familiar, but you're doing what you said you would, so I'll approve that bit.

PR #1359, opened 2024-07-16
Fix SimpleCppManager override default capabilities

feltech:
Discovered whilst working on #1202. When capabilities were explicitly given as a list in the settings, the pre-existing capability set was not reset, meaning it wasn't possible to "switch off" the default capabilities.

So `clear()` the capability set when settings are detected that should override it, before filling up the set again with those settings.

### DIFF
--- a/examples/manager/SimpleCppManager/tests/test_SimpleCppManager.py
+++ b/examples/manager/SimpleCppManager/tests/test_SimpleCppManager.py
@@ -137,6 +137,18 @@ class Test_SimpleCppManager_initialize:
         with pytest.raises(errors.NotImplementedException):
             a_fresh_simple_cpp_manager.contextFromPersistenceToken("abc")
 
+    def test_when_default_capability_missing_from_explicit_settings_then_capability_unavailable(
+        self, a_fresh_simple_cpp_manager
+    ):
+        settings = a_fresh_simple_cpp_manager.settings()
+        # Same as default, but without "resolution".
+        settings["capabilities"] = (
+            "entityReferenceIdentification,managementPolicyQueries,entityTraitIntrospection"
+        )
+        a_fresh_simple_cpp_manager.initialize(settings)
+
+        assert not a_fresh_simple_cpp_manager.hasCapability(Manager.Capability.kResolution)
+
     def test_when_capability_is_overridden_then_stub_implementations_available(
         self, a_fresh_simple_cpp_manager, a_context
     ):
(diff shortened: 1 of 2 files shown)


examples/manager/SimpleCppManager/tests/test_SimpleCppManager.py
```diff
@@ -101,6 +101,18 @@ def test_when_default_capability_then_default_capability_set_available(
         with pytest.raises(errors.NotImplementedException):
             a_fresh_simple_cpp_manager.contextFromPersistenceToken("abc")
 
+    def test_when_default_capability_missing_from_explicit_settings_then_capability_unavailable(
+        self, a_fresh_simple_cpp_manager
+    ):
+        settings = a_fresh_simple_cpp_manager.settings()
+        # Same as default, but without "resolution".
+        settings["capabilities"] = (
+            "entityReferenceIdentification,managementPolicyQueries,entityTraitIntrospection"
```
SamCrooksFoundry, line 110: [Question] Do these capabilities have to be hard coded?  Can they be dynamically read from the plugin?

SamCrooksFoundry (approved):

examples/manager/SimpleCppManager/tests/test_SimpleCppManager.py
feltech: Not quite sure what you mean by hard coded. Once the plugin is `initialize`d, the capabilities are fixed. You can re-`initialize` with different settings, as is done in this test, to change the supported capabiltiy list. The settings dict passed to `initialize` usually comes from a `.toml` config file.

feltech: Oh you mean for the purposes of the test?

I guess we _could_ 
- loop all capabilities and use `hasCapability` to find all supported capabilities
- use `kCapabilityNames` to get the stringified representation of those capabilities
- pop off an arbitarary capability from the detected supported set
- re-initialize with that reduced set
- assert that the popped capability isn't available.

Seems a hassle though. Really worth it?

SamCrooksFoundry: > You can re-initialize with different settings, as is done in this test, to change the supported capability list
This is the bit I was wondering about.  I hadn't clocked that   `settings["capabilities"]` was already initialized and the test was reinitialising it.  I had assumed that we needed to manually initialise capabilities, which would be a bit of an awkward api.

Thanks for answering the question, no changes required :)

PR #1365, opened 2024-07-19
Remove support for Python 3.7

feltech:
## Description

Part of #1340, but it's been coming for a while.

The trigger for finally doing this is the need for `os.add_dll_directory` support.

So removed Python 3.7 PyPI distributions, and all mention of it from the codebase.

The new minimum supported Python version is now 3.9. The older versions may continue to work for a time, but this is no longer guaranteed nor supported.

### DIFF
--- a/src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
+++ b/src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
@@ -117,9 +117,8 @@ class PythonPluginSystem(object):
         there was a problem loading importlib_metadata.
         """
 
-        # We opt to use the backport implementation of modern importlib to avoid
-        # needing to support 3+ code paths to cover Python 3.7 to 3.10. It is
-        # made available for 3.7 onwards (for now, at least).
+        # We opt to use the backport implementation of modern importlib
+        # since the API changes from 3.9 to 3.10.
         # Pip installs should have this module available, but other methods may not,
         # so be tolerant of it being missing.
         try:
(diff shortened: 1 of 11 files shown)


src/openassetio-python/package/openassetio/pluginSystem/PythonPluginSystem.py
```diff
@@ -117,9 +117,9 @@ def scan_entry_points(self, entryPointName):
         there was a problem loading importlib_metadata.
         """
 
-        # We opt to use the backport implementation of modern importlib to avoid
-        # needing to support 3+ code paths to cover Python 3.7 to 3.10. It is
-        # made available for 3.7 onwards (for now, at least).
+        # We opt to use the backport implementation of modern importlib
+        # since the API changes from 3.9 to 3.10. It is made available
+        # for 3.7 onwards (for now, at least).
```
foundry-markf, line 122: Although I think the 3.7 reference is to do with the importlib backport, is it worth removing the 3.7 just to purge the sourcetree of any reference? As knowing it's available from 3.7 doesn't aid anything here IMO.

feltech: Yeah good point

feltech: Removed in 75abe614

foundry-markf (approved):
LGTM, changes the usual things I would expect to see regarding Python versions.

feltech:
I've gone through the other repos in the OpenAssetIO GitHub org. I don't think we can remove Python 3.7 support from them yet, since they must support a minimum (or pinned) OpenAssetIO version.

If/when we bump the minimum supported OpenAssetIO version for a projects, we should also remember to bump the minimum supported Python version.

Not sure how to ensure we remember to do that. It's not critical, though.

PR #1368, opened 2024-07-22
Add missing shared_ptr alias tests

feltech:
## Description

Back in #918 we started ensuring that classes that have reference semantics (i.e. most objects, because Python) have a set of aliases to their associated `shared_ptr`s. To enforce this we added tests, but have not kept the tests up to date as new classes have been added.

So add the missing classes to the test.

Unfortunately there is no easy way to ensure these tests are kept up to date, other than remembering that they're there (or coming across them again months later).

### DIFF
--- a/src/openassetio-core/tests/typedefsTest.cpp
+++ b/src/openassetio-core/tests/typedefsTest.cpp
@@ -7,6 +7,7 @@
 #include <catch2/catch.hpp>
 
 #include <openassetio/Context.hpp>
+#include <openassetio/hostApi/EntityReferencePager.hpp>
 #include <openassetio/hostApi/HostInterface.hpp>
 #include <openassetio/hostApi/Manager.hpp>
 #include <openassetio/hostApi/ManagerFactory.hpp>
@@ -14,10 +15,15 @@
 #include <openassetio/log/ConsoleLogger.hpp>
 #include <openassetio/log/LoggerInterface.hpp>
 #include <openassetio/log/SeverityFilter.hpp>
+#include <openassetio/managerApi/EntityReferencePagerInterface.hpp>
 #include <openassetio/managerApi/Host.hpp>
 #include <openassetio/managerApi/HostSession.hpp>
 #include <openassetio/managerApi/ManagerInterface.hpp>
 #include <openassetio/managerApi/ManagerStateBase.hpp>
+#include <openassetio/pluginSystem/CppPluginSystem.hpp>
+#include <openassetio/pluginSystem/CppPluginSystemManagerImplementationFactory.hpp>
+#include <openassetio/pluginSystem/CppPluginSystemManagerPlugin.hpp>
+#include <openassetio/pluginSystem/CppPluginSystemPlugin.hpp>
 #include <openassetio/trait/TraitsData.hpp>
 
 namespace {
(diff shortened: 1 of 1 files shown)


SamCrooksFoundry (approved):
[Question]  Is there a reason that you went with a tuple of tuples for storing the type, pointer type and const pointer types, vs using std::add_pointer and std::add_const?

feltech:
> [Question] Is there a reason that you went with a tuple of tuples for storing the type, pointer type and const pointer types, vs using std::add_pointer and std::add_const?

The purpose of the test is to assert that appropriate aliases exist and are consistent with each other. Where could we use `std::add_pointer` et al. in this test?

Perhaps easier to add inline comments in the review, to highlight the context that you're querying. This also has the benefit that it prevents merging until the comment is addressed (resolved), and allows the comment to be closed when it is satisfactorily addressed.

src/openassetio-core/tests/typedefsTest.cpp
```diff
@@ -6,17 +6,23 @@
 #include <catch2/catch.hpp>
```
SamCrooksFoundry, line 7: Added as an inline comment, so we can use a thread, although my question isn't related to a specific line of code in this review.

> > [Question] Is there a reason that you went with a tuple of tuples for storing the type, pointer type and const pointer types, vs using std::add_pointer and std::add_const?
> 
> The purpose of the test is to assert that appropriate aliases exist and are consistent with each other. Where could we use std::add_pointer et al. in this test?

My question is not necessarily referring to a single line in this PR, and I'm happy for the PR to be merged :)

The line that caught my eye is this one from the same file:
```c++
#define CLASS_AND_PTRS(Class) std::tuple<Class, Class##Ptr, Class##ConstPtr>
```

This macro looks unsafe to me.  It would break if you passed the wrong thing into it, eg: 
```c++
std::vector<int> vec;
...
CLASS_AND_PTRS(vec::value_type)
```

But it's not the end of the world as it's just in this test file.

But that got me thinking.  This test is checking that there are aliases for the value, a const value, and a pointer, but the test code shows there's no difference between using the aliases and the return results from using the value type with std::add_pointer/std::add_const.  So my question is why have we gone to the effort of adding the aliases and the test, when their standard library functionality which provides the same.

feltech: > Added as an inline comment, so we can use a thread, although my question isn't related to a specific line of code in this review.

Thanks for that. There is a button to comment on a particular file, rather than having to choose a line in the file.  But yeah, its a bit of a limitation of GitHub code reviews that all `resolve`able review comments must be associated with at least a file.

> the test code shows there's no difference between using the aliases and the return results from using the value type with std::add_pointer/std::add_const. So my question is, why have we gone to the effort of adding the aliases and the test, when there's standard library functionality which provides the same.

Note that the pointer aliases are aliases to `shared_ptr<Class>`, not aliases to `Class*`, so I don't think `std::add_pointer` applies (unless I'm misunderstanding).

Hopefully the background will help illuminate. Littered around the C++ codebase you'll see lines like

* `OPENASSETIO_FWD_DECLARE(managerApi, ManagerInterface)` 
* `OPENASSETIO_DECLARE_PTR(Manager)`
* `OPENASSETIO_ALIAS_PTR(Manager)`

The first two expand to something like

```c++
class Manager;
using ManagerPtr = std::shared_ptr<Manager>;
using ManagerConstPtr = std::shared_ptr<const Manager>;
```

whilst the last macro is used in class declarations and expands to

```c++
using Ptr = ManagerPtr;
using ConstPtr = ManagerConstPtr;
```
that is, in this case, you can do `Manager::Ptr` and that is equivalent to `ManagerPtr`.

The `ManagerPtr` alias is most useful when forward declaring, i.e. in the `.hpp` of consuming classes.

The `Manager::Ptr` alias is most useful for writing generic code, where you can do something like

```c++
template <typename T>
typename T::Ptr castFromPyObject(PyObject* pyObject);
```

The test asserts that for expected types, if we have one alias then we also have the other, and they are equivalent.

To be fair, its a pretty pedantic test, but since these aliases are part of the public interface, testing them seems important.

SamCrooksFoundry (approved):

src/openassetio-core/tests/typedefsTest.cpp
feltech: Resolving after offline discussion and approval.

PR #1369, opened 2024-07-22
Use `std::set` for trait sets

### LINKED ISSUE #1339: Use `std::set` for C++ trait sets
## What

Switch from `std::unordered_set` to (ordered) `std::set` for trait sets in C++

## Why

C++ has set operation [algorithms](https://en.cppreference.com/w/cpp/header/algorithm) (search "Set operations").  However, these only work for _sorted_ containers.  `std::unordered_set` is not a sorted container, and so (ironically) cannot use set operations.

Most of our prototyping has been done in Python, where set operations are a part of the core language, and we have used them extensively to compare and combine trait sets (e.g. [a particularly egregious notebook](https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/blob/main/examples/generic_republish.ipynb)).

The main benefit of `unordered_set` is the better average-case time complexity.  However, for small sets the difference in performance between `std::set` and `std::unordered_set` is [negligible](https://stackoverflow.com/a/55511095/535103) (and may even favour `std::set`).  In practice we haven't seen any large trait sets in the wild - especially when you define "large" as 100k elements, like in the linked analysis!

Clearly this will be a breaking change. However, we do at least use a `TraitSet` alias everywhere, so as long as no downstream project is using `std::unordered_set` directly, or non-generic APIs, then source compatibility will be maintained.

We use the default pybind11 bindings for trait sets, which means the set is duplicated as a Python native `set` - this is the same whether it's `std::unordered_set` or `std::set`.  So the Python core library (and hence downstream Python projects) would be completely unchanged.

feltech:
## Description

Closes #1339.

C++ has set operation algorithms. However, these only work for sorted containers. `std::unordered_set` is not a sorted container, and so (ironically) cannot use set operations.

Most of our prototyping has been done in Python, where set operations are a part of the core language, and we have used them extensively to compare and combine trait sets.

The main benefit of `unordered_set` is the better average-case time complexity. However, for small sets the difference in performance between `std::set` and `std::unordered_set` is negligible (and may even favour `std::set`). In practice, we haven't seen any large 100k+ trait sets in the wild, which is where we'd expect the performance difference to show up (see links in GitHub issue for details).

Clearly this is a breaking change. However, we do at least use a `TraitSet` alias everywhere, so as long as no downstream project is using `std::unordered_set` directly, or non-generic APIs, then source compatibility will be maintained.

We use the default pybind11 bindings for trait sets, which means the set is duplicated as a Python native set. So the Python core library (and hence downstream Python projects) are unaffected.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,8 +4,10 @@ Release Notes
 v1.0.0-beta.x.x
 ---------------
 
-_This release might not be source compatible for Python hosts. See
-breaking changes, below._
+_This release breaks binary compatibility and may break source
+compatibility with C++ hosts and managers. In addition, this release may
+break source compatibility for Python hosts. See breaking changes
+section for more details._
 
 ### Breaking changes
 
@@ -23,6 +25,14 @@ breaking changes, below._
   The `openassetio.test.manager` API Compliance test suite no longer
   asserts that this is the case.
   [#1202](https://github.com/OpenAssetIO/OpenAssetIO/issues/1202)
+
+- Switched from `std::unordered_set` to `std::set` for trait sets (via
+  `TraitSet` alias) in C++. This allows C++ `std` algorithm set
+  operations to work. Python is unaffected. This is mostly a
+  source-compatible change, provided the `TraitSet` alias is used in
+  consuming code, and no `unordered_set`-specific API is used (e.g.
+  `reserve(...)`).
+  [#1339](https://github.com/OpenAssetIO/OpenAssetIO/issues/1339)
 
 ### New Features
 
--- a/doc/contributing/CODING_STANDARDS.md
+++ b/doc/contributing/CODING_STANDARDS.md
@@ -427,7 +427,7 @@ a C++ function call and those mutations persist to the next invocation.
 However, some C++ types are converted to/from native Python types,
 including primitives such as `int` and `float`, but also (by default)
 `std::string` and container types like `std::vector` (converted to
-Python `list`) and `std::unordered_set` (converted to Python `set`).
+Python `list`) and `std::set` (converted to Python `set`).
 
 In these cases, the initial conversion of the C++ default value to a
 Python object (to be stored in the function record) necessitates a copy.
--- a/src/openassetio-core/include/openassetio/trait/collection.hpp
+++ b/src/openassetio-core/include/openassetio/trait/collection.hpp
@@ -5,7 +5,7 @@
  */
 #pragma once
 
-#include <unordered_set>
+#include <set>
 #include <vector>
 
 #include <openassetio/export.h>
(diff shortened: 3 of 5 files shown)


feltech:
As expected, [ABI diff check failed](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/10044611519/job/27760046709?pr=1369):

```
===== Checking ABI of libopenassetio.so.1.0.0 =====

Leaf changes summary: 62 artifacts changed
Changed leaf types summary: 1 leaf type changed
Removed/Changed/Added functions summary: 28 Removed, 5 Changed, 28 Added functions
```

For reference, the "5 Changed" are methods where just the return type was changed. The  "28 Added"/"Removed" are where the method arguments changed.

RELEASE_NOTES.md
```diff
@@ -4,12 +4,23 @@ Release Notes
...
+compatibility with existing codebases, for both hosts and managers. See
+breaking changes section for more details._
+
 ### Breaking changes
 
 - Removed official support for Python 3.7. The minimum supported Python
   version is now Python 3.9.
   [#1365](https://github.com/OpenAssetIO/OpenAssetIO/pull/1365)
 
+- Switched to `std::set` for trait sets (via `TraitSet` alias) in C++.
```
foundry-markf, line 17: Doesn't say what you switched from in the first sentence. Should it?

unordered_set is mentioned further down.

More of a question than a problem.

resources/abi/libopenassetio.so.1.0.0.xml
foundry-markf, line 1: I have no idea how to review this file, as it has thousands of additions and deletions (which is concerning in itself for a text based file).

foundry-markf (approved):
Nothing that would stop me giving an approval

resources/abi/libopenassetio.so.1.0.0.xml
feltech: I guess you just have to go line by line :grin: 

I think we have to view these ABI snapshots as essentially checked-in binary - the file is generated ([explained more here](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/resources/abi/README.md)).  I highlighted the changes before the snapshot was updated in https://github.com/OpenAssetIO/OpenAssetIO/pull/1369#issuecomment-2243408788 .  I think that's all we can reasonably do.

RELEASE_NOTES.md
foundry-markf: The first thing that popped into my head when reading "Switched to 'std::set'" was... switched from what?

So personally I think it's useful.

feltech: OK agreed. Updated in 9db400fc

foundry-markf (approved):

Issue #999, opened 2023-07-03
Migrate `terminology.py` to C++


foundrytom:

elliotcmorris:
Terminology (and all UI releated things) seems a bit far away at the moment. Reassess when we get there as to whether this is still a good pattern to keep.

Issue #418, opened 2022-05-26
C++ logging/audit


foundrytom:
## What

Re-instante logging/audit support for C++

## Why

This was lost when we moved from the Python only implementation

feltech:
Consider https://opentelemetry.io/

PR #1371, opened 2024-07-25
Remove unused _core package

### LINKED ISSUE #1000: Remove `_core` from `openassetio-python`
## What

Remove the python package `_core`.

## Why

`_core` is dead code and no longer used.

feltech:
## Description

Closes #1000. The `_core` package contained the old debug logging and API call auditing implementation. This implementation was based on Python decorators and thus was Python-only. Work to add auditing functionality in C++ will begin with #418.

Almost all code has now been migrated to C++ and so the Python auditing implementation is unused. The exception is the `terminology` helper module, which was still using the old decorators. However, it seems rather unhelpful if this piece of minor functionality is the only thing in OpenAssetIO that is audited. The terminology utilities are also pending migration to C++ in #999, so there's even less reason to keep a Python-only implementation around.

- [ ] ~~I have updated the release notes.~~ - removal of internal package, and incomplete functionality


foundry-markf (approved):
Looks like a straightforward removal

PR #1372, opened 2024-07-25
Add kAuthError error code

### LINKED ISSUE #1009: Auth related `BatchElementError` codes
## What

Add an auth error code so that manager's have a means to inform the user that they do not have access to an asset.  We are thinking a generic auth error code, rather than a set of unauthorised/unauthenticated/permission denied error codes.

## Why

This is a precursor to having something more fleshed out, but just putting the error codes in gives managers and hosts an option in the meantime, and it's super simple.

## Acceptance Criteria

Add a new error code `kAuthError`

feltech:
## Description

Closes #1009. 

Add a `kAuthError` error code to `BatchElementError`s so that manager's have a means to inform the user that they do not (currently) have access to an action or asset.

This then allows host applications to branch differently on authentication or authorization related errors, if they wish.

### DIFF
--- a/src/openassetio-core/include/openassetio/errors/BatchElementError.hpp
+++ b/src/openassetio-core/include/openassetio/errors/BatchElementError.hpp
@@ -126,6 +126,20 @@ class BatchElementError final {
      * the manager.
      */
     kInvalidTraitSet = OPENASSETIO_BatchErrorCode_kInvalidTraitSet,
+
+    /**
+     * Error code indicating that the host is not authorized for a
+     * pariticular request.
+     *
+     * The operation may be valid if attempted again, once the
+     * authorization issue is resolved.
+     *
+     * This can be due to insufficient or incomplete authentication as
+     * well as authorization, and could be specific to a particular
+     * action, or a particular entity. The error message should provide
+     * details of the reason.
+     */
+    kAuthError = OPENASSETIO_BatchErrorCode_kAuthError,
   };
 
   /**
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -47,6 +47,10 @@ section for more details._
   `openassetio.dll`, if required.
   [#1340](https://github.com/OpenAssetIO/OpenAssetIO/issues/1340)
 
+- Added `kAuthError` to the set of possible `BatchElementError` error
+  codes.
+  [#1009](https://github.com/OpenAssetIO/OpenAssetIO/issues/1009)
+
 ### Improvements
 
 - Added singular overload of `managementPolicy` for convenience.
--- a/src/openassetio-core/include/openassetio/errors/errorCodes.h
+++ b/src/openassetio-core/include/openassetio/errors/errorCodes.h
@@ -47,3 +47,6 @@
 
 /// Failure due to a TraitSet being unknown to the manager
 #define OPENASSETIO_BatchErrorCode_kInvalidTraitSet (OPENASSETIO_BatchErrorCode_BEGIN + 6)
+
+/// Failure due to a authentication or authorization.
+#define OPENASSETIO_BatchErrorCode_kAuthError (OPENASSETIO_BatchErrorCode_BEGIN + 7)
--- a/src/openassetio-core/src/errors/exceptionMessages.cpp
+++ b/src/openassetio-core/src/errors/exceptionMessages.cpp
@@ -27,6 +27,8 @@ Str errorCodeName(BatchElementError::ErrorCode code) {
       return "invalidPreflightHint";
     case BatchElementError::ErrorCode::kInvalidTraitSet:
       return "invalidTraitSet";
+    case BatchElementError::ErrorCode::kAuthError:
+      return "authError";
   }
 
   assert(false);  // Impossible case.
--- a/src/openassetio-python/cmodule/src/errors/BatchElementErrorBinding.cpp
+++ b/src/openassetio-python/cmodule/src/errors/BatchElementErrorBinding.cpp
@@ -25,7 +25,8 @@ void registerBatchElementError(const py::module& mod) {
       .value("kEntityAccessError", BatchElementError::ErrorCode::kEntityAccessError)
       .value("kEntityResolutionError", BatchElementError::ErrorCode::kEntityResolutionError)
       .value("kInvalidPreflightHint", BatchElementError::ErrorCode::kInvalidPreflightHint)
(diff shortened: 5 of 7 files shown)


src/openassetio-core/include/openassetio/errors/BatchElementError.hpp
```diff
@@ -126,6 +126,20 @@ class BatchElementError final {
...
+
+    /**
+     * Error code indicating that the host is not authorized for a
+     * pariticular request.
+     *
+     * The operation may be valid if attempted again, once the
+     * authorization issue is resolved.
+     *
+     * This can be due to insufficient or incomplete authentication as
+     * well as authorization, and could be specific to a pariticular
```
foundry-markf, line 138: Typo: pariticular -> particular

foundry-markf (changes requested):

feltech:
Frustratingly, the libabigail ABI diff check [abidif](https://sourceware.org/libabigail/manual/abidiff.html), with our [default switches](https://github.com/OpenAssetIO/OpenAssetIO/blob/788f46646cd807ca64a12ff34014c9dfd420c28a/cmake/Testing.cmake#L93) (`--leaf-changes-only --no-default-suppression`), gives rather unhelpful information
```
Leaf changes summary: 1 artifact changed (1 filtered out)
Changed leaf types summary: 1 (1 filtered out) leaf types changed
Removed/Changed/Added functions summary: 0 Removed, 0 Changed, 0 Added function

'class openassetio::v1::pluginSystem::CppPluginSystemManagerImplementationFactory at CppPluginSystemManagerImplementationFactory.hpp:48:1' changed:
  type size hasn't changed
  there are member function changes:
```
(EOF)

... but `CppPluginSystemManagerImplementationFactory` has nothing to do with `BatchElementError::ErrorCode` enums.

If we change the switches to maximize output (`--harmless --redundant --no-default-suppression`), then buried under the mass of text we do eventually see a more expected ABI change, deep in the hierarchy
```
...
              type of 'std::function<void(long unsigned int, openassetio::v1::errors::BatchElementError)>::_Invoker_type _M_invoker' changed:
                underlying type 'void (const std::_Any_data&, unsigned long int&&, openassetio::v1::errors::BatchElementError&&)*' changed:
                  in pointed to type 'function type void (const std::_Any_data&, unsigned long int&&, openassetio::v1::errors::BatchElementError&&)':
                    parameter 3 of type 'openassetio::v1::errors::BatchElementError&&' has sub-type changes:
                      referenced type 'class openassetio::v1::errors::BatchElementError' changed at BatchElementError.hpp:49:1, as reported earlier
...
```

It seems that `--leaf-changes-only` kinda does the opposite of what we want, since its just reporting the outermost affected type, rather than the root cause. I can't find a good command-line option for `abitdiff` to get a nicer summary of the root cause.

The best I've found for this particular case is `--harmless --leaf-changes-only`, which at least reports that an enum was changed

```
Leaf changes summary: 2 artifacts changed
Changed leaf types summary: 2 leaf types changed
Removed/Changed/Added functions summary: 0 Removed, 0 Changed, 0 Added function

'enum openassetio::v1::errors::BatchElementError::ErrorCode at BatchElementError.hpp:52:1' changed:
  type size hasn't changed
  1 enumerator insertion:
    'openassetio::v1::errors::BatchElementError::ErrorCode::kAuthError' value '135'
'class openassetio::v1::pluginSystem::CppPluginSystemManagerImplementationFactory at CppPluginSystemManagerImplementationFactory.hpp:48:1' changed:
  type size hasn't changed
  there are member function changes:
```

Perhaps more experimentation with `--non-reachable-types` could yield a "flatter" diff. But that would need more time scheduled to investigate.

In the meantime, at least its flagging _something_ has broken, prompting deeper investigation and an update of the snapshot.

src/openassetio-core/include/openassetio/errors/BatchElementError.hpp
feltech: Good spot. Fixed and (force-)pushed.

resources/abi/libopenassetio.so.1.0.0.xml
foundry-markf, line 1: Just commenting that this change to the XML file is dead easy to read! Not thousands of additions and deletions. So I wonder what was different in the other PR.

foundry-markf (approved):
Looks ok.

David and I discussed the ABI comment offline, with a view to marking these in future as 'information only' or 'must get fixed'.

resources/abi/libopenassetio.so.1.0.0.xml
foundry-markf: I can see in the GitHub actions an 'ABI diff check' - going by the sound of that, I'm wondering why it didn't pick up +/- 1000s of changes when it was updated in the other PR?

The reason I was looking just now, was thinking that some aspect of ABI snapshot check/diff/update should be part of every PR, to either confirm 'nothing has broken' or 'you need to fix something'.

feltech: > I can see in the GitHub actions an 'ABI diff check' - going by the sound of that, I'm wondering why it didn't pick up +/- 1000s of changes when it was updated in the other PR? Or even prior to that.

I suspect the changes were probably harmless changes that built up over time and/or the snapshot is extremely sensitive, such that small ABI changes cascade into a huge snapshot change.

> The reason I was looking just now, was thinking that some aspect of ABI snapshot check/diff/update should be part of every PR, to either confirm 'nothing has broken' or 'you need to fix something'.

In the ideal case, the `abdiff` CI job failing is all we need to tell us that we either

* Need to update the release notes to call out a breaking change.
* Need to rethink the feature, to avoid the breaking change.

In practice, `abidiff` isn't a silver bullet, it's just an extra guardrail to try to catch unwitting ABI breaks.

I'd hoped that the output of the CI job, when there is a failure, would be useful in understanding the nature of the break. But it looks like the output obfuscates the root cause of an ABI break, either by not reporting the root cause at all, or by burying it in a deluge of output.

I'm tempted to add the `--harmless` flag to the test, to force more frequent updates of the snapshot, give more useful output when there is a break, and call out more potential breaks (even if `--harmless` changes are technically non-breaking from a pure ABI perspective, they may be breaking for other reasons).

feltech: Added https://github.com/OpenAssetIO/OpenAssetIO/issues/1373

foundry-markf (approved):

Issue #1367, opened 2024-07-22
Identify necessary breaking changes for UI Delegation


SamCrooksFoundry:
## What

This is a spike to identify any breaking changes required in the OpenAssetIO library to facilitate UI Delegation.

## Why

To add placeholders so that development can proceed over the next year without breaking existing integrations.

## Acceptance Criteria

A rough outline of how UI delegation will work. 

An outline of what changes need to be made to the API to support UI delegation, especially breaking changes.

feltech:
Added the following PRs related to this:

* https://github.com/OpenAssetIO/OpenAssetIO/pull/1382
* https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/pull/100

feltech:
Based on feedback on the aforementioned PRs, especially the decision record, it seems fairly clear that breaking changes will not be needed. 

The DR concludes that the UI delegation plugin system can be entirely distinct from the main manager plugin system.

The primary reason for breaking changes would then be in changing the public API to give explicit access to the private implementation of the manager plugin from the UI delegate plugin.  

However, this is unlikely to be necessary.  If only public (i.e. `Manager`) access is needed, then no API changes are necessary. Otherwise, the ideal solution is for plugin authors to make use of the `Context.managerState` mechanism, which again means no public API changes.

If we want to add to the API to provide the UI delegate plugin with access to the underlying `ManagerInterface` implementation of the manager plugin, then the obvious solution is to provide a new method on the public `Manager` middleware class, which returns a reference to the associated `ManagerInterface`. Providing an (accessor to a pre-existing member wouldn't change the binary layout of anything, and it wouldn't involve any `virtual` methods. So from a C++ ABI perspective it would be a patch-level update (in semver speak), along with the rest of the UI delegation functionality (in theory).

So I think it's safe to close this issue, based on current information.

PR #1382, opened 2024-08-05
DR for UI delegation plugin architecture

feltech:
## Description

Part of #1367. In order to determine the necessary changes to the existing API in order to support UI delegation, we must first decide how UI delegation implementations should be loaded into an application.

So add a decision record that theorises an implementation based on prior art in Katana and FnAssetAPI, and contrasts this with the naive implementation that extends the existing plugin scaffolding.

## Reviewer Notes

[Rendered output](https://github.com/OpenAssetIO/OpenAssetIO/blob/9e84d0ade9921f5967f85e37bde83cd331edc266/doc/decisions/DR025-UI-delegation-plugin-interface.md)

Keep an eye out for, e.g. 
- Things not explained well that need rephrasing or additional information
- Pros and cons that may have been missed
- Entirely different approaches that should be considered

### DIFF
--- a/doc/decisions/DR025-UI-delegation-plugin-interface.md
+++ b/doc/decisions/DR025-UI-delegation-plugin-interface.md
@@ -1,0 +1,236 @@
+# DR025 UI delegation plugin interface
+
+- **Status:** Proposed
+- **Impact:** High
+- **Driver:** @feltech
+- **Approver:** The OpenAssetIO Community
+- **Due date:** 2024-08-12
+- **Outcome:** Manager plugins and UI delegate plugins will be separate
+  plugins with their own plugin systems. The UI delegate plugin will be
+  provided a reference to the manager implementation.
+
+## Background
+
+OpenAssetIO is an abstraction that aims to be minimally opinionated.
+Asset-managed pipelines don’t conform to a single structure — each
+facility will vary in its workflow processes and backend systems, often
+significantly. The OpenAssetIO core library provides an API for common
+operations that a host application may wish to perform, with no
+assumptions made about how that data is referenced or structured on the
+backend.
+
+However, some user-facing operations require knowledge of how a
+particular pipeline works so that the appropriate interface can be shown
+to the user.
+
+For example, using a dedicated asset-centric browser UI, rather than
+browsing for files on a file system, could provide a richer and more
+relevant interface for artists, by filtering for approved assets that
+are relevant for their current task, and presenting useful metadata.
+
+A third-party vendor attempting to implement this in a generic way that
+works across all potential pipeline setups is a daunting task, and it is
+very unlikely to warrant expending engineering effort to do optimally
+when it is far from the primary purpose of the software.
+
+The engineers best placed to implement such an interface are the
+maintainers of the asset management system itself. However, typically
+this means separate secondary applications or, at best, surface-level
+patching of application UIs, without deep integration into / replacement
+of the existing UI layout.
+
+If a third-party vendor can provide UI hooks into their software, such
+that appropriate UI functionality can be _delegated_ to a plugin, then
+we can have the best of both worlds — a native level integration of
+bespoke interfaces that are optimized for the user's context.
+
+### Aims
+
+The aim of this decision is not to fully design the architecture for UI
+delegation in OpenAssetIO. The aim is to decide on broad architecture,
+with a focus on the fundamentals that must be fixed in place to enable
+future evolution without requiring significant changes by host and
+manager maintainers. That is, the primary focus is on the plugin
+architecture and the (breaking) changes required to support it.
+
+### Prior art
+
+The precursors to OpenAssetIO provide some prior art in the area of UI
+delegation.
+
+#### Katana
+
+[Katana](https://www.foundry.com/products/katana) is a look development
+and lighting tool by Foundry, and its AssetAPI was the original
+inspiration for OpenAssetIO. The AssetAPI and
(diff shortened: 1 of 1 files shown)


doc/decisions/DR025-UI-delegation-plugin-interface.md
```diff
@@ -0,0 +1,184 @@
...
+are relevant for their current task, and presenting useful metadata.
+
+A third-party vendor attempting to implement this in a generic way that
+works across all potential pipeline setups is a daunting task, and it is
+very unlikely to warrant expending engineering effort to do optimally
+when it is far from the primary purpose of the software.
+
+The engineers best placed to implement such an interface are the
+maintainers of the asset management system itself. Indeed, this is
+commonplace. However, typically this means separate secondary
```
foundry-markf, line 37: > Indeed, this is commonplace.

For the intended audience, is this a known fact?

doc/decisions/DR025-UI-delegation-plugin-interface.md
```diff
@@ -0,0 +1,184 @@
...
+
+A single logical plugin using the existing plugin system, with
+additional method(s) added to the existing `ManagerInterface` class.
+I.e. a `getUIDelegate` (or similar) method returns a UI delegate object,
+which is responsible for creating/modifying UI elements.
+
+#### Pros
+
+- Choosing a different language (C++/Python) for the UI vs. core methods
+  is possible using the Hybrid Plugin System with the addition of a new
```
foundry-markf, line 167: Is the hybrid plugin system argument enough to prefer this option, as a '1st class feature' of OpenAssetIO?

Or would you rather let the hybrid system settle in a bit?

foundry-markf (commented):
Sounded very much like what you presented earlier.

doc/decisions/DR025-UI-delegation-plugin-interface.md
feltech: It allows mixing and matching Python and C++, but is a bit more involved and tightly coupled compared to the other solution of having an entirely different plugin.  I guess I should expand on this.

doc/decisions/DR025-UI-delegation-plugin-interface.md
feltech: Honestly not sure :see_no_evil: .  Every manager integration I've seen has at least something, e.g. a menu item.

foundry-markf: It sounds like a well known fact.

If it's based on just your observation, rather than something you can reference, perhaps it could be rephrased as something like

"Based on experiences with X, Y, Z, this is commonplace."

doc/decisions/DR025-UI-delegation-plugin-interface.md
foundry-markf: I was just thinking about it, because I imagine you're going to shout loudly about the Hybrid Plugin system to users, and then if it's mentioned here, but ultimately not used, that would raise an eyebrow for me. I get that it might not be the best fit, but the expansion you mention will probably cover that to make it plain for other readers.

doc/decisions/DR025-UI-delegation-plugin-interface.md
feltech: Just dodged this by removing the statement as part of 071bb98e :grin:

doc/decisions/DR025-UI-delegation-plugin-interface.md
feltech: Expanding a bit more on the drawbacks of the hybrid plugin system for this particular use case as part of 071bb98e

foundry-markf (approved):
You've resolved all my queries, so LTGM

doc/decisions/DR025-UI-delegation-plugin-interface.md
```diff
@@ -0,0 +1,194 @@
...
+2021 when it formed the basis of OpenAssetIO.
+
+FnAssetAPI was entirely Python-based for most of its life. C++
+functionality was begun but not proven out, and the approach was not
+adopted for OpenAssetIO's C++ architecture.
+
+The UI delegation layer of FnAssetAPI took a slightly different approach
+to Katana. In FnAssetAPI, a single plugin system was used to load both
+the `ManagerInterfaceBase`-derived core interface and a
+`ManagerUIDelegate`-derived UI delegate. That is, a single plugin
```
tomc-cinesite, line 100: Fwiw, iirc, this was purely done for convenience - as it was possible, in Python, not to have a UI framework as a dependency unless it was needed. The plan for C++ was that they'd be separate due to the noted con that its difficult to manage c++ dependencies at runtime.

doc/decisions/DR025-UI-delegation-plugin-interface.md
```diff
@@ -0,0 +1,194 @@
...
+- Relying on the hybrid plugin system to provide multi-language support
+  adds unnecessary boilerplate to the implementation of the plugin when
+  its functionality is higher level than (and more akin to a consumer
+  of) the core API.
+- Relying on the hybrid plugin system means that support for UI
+  delegation must be added to the C++ API first. So it is not possible
+  to add Python support first and C++ support later.
+- Definitely requires breaking changes (though only additions so should
+  be source compatible).
+- Tight coupling of UI and headless concerns means care must be taken,
```
tomc-cinesite, line 187: For me, this would be a deal breaker, its critical to have a UI dependency free resolver.

SamCrooksFoundry (approved):

tomc-cinesite (approved):

PR #1383, opened 2024-08-05
Manager doc update

brycegbrazen:
Remove temp doc message and link to OpenAssetIO Manager Python and Plugin for Ayon

## Description

### DIFF
--- a/doc/doxygen/src/Examples.dox
+++ b/doc/doxygen/src/Examples.dox
@@ -73,7 +73,12 @@
  * that some example @ref asset_management_system has a plugin,
  * installed on @ref plugin_path_var).
  *
- * We will be providing an example manager implementation soon!
+ * Check out <a style="font-weight:bold" href="https://github.com/OpenAssetIO/Template-OpenAssetIO-Manager-Python"
+ * target="_blank">Python Manager Template</a> for an example
+ * minimalist plugin.
+ *
+ * Also, for a "real world" manager plugin, check out the <a style="font-weight:bold" href="https://github.com/ynput/ayon-openassetio-manager-plugin"
+ * target="_blank">Plugin for Ayon</a>.
  *
  * @code{.py}
  * availableManagers = managerFactory.availableManagers()


doc/doxygen/src/Examples.dox
```diff
@@ -64,7 +64,12 @@
  * that some example @ref asset_management_system has a plugin,
  * installed on @ref plugin_path_var).
  *
- * We will be providing an example manager implementation soon!
+ * Check out <a href="https://github.com/OpenAssetIO/Template-OpenAssetIO-Manager-Python"
```
brycegbrazen, line 67: Not sure if this is the best way to reference these links. Also, they are not bold right now so they are hard to see.

feltech: I think that's fine. These are both linked on the home page too ("An Introduction").

feltech, line 67: [fix] I'd be happy just keeping the removal and no need to link the Template or Ayon.

If you want to keep these links, can we change `OpenAssetIO Manager Python` to `Python Manager Template`?

feltech (changes requested):

doc/doxygen/src/Examples.dox
brycegbrazen: I'm personally partial to adding additional cross-references in documentation so people can more easily find things. Let me know if you disagree though!

brycegbrazen: Fixed the styling so they are bold 👍

feltech (approved):

feltech:
Fine for this case, but in general we'd avoid commits that just fix issues with previous commits in the same PR, e.g. in this case 0d2cfb5a5dbee9861adc29410224e468f0f2217e. 

For a small PR like this you could just edit the commit and force push. 

But more generally, and for future reference, we make use of git [autosquash](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash),  In this particular example the commit title would be `fixup! [Docs] Improve manager documentation` or `squash! [Docs] Improve manager documentation`. You can then push that commit, and put the commit hash in the PR comment thread, providing a handy link for the reviewer to see how the comment was addressed. When the PR is ready to merge you can then `git rebase -i --autosquash` to get rid of the `fixup!`s, then force-push.  See also [contributing docs](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/contributing/PULL_REQUESTS.md#our-pull-request-process).

feltech:
I've taken the liberty of squashing and rebasing.

PR #1387, opened 2024-08-06
Use enum for BatchElementError codes

### LINKED ISSUE #1060: Consistent C/C++ shared enumerations
## What

Use C-style namespaced enums instead of macros in shared C/C++ headers.

## Why

In #1056 a new enumeration was added for `Access` categories, and left as a C-style but C++-namespaced `enum`.

In #557 we added macros as a way to share enum values between C and C++, when we could just have easily used enums.

We should be consistent regardless, and probably just use `enum` for everything as a cross-language common mechanism.

## Acceptance Criteria

Remove batch element error code macros
Add enum to the internal cpp namespace

## Note

Duplicate values for C

feltech:
## Description

Closes #1060. Add `BatchElementError` error codes as a C-style `enum` to the `internal::errors` namespace, and use these instead of macros in the C++ API. Hence remove the macros.

The (woefully incomplete and due for a rethink) C API only used one of the macro values, and it was trivial, so just (effectively) duplicate the value.


foundry-markf (approved):
Looks ok to me

Issue #1353, opened 2024-07-12
Default libstdc++ ABI to match system


feltech:
## What

Do not force the libstdc++ ABI away from the system defaults, but leave the option available.

## Why

When the OpenAssetIO build system and CI was first set up, VFX Reference Platform CY22 was the latest.  It was still CentOS 7 based and specified that the deprecated libstdc++ ABI should be used. We decided that CY22 would be our default target. 

However, the closest we could get on GitHub CI was Ubuntu 20.04, which matches the CY22 GCC version (9), but defaults to the non-deprecated libstdc++ ABI.  And so, we decided to configure CMake to force the deprecated ABI, unless explicitly toggled.

However, this is not a valid default for CY23+, which specify that the non-deprecated ABI should be used.

The closest we can get to "automatically" determining whether the deprecated ABI should be used or not, is to use whatever is the system default ABI.  So if someone is building on CentOS 7, it'll be the deprecated ABI, and if someone is building on Rocky 8, it'll be the non-deprecated ABI.

With our CI on Ubuntu 20.04, such a scheme will cause us to diverge more from CY22. We _could_ explicitly use the deprecated ABI, but probably a better option would be to upgrade to Ubuntu 22.04, which matches CY23+ and will be done in a separate issue (#812).

## Acceptance Criteria

Don't set `_GLIBCXX_USE_CXX11_ABI ` by default

feltech:
Tried out the changes on my fork (https://github.com/feltech/OpenAssetIO/pull/19) - it's mostly fine, but TraitGen integration tests fail, because its CMake hardcodes use of the deprecated ABI. That CI job then stops before testing other projects, but Test-CMake may have the same issue.

One possibility would be to augment the OpenAssetIO CMake config file with a variable specifying whether the deprecated libstdc++ ABI is expected or not, so that consumers can read and branch on it.

Issue #1351, opened 2024-07-12
Improve support for more modern VFX Reference Platforms


feltech:
## What

Use a more recent VFX Reference Platform (2023+) as the default distribution target for OpenAssetIO, whilst making a best-effort to retain support for older (in-use) target platforms.

## Why

All our CMake, CI and PyPI defaults assume VFX Reference Platform 2022 is the predominant standard. The world has moved on and now the defaults make less sense to average consumers.

For example

* Forcing the deprecated `libstdc++` ABI on Linux means the default build is incorrect for anyone building OpenAssetIO for VFX CY23+.
* Our use of the [manylinux2014 container](https://github.com/pypa/manylinux?tab=readme-ov-file#manylinux2014-centos-7-based-glibc-217) for [cibuildwheel](https://github.com/pypa/cibuildwheel), which is CentOS 7 based, means that, by default, a `pip install openassetio` from PyPI is only compatible with C++ manager plugins built against the deprecated `libstdc++` ABI, i.e. is incompatible with VFX CY23+, even for Python 3.10+.

### Tasks
- [x] https://github.com/OpenAssetIO/OpenAssetIO/issues/1353
- [x] https://github.com/OpenAssetIO/OpenAssetIO/issues/1352
- [x] https://github.com/OpenAssetIO/OpenAssetIO/issues/984
- [x] https://github.com/OpenAssetIO/OpenAssetIO/issues/812
- [x] https://github.com/OpenAssetIO/OpenAssetIO/issues/1354
- [x] https://github.com/OpenAssetIO/OpenAssetIO/issues/1355
- [x] https://github.com/OpenAssetIO/OpenAssetIO/issues/1389

feltech:
Closing as all sub-tasks are now done.

PR #1388, opened 2024-08-07
Default libstdc++ ABI based on Python version

### LINKED ISSUE #1352: Use non-deprecated libstdc++ ABI in Python wheels

feltech:
## Description

Closes #1352. When we detect that we're likely building using GCC for VFX Reference Platform CY23 or above, we should use the libstdc++ C++11 ABI by default, otherwise we use the deprecated ABI.

## Reviewer notes

This is part of a broader epic https://github.com/OpenAssetIO/OpenAssetIO/issues/1351 and subsequent tasks in that epic may mean this solution needs to be refactored, in particular https://github.com/OpenAssetIO/OpenAssetIO/issues/1353

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -18,7 +18,7 @@ concurrency:
 
 jobs:
   build_wheels:
-    name: Build wheels on ${{ matrix.config.os }} - ${{matrix.python-build}}
+    name: Build wheels on ${{ matrix.config.os }} for Python ${{matrix.python_version}}
     runs-on: ${{ matrix.config.os }}
     strategy:
       fail-fast: false
@@ -31,14 +31,19 @@ jobs:
           - os: windows-2019
             preamble: call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat" x64
             shell: cmd
-          # These are the platform build strings provided to
-          # cibuildwheel, with wildcarding. See
-          # https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip
-        python-build: ['cp39*64', 'cp310*64', 'cp311*64']
+        # We matrix the python build here ourselves, rather than
+        # letting cibuildwheel do its regular python matrix, as we
+        # want to allow a full release pipeline to run "vertically" on
+        # CI, on individual agents, rather than building every python
+        # version sequentially on each platform. This is done less for
+        # performance reasons (although it does help) and more so a
+        # single failing python version won't interrupt every other
+        # deploy on that platform. It also helps by ensuring that the
+        # `bootstrap_platform` workflow uses the appropriate Python
+        # version.
+        python_version: ["3.9", "3.10", "3.11"]
     defaults:
       run:
-        # Annoyingly required here since `matrix` isn't available in the
-        # `shell` property of individual steps.
         shell: ${{ matrix.config.shell }}
 
     steps:
@@ -46,6 +51,14 @@ jobs:
 
       - name: Bootstrap
         uses: ./.github/bootstrap_platform
+
+      - name: Python version to CIBW version format
+        # These are the platform build strings provided to
+        # cibuildwheel, with wildcarding. E.g. "cp311*64". See
+        # https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip
+        run: >
+          echo CIBW_BUILD="cp"$(printf ${{ matrix.python_version }} | sed 's/\.//')"*64" >> $GITHUB_ENV
+        shell: bash
 
       - name: Build wheels
         run: |
@@ -58,17 +71,19 @@ jobs:
           # different, and are containerised fully, therefore they get
           # their toolchain path from pyproject.toml
           CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/.conan/conan_paths.cmake
-          # We matrix the python build here ourselves, rather than
-          # letting cibuildwheel do its regular python matrix, as we
-          # want to allow a full release pipeline to run "vertically" on
-          # CI, on individual agents, rather than building every python
-          # version sequentially on each platform. This is done less for
-          # performance reasons (although it does help) and more so a
-          # single failing python version won't interrupt every other
-          # deploy on that platform.
-          CIBW_BUILD: ${{ matrix.python-build }}
           CIBW_SKIP: '*musllinux* *arm64*'
-          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
+          # Python 3.9 implies VFX Reference Platform CY22, which
+          # specifies glibc 2.17 and the deprecated libstdc++ ABI, and
+          # is typically associated with CentOS 7. The manylinux2014
+          # container is CentOS 7 based so is a great fit in this case.
+          # However, for CY23+, i.e. Python 3.10+, glibc is 2.28
+          # and the non-deprecated libstdc++ ABI should be used. In this
+          # case, manylinux_2_28 must be used to support the
+          # non-deprecated libstdc++ ABI, i.e. so that compiler macro
+          # `-D_GLIBCXX_USE_CXX11_ABI=1` will work. See
+          # https://github.com/pypa/manylinux and
+          # https://vfxplatform.com
+          CIBW_MANYLINUX_X86_64_IMAGE: ${{ matrix.config.python_version == '3.9' && 'manylinux2014' || 'manylinux_2_28' }}
           PIP_VERBOSE: 1
           # Required as we make use of c++17 features
           MACOSX_DEPLOYMENT_TARGET: 10.15
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -199,7 +199,7 @@ jobs:
           OPENASSETIO_DLL_PATH: ${{ github.workspace }}/dist/bin
 
   cpp-manager-plugin:
-    name: C++ manager plugin on ${{ matrix.config.os }}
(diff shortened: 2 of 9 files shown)


feltech:
(Deliberately) failing CI job: https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/10283936254/job/28458899280?pr=1388

```
      debug: CppPluginSystem: Failed to open library '/home/runner/work/OpenAssetIO/OpenAssetIO/dist/plugins/SimpleCppManager.so': /home/runner/work/OpenAssetIO/OpenAssetIO/dist/plugins/SimpleCppManager.so: undefined symbol: _ZN11openassetio2v110managerApi16ManagerInterface4infoB5cxx11Ev
```

.github/workflows/build-wheels.yml
```diff
@@ -31,14 +31,19 @@ jobs:
...
-          # These are the platform build strings provided to
-          # cibuildwheel, with wildcarding. See
-          # https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip
-        python-build: ['cp39*64', 'cp310*64', 'cp311*64']
+        # We matrix the python build here ourselves, rather than
+        # letting cibuildwheel do its regular python matrix, as we
+        # want to allow a full release pipeline to run "vertically" on
+        # CI, on individual agents, rather than building every python
+        # version sequentially on each platform. This is done less for
+        # performance reasons (although it does help) and more so a
```
foundry-markf, line 39: [suggestion] 'and more so a' change to 'but more so a'

.github/workflows/build-wheels.yml
```diff
@@ -58,17 +71,19 @@ jobs:
...
-          # CI, on individual agents, rather than building every python
-          # version sequentially on each platform. This is done less for
-          # performance reasons (although it does help) and more so a
-          # single failing python version won't interrupt every other
-          # deploy on that platform.
-          CIBW_BUILD: ${{ matrix.python-build }}
           CIBW_SKIP: '*musllinux* *arm64*'
-          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
+          # Python 3.9 implies VFX Reference Platform CY22, which
+          # specifies glibc 2.17 and the deprecated libstdc++ ABI, and
```
foundry-markf, line 76: [question] I've not heard it called 'deprecated' before. Have you got a reference?

https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html does not mention deprecated.

I've heard 'new' and 'old' ABIs referred to.
It's not even you can say C++11 onward, as we used C++11 in the old ABI.

.github/workflows/build-wheels.yml
```diff
@@ -58,17 +71,19 @@ jobs:
...
+          # is typically associated with CentOS 7. The manylinux2014
```
foundry-markf, line 77: CentOS 7 and older. e.g. Ubuntu was once using that ABI too.

What I'm calling out is that it's not specific to CentOS version 7.

.github/workflows/build-wheels.yml
```diff
@@ -58,17 +71,19 @@ jobs:
...
+          # container is CentOS 7 based so is a great fit in this case.
+          # However, for CY23+, i.e. Python 3.10+, glibc is 2.28
+          # and the non-deprecated libstdc++ ABI should be used. In this
```
foundry-markf, line 80: Another use of deprecated, in the negated sense.

.github/workflows/integrations.yml
```diff
@@ -208,12 +208,19 @@ jobs:
...
             site-packages: Lib/site-packages
+            python_version: "3.9"
             shell: cmd
           - os: ubuntu-20.04
             site-packages: lib/python3.9/site-packages
+            python_version: "3.9"
+            shell: bash
+          - os: ubuntu-20.04
+            site-packages: lib/python3.10/site-packages
+            python_version: "3.10"  # CY23
```
foundry-markf, line 219: Q: do you support CY24, Python 3.11?

Asking as it's relevant to Nuke 16.

.github/workflows/integrations.yml
```diff
@@ -250,7 +257,7 @@ jobs:
 
           cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo
           -DOPENASSETIO_ENABLE_TESTS=ON
-          -DOPENASSETIO_GLIBCXX_USE_CXX11_ABI=OFF
+          -DOPENASSETIO_GLIBCXX_USE_CXX11_ABI=${{ matrix.config.python_version == '3.9' && 'OFF' || 'ON' }}
```
foundry-markf, line 260: I don't understand this syntax.

How does
`matrix.config.python_version == '3.9' && 'OFF'` evaluate?

RELEASE_NOTES.md
```diff
@@ -66,6 +66,13 @@ section for more details._
 
 ### Improvements
 
+- Updated the default libstdc++ ABI (i.e. CMake option
+  `OPENASSETIO_GLIBCXX_USE_CXX11_ABI`) to be based on the detected
+  Python version (if Python is enabled) such that it is consistent with
+  the VFX Reference Platform. This means Python wheels for Python 3.10+
+  will now use the non-deprecated ABI.
```
foundry-markf, line 73: Another use of deprecated.

foundry-markf (changes requested):
In our internal builds, we simply choose the appropriate compiler for what we expect the std::string ABI to be.

i.e. if we build for vfx22, then it's the old ABI.
if we build for vfx23, then it's the new ABI.

We don't fiddle with anything, nor set any extra CMake variables. Just take the default in the compiler on the distro version it's default state.

I may be missing a detail on GitHub actions, but it feels like it should be as simple as 'pick the right distro version and compiler version for each refspec'.

.github/workflows/build-wheels.yml
feltech: That's just me, it feels wrong to say "new ABI" now that it's 9 years since it was introduced and RHEL/CentOS 7 is EOL :grimacing: To the best of my knowledge, the "old ABI"  its no longer the default anywhere.

If its going to raise eyebrows using the word "deprecated", then I can change the language.

.github/workflows/build-wheels.yml
feltech: Yeah I was careful to say "typically associated with" since that's predominant in VFX - it feeds into the next point about the manylinux2014 container being based on CentOS 7.

.github/workflows/integrations.yml
feltech: Yeah we build wheels for 3.9, 3.10 and 3.11 and publish them to PyPI (`build-wheels.yml`, `deploy-pypi.yml`)

.github/workflows/integrations.yml
feltech: It's actually [documented by GitHub as a "ternary operator"](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/expressions#example)

This is fairly idiomatic ternary syntax in Javascript, in my experience, and also has an equivalent in Python
 ```
>>> True and "its truthy" or "its falsey"
'its truthy'
>>> False and "its truthy" or "its falsey"
'its falsey'
```

feltech:
> In our internal builds, we simply choose the appropriate compiler for what we expect the std::string ABI to be.
> 
> i.e. if we build for vfx22, then it's the old ABI. if we build for vfx23, then it's the new ABI.
> 
> We don't fiddle with anything, nor set any extra CMake variables. Just take the default in the compiler on the distro version it's default state.
> 
> I may be missing a detail on GitHub actions, but it feels like it should be as simple as 'pick the right distro version and compiler version for each refspec'.

Indeed, and that's more-or-less the plan for https://github.com/OpenAssetIO/OpenAssetIO/issues/1353

The wrinkle with that is the downstream projects CI/tests that need to be updated, which are currently assuming a default of old/deprecated ABI. So this PR was written to not step on any toes.

To be fair, I thought this PR would be much simpler than it turned out to be, and in retrospect doing #1353 first would have simplified things significantly.

It would simplify things even more if we just dropped support for CY22, which is very very tempting :thinking:

.github/workflows/integrations.yml
foundry-markf: Bit bashy too then. Thanks for the link.

Logically, `<something> && OFF` just feels wrong :P

.github/workflows/build-wheels.yml
foundry-markf: My main issue with deprecated is that I'm not sure it is. No idea how long it will hang around for.

With that link above they do say
> The _GLIBCXX_USE_CXX11_ABI macro controls whether the declarations in the library headers use the old or new ABI

so new/old is the documented wording from GNU.

.github/workflows/build-wheels.yml
foundry-markf: Fair point on the wording.
An EL7 derivative is what most VFX users would be aware of, and commonly use, so I'll resolve this.

.github/workflows/integrations.yml
foundry-markf: I asked the question as in this snippet, I only see Python 3.9 and 3.10 mentioned for Linux. Should 3.11 be here too?

feltech: Oh that was just to keep the build matrix down and reduce CI wait time, by focusing testing just over each side of the ABI fence.

Ideally yeah every CI test would run on every platform for every Python version, but pragmatically it adds a lot of wait time to CI for not much extra coverage. Other Python versions are tested in other CI jobs, albeit less thoroughly (just unit tests).

foundry-markf: Understood!
> by focusing testing just over each side of the ABI fence.

feltech:
Closing this in favour of uber PR https://github.com/OpenAssetIO/OpenAssetIO/pull/1391

PR #1391, opened 2024-08-09
Update default toolchain to CY24 and drop support for CY22

feltech:
## Description

Addresses much of #1351. 

This follows on from an attempt to maintain support for CY22 whilst also fixing the libstdc++ ABI for CY23 and CY24 in #1388. That attempt introduced a fair amount of boilerplate in order to work around the existing build features of OpenAssetIO.

So rather than introduce functionality only for it to be deprecated very soon, this uber PR attempts to modernise the OpenAssetIO build and CI infrastructure in one fell swoop.

- Support for CY22 is ultimately removed. This means Python 3.9 and old/deprecated libstdc++ ABI no longer needs build support.
- The default choice for CI is now CY24. This means GitHub runners use Ubuntu 22.04 / Windows 2022 / MacOS deployment target 11; and our `openassetio-build` Docker image inherits from the `aswf/ci-base:2024` image.
- Hence the `OPENASSETIO_GLIBCXX_USE_CXX11_ABI` toggle can be removed.
- Hence the `importlib_metadata` backport dependency can be removed.
- Hence the libstdc++ ABI used for Python 3.10+ wheels is trivially fixed by using a different base image for `cibuildwheel`.
- Hence no need to work around the deprecation of Node 16 in favour of Node 20 on GitHub runners due to glibc version being too old.

A few small issues were revealed during the upgrade, which have also been fixed.

## Reviewer Notes

Downstream projects must be updated to work with the new defaults. PRs for those pending.

I realised that I'd previously broken the versioning scheme of our `openassetio-build` Docker image - it was supposed to be `<VFX Ref. Plat. year>.<version>` and I'd pushed a few as `<current year>.<version>`. In particular, the current (pre this PR) image, i.e. `:latest` is `2024.4`, despite being CY2022 (CentOS 7) based. I've done some tidying and will finish off once this PR is merged. But that's why the image tag starts at `2024.5`. See https://github.com/OpenAssetIO/OpenAssetIO/pkgs/container/openassetio-build/versions


.github/workflows/build-instructions.yml
```diff
@@ -18,23 +18,23 @@ concurrency:
...
+    name: CY2024
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v4
 
     - name: Build and install (Docker)
       run: |
-        docker run -v `pwd`:/src ghcr.io/openassetio/openassetio-build bash -c '
+        docker run -v `pwd`:/src ghcr.io/openassetio/openassetio-build:2024.5.rc1 bash -c '
```
feltech, line 29: [fix] Drop commit that adds rc1 tag before merge

.github/workflows/integrations.yml
```diff
@@ -110,12 +110,15 @@ jobs:
...
       matrix:
         config:
-        - os: windows-2019
+        - os: windows-2022
           site-packages: Lib/site-packages
-        - os: ubuntu-20.04
-          site-packages: lib/python3.9/site-packages
+        # TODO(DF): Disabled pending upstream changes to use system
+        # default libstdc++ - i.e. dropping CY22 support in favour of
+        # CY24.
```
foundry-markf, line 117: Why is this a TODO?

resources/abi/libopenassetio-python.so.1.0.0.xml
foundry-markf, line 1: Not reviewed

resources/abi/libopenassetio.so.1.0.0.xml
foundry-markf, line 1: Not reviewed

foundry-markf (commented):
Generally, I see a lot of repetition. Ubuntu version, MSVC version and path, etc. Guess there's nothing can be done about that? It's just a broad change when those update.

I had to look up the std::memcpy fix, as it stood out as 'why's that there?' :)

Just one question from the comments.

.github/workflows/integrations.yml
feltech: The TODO is to un-comment those lines, which is done in https://github.com/OpenAssetIO/OpenAssetIO/pull/1392 (which can't be merged until downstream projects are updated).  I guess there was no real reason to add the TODO tag, it just helps draw attention in case it gets forgotten.

foundry-markf: OK, I'll check that the commenting out is undone in the next PR then :)

feltech:
> Generally, I see a lot of repetition. Ubuntu version, MSVC version and path, etc. Guess there's nothing can be done about that? It's just a broad change when those update.

That's a very good point there _might_ be a way to per .yml file variables for re-use, but I'm pretty certain there's no way to have  global variables/snippets that can be reused and exploited for bulk updates.

Having said that, I will do some reading in GitHub docs to see if any duplication can be reduced.

foundry-markf (approved):

feltech:
> > Generally, I see a lot of repetition. Ubuntu version, MSVC version and path, etc. Guess there's nothing can be done about that? It's just a broad change when those update.
> 
> That's a very good point there _might_ be a way to per .yml file variables for re-use, but I'm pretty certain there's no way to have global variables/snippets that can be reused and exploited for bulk updates.
> 
> Having said that, I will do some reading in GitHub docs to see if any duplication can be reduced.

I tried with commit 2653ca124fd18ccbf753e0487b1a0276ffac912b

> Use the [GitHub Organisation
Variables](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/variables#creating-configuration-variables-for-an-organization)
feature to configure top-level [OpenAssetIO Org
variables](https://github.com/organizations/OpenAssetIO/settings/variables/actions)
that define default GitHub runners, Python versions, etc, so that we
remove duplication and have an easier upgrade path in future.

But it doesn't work. For (security?) reasons, [pull requests from a fork do not have access to global variables](https://github.com/orgs/community/discussions/44322) :cry:

feltech:
Added 3e88095 updating Docker image to pull in USD and its dependencies. Tested in https://github.com/OpenAssetIO/usdOpenAssetIOResolver/pull/43

foundry-markf (approved):
Discussed the new TODO regarding Docker (see ASWF Slack wg-ci post).

My only comment is that _our_ builds of USD do have _more_ shared dependencies than what I see being copied (more than Boost, MaterialX, tbb). But, it would heavily depend upon how those other dependencies are configured in the Docker container.

If what you've got works to your satisfaction, all good.

For reference, the other dynamic dependencies we deal with are GLEW, OpenEXR, imath, OCIO, OpenVDB. (Beyond what I'd consider the minimum of Python, Qt and PySide for any imaging enabled USD builds)

feltech:
> If what you've got works to your satisfaction, all good.
> 
> For reference, the other dynamic dependencies we deal with are GLEW, OpenEXR, imath, OCIO, OpenVDB. (Beyond what I'd consider the minimum of Python, Qt and PySide for any imaging enabled USD builds)

That's interesting. The `openassetio-build` image is based off the `aswf/ci-base` image, which [already contains](https://hub.docker.com/r/aswf/ci-base) GLEW, and also already contains TBB and Boost. I tried removing the `COPY` of TBB and Boost from `vfxall` and it worked fine (updated https://github.com/OpenAssetIO/usdOpenAssetIOResolver/pull/43). I wonder if it was due to the mismatch of `:2024` between `ci-base` and `ci-vfxall`, which has since been resolved (by aforementioned TODO).

I confirmed that OpenEXR, imath, OCIO and OpenVDB are indeed absent from the image. So either they're statically linked, or they're optional dependencies.

I think until there's a use-case for building/testing with these other dependencies present, we can leave them out.

foundry-markf (approved):
Rechecked the latest Docker changes

PR #1402, opened 2024-08-22
[Release] Bump version to v1.0.0-rc.1.0 

### LINKED ISSUE #1401: Release 1.0.0-rc.1.0
## What

Release OpenAssetIO v1.0.0-rc.1.0

## Why

This release finalises all breaking changes to the API that are needed for a v1.0.0 release. This will be the first release candidate for v1.0.0. 

We remain a release candidate until verified by downstream adopters.

This is not everything needed to complete a v1.0 - additional non-breaking changes are planned for 1.0.0.rc.1.1 or 1.0.1.  Beyond this, there are plenty of issues to address and features to add for 1.1.0. or 2.0.0.

feltech:
## Description

Closes #1401. Bump from beta to release candidate. 

Remove "beta" version introspection functions. We could have adapted these somehow, but that would be a lot of work for no current use-case.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,13 +1,12 @@
 Release Notes
 =============
 
-v1.0.0-beta.x.x
+v1.0.0-rc.x.x
 ---------------
 
-_This release breaks binary compatibility and may break source
-compatibility with C++ hosts and managers. In addition, this release may
-break source compatibility for Python hosts. See breaking changes
-section for more details._
+_This release breaks binary compatibility with C++ hosts and managers.
+It may break source compatibility for hosts and managers in both C++ and
+Python. See breaking changes section for more details._
 
 ### Breaking changes
 
@@ -52,6 +51,9 @@ section for more details._
   will use the system default libstdc++ ABI, rather than defaulting to
   the old ABI.
   [#1353](https://github.com/OpenAssetIO/OpenAssetIO/issues/1353)
+
+- Removed support for querying the "beta" version number of the library.
+  [#1401](https://github.com/OpenAssetIO/OpenAssetIO/issues/1401)
 
 ### New Features
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,7 +1,7 @@
 Release Notes
 =============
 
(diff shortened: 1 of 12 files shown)


foundry-markf (approved):
I'll admit I don't understand why a prerelease identifier needs anything more than a monotonically increasing integer, rather than an x.y. But, you're not changing from previous behaviour here, so just my observation.

As there was a mechanism to query a beta major and minor version, why isn't there a mechanism to query a release candidate major and minor version?

I'm approving regardless of the above, because the change is doing what it says it does.

feltech:
> I'll admit I don't understand why a prerelease identifier needs anything more than a monotonically increasing integer, rather than an x.y. But, you're not changing from previous behaviour here, so just my observation.

We do have a [note in the README](https://github.com/OpenAssetIO/OpenAssetIO?tab=readme-ov-file#pre-releases) about this. It's to be able to flag major breaking vs. non-breaking changes. When we originally discussed it, it was partially so as a team we  could practice applying semver logic. We knew we'd have a fairly long beta period, during which time real-world applications would start using OpenAssetIO, and so semver seemed valuable.  As it turns out, I've not come across a  case where the pre-release major/minor distinction has been useful, though.

> As there was a mechanism to query a beta major and minor version, why isn't there a mechanism to query a release candidate major and minor version?

Basically, laziness/reducing maintenance burden. I don't think we properly considered the usefulness of exposing pre-release version. From the commit message
```

We could reword to "rc" but then rc 1.0 would seem lower than beta 2.0,
unless we expose the "pre-release phase" for introspection too... it all
gets a bit messy and requires proper design.

Instead, take the easy way out and remove introspection of pre-release
version numbers altogether. But keep the pre-release tag in the full
version string, so the information remains available for those that wish
to parse it.
```

Do you think there's value in keeping it, after all?

foundry-markf:
> Do you think there's value in keeping it, after all?

Only if there's evidence that people have used it.
(They can also do their own version number parsing if it's really desired)

feltech:
Added 27c84c19 with some changes to docs to stop referring to alpha/beta.

foundry-markf (approved):

PR #1404, opened 2024-08-22
Add link to Ftrack plugin

### LINKED ISSUE #1399: Add ftrack link to examples docs
## What

Add a link to the ftrack plugin alongside other links to pre-existing examples. I.e. https://github.com/ftrackhq/ftrack-openassetio-manager

## Why
 
Ftrack is a commercial asset management system that was instrumental in the ancestry of OpenAssetIO and continues to provide guidance via the TSC. The OpenAssetIO team started an experimental ftrack plugin, which has since been forked and maintained by the ftrack team themselves. However, we do not mention or link to it in our docs, despite ftrack being a key collaborator. Not to mention ftrack support could help drive adoption.

## Acceptance Criteria

* Link to ftrack plugin from `examples/README.md`
* Update API docs landing page to link to `examples/README.md`

feltech:
## Description

Closes #1399. Add a link to the Ftrack manager plugin in the `examples/README.md`. Also update the Doxygen API docs main page to link to this README rather than maintaining a separate list of examples.

## Reviewer Notes

To build the docs
```
cd docs/doxygen
make
```
The docs will be placed in a `html` directory under `docs/doxygen`.

### DIFF
--- a/.github/markdown-link-check-config.json
+++ b/.github/markdown-link-check-config.json
@@ -1,0 +1,7 @@
+{
+  "ignorePatterns": [
+    {
+      "pattern": "^https://www.ftrack.com"
+    }
+  ]
+}
--- a/doc/doxygen/src/MainPage.dox
+++ b/doc/doxygen/src/MainPage.dox
@@ -6,13 +6,8 @@
  * * 📐 Get off to a good start with the <a
      href="https://github.com/OpenAssetIO/Template-OpenAssetIO-Manager-Python">Python
      manager template.</a>
- * * 📖 See some examples: <a
-     href="https://github.com/ynput/ayon-openassetio-manager-plugin/tree/main#readme">Ayon
-     manager plugin</a>, <a
-     href="https://github.com/OpenAssetIO/usdOpenAssetIOResolver#readme">USD
-     Asset Resolver prototype</a>, <a
-     href="https://github.com/OpenAssetIO/otio-openassetio">OTIO media
-     linker</a>
+ * * 📖 See some <a href="https://github.com/OpenAssetIO/OpenAssetIO/tree/main/examples#readme">examples</a>
+ *   in the OpenAssetIO ecosystem, including USD, OTIO, Ftrack, Ayon, and more.
  * * 📖 See a <a
      href="https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/blob/main/examples/hello_openassetio.ipynb">Hello World</a> tutorial example, with more tutorial examples to be found in <a
      href="https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/tree/main/examples/">OpenAssetIO-MediaCreation</a>
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -64,6 +64,10 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: gaurav-nelson/github-action-markdown-link-check@v1
+      with:
+        # Suppress ftrack.com, which errors with http 520 when accessed
+        # via GitHub CI, for unknown reasons.
+        config-file: .github/markdown-link-check-config.json
 
   # Note: in order to keep an `actions/cache` cache up to date, we must
   # use the approach detailed in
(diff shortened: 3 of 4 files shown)


feltech:
Interestingly, `https://www.ftrack.com/en/studio` [fails markdown-link-check](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/10511333663/job/29152549059) with a HTTP 520 error on GitHub CI. A 520 is apparently Cloudflare-specific and just means "unknown server error", so no help there.  The error does not occur when I run `markdown-link-check` locally.  I wonder if its a localisation thing (noting the `/en` in the URL)? No big deal to use the top-level URL `https://www.ftrack.com` .

feltech:
> Interestingly, `https://www.ftrack.com/en/studio` [fails markdown-link-check](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/10511333663/job/29152549059) with a HTTP 520 error on GitHub CI. A 520 is apparently Cloudflare-specific and just means "unknown server error", so no help there. The error does not occur when I run `markdown-link-check` locally. I wonder if its a localisation thing (noting the `/en` in the URL)?
> 
> EDIT: [Same issue](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/10521646519/job/29152701379?pr=1404) with `https://www.ftrack.com` 🤔

I ended up just adding a suppression for ftrack.com :cry: . There are more options to try to get it to work, e.g. experimenting with HTTP headers. However, the time spent trying to get this to work (especially as its not reproducible locally) doesn't seem worth the effort.

.github/markdown-link-check-config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "ignorePatterns": [
+    {
+      "pattern": "^https://www.ftrack.com"
```
foundry-markf, line 4: [question] Can this suppression be more specific to the 520 error? i.e. if a 500 error is returned, then it's not ignored.

doc/doxygen/src/MainPage.dox
```diff
@@ -6,13 +6,8 @@
...
      href="https://github.com/OpenAssetIO/Template-OpenAssetIO-Manager-Python">Python
      manager template.</a>
- * * 📖 See some examples: <a
-     href="https://github.com/ynput/ayon-openassetio-manager-plugin/tree/main#readme">Ayon
-     manager plugin</a>, <a
-     href="https://github.com/OpenAssetIO/usdOpenAssetIOResolver#readme">USD
-     Asset Resolver prototype</a>, <a
-     href="https://github.com/OpenAssetIO/otio-openassetio">OTIO media
-     linker</a>
+ * * 📖 See some  <a href="https://github.com/OpenAssetIO/OpenAssetIO/tree/main/examples#readme">examples</a>
```
foundry-markf, line 9: Nit: Was it intentional to add two spaces after "See some" and the anchor?

foundry-markf (approved):
My queries are not blocking giving an approval.

.github/markdown-link-check-config.json
feltech: We can mark specific status codes as "OK" in [the config](https://github.com/tcort/markdown-link-check#config-file-format) but it applies to _all_ URLs. I was tempted to do that, but went with targeting the specific URL instead.

There is an argument that _any_ 5xx code should be "OK" from the perspective of just checking that links are valid. But there are cases where a malformed link, coupled with poor validation on the server side, could trigger a server error code path. So on balance I thought blanket suppressing the specific link, rather than error code, was the lesser evil.

Happy to hear any thoughts on this?

doc/doxygen/src/MainPage.dox
feltech: Good catch. Fixed and took the liberty of amending and force pushing (d9f78a03).

.github/markdown-link-check-config.json
foundry-markf: I imagine it's whether such as in https://github.com/tcort/markdown-link-check that says a 200 return code is the only way to say a link is alive or not.
But as it's just checking Markdown, I'm not going to make a fuss. (I just had it in my mind pytest tests for `requests` where it's useful to check the return code)

foundry-markf (approved):

PR #1418, opened 2024-09-20
Don't export pybind11 exception types

### LINKED ISSUE #1416: Mitigation for pybind11 exceptions being forcibly exported
## What

Decide whether to mitigate and if so design and implement a mitigation for exception types defined by pybind11 being exported by the pybind11 library (i.e. `default` visibility).

## Why

Ideally, all OpenAssetIO dependencies are private and do not leak to the host application. This includes pybind11, where it would be ideal if OpenAssetIO's Python bindings are independent from the host application, i.e. where the host could use a different version of pybind11 if they wished, and everything would still work.

However, it turns out that pybind11 exports it's exception classes. In particular, the `pybind11::error_already_set` exception type that is used to carry Python exception data through a C++ exception (and potentially back to Python again). 

For example, the ABI of the `error_already_set` type was changed radically between pybind11 2.9 and 2.10. If OpenAssetIO is built with 2.10 and the host is built with 2.9, and the exception is caught in the host (even as a `std::exception`), and `.what()` is called on it, the wrong implementation of `.what()` might be executed, giving garbled text.

The only solution for now is for consumers to ensure OpenAssetIO is built with the same version of pybind11 as the host application.

## Notes

One quick solution to this is to set `-DPYBIND11_EXPORT_EXCEPTION=""`, but this may have other unwanted side-effects - e.g. if we needed to catch a pybind11-specific exception in `openassetio-python.so` (Python<->C++ bridge) that was thrown by `_openassetio.so` (Python extension module).

Another option is to change the `pybind11` namespace. This isn't trivial to do. We can completely change `pybind11` to something else (by setting `PYBIND11_NAMESPACE`), and update all our code usages. But ideally the pybind11 authors would add an `inline` namespace to protect against ABI changes, as we do.

feltech:
## TODO

- Double-check Windows.
- Automated test?

## Description

Closes #1416.

For example, the ABI of the `error_already_set` type was changed radically between pybind11 2.9 and 2.10. If OpenAssetIO is  built with 2.10 and the host is built with 2.9, and the exception is caught in the host (even as a `std::exception`), and `.what()` is called on it, the wrong implementation of `.what()` may be executed, giving garbled text. This exact scenario was seen in practice.

Disabling the export of the pybind11 exception will cause problems if we ever need to catch a pybind11-specific exception in a different DSO. We don't currently have that use-case, however.

So disable the export of pybind11 exceptions by overriding its `PYBIND11_EXPORT_EXCEPTION` macro wherever the `pybind11::module` CMake target is linked.

Host applications can still catch the pybind11 exception as a `std::exception`, and safely call `.what()` on it to extract the error
message, and the correct implementation will be used. 

## Reviewer Notes

The alternative is to insist that hosts build with the same pybind11 as is used to build OpenAssetIO, which isn't entirely unreasonable.

The root cause was discovered using a gdb debug session of a gcc 11 build, which revealed that the OpenAssetIO pybind11 was used to construct the exception object, but the host pybind11 was used for its vtable pointer.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -15,6 +15,13 @@ managers._
   object workflows.
   [#1170](https://github.com/OpenAssetIO/OpenAssetIO/issues/1170)
 
+### Bug fixes
+
+- pybind11 exception class symbols are no longer exported from the
+  OpenAssetIO Python extension module, so that hosts using a different
+  version of pybind11 do not conflict.
+  [#1416](https://github.com/OpenAssetIO/OpenAssetIO/issues/1416)
+
 v1.0.0-rc.1.0
 ---------------
 
--- a/cmake/ThirdParty.cmake
+++ b/cmake/ThirdParty.cmake
@@ -79,6 +79,13 @@ if (OPENASSETIO_ENABLE_PYTHON)
 
     # pybind11 for C++ Python bindings.
     find_package(pybind11 2 REQUIRED)
+    # TODO(DF): Remove if/when fixed upstream: https://github.com/pybind/pybind11/issues/5359
+    # Ensure the pybind11 exception types are not exported. If a host
+    # uses a different version of pybind11 then we can get bad ODR bugs.
+    get_target_property(_pybind11_defs pybind11::module INTERFACE_COMPILE_DEFINITIONS)
+    list(APPEND _pybind11_defs PYBIND11_EXPORT_EXCEPTION=)
+    set_target_properties(
+        pybind11::module PROPERTIES INTERFACE_COMPILE_DEFINITIONS "${_pybind11_defs}")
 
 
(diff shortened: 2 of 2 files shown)


feltech:
Just hit an issue with this solution. For some reason it caused a negative refcount error (with CPython in debug mode). The exception was picked up by the application's pybind11 as it propagated back into Python. 

Specifically, the `translate_exception` function in pybind11's `internals.h` attempted to `restore()` the `error_already_set`, where the version of pybind11 attempting to do this is the application's not OpenAssetIO's.

feltech:
Since this fix doesn't fully solve the problem, I'll close this. The issue must be fixed upstream (see parent issue). The only workaround for now is to use a consistent version of pybind11 across all components of a host application.

PR #1421, opened 2024-10-28
Downgrade pybind11 used in CI

### LINKED ISSUE #1420: Downgrade pybind11 requirement to 2.9
## What

Downgrade pybind11 version used in build instructions / CI to v2.9.

## Why

Existing known integrations use pybind11 v2.9.

It turns out that pybind11 leaks symbols, and introduces breaking changes in minor versions - see #1416.

pybind11 v2.9 is missing some features that we assume from v2.10 - see https://github.com/OpenAssetIO/OpenAssetIO-TraitGen/issues/55

So we should use v2.9 by default in CI and build instructions, to better simulate downstream platforms.

feltech:
## Description

Closes #1420.

Existing known integrations use pybind11 v2.9. If we test with a newer version, then we run the risk of using features that are not available in older versions.

Ideally this wouldn't be a problem, because pybind11 is a private dependency. However, it becomes a problem when a host application also uses pybind11, for a couple of reasons.

Firstly, build systems may not like building different components of an application with different dependency versions (e.g. Conan), even if those dependencies are private to the component.

Secondly, pybind11 is not a truly private dependency. If a host application also uses pybind11, then they will interact, especially through exceptions, which are exported symbols with no ABI namespace to disambiguate (in current versions of pybind11).

An example of a feature that is not supported in older pybind11 versions is the native support of `frozenset`, which was used in TraitGen in OpenAssetIO/OpenAssetIO-TraitGen#55 and triggers an exception during Python->C++ conversion (see OpenAssetIO/OpenAssetIO-TraitGen#94 to address this).

- [ ] ~~I have updated the release notes.~~ - build system/instructions only.


feltech:
Turns out for CI to pass https://github.com/OpenAssetIO/OpenAssetIO-TraitGen/issues/94 will have to be done first.

Also interesting that the ABI diff check CI job didn't fail. We deliberately don't test the Python extension module, reasoning that nothing in it is public. But of course that isn't true due to pybind11, in particular its exported exception types.

foundry-markf (approved):
Reasonable change

PR #1422, opened 2024-10-28
C++->Python exception "raise from" support

### LINKED ISSUE #1419: Support chaining exceptions in Python exceptions translator
## What

Support chaining exceptions (`raise from`) during C++->Python exception translation

## Why

If a C++ exception occurs in a pybind11-bound C++ function, then we attempt to translate the C++ exception to a Python exception. As part of this we execute Python statements, in-particular we import `openassetio._openassetio.errors`. Attempting to do this import whilst the Python error indicator is set causes a fatal error in Python.

This can happen if e.g. in the call stack of a pybind11-bound function, some C++ code uses the CPython API, which triggers a Python exception, but then the C++ code goes on to throw its own exception too.

pybind11 handles this case in the default exception translator by checking if `PyErr_Occurred()` then doing the CPython equivalent of `raise from`. The OpenAssetIO exception translator should do similar.

feltech:
## Description

Closes #1419. If a C++ exception occurs in a pybind11-bound C++ function, then we attempt to translate the C++ exception to a Python exception. As part of this we execute Python statements, in-particular we import `openassetio._openassetio.errors`. Attempting to do this import whilst the Python error indicator is set causes a fatal error in Python.

This can happen if e.g. in the call stack of a pybind11-bound function, some C++ code uses the CPython API, which triggers a Python exception, but then the C++ code goes on to throw its own exception too.

So handle this case by fetching the pre-existing exception, and setting the `__cause__` and `__context__` of the newly translated C++->Python exception.

The code duplicates the logic in `pybind11::raise_from`. We can't use that function directly, since it only supports simple exceptions that can be created with `PyErr_SetString` (i.e. doesn't support `BatchElementError`, whose constructor takes arguments other than a string `message`).

### DIFF
--- a/src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
+++ b/src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright 2013-2023 The Foundry Visionmongers Ltd
+// Copyright 2023-2024 The Foundry Visionmongers Ltd
+#include <cassert>
 #include <cstddef>
 #include <exception>
 #include <string_view>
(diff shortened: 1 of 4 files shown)


src/openassetio-python/cmodule/src/errors/exceptionsBinding.cpp
```diff
@@ -51,7 +81,23 @@ void setPyException(const Exception &exception, const py::module_ &pyModule,
...
+
+  // Set the __cause__ of our new exception if raising from a previous
+  // exception.
+  if (fromExcVal != nullptr) {
+    PyObject *excVal = nullptr;
+    PyObject *excType = nullptr;
+    PyObject *excTraceback = nullptr;
+    PyErr_Fetch(&excType, &excVal, &excTraceback);
+    PyErr_NormalizeException(&excType, &excVal, &excTraceback);
+    Py_INCREF(fromExcVal);  // Bump to own two references, stolen by SetCause and SetContext.
```
foundry-markf, line 96: To confirm the logic here, `PyErr_Fetch` adds a reference to each of type, val, and traceback. Type and traceback are decref'd in the lambda above. So val (returned from the lambda) has ref=1. You then increment it here, it's ref=2. And then `PyException_SetCause` and `PyException_SetContext` steal 2 references, so we're now back to val has none. 

But reading `PyErr_Restore`, says each value requires a valid reference, as it takes it away. https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Restore

So I've lost track of intentions on references here.

feltech: I think you've mixed up `fromExcVal` and `excVal` - the exception values from two different exceptions. I.e. the `PyErr_Restore` below steals a reference from `excVal`, not `fromExcVal`.

This code is mostly a copy-paste from pybind11's `raise_from` implementation.

foundry-markf: I was originally looking at the `PyErr_Fetch` on line 61 in the lambda, and counting from there, which assigns to `fromExcVal`. Bit confusing when there are two `PyErr_Fetch`s in play.

But, if you're not seeing any reference leaks, then I could quite possibly have miscounted from just eyeballing.

feltech: Is then the missing piece of the puzzle the `PyErr_SetObject` inbetween on line 86? This adds a second exception into the mix.

I.e. we execute `PyErr_SetObject`, then (if there was a `fromExcVal`) we immediately `Fetch` that new exception, in order to `SetCause` and `SetContext` on it to the previous exception (`fromExcVal`), before `Restore`ing back the new exception.

foundry-markf: I would've said no to the `PyErr_SetObject` as it doesn't seem to interact with `fromExcVal` returned from the earlier lambda.

As we discussed earlier, tracking references isn't the best tooled up in Python, so I don't know if there's a best route forward on here, other than mark it mentally as 'something to watch'.

foundry-markf: Resolved offline.
I *think* (it's been a few days) that I mistaook the `PyErr_Restore` on line 99 for also using `fromExcVal`, but it doesn't.

So ref count of that from the lambda is 1, it's added to here, and then taken away twice in SetCause and SetContext.

We're square with the universe again.

foundry-markf (approved):

PR #1425, opened 2024-11-12
`EntityReference` associative container support

### LINKED ISSUE #573: [Core] `EntityReference` container interoperability

feltech:
## Description

Closes #573. Entity references used to be simple strings. Now they are a strong type and lose the conveniences that come with strings.

In particular, they could not be used as keys in associative containers, in Python nor C++.

So add operators and hashing in both C++ and Python that allow entity references to be used in containers.

For completeness, add the similar operators that are not technically needed for container support, but may be useful in other situations. In particular noting that Python infers partner operators (e.g. `>` from `<`) whereas C++ (17) does not. So adding explicit support for these to C++ provides parity with Python functionality.

### DIFF
--- a/src/openassetio-core/include/openassetio/EntityReference.hpp
+++ b/src/openassetio-core/include/openassetio/EntityReference.hpp
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright 2022 The Foundry Visionmongers Ltd
+// Copyright 2024 The Foundry Visionmongers Ltd
 #pragma once
 
 #include <utility>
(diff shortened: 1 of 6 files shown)


src/openassetio-core/include/openassetio/EntityReference.hpp
```diff
@@ -41,16 +41,68 @@ class EntityReference final {
...
+   * @return `true` if contents are not equal, `false` otherwise.
+   */
+  [[nodiscard]] bool operator!=(const EntityReference& other) const { return !(*this == other); }
+
+  /**
+   * Compare this reference with another.
+   *
+   * @param other Entity reference to compare against.
+   *
+   * @return `true` if this reference is less than other, `false`
```
foundry-markf, line 68: Is it worth defining what "less than" means for an Entity reference in the public docs?

Is it equivalent to "less than" for a std::string? (lexicographical compare if I recall) I suspect so givn the implementation.

src/openassetio-core/include/openassetio/EntityReference.hpp
```diff
@@ -62,8 +114,34 @@ class EntityReference final {
...
+    // Small coprime shift distances to spread out the bits.
+    constexpr std::size_t kLShift = 6;
+    constexpr std::size_t kRShift = 2;
+    // Seed incorporating type hash, assuming initial seed of 0
+    // (note: hash_code() is not constexpr in C++17).
+    static const std::size_t kSeed = typeid(openassetio::EntityReference).hash_code() + kInvPhi;
+    // Precompute rhs of hash equation.
+    static const std::size_t kBitMixer = kInvPhi + (kSeed << kLShift) + (kSeed >> kRShift);
+
+    return kSeed ^ (std::hash<openassetio::Str>{}(ref.toString()) + kBitMixer);
```
foundry-markf, line 144: As the comparison operators just work on the member string, should this hash not be the hash of that string? (in terms of consistency)

(I also assume, as can't see the code in this PR, that `ref.toString()` returns `ref.entityReferenceString_`?)

foundry-markf (approved):
A few questions, but otherwise LGTM

src/openassetio-core/include/openassetio/EntityReference.hpp
feltech: > As the comparison operators just work on the member string, should this hash not be the hash of that string? (in terms of consistency)

As in, don't add the typeid to the hash?  

I think its consistent with the operators, in that you can't compare a string to an `EntityReference` directly (only if you explicitly call `.toString()`).  

I thought it a good idea to add typeid to the hash more for Python than for C++, where heterogeneous containers are a possibility (e.g. a `set` with both an `EntityReference` and a string). 

As it turns out, the C++ `std::hash` of a `std::string` gives a different value than the Python `__hash__` of a `str`, so despite there being a test adding both an `EntityReference` and a `str` to a container, there's fundamentally no chance of a clash. 

I did consider augmenting the tests to add a (test only) Python binding of `std::string` to show how they could coexist with `EntityReference`s in an associative container without clashing. However, it seemed like a lot of effort to test such a theoretical case.  Perhaps I should revisit that, if only to prove the point of including the typeid in the hash?

There's also an argument that the hash of an `EntityReference` may have other unforeseen uses where keeping those hashes distinct from the hash of a `std::string` would be useful. It just seems a best practice to disambiguate.

Happy to reconsider if there's a compelling reason?

> (I also assume, as can't see the code in this PR, that ref.toString() returns ref.entityReferenceString_?)

Yeah exactly.

src/openassetio-core/include/openassetio/EntityReference.hpp
feltech: Good shout - littered the docstrings with "lexicographically" in ca44851c

src/openassetio-core/include/openassetio/EntityReference.hpp
foundry-markf: > I think its consistent with the operators, in that you can't compare a string to an EntityReference directly

I hadn't considered that - I was just thinking that if an Entity Reference _is_ a string (by composition), then maybe the hash should be the same.

Although the 'data' should be the same (which is what is in my head), the semantics of comparing two types should give a different value.

Resolving.

foundry-markf (approved):
LGTM

Issue #1390, opened 2024-08-09
Upgrade Clang based static analysis and linting tools


feltech:
## What

Update to the latest feasible `clang-tidy` and `clang-format` tools

## Why

At time of writing GitHub CI is still on v12 when v18 is the latest stable release. Local usage of later `clang-tidy` versions shows that they do find more and genuine issues, as well as fix bugs in older versions, especially false negatives.

```[tasklist]
### Tasks
- [ ] Remove clang-tidy suppressions from external macros e.g. assert, Catch2
```

feltech:
See comments on https://github.com/OpenAssetIO/OpenAssetIO/issues/812 for some initial investigation re. upgrading Clang tooling.

feltech:
Interestingly, I found that I've been running clang-tidy 13 locally (vs. 12 on CI).  When I try clang-tidy 12 I get unknown compiler option errors (`-flto=auto`). My theory why CI works is that even though clang-tidy 12 is used, clang itself is newer ([v13-v15](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md)).  Since I'm using conda locally, I can't have a mismatched clang and clang-tidy version.

One more reason to upgrade the clang tooling.

foundry-markf:
Can you grab them from PyPi instead? 

https://pypi.org/project/clang-tidy/ v19
https://pypi.org/project/clang-format/ v19

feltech:
> Can you grab them from PyPi instead?
> 
> https://pypi.org/project/clang-tidy/ v19 https://pypi.org/project/clang-format/ v19

Yeah, I noticed this last year (https://github.com/OpenAssetIO/OpenAssetIO/issues/812#issuecomment-1458226505), should make adopting them on CI much easier.  The real work is in fixing (or suppressing) all the issues it raises.

Issue #1196, opened 2023-11-10
Strategy for out-of-bounds index given to success/error callbacks


foundrytom:
## What

Bounds check indices passed to success/error callbacks in convenience signatures. 

## Why

Avoid segfaults and undefined behavior by accessing out of bounds elements.

## Acceptance Criteria

If an out of bounds index is passed to a success/error callback, we should throw a nice exception with a human friendly `.what()`.

elliotcmorris:
We should likely use .at() here rather than direct access. The manager is the one in control of the index here, it's a safety concern.

PR #1426, opened 2024-11-28
Support cppcoreguidelines Clang-Tidy checks

feltech:
## Description

In a quest to add static analysis checks for unsafe array access as part of #1196, the cppcoreguidelines set of Clang-Tidy checks seemed to fit the bill. However, it turns out that there is only a check for `std::array` and not for e.g. `std::vector`, which is the main cause for concern (i.e. callbacks defined in `ManagerConveniences.cpp` that take a `std::vector` index from an untrusted plugin source).

Despite this, many of the other checks in cppcoreguidelines are valuable.

So enable the cppcoreguidelines set of checks in the `.clang-tidy` config, and fix the issues that are flagged.

Whilst doing this, it became apparent that `LoggerInterface` and `ManagerImplementationFactoryInterface` had pure virtual destructors, when that is not needed. So `default` them, meaning implementors are not forced to implement a destructor if they don't need to.

There are some limitations to the cleverness of the cppcoreguidelines checks that are fixed in later versions, but updating the Clang-Tidy version used in CI is another issue (#1390).

So suppress cases where the rule of 5 shouldn't apply because the class is a single defaulted destructor, but the `default` is specified in the source file, rather than header. This is fixed in later versions.

Also suppress many cases flagged in system headers via use of a macro from the system header, e.g. the `assert` macro, Catch2 test macros. This is also fixed in later versions.

Dealing with rule of 5 violations led to some discoveries of classes with dangerously defined default copy/move operations.

So explicitly delete move and assignment of `TraitsData` instances. Explicitly delete all copy and move of `FileUrlPathConverter`. Explicitly delete all copy and move of (private) `Regex` and its `Regex::Match` helper struct (though allow move of `Match` as a requirement for returning in a `std::optional`). Adding support for copy/move to these types may be revisited in the future.

Arbitrarily add some missing `#include`s for headers that are otherwise brought in via transitive includes, as flagged by latest Clang-Tidy's `include-cleaner` check (not yet enforced on CI, see note above about upgrading).

## Reviewer Notes

Added a task to #1390 to ensure the many suppressions of third-party macros are removed when moving to newer Clang-Tidy


feltech:
Result of ABI diff check CI job

```
 ===== Checking ABI of libopenassetio.so.1.0.0 =====

Leaf changes summary: 4 artifacts changed
Changed leaf types summary: 0 leaf type changed
Removed/Changed/Added functions summary: 4 Removed, 0 Changed, 0 Added function

4 Removed functions:

  [D] 'method virtual openassetio::v1::hostApi::ManagerImplementationFactoryInterface::~ManagerImplementationFactoryInterface()'    {_ZN11openassetio2v17hostApi37ManagerImplementationFactoryInterfaceD0Ev}
    note that this removes an entry from the vtable of class openassetio::v1::hostApi::ManagerImplementationFactoryInterface
  [D] 'method virtual openassetio::v1::hostApi::ManagerImplementationFactoryInterface::~ManagerImplementationFactoryInterface()'    {_ZN11openassetio2v17hostApi37ManagerImplementationFactoryInterfaceD2Ev, aliases _ZN11openassetio2v17hostApi37ManagerImplementationFactoryInterfaceD1Ev}
    note that this removes an entry from the vtable of class openassetio::v1::hostApi::ManagerImplementationFactoryInterface
  [D] 'method virtual openassetio::v1::log::LoggerInterface::~LoggerInterface()'    {_ZN11openassetio2v13log15LoggerInterfaceD0Ev}
    note that this removes an entry from the vtable of class openassetio::v1::log::LoggerInterface
  [D] 'method virtual openassetio::v1::log::LoggerInterface::~LoggerInterface()'    {_ZN11openassetio2v13log15LoggerInterfaceD2Ev, aliases _ZN11openassetio2v13log15LoggerInterfaceD1Ev}
    note that this removes an entry from the vtable of class openassetio::v1::log::LoggerInterface
```

I think this is a false positive, since the definition (`= default`) of the destructors has moved from the source file to the header. Does this really affect the vtable?

The release notes were updated to remove the note on remaining binary compatible, out of an abundance of caution.

src/openassetio-core-c/tests/hostApi/ManagerTest.cpp
```diff
@@ -187,7 +187,7 @@ SCENARIO("A host calls Manager::identifier") {
     // Create mock ManagerInterface to inject and assert on.
     const managerApi::ManagerInterfacePtr mockManagerInterfacePtr =
         std::make_shared<MockManagerInterface>();
-    auto& mockManagerInterface = static_cast<MockManagerInterface&>(*mockManagerInterfacePtr);
+    auto& mockManagerInterface = dynamic_cast<MockManagerInterface&>(*mockManagerInterfacePtr);
```
foundry-markf, line 190: Just calling out I didn't see casts (particularly dynamic_casts) appear in the description, so checking if this needs some words to explain to readers?

src/openassetio-core/src/utils/Regex.hpp
```diff
@@ -41,6 +41,14 @@ class Regex {
    public:
     explicit Match(const pcre2_code* code);
 
+    ~Match() = default;
+    Match(const Match&) = delete;
```
foundry-markf, line 45: As there's a custom constructor, should the default ctor be deleted?

foundry-markf: Same question to `Regex` below.

foundry-markf: DeathwatchedSimpleCppType too?

foundry-markf (commented):
Just a couple of questions, but I'd likely approve after a bit of discussion.

src/openassetio-core-c/tests/hostApi/ManagerTest.cpp
feltech: Yeah there are a couple of changes to satisfy cppcoreguidelines that I didn't put in the description (e.g. uninitialised variables), since I thought they didn't really need further explanation. Do you think this change is controversial enough to warrant flagging explicitly in the description/commit?

foundry-markf: `dynamic_cast` tends to catch my eyes as to why is this needed. Unless all the consumers are well aware of this? Maybe not an issue as the library is compiled with RTTI, but that tends to be a sticky point IMO.

src/openassetio-core/src/utils/Regex.hpp
feltech: The default constructor is implicitly deleted in these cases, I believe.

From [cppreference](https://en.cppreference.com/w/cpp/language/default_constructor)

> If **there is no user-declared constructor** or constructor template for a class type, the compiler will implicitly declare a default constructor as an inline public member of its class. 

To double check: https://godbolt.org/z/n1o6dG4jc

Do you still think its worth explicitly `delete`ing to be explicit? cppcoreguidelines doesn't seem to mind (at least, the Clang-Tidy checks don't).

src/openassetio-core-c/tests/hostApi/ManagerTest.cpp
feltech: I think its reasonable that casting from base to derived should be a `dynamic_cast` for safety, as a general rule. 

In this case we can be certain of the provenance of the object, so perhaps `static_cast` isn't so bad after all? 

This does only affect unit tests so is not so important.

Indeed OpenAssetIO will have all sorts of problems without RTTI.

Otherwise, I'll add a note about it in the commit message when squashing down the commits.

foundry-markf: Ahh there you go, hadn't realised it was intentionally from base to derived. All good!

src/openassetio-core/src/utils/Regex.hpp
foundry-markf: Having done more reading on it, and looking at your Compiler Explorer link, I'm satisfied it's deleted.

I was merely wondering if it made more sense to be explicit, but seems C++ handles this for us.

foundry-markf (approved):
LGTM, nice cleanup

feltech:
> Result of [ABI diff check CI job](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/12065387117/job/33644046514)
> 
> <snip> 

Updated to move the definition of those destructors back into the source file.  They were the only virtual methods of those classes, making vtable linker resolution follow "inline" rules when `default`ed in the header. This shouldn't be an issue, but in the past we found Windows MSVC could get linker errors unless something is defined in a source file. So out of an abundance of caution, and because the cost is very little, let's play it safe and move them back.

This means the only potential ABI change is the removal of `= 0`, which does still trigger an [ABI diff check job failure](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/12068241519/job/33652922962?pr=1426)

```
===== Checking ABI of libopenassetio-python.so.1.0.0 =====

Leaf changes summary: 1 artifact changed
Changed leaf types summary: 0 leaf type changed
Removed/Changed/Added functions summary: 0 Removed, 0 Changed, 0 Added function
Function symbols changes summary: 0 Removed, 0 Added function symbol not referenced by debug info
Variable symbols changes summary: 1 Removed, 0 Added variable symbol not referenced by debug info

1 Removed variable symbol not referenced by debug info:

  [D] _ZTSN11openassetio2v17hostApi37ManagerImplementationFactoryInterfaceE

gmake[3]: *** [src/openassetio-python/CMakeFiles/openassetio.test.abi.openassetio-python-bridge.dir/build.make:73: src/openassetio-python/CMakeFiles/openassetio.test.abi.openassetio-python-bridge] Error 12
gmake[2]: *** [CMakeFiles/Makefile2:840: src/openassetio-python/CMakeFiles/openassetio.test.abi.openassetio-python-bridge.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:847: src/openassetio-python/CMakeFiles/openassetio.test.abi.openassetio-python-bridge.dir/rule] Error 2
gmake: *** [Makefile:426: openassetio.test.abi.openassetio-python-bridge] Error 2

3/3 Test #3: openassetio.test.abi.openassetio-core ............***Failed    6.33 sec
[ 90%] Built target openassetio-core

===== Checking ABI of libopenassetio.so.1.0.0 =====

Leaf changes summary: 1 artifact changed
Changed leaf types summary: 0 leaf type changed
Removed/Changed/Added functions summary: 0 Removed, 0 Changed, 0 Added function
Function symbols changes summary: 0 Removed, 0 Added function symbol not referenced by debug info
Variable symbols changes summary: 0 Removed, 1 Added variable symbol not referenced by debug info

1 Added variable symbol not referenced by debug info:

  [A] _ZTVN11openassetio2v17hostApi37ManagerImplementationFactoryInterfaceE
```

The symbol supposedly removed from `libopenassetio-python` demangles to `typeinfo name for openassetio::v1::hostApi::ManagerImplementationFactoryInterface`.

The symbol supposedly added to `libopenassetio` demangles to `vtable for openassetio::v1::hostApi::ManagerImplementationFactoryInterface`

I'm not sure what this means exactly for ABI compatibility. It's interesting that `LoggerInterface` does not cause a similar ABI dump change, despite being structurally very similar.

Perhaps best to keep the change to the release notes calling out that it's not necessarily binary compatible.

PR #1435, opened 2025-01-03
Update macOS runner to macos-13

### LINKED ISSUE #1434: Update to non-deprecated GitHub macOS runner version
## What

Update macOS CI jobs to use `macos-13`

## Why

GitHub CI has started rejecting macOS 12, e.g. from https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/12597916489/job/35111779833?pr=1433

> The macOS-12 environment is deprecated, consider switching to macOS-13, macOS-14 (macos-latest) or macOS-15. For more details, see https://github.com/actions/runner-images/issues/10721

[VFX Reference Platform](https://vfxplatform.com/#footnote-macos) CY2024 only specifies a Minimum Deployment Target of 11.0 (CY2025 specifies 12.0). So we cannot match our GitHub runner to VFX CY2024. Nearest we can do is target the lowest available macOS version.

feltech:
## Description

Closes #1434. GitHub CI has started rejecting macOS 12, i.e.

> The macOS-12 environment is deprecated, consider switching to macOS-13, macOS-14 (macos-latest) or macOS-15. For more details, see actions/runner-images#10721

VFX Reference Platform CY2024 only specifies a Minimum Deployment Target of 11.0 (CY2025 specifies 12.0). So we cannot match our GitHub runner to VFX CY2024. Nearest we can do is target the lowest available macOS version.


foundry-markf:
I did a search through GitHub's web interface, and couldn't find `macosx-version-min=11.0` (or 12.0) specified anywhere.

Although you couldn't check with these actions that the min version is honoured, you can at least compile against it, as there are checks in the Xcode APIs if you're using functions that are not supported on the older versions.

feltech:
> I did a search through GitHub's web interface, and couldn't find `macosx-version-min=11.0` (or 12.0) specified anywhere.
> 
> Although you couldn't check with these actions that the min version is honoured, you can at least compile against it, as there are checks in the Xcode APIs if you're using functions that are not supported on the older versions.

Yeah we do `set(CMAKE_OSX_DEPLOYMENT_TARGET 11)` in CMake, as well as the `MACOSX_DEPLOYMENT_TARGET` env var for `cibuildwheel` (for creating Python wheels), so I hope that adds some validation.

foundry-markf (approved):
As the min version is set (I hadn't spotted it in CMake variables), then this should be fine, and will complain at compile/link time of anything untoward.

It's a pity we can't validate it on the desired machinery, but we can do that in other ways in the consumers of OpenAssetIO.

PR #1436, opened 2025-01-03
Upgrade to clang-tidy/clang-format v19

### LINKED ISSUE #1390: Upgrade Clang based static analysis and linting tools

feltech:
## Description

Closes #1390. Upgrade to the latest available `clang-tidy` and `clang-format` packages at time of writing.

Use `pip` to install these, since the GitHub ubuntu-22.04 runner maxes out at v15, whereas `pip` currently provides up to v19.

Update code to satisfy the new clang-tidy version, mostly adding lots of `#include`s. There are also a handful of non-clang-tidy changes, where CLion flagged some linting issues of its own.

Perhaps worth a shout that, in addition to clang tooling, the version of CMake is bumped to v3.31, so that `generate_export_header()` will add clang-tidy suppression comments to the generated header ([CMake issue](https://discourse.cmake.org/t/the-generateexportheader-should-prevent-clang-tidy-warning/10620)).  This is not reflected in the `cmake_minimum_required()` call in `CMakeLists.txt`, since it's only a requirement when linting.

## Reviewer Notes

The changes shouldn't introduce any source or binary incompatibilities, as evidenced (in part) by the ABI diff check CI job.

Any C API changes aren't too important, since the C API is woefully incomplete and needs revisiting as a whole.

See commit message for some details on choices made to satisfy clang-tidy.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022 Google LLC
-# Copyright 2013-2022 The Foundry Visionmongers Ltd
+# Copyright 2013-2025 The Foundry Visionmongers Ltd
 
 # Modified from: https://raw.githubusercontent.com/googleapis/google-cloud-cpp/main/.clang-tidy
 # See: https://releases.llvm.org/10.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/list.html
@@ -74,7 +74,7 @@ CheckOptions:
     - { key: readability-identifier-naming.ProtectedMemberSuffix,  value: _          }
     - { key: readability-identifier-naming.EnumConstantCase,         value: CamelCase }
     - { key: readability-identifier-naming.EnumConstantPrefix,       value: k         }
-    - { key: readability-identifier-naming.ConstexprVariableCase,    value: Camel_Snake_Case }
+    - { key: readability-identifier-naming.ConstexprVariableCase,    value: CamelCase }
     - { key: readability-identifier-naming.ConstexprVariablePrefix,  value: k         }
     - { key: readability-identifier-naming.GlobalConstantCase,       value: CamelCase }
     - { key: readability-identifier-naming.GlobalConstantPrefix,     value: k         }
@@ -92,3 +92,12 @@ CheckOptions:
     # OpenAssetIO define a virtual default destructor just to make them
     # polymorphic.
     - { key: cppcoreguidelines-special-member-functions.AllowSoleDefaultDtor,  value: 1 }
+    # Suppressions for `#include` checks.
+    # * Python.h, pystate.h, pytypedefs.h, object.h, pyerrors.h,
+    #   abstract.h - Python.h is the umbrella header.
+    # * toml++/ - toml.h is the umbrella header.
+    # * ada/ - ada.h the umbrella header.
+    # * pybind11/ - pybind11.h is the umbrella header. Other headers add
+    #   functionality (e.g. stl.h), but may be falsely flagged as
+    #   unused.
+    - { key: misc-include-cleaner.IgnoreHeaders,  value: "Python.h;pystate.h;pytypedefs.h;object.h;pyerrors.h;abstract.h;toml\\+\\+/.*;ada/.*;pybind11/.*" }
--- a/src/openassetio-core/src/pluginSystem/CppPluginSystem.cpp
+++ b/src/openassetio-core/src/pluginSystem/CppPluginSystem.cpp
@@ -1,23 +1,32 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright 2024 The Foundry Visionmongers Ltd
+// Copyright 2024-2025 The Foundry Visionmongers Ltd
 // Copyright Contributors to the OpenImageIO project.
 // Much of the cross-platform code below is taken and modified from the
 // OpenImageIO project.
 
+#include <algorithm>
+#include <cstddef>
+#include <exception>
+#include <filesystem>
+#include <iterator>
+#include <memory>
+#include <string>
+#include <string_view>
+#include <utility>
 #ifdef _WIN32
 #include <windows.h>
 #else
 #include <dlfcn.h>
 #endif
 
-#include <filesystem>
+#include <fmt/core.h>
 
-#include <fmt/format.h>
-
+#include <openassetio/export.h>
 #include <openassetio/errors/exceptions.hpp>
 #include <openassetio/log/LoggerInterface.hpp>
 #include <openassetio/pluginSystem/CppPluginSystem.hpp>
 #include <openassetio/pluginSystem/CppPluginSystemPlugin.hpp>
+#include <openassetio/typedefs.hpp>
 
 namespace openassetio {
 inline namespace OPENASSETIO_CORE_ABI_VERSION {
@@ -113,7 +122,7 @@ void CppPluginSystem::scan(const std::string_view paths) {
         paths.substr(pathsStartIdx, pathsEndIdx - pathsStartIdx);
 
     // Check the provided path is actually a searchable directory.
-    if (!std::filesystem::is_directory(directoryPath)) {
+    if (!is_directory(directoryPath)) {
       logger_->debug(fmt::format("CppPluginSystem: Skipping as not a directory '{}'",
                                  directoryPath.string()));
       continue;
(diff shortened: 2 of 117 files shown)


src/openassetio-core/tests/utils/RegexTest.cpp
```diff
@@ -1,9 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright 2023 The Foundry Visionmongers Ltd
-#include <fmt/format.h>
+// Copyright 2025 The Foundry Visionmongers Ltd
```
foundry-markf, line 2: Surely 2023-2025 for consistency with other copyright updates?

src/openassetio-core/tests/utils/PrintableTest.cpp
```diff
@@ -1,27 +1,31 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright 2024 The Foundry Visionmongers Ltd
+// Copyright 2025 The Foundry Visionmongers Ltd
```
foundry-markf, line 2: 2024-2025?

src/openassetio-core/tests/typedefsTest.cpp
```diff
@@ -1,8 +1,10 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2022 The Foundry Visionmongers Ltd
```
foundry-markf, line 2: Unusually not updated copyright

foundry-markf: I'm now spotting a few of these. Is there a guideline for updating/not updating?

src/openassetio-core/tests/utils/RegexTest.cpp
foundry-markf: I'm seeing a number of updates like this, just changing the year, while others have a range. What is the update policy?

src/openassetio-core/src/utils/formatter.cpp
```diff
@@ -98,28 +104,30 @@ auto fmt::formatter<openassetio::trait::TraitSet>::format(
...
-  std::string operator()(openassetio::Int arg) const { return std::to_string(arg); }
-  std::string operator()(openassetio::Bool arg) const { return arg ? "True" : "False"; }
+  std::string operator()(openassetio::Str arg) const;
+  std::string operator()(const openassetio::Float arg) const { return std::to_string(arg); }
+  std::string operator()(const openassetio::Int arg) const { return std::to_string(arg); }
+  std::string operator()(const openassetio::Bool arg) const { return arg ? "True" : "False"; }
 };
+std::string ToStringVisitor::operator()(openassetio::Str arg) const {
+  return fmt::format("'{}'", arg);
+}
```
foundry-markf, line 125: This made me stop and think... I'm not sure why the implementation got pulled outside of the struct, as it doesn't look all that different to the other overloads, with the exception of using `fmt`.

src/openassetio-core/src/pluginSystem/CppPluginSystem.cpp
```diff
@@ -158,7 +164,7 @@ const CppPluginSystem::PathAndPlugin& CppPluginSystem::plugin(const Identifier&
 CppPluginSystem::MaybeIdentifierAndPlugin CppPluginSystem::maybeLoadPlugin(
     const std::filesystem::path& filePath) {
   // Check the proposed path is actually a file.
-  if (!std::filesystem::is_regular_file(filePath)) {
+  if (!is_regular_file(filePath)) {
```
foundry-markf, line 167: This (and similar) changes dropping the `std::filesystem::` namespace puzzled me, as I couldn't see a `using namespace`, and it's used in the signature.

I'm supposing that because the argument is of type `std::filesystem::path` that the compiler can resolve it from all possible names of `is_regular_file`?

It was a bit of a surprise change.

feltech:
@foundry-markf Thanks for the review. I didn't actually realise I'd put up this PR - I thought I'd just done it on my fork, for testing. I'm still in the process of making the necessary code changes to satisfy clang-tidy v19.

I've changed the status of this PR to Draft for now. That said, I'll take a look at your comments asap.

.clang-tidy
```diff
@@ -74,7 +74,7 @@ CheckOptions:
     - { key: readability-identifier-naming.ProtectedMemberSuffix,  value: _          }
     - { key: readability-identifier-naming.EnumConstantCase,         value: CamelCase }
     - { key: readability-identifier-naming.EnumConstantPrefix,       value: k         }
-    - { key: readability-identifier-naming.ConstexprVariableCase,    value: Camel_Snake_Case }
+    - { key: readability-identifier-naming.ConstexprVariableCase,    value: CamelCase }
```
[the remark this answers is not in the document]
feltech: Clang-Tidy v12 was bugged in that it would not flag many violations of this rule. This change conforms the config to the reality in the codebase. IIRC there are a couple of variables that are Snake_Case, which I think will need suppressions (for now), but I've not got around to tackling them just yet.

src/openassetio-core/tests/utils/RegexTest.cpp
feltech: Yeah this is wrong.  I set up my IDE (CLion) to auto-update Copyright years, but it's hit and miss. Sometimes it doesn't update it (e.g. if the change was due to an automatic refactor), and when it does update it, it doesn't introduce a range when it should (like in this case you highlight).  

I am aware of it, and need to do a pass over all modified files and fix the headers.  Another artifact of the fact that I PR'd this accidentally and too early.

src/openassetio-core/src/utils/formatter.cpp
feltech: Oh this is an interesting one, not sure about that, I might have accidentally applied the wrong automated refactor :worried: :roll_eyes:

src/openassetio-core/src/pluginSystem/CppPluginSystem.cpp
feltech: This is a result of CLion's linter (rather than Clang-Tidy) recommending the use of ADL.  I guess its a personal preference, but I liked the conciseness.

In general my process for these changes is to go through the errors/warnings/hints highlighted by CLion inline in the source file, which are largely Clang-Tidy issues, but occasionally CLion's own hints, which I've ad-hoc also applied.

Do you think I should avoid such ADL refactors?

foundry-markf: It came as a surprise, but if all compilers can handle it, I don't disagree with the concise nature of it.
Maybe I don't look at enough variety of C++ code these days.

If it doesn't conflict with other linters (not everyone uses CLion) and compilers, then I don't see a reason to change.

src/openassetio-core/src/utils/formatter.cpp
feltech: Yeah this was indeed an accidental auto-refactor, whoops! Reverted as part of subsequent work, force-pushed.

src/openassetio-core/tests/utils/RegexTest.cpp
feltech: All Copyright headers of modified files should now be correct, i.e. IIRC they're all a range like `Copyright YYYY-2025` (don't think there's any file newly created in 2025).

src/openassetio-core/src/pluginSystem/CppPluginSystemManagerImplementationFactory.cpp
```diff
@@ -46,8 +51,7 @@ CppPluginSystemManagerImplementationFactory::CppPluginSystemManagerImplementatio
     : CppPluginSystemManagerImplementationFactory{
           // getenv returns nullptr if var not set, which cannot be
           // used to construct a std::string.
-          // False positive in clang-tidy-12:
-          // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg)
+          // NOLINTNEXTLINE(*-suspicious-stringview-data-usage)
           [paths = std::getenv(kPluginEnvVar.data())] { return paths ? paths : ""; }(),
```
foundry-markf, line 55: I recall writing something, but then can't see it now for this. Must've lost it in a refresh.

The *-suspicious-stringview-data-usage is for the use of `std::getenv(kPluginEnvVar.data())`.

There is a std::getenv_s (https://en.cppreference.com/w/c/program/getenv) which should get around that, but at the cost of more code.

This is a suggestion more than anything, if it appeals.

foundry-markf (approved):
Looks good. Only one suggestion that doesn't stop me approving.

src/openassetio-core/src/pluginSystem/CppPluginSystemManagerImplementationFactory.cpp
foundry-markf: Drat.
end of line.

PR #1438, opened 2025-01-09
[Docs] Update API docs URL to docs.openassetio.org

### LINKED ISSUE #1400: Migrate API docs links to use docs.openassetio.org
## What

Migrate all references of `openassetio.github.org` to instead use `docs.openassetio.org`

## Why

ASWF has set up a CNAME record for us for `docs.openassetio.org`, and this looks more polished.

feltech:
## Description

Closes #1400. ASWF has set up a CNAME record for us for docs.openassetio.org, and this looks more polished.

## Reviewer notes

Note that we have a CI job, markdown-link-check, that will check the URLs in markdown files.

### DIFF
--- a/src/openassetio-python/package/openassetio/__init__.py
+++ b/src/openassetio-python/package/openassetio/__init__.py
@@ -1,5 +1,5 @@
 #
-#   Copyright 2013-2024 The Foundry Visionmongers Ltd
+#   Copyright 2013-2025 The Foundry Visionmongers Ltd
 #
 #   Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


src/openassetio-python/package/openassetio/__init__.py
```diff
@@ -65,7 +65,7 @@
 -----------------
 
 The documentation for OpenAssetIO can be found here:
-   https://openassetio.github.io/OpenAssetIO.
+   https://docs.openassetio.org/OpenAssetIO.
```
foundry-markf, line 68: Nit: Do you want a full stop at the end of this, as the URL is on a line of it's own?

foundry-markf (approved):
Tried the new URL, works.

I did have a comment on the non-URL change on DR023-Versioning-traits-and-specifications-method.md but the change is not shown in the diff if you hide whitespace, so it's obviously that, but I can't see where it is -

src/openassetio-python/package/openassetio/__init__.py
feltech: Good spot, removed, took the liberty of force-pushing the update.

foundry-markf (approved):

PR #1444, opened 2025-03-11
[CI] Bump actions/upload-artifact to v4

feltech:
`actions/upload-artifact` v3 is now unavailable on GitHub CI and is causing builds to fail, e.g. https://github.com/OpenAssetIO/OpenAssetIO/pull/1443

We had previously skipped a dependabot update due to NodeJS version requirement conflict with the old VFX22 Docker image, see #1232. Since we're now testing on VFX24 this shouldn't be a problem.

### DIFF
--- a/.github/workflows/build-vfx-reference-platform.yml
+++ b/.github/workflows/build-vfx-reference-platform.yml
@@ -45,7 +45,7 @@ jobs:
         cmake --install build
 
     - name: Upload archive
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         name: openassetio-CY2024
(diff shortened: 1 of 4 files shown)


feltech:
We also have a problem due to [a change in behaviour in v4](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes)

> Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.

This will need a little refactoring in the CI.

hutchinson (approved):

Issue #1445, opened 2025-03-12
Implement UI delegation proposal


feltech:
## What

Implement the UI delegation proposal outlined in https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/pull/100

## Why

UI delegation is a big piece of the puzzle required for a compelling OpenAssetIO workflow in artist-facing DCC tools.

The proposal has been validated in Jupyter, Katana and Nuke, by prototyping with mocked-up UIs and Flow Production Tracking.

The proposal hasn't attracted much feedback. It was decided in the March 2025 TSC to get the proposal implemented as-is.

## Acceptance Criteria

* OpenAssetIO - UI delegation support similar to the proposal prototype, but at full engineering quality, with docs and tests.
 * OpenAssetIO-TraitGen - (minor update) should allow UI-specific traits in its schema.
 * OpenAssetIO-MediaCreation
     - First tranche of UI-specific traits.
     - A Jupyter Notebook illustrating the UI delegation API.
  * OpenAssetIO-Manager-BAL - provide a UI delegate plugin alongside the manager plugin, suitable for DCC authors to test UI delegation without access to "real" implementations.
  * KatanaOpenAssetIO - prototype-quality (possibly separate branch) Widget Delegate adapter to support OpenAssetIO UI delegates.
      - Validation of UI delegation in Katana using BAL manager/UI.

## Implementation notes

- OpenAssetIO-Manager-BAL - can potentially build on the [previous prototype](https://github.com/feltech/OpenAssetIO-Manager-BAL/tree/prototype/OpenAssetIO/1403-uiDelegation).
- KatanaOpenAssetIO - can potentially build on the previous prototype (private code needs publishing)
- OpenAssetIO-MediaCreation - should use (and tweak if necessary) the [existing Jupyter Notebook](https://github.com/OpenAssetIO/OpenAssetIO-MediaCreation/pull/100).

PR #1451, opened 2025-03-28
Add "isPython" to `HostInterface.info()`

feltech:
## Description

Part of #1445. OpenAssetIO is a mostly seamless dual-language library, but for some access patterns it may be necessary for managers to be aware that the host is written in Python to ensure correct behaviour.

For example, UI delegation may require UI elements to be passed to the host as Python object wrappers around native UI elements.

### DIFF
--- a/src/openassetio-python/package/openassetio/hostApi/HostInterface.py
+++ b/src/openassetio-python/package/openassetio/hostApi/HostInterface.py
@@ -1,0 +1,44 @@
+#
+#   Copyright 2025 The Foundry Visionmongers Ltd
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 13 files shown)


hutchinson (approved):

src/openassetio-python/package/openassetio/hostApi/HostInterface.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def info(self):
+        """
+        Override base class implementation to add
+        @fqref{constants.kInfoKey_IsPython} "kInfoKey_IsPython".
+
+        @see @fqref{hostApi.HostInterface.info} "HostInterface.info".
+        """
+        out = super().info()
+        out[constants.kInfoKey_IsPython] = True
```
hutchinson, line 43: question: one thought - should the `dict` be checked for the presence of the key before potentially modifying it? feels like overkill I think.

feltech: The superclass is the C++ base class, which just returns an empty dict. We could potentially add a conflicting value to the base class's `info()` in future, but I'm not sure what the use-case for that would be. At least its something we control and can change at a later date if we do need to, without breaking anything.

PR #1454, opened 2025-04-10
Reduce severity of CppPluginSystem search path warning

feltech:
## Description

~~Closes # (issue)~~

Typically, a host application will want to enable both Python and C++ manager plugins (via the Hybrid plugin system).

Python plugins support either `OPENASSETIO_PLUGIN_PATH` environment variable or Python entry-point plugins
(`importlib.metadata.entry_points)`. C++ plugins only support `OPENASSETIO_PLUGIN_PATH`.

We would like to warn the user when there is no possibility of loading a manager plugin, because neither entry-point nor
`OPENASSETIO_PLUGIN_PATH` is configured.

Conversely, we do _not_ want to issue this warning if one of these mechanisms is available.

However, we had a situation where, if `OPENASSETIO_PLUGIN_PATH` is not set, because the user expects a Python entry-point plugin to be used instead, then the C++ plugin system would still warn that no plugins will load. In addition, the log message did not identify that it was coming from the `CppPluginSystem`, like other logs do.

So update the C++ log message to prefix with `CppPluginSystem`, making it consistent with all other logs.

Also reduce the severity of the logs of both Python and C++ plugin systems down to `kDebug`, since in general the fact that one plugin system cannot find a plugin does not mean that all plugin systems will fail, and so we should not be spamming the user's logs with warnings.

## Reviewer Notes

Noticed this whilst working on some updates to KatanaOpenAssetIO.

### DIFF
--- a/src/openassetio-core/src/pluginSystem/CppPluginSystem.cpp
+++ b/src/openassetio-core/src/pluginSystem/CppPluginSystem.cpp
@@ -155,14 +155,22 @@ void CppPluginSystem::scan(const std::string_view paths, const std::string_view 
   if (!paths.empty()) {
     scanPaths(paths);
   } else {
-    // NOLINTNEXTLINE(*-suspicious-stringview-data-usage)
-    const std::string_view pathsFromEnvVar = [maybePaths = std::getenv(pathsEnvVar.data())] {
-      return maybePaths == nullptr ? "" : maybePaths;
+    const std::string_view pathsFromEnvVar = [&] {
+      if (pathsEnvVar.empty()) {  // Defend against uninitialised string_view.
+        return "";
+      }
+      // NOLINTNEXTLINE(*-suspicious-stringview-data-usage)
+      const char* maybePaths = std::getenv(pathsEnvVar.data());
+      if (maybePaths == nullptr) {
+        return "";
+      }
+      return maybePaths;
     }();
 
     if (pathsFromEnvVar.empty()) {
-      logger_->warning(fmt::format(
-          "No search paths specified, no plugins will load - check ${} is set", pathsEnvVar));
+      logger_->debug(fmt::format(
+          "CppPluginSystem: No search paths specified, no plugins will load - check ${} is set",
+          pathsEnvVar));
     } else {
       scanPaths(pathsFromEnvVar);
     }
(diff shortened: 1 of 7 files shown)


src/openassetio-core/src/pluginSystem/CppPluginSystem.cpp
```diff
@@ -161,8 +161,9 @@ void CppPluginSystem::scan(const std::string_view paths, const std::string_view
     }();
 
     if (pathsFromEnvVar.empty()) {
-      logger_->warning(fmt::format(
-          "No search paths specified, no plugins will load - check ${} is set", pathsEnvVar));
+      logger_->debug(fmt::format(
+          "CppPluginSystem: No search paths specified, no plugins will load - check ${} is set",
+          pathsEnvVar));
```
hutchinson, line 166: Does there need to be a `nullptr` check on `pathsEnvVar` or will `fmt` handle null with `${}`? Otherwise this change is good to go.

hutchinson (approved):

src/openassetio-core/src/pluginSystem/CppPluginSystem.cpp
feltech: Good point. Since the previous refactor (#1450), the env var name is now a parameter, and part of the public interface.

Trying out in Compiler Explorer:

```
std::getenv(std::string_view{}.data());

Program returned: 139
Program terminated with signal: SIGSEGV
```

```
fmt::print("{}", std::string_view{}.data());

terminate called after throwing an instance of 'fmt::v8::format_error'
  what():  string pointer is null
```

So if the user passes an uninitialised `std::string_view`, then we have a problem.

All usages within OpenAssetIO set `pathsEnvVar` to point to a `constexpr` string literal, so there is no issue. But theoretically a host application could use the `CppPluginSystem` independently and potentially trigger a segfault.

However, a couple of things make me hesitant to add a check for this:

* Despite being technically public, `CppPluginSystem` is really an internal implementation detail and I don't expect any host application to ever want to use it directly. Internal usages are fully unit tested, so we should be protected against misuse within OpenAssetIO.
* If we add a check for uninitialised `std::string_view` here, then we're opening Pandora's box, since we should also add checks in every other place where a `std::string_view` is accepted as a parameter.

Perhaps it was/is a mistake in general to use `std::string_view` as parameters in public-facing APIs, for this reason.

Do you think we should still go ahead and add a check for uninitialised `std::string_view` here, and elsewhere?  Or switch to `std::string` everywhere for public-facing arguments?

Regardless, this is an issue that predates this PR so should be a separate issue if we decide to address it.

feltech: Added ddbe53ce with a fix, including a test that does indeed reproduce a segfault without the fix.

hutchinson (approved):

PR #1461, opened 2025-06-10
[Core] Move macro definitions to new macros.hpp header

### LINKED ISSUE #1437: Split macros out of typdefs header
## What

Move macros such as `OPENASSETIO_FWD_DECLARE` out of the `typedefs.hpp` header and into its own `macros.hpp` header.

## Why

Consumers of OpenAssetIO may wish to forward declare OpenAssetIO types in their headers, but without the compiler burden of pulling in standard library types. Currently `typedefs.hpp` defines both utility macros as well as various aliases of standard library types. So a downstream project must include potentially unused (and large) standard library headers, just to get access to our macros.

Hofynea:
## Description
Moves macros from typedefs.hpp to a new macros.hpp.
typedefs.hpp now includes macros.hpp to keep existing behavior.

Closes #1437 

- [ ] I have updated the release notes. (Not needed for this change)
- [ ] I have updated all relevant user documentation. (No user-facing impact)


src/openassetio-core/include/openassetio/macros.hpp
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2013-2023 The Foundry Visionmongers Ltd
+#pragma once
+
+#define OPENASSETIO_DECLARE_PTR(Class)             \
+  /* NOLINTNEXTLINE(bugprone-macro-parentheses) */ \
+  class Class;                                     \
+  using Class##Ptr = std::shared_ptr<Class>;       \
+  using Class##ConstPtr = std::shared_ptr<const Class>;
```
feltech, line 9: [fix] Cpplint [is complaining](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/15581247787/job/43882282699?pr=1461) that we should

> Add #include <memory> for shared_ptr<>  [build/include_what_you_use]

which would defeat the whole point of this file :roll_eyes: 

So we should suppress the linter in this case. I believe a `// NOLINT(build/include_what_you_use)` comment at the end of the line will do it (clang-tidy and cpplint use the same format for suppressions).

Cpplint might complain about other lines too. If you're building OpenAssetIO locally, it would be worth using `--preset lint` in CMake, or enabling the [linter CMake options](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/BUILDING.md#cmake-options) manually.  

Otherwise you can push again and see if CI fails, but this may take a while, especially as you'll have to wait for me to start a CI run, so it'll be a slow feedback loop.

src/openassetio-core/include/openassetio/macros.hpp
feltech, line 1: [fix] For the commits themselves

* Please squash all the commits down to one commit. In general we avoid merging commits where the codebase is in a broken or incomplete state, to help with git archaeology and bisecting.
    - If the intention was always to squash them down after review, then you can flag this by prefixing the commit message with `fixup:` or `squash:`, or using [autosquash format](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash).
* You need to sign off the commits (see [DCO check](https://github.com/OpenAssetIO/OpenAssetIO/pull/1461/checks?check_run_id=43876821432)). Its just an extra line at the bottom of the commit message.
* Commit message titles are good, thanks for using the correct component tag.  But please reference the GitHub issue in the body of the commit message. Just something like `Closes #1437.` will do in this case.

feltech (changes requested):
Thanks! Looking good, just a few minor niggles.

feltech:
Minor point, please don't Resolve review comments, that's up to the reviewer once they've checked its been fixed. 

Just reply to say its been fixed, with any relevant details (e.g. with a commit hash if it was fixed in a new commit, though I note you force-pushed the fixes, which is fine for a small PR like this).

src/openassetio-core/include/openassetio/macros.hpp
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Assumes the surrounding namespace is valid for the given class.
+ *
+ * Declares both non-const and const variants as `ClassPtr` and
+ * `ClassConstPtr`, respectively.
+ */
+#define OPENASSETIO_DECLARE_PTR(Class)             \
+  /* NOLINTNEXTLINE(bugprone-macro-parentheses) */ \
+  class Class;                                     \
+  using Class##Ptr = std::shared_ptr<Class>;       \ // NOLINT(build/include_what_you_use)
```
feltech, line 15: [fix] Looks like CI is failing because the comment comes after the `\`.  I think you'll have to put the comment in `/* ... */` before the `\`.

feltech (changes requested):

Hofynea:
To be very honest I've never seen anything like this before so I didn't know how to properly follow PR etiquette. I should've asked I guess. Now I know. Thanks for the heads-up! 😊

src/openassetio-core/include/openassetio/typedefs.hpp
```diff
@@ -8,8 +8,7 @@
 #include <utility>
 #include <vector>
 
-#include <openassetio/export.h>
```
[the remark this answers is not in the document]
Hofynea: Fixed — re-added `export.h` include to restore `OPENASSETIO_CORE_ABI_VERSION` reference.

src/openassetio-core/include/openassetio/macros.hpp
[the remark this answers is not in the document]
Hofynea: Fixed — amended and force-pushed with a verified GPG signature.  
I had initially confused "signed-off-by" with actual GPG signing — thanks for the clarification!  
Now I finally know what "Verified" after the commit message actually stands for 😄

src/openassetio-core/include/openassetio/macros.hpp
feltech: I'm kinda surprised this works, i.e. that the NOLINT applies to the following line too, but cpplint isn't complaining about this specific issue any more.

However, now both cpplint and clang-tidy [is complaining](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/15619139028/job/44036846770?pr=1461), I think because there needs to be two spaces before the `// NOLINT`.

Again, it would save a lot of time if you could build locally to check the linters are happy before pushing.

src/openassetio-core/include/openassetio/typedefs.hpp
```diff
@@ -8,8 +8,8 @@
 #include <utility>
 #include <vector>
 
+#include <openassetio/macros.hpp>
 #include <openassetio/export.h>
```
feltech, line 12: [fix] CI linters [are complaining](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/15619139028/job/44036846770?pr=1461).  Clang-Tidy is complaining I think because the headers are not in alphabetical order. Cpplint is complaining because it wants C headers before C++ headers.

I think just reversing the order here will work.

feltech: Looks like you've renamed `export.h` to `export.hpp`, which doesn't exist and is failing CI.

src/openassetio-core/include/openassetio/macros.hpp
feltech, line 1: [fix] Could we perhaps reword the commit message - "Apply review suggestions to macro refactor" is not super descriptive :wink:

feltech, line 1: [consider] I should have brought this up sooner, but this is a useful change that I think should be called out in RELEASE_NOTES.md.  Something simple under the top `## Improvements` section, e.g.

> Split common C++ macros from `typedefs.hpp` into a separate zero-dependency `macros.hpp` header.

followed by a link to the issue (i.e. similar to other bullets in the release notes).

feltech (changes requested):

feltech:
@Hofynea I hope you don't mind, but I really wanted this change to make it into the next release, so I took the liberty of pushing some changes to get this over the line.

- Rebased on the latest `main`.
- Fixed remaining compiler/linter errors.
- Updated the commit message.

feltech (approved):

Hofynea:
@feltech, I don’t mind at all. We’re in the middle of a move, so I’ve been juggling a few things and didn’t get to those last changes in time. I really appreciate your help closing this out — I learned a lot working on it.

PR #1464, opened 2025-06-17
[CI] Add markdown-link-check-suppression for lists.aswf.io

feltech:
The lists.aswf.io link in the README has been consistently failing the markdown-link-check CI job for several days with

> [✖] https://lists.aswf.io/g/openassetio-discussion → Status: 403

i.e. a "Forbidden" error. The link works fine in a local browser and in `curl`, so it must be that GitHub runners are blocked, somehow.

So add a suppression to stop this from failing CI.

### DIFF
--- a/.github/markdown-link-check-config.json
+++ b/.github/markdown-link-check-config.json
@@ -5,6 +5,9 @@
     },
     {
       "pattern": "^https://www.gnu.org"
+    },
+    {
+      "pattern": "^https://lists.aswf.io"
     }
   ]
 }


feltech:
:angry: And now suddenly markdown-link-check fails for github.com URLs with 429 Too Many Requests. Hopefully its just transient, but already tried to re-run once and it had the same issue.

feltech:
OK, all good. Given this is just a tiny CI config change, then being pragmatic I'll use admin powers to merge without a review :worried:

PR #1465, opened 2025-06-23
1103 ensure pragma once is at top of files

### LINKED ISSUE #1103: Ensure #pragma once is at top of files.

Gmaj9:
## Description

Closes #1103.

This PR ensures the `#pragma once` directive is at the top of all headers in `src`. (For this, I used a scope in CLion to filter for `.h` and `.hpp` files and checked them one by one.)

There are also nitpicky changes to make the spacing around `#pragma once` consistent (`#pragma once` now occurs on the line directly below the copyright comment and any file docstring and a newline follows after `#pragma once` before any `#include` declarations).

## Reviewer Notes

There are two commits: one moves `#pragma once` and the second tweaks whitespace.

Since this is small, does anything need updating per the checkboxes above?

## Test Instructions

I have test failures on my machine but they failed the same way before and after my changes. (They look to be because my Python version and something else with my environment/machine is slightly off, which I had discussed in Slack with @feltech.)

### DIFF
--- a/src/openassetio-core-c/src/managerApi/CManagerInterfaceAdapter.hpp
+++ b/src/openassetio-core-c/src/managerApi/CManagerInterfaceAdapter.hpp
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright 2013-2024 The Foundry Visionmongers Ltd
+// Copyright 2013-2025 The Foundry Visionmongers Ltd
 #pragma once
+
 #include <string>
 
 #include <openassetio/c/export.h>
--- a/src/openassetio-core-c/tests/StringViewReporting.hpp
+++ b/src/openassetio-core-c/tests/StringViewReporting.hpp
@@ -1,10 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright 2013-2024 The Foundry Visionmongers Ltd
+// Copyright 2013-2025 The Foundry Visionmongers Ltd
 /**
  * Comparison and stream operators to simplify asserting and reporting
  * of StringView instances during tests.
  */
 #pragma once
+
 #include <openassetio/c/StringView.h>
 #include <openassetio/c/namespace.h>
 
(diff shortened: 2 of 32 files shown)


src/openassetio-core-c/src/managerApi/CManagerInterfaceAdapter.hpp
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2013-2024 The Foundry Visionmongers Ltd
```
feltech, line 2: [fix] Here and elsewhere, if we've touched the files, even for something so minimal, we should really update the copyright date range. So e.g. (looking at a couple of examples) "2013-2024" becomes "2013-2025" and "2022" becomes "2022-2025".

feltech (changes requested):
Thanks for this!  Looking good other than minor gripe re. copyright years. 

Thanks for splitting the commits and the descriptive commit messages that match the guidelines, surprisingly rare to see!

src/openassetio-core-c/src/managerApi/CManagerInterfaceAdapter.hpp
Gmaj9: OK, I'll address that too.

Gmaj9: It should be good now.

feltech (changes requested):
Code all looks good!  Last [minor gripe](https://github.com/OpenAssetIO/OpenAssetIO/pull/1465#discussion_r2163536880) re. commit structure.

PR #1469, opened 2025-09-11
Add project docs as per ASWF guidelines

feltech:
## Description

In preparation for the project moving to the ASWF incubation phase, add various docs at the root level covering project governance and contribution guidelines, as listed in the annual review slide template (available [here](https://tac.aswf.io/process/review_cycle.html)) listing requirements for an Incubation stage project.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,62 +1,74 @@
 # Roadmap
 
-The OpenAssetIO project is working towards an initial v1.0.0 release of
-the core API.
+The OpenAssetIO project has reached a stable v1 release of the core API.
 
-We are looking for early adopters who can help fine tune its final form,
-and validate key design choices, and help flesh out asset types and
-their properties.
-
-The project is functional, and moving forward we expect minimal churn in
-the API itself as we complete the implementation of language specific
-features.
+However, there are still several features planned for the core API, and
+the broader ecosystem remains at in a broadly beta-level state of
+maturity.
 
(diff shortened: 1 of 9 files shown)


ROADMAP.md
```diff
@@ -1,62 +1,74 @@
...
+OpenAssetIO is adopted by hosts.
 
-- [ ] Trait/Specification versioning support
-- [ ] Debug trace logging support.
-- [x] Entity introspection API methods.
-- [x] C++ Plugin System
-- [x] Hybrid C++/Python manager bridge.
+Given that many asset management systems are bespoke and private, the
+main focus is on encouraging adoption by host applications, both
+commercial and open source.
```
rfanner, line 70: Nits. Feel free to ignore if they don't resonate:

```OpenAssetIO has the "chicken and egg" problem with respect to adoption:
Despite the longer-term benefits that would be realised, first-adopters on the *host* side are faced with a leap of faith if there is limited manager support, and likewise *asset managers*/*studios* may also face this leap of faith if there is limited host support.```

I think the host application observation feels on point for the core library, but perhaps it's less applicable to the overall ecosystem? I've tried to take a stab at capturing what we've been doing over the last few quarters below.

```Part of the project plan is therefore to lower these barriers for first-adopters by providing more official integrations and integration examples, whether directly on the project or via collaboration.```

Thoughts?

rfanner (approved):
Thanks for sorting these items out, David!

PR #1495, opened 2026-04-21
[Python] Support Python 3.13

adambowen-foundry:
## Description

Closes #1487.

Python 3.13 is required by VFX Reference Platform CY26. Currently the source does not compile because `_Py_IsFinalizing` has been made part of the stable ABI and renamed to `Py_IsFinalizing`. We update the references to the new name and maintain compatibility with older Python versions using a macro.

The tests pin specific setuptools and pip versions that are not compatible with Python 3.13, so also update those to allow tests to be run against the new Python version.

Update the CI to build Python 3.13 wheels, but do not update the toolchain or MacOS minimum deployment target in line with CY26. In particular GCC14 is only officially available from 24.04 which would further bring the glibc version out of alignment with the reference platform. Note that the wheels are built with `manylinux_2_28` by `cibuildwheel` on Linux, which is GCC14 and glibc 2.28 compatible anyway.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -37,7 +37,8 @@ jobs:
           # prefix here so that it can also be used in the file name of
           # the uploaded archive (i.e. wildcard `*` is not allowed).
           # See https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip
-        python-build: ['cp310', 'cp311', 'cp312']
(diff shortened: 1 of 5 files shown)


feltech (approved):
LGTM

feltech:
No way, Copilot has nothing to say? I didn't think that was even possible :laughing:

PR #1506, opened 2026-05-21
Reorder Manager pybind overload registrations

### LINKED ISSUE #1406: Reorder pybind11 method overloads by likelihood
## What

Re-order overloads of methods in pybind11 bindings (esp. for `Manager`) such that the most likely overload is listed first.

## Why

When given a method name and arguments, pybind11 will search all overloads linearly until it finds a match for the argument list. So having the most likely methods listed first is better for performance. Currently this isn't the case for most methods, which is a small pessimisation.

marivenkatakrishnan:
## Description

Reorders the existing pybind11 overload registrations for `hostApi.Manager` in the below order:

  1. Singular default
  2. Singular explicit variant/result
  3. Singular explicit exception
  4. Batch default
  5. Batch explicit variant/result
  6. Batch explicit exception
  7. Callback

  No public API, behaviour are changed.

Closes #1406 

 Release notes and user documentation were not updated because this is an internal binding registration.

## Reviewer Notes

pybind11 resolves overloads linearly in registration order. This PR changes `Manager` overload bindings only; it should make without changing API behaviour.

## Test Instructions

 Verify the diff is limited to reordering overload registrations in `src/openassetio-python/cmodule/src/hostApi/ManagerBinding.cpp with no extra addition or deletion.

### DIFF
--- a/src/openassetio-python/cmodule/src/hostApi/ManagerBinding.cpp
+++ b/src/openassetio-python/cmodule/src/hostApi/ManagerBinding.cpp
@@ -127,6 +127,22 @@ void registerManager(const py::module& mod) {
           },
           py::arg("traitSet"), py::arg("defaultEntityReferenceAccess"),
           py::arg("context").none(false), py::call_guard<py::gil_scoped_release>{})
+      .def("defaultEntityReference",
+           py::overload_cast<const trait::TraitSet&, access::DefaultEntityAccess,
+                             const ContextConstPtr&,
+                             const Manager::BatchElementErrorPolicyTag::Variant&>(
+               &Manager::defaultEntityReference),
+           py::arg("traitSet"), py::arg("defaultEntityReferenceAccess"),
+           py::arg("context").none(false), py::arg("errorPolicyTag"),
+           py::call_guard<py::gil_scoped_release>{})
+      .def("defaultEntityReference",
+           py::overload_cast<const trait::TraitSet&, access::DefaultEntityAccess,
+                             const ContextConstPtr&,
+                             const Manager::BatchElementErrorPolicyTag::Exception&>(
+               &Manager::defaultEntityReference),
+           py::arg("traitSet"), py::arg("defaultEntityReferenceAccess"),
+           py::arg("context").none(false), py::arg("errorPolicyTag"),
+           py::call_guard<py::gil_scoped_release>{})
       .def(
           "defaultEntityReference",
           [](Manager& self, const trait::TraitSets& traitSets,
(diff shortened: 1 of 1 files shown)


feltech (changes requested):
This is great, nice and consistent, and minor perf boost.

Copilot seems to have caught a genuine issue with arg naming (which I missed!).

Also please can you re-word your commit title to prefix a [component name](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/contributing/COMMITS.md#commit-messages).  I think `[Python]` is probably the correct one here (tricky to decide between `[Core]` and `[Python]` when its to do with bindings).

feltech:
The markdown-link-check CI action is (once again) failing on en.cppreference.com. Nothing to do with this PR, and a false positive anyway - the link works find locally, so they must be blocking GitHub CI runner machines.  We'll push a suppression for that in a separate PR.

feltech:
Also apologies for taking so long to review this, it's been holiday time in these parts.

marivenkatakrishnan:
> Also apologies for taking so long to review this, it's been holiday time in these parts.

No Issues at all, i have also got occupied, updated the PR with the copilot suggesstion, Kindly review the PR.

feltech (changes requested):
The C++ linting CI check found a genuine formatting issue - please run clang-format on the sources.

Please can you squash the commits into one? For future reference, for review fixes we would encourage use of `fixup!`s as mentioned - albeit briefly - in the [contribution docs](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/contributing/COMMITS.md#commit-messages). It also looks like that second commit is missing the DCO "Signed-off-by", but that won't be a problem if it's squashed away.

Also, as mentioned in my previous review comment, 

> Also please can you re-word your commit title to prefix a [component name](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/contributing/COMMITS.md#commit-messages).  I think `[Python]` is probably the correct one here (tricky to decide between `[Core]` and `[Python]` when its to do with bindings).

feltech (changes requested):
Thanks. Nearly there!  

Please can you rebase on the latest `main` branch.

And update your commit message to at least have the issue number, just `Closes #1406` would be sufficient, though a more elaborate message explaining the reason for change would be appreciated - a near copy-paste of the "Why" section from the issue is a good strategy (GitHub _might_ not always be available for reference 😉 ).

marivenkatakrishnan:
Sorry couldn't complete the left over due to a viral infection and subsequent hospitalisation, Kindly let me know if anything is needed.

feltech (approved):
LGTM! Thanks for sticking with it!

marivenkatakrishnan:
thank you for your guidance,

PR #1509, opened 2026-06-03
[Core]  Throw descriptive error if identifier not found in config file

### LINKED ISSUE #1508: Throw descriptive error if identifier not found in config file
## What

If a config file (located via `OPENASSETIO_DEFAULT_CONFIG`) does not contain a top-level `manager.identifier`, raise a `ConfigurationException` with message 

> Configuration file is missing 'identifier' field in its [manager] section at '${configPath}'

(or similar).

## Why

Currently, if a TOML configuration file is found and parsed successfully, but does not contain a `manager.identifier` field, the identifier is assumed to be an empty string, and we get an unhelpful exception

```
Failed to create the default OpenAssetIO Manager (Unexpected): HybridPluginSystem: No plug-in registered with the identifier ''
```

Debugging this is tricky without knowing where to look and what the correct field should be (e.g. someone could assume `id` or `name` etc), and at time of writing the documentation for config file structure is [rather buried](https://docs.openassetio.org/OpenAssetIO/classopenassetio_1_1v1_1_1host_api_1_1_manager_factory.html#a0e8c895c03aea2ad2bd5d8aa4764a0ea).

We should detect early that the identifier is not set and throw a helpful exception saying what the problem is and how to fix it.

## Notes

At time of writing, the root cause is in `src/private/include/openassetio/private/hostApi/factory.hpp`, specifically `identifierAndSettingsFromConfigFile`, where we use `.value_or("")`:

```cpp
identifier = config["manager"]["identifier"].value_or("");
```

The git history has no justification for this, so we're free to change it.

## Acceptance Criteria

- Throw a `ConfigurationException` as described above, if there is no `identifier` in the config file.
- Update docstrings for public methods that make use of `identifierAndSettingsFromConfigFile`.
- Add tests for this error case for all public methods that make use of `identifierAndSettingsFromConfigFile`

raghav3615:
Closes #1508
Description

Update `factory::identifierAndSettingsFromConfigFile` to explicitly check for the presence of the `manager.identifier` field in the TOML configuration file. If the field is missing, it now raises a `ConfigurationException` with a descriptive error message indicating the file path and the missing field. This avoids a confusing downstream exception where a missing identifier was silently parsed as an empty string.

Test Instructions

Python tests verify the new `ConfigurationException` is raised with the correct descriptive message when a config file lacking the `manager.identifier` field is loaded via `ManagerFactory` and `UIDelegateFactory`.

### DIFF
--- a/src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
+++ b/src/openassetio-core/include/openassetio/hostApi/ManagerFactory.hpp
@@ -219,7 +219,8 @@ class OPENASSETIO_CORE_EXPORT ManagerFactory final {
    *
    * @throws errors.ConfigurationException if there are errors occur
    * whilst loading the TOML file referenced by the @ref
-   * default_config_var env var.
+   * default_config_var env var, or if it is missing the
+   * `manager.identifier` field.
    */
   [[nodiscard]] static ManagerPtr defaultManagerForInterface(
       const HostInterfacePtr& hostInterface,
@@ -271,7 +272,8 @@ class OPENASSETIO_CORE_EXPORT ManagerFactory final {
    * config file does not exist at the path provided in @p configPath.
    *
    * @throws errors.ConfigurationException if there are errors occur
-   * whilst loading the TOML file.
+   * whilst loading the TOML file, or if it is missing the
+   * `manager.identifier` field.
    */
   [[nodiscard]] static ManagerPtr defaultManagerForInterface(
       std::string_view configPath, const HostInterfacePtr& hostInterface,
--- a/src/openassetio-python/tests/cmodule/gil/test_uidelegatefactory_gil.py
+++ b/src/openassetio-python/tests/cmodule/gil/test_uidelegatefactory_gil.py
@@ -151,7 +151,7 @@ class Test_UIDelegateFactory_gil:
     ):
         mock_ui_delegate_impl_factory.mock.instantiate.return_value = mock_ui_delegate_interface
         config_path = tmp_path / "ui_delegate.toml"
-        config_path.write_text('[ui_delegate]\nidentifier = "something"')
+        config_path.write_text('[manager]\nidentifier = "something"')
 
         # First overload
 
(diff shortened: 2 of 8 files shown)


raghav3615:
@feltech can you please review this

feltech (changes requested):
This is great! You jumped on this quick 😉 

Solution looks good. Great attention to detail with the test cases too. 

It seems your changes have found a bug in an existing test `Test_UIDelegateFactory_gil.test_defaultUIDelegateForInterface`, which creates an invalid stub config file.

It's good that you've pushed fixes as separate commits 👍.  However, these should be squashed before merge. So a good idea to use `fixup!` or `squash!` commits ([autosquash-style](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash)) for fixes to an already-pushed commit, as (poorly) documented in the [contribution docs](https://github.com/OpenAssetIO/OpenAssetIO/blob/main/doc/contributing/COMMITS.md#commit-messages).

I would suggest trying to run tests and linters locally if at all possible, since I (or another maintainer) has to approve every CI run, so the turnaround time could get rather long.

raghav3615:
hey @feltech can you review it now, as it would be difficult for me to use run the CI locally

feltech (changes requested):
Latest changes look fine, pending CI checks.

I note that to get the PR up to date with `main`, you merged `main` into the PR branch.  Being a bit picky, but if you look at (most of) our git history, there are no merges from main into PR branches, we always rebase on `main` and force-push, to keep the history clean.

raghav3615:
@feltech I clean the commit history and fix the 2 CI fails, can you check now?

raghav3615:
@feltech can you run the ci, so I can check if it fails or not

feltech (changes requested):
Looking good, thanks for sticking with it.  Almost good to go, just need to GPG sign your commit, as mentioned in the GitHub checks

> Commits must have verified signatures

feltech (approved):
Given

* It's been more than a week since my last review.
* This PR is finished, just needed GPG signing.
* It's a really useful little fix.

I've taken the liberty of rebasing the branch on the latest `main` myself, which also has the side-effect of GPG signing the commit with my key.

